PR #11, opened 2017-12-08
Add missing files (previously ignored by gitignore)

gustavla:
Add missing files.


igiloh:
You're missing this files: `src/unity/python/README.md`.   
Other than that seems legit.

PR #12, opened 2017-12-08
Add missing files

znation:
These were ignored by .gitignore but need to be in the repo.

Proposed as an alternative to #11 -- this commit doesn't add back quite as many files, excluding those I would deem unnecessary, like HTML docs of dependencies.


znation:
Ok, looks like this one actually still breaks the build. Going with #11 since it's more complete; will address some issues there.

PR #36, opened 2017-12-09
Fix issue in inline code snippets

BalestraPatrick:
The previous syntax for the install issues is inconsistent a probably wrong, so I fixed by adding quotes to the error message and include the full title sentence in a code snippet.


TobyRoseman:
Thank you @BalestraPatrick

PR #40, opened 2017-12-10
Cleanup

delirious-lettuce:
1. Use `is` or `is not` for comparisons to `None`: 124f130
    > Comparisons to singletons like None should always be done with is or is not, never the equality operators.
2. Simplify boolean variable checks: 63a8f26
    > Don’t compare boolean values to True or False using ==:
    >
    > Yes:
    > if greeting:
    >
    > No:
    > if greeting == True:
    >
    > Worse: 
    > if greeting is True:

3. Remove duplicate dictionary key/value pairs: b3ddc2f

[(source for 1 & 2)](http://pep8.org/#programming-recommendations)

### DIFF
--- a/scenario-tests/tests/driver_test/driver_test.py
+++ b/scenario-tests/tests/driver_test/driver_test.py
@@ -9,4 +9,4 @@ class DriverTest(unittest.TestCase):
         raise unittest.SkipTest("expected skip")
 
     def test_driver_passthrough(self):
-        self.assertTrue(os.environ.get("TEST_ENVIRON_SET") == None)
+        self.assertTrue(os.environ.get("TEST_ENVIRON_SET") is None)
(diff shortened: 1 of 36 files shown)


srikris:
Thanks for the PR. We really appreciate the effort in helping improve this project. For the ease of reviewing, can we split this PR into 3 PRs for each of the cleanup items:

- Use is or is not for comparisons to None
- Simplify boolean variable checks
- Remove duplicate dictionary key/value pairs

In general, we would prefer smaller PRs that fix specific things over larger PRs that do multiple things. It makes it easier to review and test. The easiest way to do that would be to make 3 branches and cherry pick each of the commits and make a PR for each of those 3 branches. Once you do that, I will close this PR.

Thanks!

delirious-lettuce:
@srikris ,

My mistake, I will split this into three PRs as you requested. I'll close this one now to save you a click!

PR #42, opened 2017-12-10
fix broken link

BL4CKvGHOST:
**../sframe/sframe-intro.html not exitsts!.**

changed from `../sframe/sframe-intro.html` to `../sframe/sframe-intro.md`


znation:
Thanks for fixing this, @BL4CKvGHOST!

PR #44, opened 2017-12-10
Remove duplicated README.md

igiloh:
The file src/unity/python/README.md seems to be a duplicate of the main REDAME.md file, but without the latest updates.


znation:
@igiloh That file is needed for the description on PyPI. Can we make it a symbolic link, or a hard link, instead of a copy? (Are links supported in `git` in a cross-platform way?)

igiloh:
Good question. I the same idea, but I'm not sure...
I'll google it.

igiloh:
It's seems that symlinks are generally allowed in git, but:
> Git can handle symlinks just fine as long as the operating system used by all developers supports them.

So we'd better first test on all supported OSes before taking this route. I would suspect windows 10 WSL might give us some problems...

Sources: https://stackoverflow.com/questions/954560/how-does-git-handle-symbolic-links
https://stackoverflow.com/questions/15465436/git-how-to-handle-symlinks

srikris:
The file for PyPI must be a README.rst file and the content should be only a subset of what we have in the github README.md. We should not use a symlink.

src/unity/python/README.md
```diff
@@ -1,132 +0,0 @@
-<img align="right" src="https://docs-assets.developer.apple.com/turicreate/turi-dog.svg" alt="Turi Create" width="100">
-
-# Turi Create 
-
-Turi Create simplifies the development of custom machine learning models. You
-don't have to be a machine learning expert to add recommendations, object
-detection, image classification, image similarity or activity classification to
```
srikris: Rename this file to README.rst

src/unity/python/README.md
```diff
@@ -1,132 +0,0 @@
...
-your app.
-
-* **Easy-to-use:** Focus on tasks instead of algorithms
-* **Visual:** Built-in, streaming visualizations to explore your data
-* **Flexible:** Supports text, images, audio, video and sensor data
-* **Fast and Scalable:** Work with large datasets on a single machine
-* **Ready To Deploy:** Export models to Core ML for use in iOS, macOS, watchOS, and tvOS apps
-
-Example: Image classifier with a few lines of code
```
srikris: This section is not required in the pypi readme.

src/unity/python/README.md
```diff
@@ -1,132 +0,0 @@
...
--------------------
-
-* Python 2.7 (Python 3.5+ support coming soon)
-* x86\_64 architecture
-* macOS 10.11+, Linux with glibc 2.12+ (including WSL on Windows 10)
-
-Installation
-------------
-
-For detailed instructions for different varieties of Linux see [LINUX\_INSTALL.md](https://github.com/apple/turicreate/LINUX_INSTALL.md).
```
srikris: This section is not required in the pypi readme.

src/unity/python/README.md
```diff
@@ -1,132 +0,0 @@
...
--------------
-
-The package [User Guide](https://apple.github.io/turicreate/docs/userguide) and [API Docs](https://apple.github.io/turicreate/docs/api) contain
-more details on how to use Turi Create.
-
-GPU Support
------------
-
-By default, `turicreate` takes a dependency on the default installation of
-`mxnet`. To enable GPU support after installation of the `turicreate` package,
```
srikris: This section is not required in the pypi readme.

src/unity/python/README.md
```diff
@@ -1,132 +0,0 @@
...
-source ~/venv/bin/activate
-
-# Install Turi Create in the new virtual environment, pythonenv
-(venv) pip install -U turicreate
-```
-
-Documentation
--------------
-
-The package [User Guide](https://apple.github.io/turicreate/docs/userguide) and [API Docs](https://apple.github.io/turicreate/docs/api) contain
```
srikris: This section in the PyPI readme should be modified to point to the Github too and we can make it bullet points (see https://github.com/apple/coremltools/blob/master/README.rst#more-information for an example)

znation:
@srikris:

> The file for PyPI must be a README.rst file

Not sure that's true, see http://peterdowns.com/posts/first-time-with-pypi.html -  but perhaps this page is incorrect? This is the only article I can find that claims README.md actually works. Unfortunately it was also the first article I came across when looking for how to properly set up the package for PyPI, so that's how we ended up with a README.md here. I'm fine changing it to rst, and it sounds like that will probably work better, but doesn't the syntax for things like links need to change somewhat if we do that?

src/unity/python/README.md
znation: Why not?

znation:
Also, what's the general guideline for what kinds of content should go in the PyPI readme? I assumed everything from the GitHub readme should go in there, except for anything about building from source or contributing - the install instructions for Linux, GPU, etc. seem just as relevant there as on GitHub.

srikris:
@znation I suppose the .md file might work too. I've only done .rst in the past.

With respect to the REAMDE.md, the intro content can be the same and we only need documentation content and a pointer to the source repo. I've commented on this PR on which sections I think can be skipped for the PyPI readme.

znation:
@srikris Sounds good. 👍

gustavla:
I have only used reST on PyPI and I'm unaware of Markdown working there. However, GitHub supports both md and rst, so we can always switch to using rst for the main readme (pandoc can easily convert it). Keeping two separate readmes up-to-date is more work, although I agree with @srikris that a leaner readme is nicer on PyPI.

znation:
@gustavla Good to know. I think it probably makes the most sense for us to just maintain two different READMEs, a `.rst` for PyPI that's pared down and links to GitHub, and the full `.md` for GitHub.

src/unity/python/README.md
igiloh: I think system requirements are actually relevant, aren't they?

src/unity/python/README.md
igiloh: Done

src/unity/python/README.md
igiloh: I've changed the file according to what agreed here in the conversation. Please review again.

igiloh:
I've changed the file according to what agreed here in the conversation. Please review again.

znation:
@igiloh The contents of the `README.rst` still look to me like Markdown format. See http://docutils.sourceforge.net/rst.html to convert it to ReStructuredText syntax.

gustavla:
@znation is right, the syntax is still Markdown (the links in particular). If you have pandoc:

```
pandoc --from=markdown --to=rst --output=README-fixed.rst README.rst
```

igiloh:
Thanks @gustavla! I still had to do some manual conversions, but it indeed saved me some time.

@znation @srikris - please review for both format and content.

src/unity/python/README.md
```diff
@@ -1,132 +0,0 @@
-<img align="right" src="https://docs-assets.developer.apple.com/turicreate/turi-dog.svg" alt="Turi Create" width="100">
```
znation: You can delete this file entirely.

src/unity/python/README.rst
```diff
@@ -0,0 +1,35 @@
+Turi Create
+===========
+
+Turi Create simplifies the development of custom machine learning
+models. You don’t have to be a machine learning expert to add
+recommendations, object detection, image classification, image
+similarity or activity classification to your app.
+
+Supported Platforms Turi Create supports:
```
znation: Change this to just say `Supported Platforms` - the rest is redundant since this whole document is about Turi Create.

src/unity/python/README.md
igiloh: I did. GH has a very strange way of showing it in the diff here...
if you'll checkout my branch and do `git log --name-status` you can see it's deleted.

src/unity/python/README.rst
igiloh: Oh, it's a typo, probably from the md-->rst conversion.
There should be a \<newline> there, as in:

Supported Platforms 
--------------------

Turi Create supports:
mac ...

PR #47, opened 2017-12-10
[AC] Bugfix - the last session was sometimes not counted

igiloh:
In the sequence_iterator's prep_data() function, we've only counted the last session of each dataset
if it wasn't divisible by chunk_len.
This could theoretically lead to setting num_gpus (which depends on the number of sessions) incorrectly, and even zero.

I've fixed it so the last session is always counted.


igiloh:
I've thought of that - but since the `session_id` column is allowed to be either string or int - this could be a bit problematic.

PR #50, opened 2017-12-10
Simplify boolean variable checks

delirious-lettuce:
(cherry picked from commit 63a8f26)

> Don’t compare boolean values to True or False using ==:
> 
> Yes:
> if greeting:
>
> No: 
> if greeting == True:
>
> Worse: 
> if greeting is True:

[source](http://pep8.org/#programming-recommendations)

### DIFF
--- a/src/unity/python/turicreate/data_structures/gframe.py
+++ b/src/unity/python/turicreate/data_structures/gframe.py
@@ -87,7 +87,7 @@ class GFrame(SFrame):
         if not isinstance(column_name, str):
             raise TypeError("Invalid column name: must be str")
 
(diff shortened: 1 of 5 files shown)


gustavla:
Thanks @delirious-lettuce for the PR! I agree with this style change and I think it improves readability a lot.

Since we currently don't have automatic tests, we have to make sure to run the full test suite locally, since some of these changes are not strictly equivalent. For instance, if `x` is `[1]` then `bool(x)` is True but `x == True` is not. I will kick off the tests now and write another comment when it's done.

gustavla:
Tests are OK.

PR #54, opened 2017-12-11
Docs: add notes on conda install, fix whitespace

pavopax:
I have conda installed, which created some issues when trying to install turi using system pip, as recommended.

I provide some notes for conda users which may be helpful.

(I think this assumes I installed system pip in the past. Let's discuss more?)

### DIFF
--- a/INSTALL_ISSUES.md
+++ b/INSTALL_ISSUES.md
@@ -16,3 +16,24 @@ You need to install Blas and Lapack. See [LINUX\_INSTALL.md](LINUX_INSTALL.md) f
 ### `I get an error "ImportError: liblapack.so.3: Cannot open shared object file: No such file or directory"`
 
 You need to install Blas and Lapack. See [LINUX\_INSTALL.md](LINUX_INSTALL.md) for details.
+
+# Have conda?
+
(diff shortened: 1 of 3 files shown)


srikris:
I'm surprised that you ran into issues with conda. Could you elaborate more?

igiloh:
Hi @pavopax, thank you for pointing this out,
Have you tried using conda's built-in virtualenv, using `conda create`?

I've just tried these steps, and they seem to work on my system:
```
conda create -n venv python=2.7 anaconda
source activate venv
pip install --upgrade pip
pip install turicreate
python -c "import turicreate as tc ; print tc.__version__"
```

igiloh:
Thanks for the input @pavopax !
I wonder whether we should add a conda option for the main README.md. My only concern is that extending the readme with too many options might make it more confusing.

pavopax:
personally, I was confused by "system pip" which to me basically meant "don't use Conda"

but that may be just me

in terms of what to recommend, I am no expert and you/the team would know better, so I have no further recommendations! thanks for the follow-ups!

igiloh:
Thank you for all of your inputs @pavopax.
I'll close this PR now and start a new one with the relevant documentation updates.

Issue #52, opened 2017-12-10
Consistent min/max values for y-axis of diagrams in documentation


frogg:
It would be great if the diagrams had consistent min/max values. Especially the 3-axial gyroscope is really hard to evaluate/compare (walking vs. sitting) on this page: https://github.com/apple/turicreate/blob/master/userguide/activity_classifier/introduction.md

frogg:
Thanks for fixing, now this makes a lot more sense when just glancing at the diagrams :)

PR #56, opened 2017-12-11
Fix plot scale in AC UG

alonpal:
Set the scale of the "sitting" example in the activity classifier user guide to the same scale as the "walking" example.
Fixes https://github.com/apple/turicreate/issues/52.


igiloh:
Seems great. The difference between the two activities is much more apparent this way.

PR #62, opened 2017-12-11
Fix formatting errors detailed in #39

### LINKED ISSUE #39: String formatting errors
https://github.com/apple/turicreate/blob/7f07ce795833d0c56c72b3a1fb9339bed6d178d1/src/unity/python/turicreate/cython/cy_pylambda_workers.pyx#L193

```python
# current ^^^
for i in range(n):
    if lcd.input_keys[0][i].size() != n_keys:
        raise ValueError("Row %d does not have the correct number of rows (%d, should be %d)"
                         % (lcd.input_keys[0][i].size(), n))

# the problem is that there are only two arguments provided but the string is expecting three
>>> raise ValueError("Row %d does not have the correct number of rows (%d, should be %d)"
...                  % (22, 44))
  File "<stdin>", line 2, in <module>
TypeError: not enough arguments for format string

# if I am reading this correctly, shouldn't the first argument in the tuple be `i`?
raise ValueError("Row %d does not have the correct number of rows (%d, should be %d)"
                 % (i, lcd.input_keys[0][i].size(), n))
                    ^
                    ^
                    ^
```

---
https://github.com/apple/turicreate/blob/7f07ce795833d0c56c72b3a1fb9339bed6d178d1/src/nnvm/tvm/python/tvm/contrib/xcode.py#L189

```python
>>> rpc_root = 'TEST'
>>> raise RuntimeError("Cannot find tvmrpc.xcodeproj in %s," +
...                    (" please set env TVM_IOS_RPC_ROOT correctly" % rpc_root))
  File "<stdin>", line 2, in <module>
TypeError: not all arguments converted during string formatting

# this error can be fixed by removing two brackets and the `+` sign
>>> raise RuntimeError("Cannot find tvmrpc.xcodeproj in %s,"
...                    " please set env TVM_IOS_RPC_ROOT correctly" % rpc_root)
  File "<stdin>", line 2, in <module>
RuntimeError: Cannot find tvmrpc.xcodeproj in TEST, please set env TVM_IOS_RPC_ROOT correctly
```

---
https://github.com/apple/turicreate/blob/7f07ce795833d0c56c72b3a1fb9339bed6d178d1/scenario-tests/tests/single_test_driver.py#L11

```python
# current ^^^
def print_help():
    print("""\
{} [sub_directory] [Optional output xunit xml path prefix]

This test runs the scenario test located in the sub directory.
Every python file in the sub directory will be executed against pytest.

Optionally, a setup.py file may exist inside the subdirectory in which case
setup.py is effectively "sourced" before any of the tests are run. This allows
setup.py to modify environment variables which will be picked up by the
tests.

For instance, if the PATH variable is changed in setup.py a different python
environment may be used to run pytest
""" % sys.argv[0])

# the problem above is that the new style `{}` placeholder has been used instead of `%s`
>>> '{}' % 'TEST'
TypeError: not all arguments converted during string formatting
>>> '%s' % 'TEST'
'TEST'
>>> '{}'.format('TEST')
'TEST'

# since the project seems to prefer the old style VS new style,
# simply replacing `{}` with `%s` should fix this error
def print_help():
    print("""\
%s [sub_directory] [Optional output xunit xml path prefix]
^^
^^
^^
```

delirious-lettuce:
Fix #39

### DIFF
--- a/scenario-tests/tests/single_test_driver.py
+++ b/scenario-tests/tests/single_test_driver.py
@@ -8,7 +8,7 @@ from setup_reporter import try_execfile
 
 def print_help():
     print("""\
-{} [sub_directory] [Optional output xunit xml path prefix]
+%s [sub_directory] [Optional output xunit xml path prefix]
 
 This test runs the scenario test located in the sub directory.
 Every python file in the sub directory will be executed against pytest.
--- a/src/nnvm/tvm/python/tvm/contrib/xcode.py
+++ b/src/nnvm/tvm/python/tvm/contrib/xcode.py
@@ -186,8 +186,8 @@ def popen_test_rpc(host,
         rpc_root = os.path.join(curr_path, "../../../apps/ios_rpc")
     proj_path = os.path.abspath(os.path.join(rpc_root, "tvmrpc.xcodeproj"))
     if not os.path.exists(proj_path):
-        raise RuntimeError("Cannot find tvmrpc.xcodeproj in %s," +
-                           (" please set env TVM_IOS_RPC_ROOT correctly" % rpc_root))
+        raise RuntimeError("Cannot find tvmrpc.xcodeproj in %s,"
+                           " please set env TVM_IOS_RPC_ROOT correctly" % rpc_root)
     with open(os.path.join(rpc_root, "rpc_config.txt"), "w") as fo:
         fo.write("%s %d %s\n" % (host, port, key))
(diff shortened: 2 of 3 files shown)


znation:
@delirious-lettuce Thanks for fixing these. Looks great!

delirious-lettuce:
@znation ,

No problem, thanks for your time!

PR #65, opened 2017-12-12
Introduce seed in image classifier and stabilize unit test

gustavla:
The unit test `ImageClassifierTest::test_export_coreml_with_predict` failed with high probability. To address this, this PR increases the test tolerance and makes model creation optionally deterministic by exposing a seed parameter.

I changed the tolerance from 0.005 to the much higher 0.02. This is after running 25 trials and collecting some statistics. Only 3 times was it below 0.005 and the highest was around 0.015.

Note however that even though this tolerance seems awfully high, separate tests indicate that the error is many orders of magnitude smaller when the network is fed natural images (instead of randomly generated input as in this unit test).

### DIFF
--- a/src/unity/python/turicreate/toolkits/_supervised_learning.py
+++ b/src/unity/python/turicreate/toolkits/_supervised_learning.py
@@ -295,7 +295,8 @@ def print_validation_track_notification():
 
 
 def create(dataset, target, model_name, features=None,
-           validation_set='auto', verbose=True, distributed='auto', **kwargs):
+           validation_set='auto', distributed='auto',
+           seed=None, verbose=True, **kwargs):
     """
     Create a :class:`~turicreate.toolkits.SupervisedLearningModel`,
 
@@ -332,6 +333,10 @@ def create(dataset, target, model_name, features=None,
     distributed: env
         The distributed environment
 
+    seed : int, optional
+        Seed for random number generation. Set this value to ensure that the
+        same model is created every time.
+
     verbose : boolean
         whether print out messages during training
 
@@ -347,7 +352,7 @@ def create(dataset, target, model_name, features=None,
             if dataset.num_rows() >= 100:
                 if verbose:
                     print_validation_track_notification()
-                dataset, validation_set = dataset.random_split(.95)
+                dataset, validation_set = dataset.random_split(.95, seed=seed)
             else:
                 validation_set = None
         else:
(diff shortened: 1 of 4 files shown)


src/unity/python/turicreate/toolkits/_supervised_learning.py
```diff
@@ -295,7 +295,8 @@ def print_validation_track_notification():
 
 
 def create(dataset, target, model_name, features=None,
-           validation_set='auto', verbose=True, distributed='auto', **kwargs):
+           validation_set='auto', distributed='auto',
```
znation: I think this is potentially a breaking API change, if callers were using positional arguments for `verbose` (which is now `distributed`) or `distributed` (which is now `seed`). Can we keep the order the same as before and add `seed` at the end of the positional arguments?

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
```diff
@@ -36,6 +36,7 @@ def create(dataset, target, features=None,
     max_iterations = _DEFAULT_SOLVER_OPTIONS['max_iterations'],
     class_weights = None,
     validation_set = 'auto',
+    seed=None,
```
znation: Same comment here: can we preserve API compatibility?

src/unity/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -27,7 +27,8 @@
 from turicreate.toolkits._internal_utils import (_raise_error_if_not_sframe,
                                                  _numeric_param_check_range)
 
-def create(dataset, target, feature = None, model = 'resnet-50', max_iterations = 10, verbose = True):
+def create(dataset, target, feature = None, model = 'resnet-50',
```
znation: Same comment here.

gustavla: This is a great point. I agree and will push a change.

Done.

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
gustavla: Done.

src/unity/python/turicreate/toolkits/_supervised_learning.py
gustavla: Done.

PR #68, opened 2017-12-12
Minor fixes

kant:
For clearance's sake.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -34,8 +34,8 @@ be sure to install virtualenv using your system pip (conda pip is not supported)
 
     ./configure
 
-Running configure will create two sub-directories, release/ and
-debug/ . cd into src/unity under either of these directories and running make will build the
+Running configure will create two sub-directories, `release/` and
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -44,10 +44,10 @@ process. For instance:
     cd debug/src/unity
     make -j 4
 
-will perform up to 4 build tasks in parallel. When building in release/ mode,
+will perform up to 4 build tasks in parallel. When building in `release/` mode,
 Turi Create does require a large amount of memory to compile with the
 heaviest toolkit requiring 1GB of RAM. Where K is the amount of memory you
-have on your machine in GB, we recommend not exceeding make -j K
+have on your machine in GB, we recommend not exceeding make `-j` K
```
TobyRoseman: "make" and "K" are part of the command. I think it should be: `make -j K`

TobyRoseman:
The changes here have gone in as part #203.

BUILD.md
kant: that's ok

PR #73, opened 2017-12-12
assertEquals -> assertEqual (fix deprecation)

gustavla:
`assertEquals` (with a trailing `s`) is deprecated in documentation already in Python 2.7, and emits run-time warnings since Python 3.2. This is a huge source of clutter in the unit tests on 3.2+.

This also renames a custom assert function in the JSON tests (`_assertEquals` -> `_assertEqual`) for consistency.

Passes all tests.


srikris:
@gustavla Can we merge this?

PR #75, opened 2017-12-12
Added Setup And Install Instructions For Windows

JCA122204:
I added them because they are slightly different from the original instructions

I did leave the original instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,14 +87,22 @@ The method for installing *Turi Create* follows the
 To create a Python virtual environment called `venv` follow these steps:
 
 ```shell
-# Create a Python virtual environment
+# Create a Python virtual environment on Windows
+cd C:\Users\{Username}\Desktop
+virtualenv venv
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,14 +87,22 @@ The method for installing *Turi Create* follows the
 To create a Python virtual environment called `venv` follow these steps:
 
 ```shell
-# Create a Python virtual environment
+# Create a Python virtual environment on Windows
```
znation: I don't think this style of file path will work in WSL (which is our currently supported Windows environment). WSL is a Linux subsystem on top of Windows, and uses Linux-style file paths including `~` for home directory, so the same instructions as for Linux and macOS should work. Have you tried the existing instructions there?

znation:
@JCA122204 Following up on this -- were you able to test the existing instructions, and the modifications here, on WSL? I don't expect our installation instructions to work outside of WSL in any case. Thanks!

znation:
Closing due to inactivity.

PR #78, opened 2017-12-13
Doctest quote characters

delirious-lettuce:
There were three missing double quotes which this PR fixes.

I also replaced two single quotes with double-quotes for readability.

> [When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.](http://pep8.org/#string-quotes)


delirious-lettuce:
@znation ,

No problem!

PR #79, opened 2017-12-13
Raise `RuntimeError` instead of bare strings

### LINKED ISSUE #38: Raise specific exceptions?
https://github.com/apple/turicreate/blob/7f07ce795833d0c56c72b3a1fb9339bed6d178d1/src/unity/python/turicreate/util/config.py#L55

```python
>>> raise "Please set the TEMP environment variable"
TypeError: exceptions must be old-style classes or derived from BaseException, not str
```

https://github.com/apple/turicreate/blob/7f07ce795833d0c56c72b3a1fb9339bed6d178d1/src/unity/python/turicreate/test/test_nearest_neighbors.py#L1591

```python
>>> raise "Unknown distance"
TypeError: exceptions must be old-style classes or derived from BaseException, not str
```

delirious-lettuce:
Fix #38 

I found two more instances of raising strings (other than the two fixed by this PR) but they are both under `deps` so I was hesitant to touch them.

### DIFF
--- a/src/unity/python/turicreate/test/test_nearest_neighbors.py
+++ b/src/unity/python/turicreate/test/test_nearest_neighbors.py
@@ -1588,7 +1588,7 @@ class NearestNeighborsSparseQueryTest(unittest.TestCase):
             elif distance == 'manhattan':
                 ans = manhattan(query, ref)
             else:
-                raise "Unknown distance"
+                raise RuntimeError("Unknown distance")
             self.assertAlmostEqual(score, ans)
 
     def test_query_distances(self):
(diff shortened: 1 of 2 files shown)


delirious-lettuce:
https://github.com/apple/turicreate/blob/fe1c1315cc7730bf05452f13bb1fff77e6cfcaf9/deps/src/boost_1_65_1/tools/build/src/tools/msvc.py#L348

https://github.com/apple/turicreate/blob/fe1c1315cc7730bf05452f13bb1fff77e6cfcaf9/deps/src/boost_1_65_1/tools/build/test/BoostBuild.py#L593

PR #82, opened 2017-12-13
Update README.md - user may use conda venv

igiloh:
Following discussion in https://github.com/apple/turicreate/pull/54, I updated that readme that user may use any `virtualenv` and `pip` installations, including `conda` ones. 
This is also to further clarify that the user doesn't have to user the "system" pip per-se.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,6 @@ For detailed instructions for different varieties of Linux see [LINUX\_INSTALL.m
 For common installation issues see [INSTALL\_ISSUES.md](INSTALL_ISSUES.md).
 
 We recommend using virtualenv to use, install, or build Turi Create. 
-Be sure to install virtualenv using your system pip.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,28 +76,31 @@ For detailed instructions for different varieties of Linux see [LINUX\_INSTALL.m
...
 cd ~
 virtualenv venv
-```
 
-To activate your new virtual environment and install `Turi Create` in this environment, follow these steps:
-```shell
 # Activate your virtual environment
 source ~/venv/bin/activate
+```
+Alternatively, if you are using Anaconda, you may use its virtual environment:
```
znation: I'd link to [Anaconda](https://www.anaconda.com/what-is-anaconda/) here.

igiloh: Done

PR #86, opened 2017-12-13
Fix typo: `deamon` -> `daemon`

delirious-lettuce:
I just came across this by accident but in four different places `daemon` was being spelled as `deamon`.

* https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Process.daemon

```python

Python 2.7.12 (default, Nov 20 2017, 18:23:56) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from multiprocessing import Process
>>> p = Process()
>>> p.deamon = True
>>> p.daemon
False

Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from multiprocessing import Process
>>> p = Process()
>>> p.deamon = True
>>> p.daemon
False
```

### DIFF
--- a/src/nnvm/tvm/python/tvm/contrib/rpc.py
+++ b/src/nnvm/tvm/python/tvm/contrib/rpc.py
@@ -89,7 +89,7 @@ def _listen_loop(sock):
             conn.sendall(struct.pack("@i", RPC_MAGIC))
         logging.info("Connection from %s", addr)
         process = multiprocessing.Process(target=_serve_loop, args=(conn, addr))
-        process.deamon = True
+        process.daemon = True
         process.start()
         # close from our side.
         conn.close()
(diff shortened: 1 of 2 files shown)


znation:
Hmm, I'm not sure where this part of the source code is used. I think this code is currently unused and should be deleted instead of modified. I will look into it.

delirious-lettuce:
@znation ,

Ok, no problem.

srikris:
@delirious-lettuce Your fix revealed  that we probably should delete this code entirely. Thanks! We will take care of that in #87

delirious-lettuce:
@srikris ,

Ok, sounds good! Thanks!

Issue #57, opened 2017-12-11
unable to read from/write to a non public S3 bucket


davidswaven:
It seems not possible to use method turicreate.aws.set_credentials() described in documentation https://apple.github.io/turicreate/docs/api/generated/turicreate.SFrame.html

There is no aws object within turicreate.
It seems also not possible to use environment variables, either when set in the shell environment, or before calling the load_sframe method.

This is the code i'm using to get an SFrame in S3. 
It works well with the sframe library, but fails with turicreate

*def getSFrame(s3SFramePath):
    import os
    import turicreate as tc
    global S3_ACCESS_KEY, S3_SECRET_KEY
    os.environ['AWS_ACCESS_KEY_ID'] = S3_ACCESS_KEY
    os.environ['AWS_SECRET_ACCESS_KEY'] = S3_SECRET_KEY
    return tc.load_sframe(s3SFramePath)*

srikris:
Thanks for reporting this. We will look into this issue and update the instructions on using S3.

znation:
Adding the `bug` tag since it looks like there is a real bug here as well. Even when credentials are set via env variables, reading from buckets doesn't always work.

znation:
The fix to support s3 regions other than the default is in #90. Leaving this issue open for now, to update the documentation as well.

afranklin:
Update: Bug resolved; waiting on docs.

znation:
Docs updated with #183.

davidswaven:
hi, any idea when this will be released ? thanks

znation:
@davidswaven Stay tuned - I can't give an exact date but we're putting together a 4.1 release including all current fixes.

srikris:
We are hoping for this week.

afranklin:
Turi Create 4.1 is now available.

davidswaven:
I'm able to make it work from my laptop. Thanks.

Unfortunately, from an EC2 linux server that has an IAM role, I have no way to access my SFrame stored in S3 bucket. 
If I don't provide any environment variable, I get the following (and expected) error:
`KeyError('No access key found. Please set the environment variable AWS_ACCESS_KEY_ID.',)`
If i provide an access & secure key that has access to the bucket, I get the following (but unexpected) error:
IOError: s3://{my-bucket}/{my_sframe_folder_path} not found.: iostream error

The same code was working in lib sframe 2.1

znation:
[...]

PR #90, opened 2017-12-14
Fix S3 Region

ylow:
S3 Regions are not set in the S3 client. We need to set it to support
anything outside of us-east-1.

### DIFF
--- a/src/fileio/get_s3_endpoint.cpp
+++ b/src/fileio/get_s3_endpoint.cpp
@@ -5,6 +5,7 @@
  */
 #include <vector>
 #include <string>
+#include <map>
 #include <boost/algorithm/string/predicate.hpp>
 #include <fileio/get_s3_endpoint.hpp>
 #include <fileio/get_s3_endpoint.hpp>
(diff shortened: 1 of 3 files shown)


src/fileio/get_s3_endpoint.cpp
```diff
@@ -15,15 +16,32 @@ namespace fileio {
...
+  "s3-eu-central-1.amazonaws.com",
   "s3-ap-southeast-1.amazonaws.com",
   "s3-ap-southeast-2.amazonaws.com",
   "s3-ap-northeast-1.amazonaws.com",
+  "s3-ap-northeast-2.amazonaws.com",
   "s3-sa-east-1.amazonaws.com",
+  "s3-ap-south-1.amazonaws.com",
 };
 
+const std::map<std::string, std::string> AWS_S3_ENDPOINT_TO_REGION {
```
TobyRoseman: It looks like there are other S3 regions. Do we want to add more here?

http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

ylow:
Hmm... The oddity is that the AWS C++ SDK only has a smaller set of regions.

https://github.com/aws/aws-sdk-cpp/blob/3e979d7bc272f77ac179b4849f7386027201c1a2/aws-cpp-sdk-ec2/include/aws/ec2/model/Region.h

znation:
@ylow @TobyRoseman Do we have a decision here - should we take the larger list, from AWS S3 documentation, or the smaller list, from AWS C++ SDK documentation? Do we know whether the larger list includes regions that do actually exist in S3 that we're missing, and whether we need to add them in order for buckets in those regions to work?

ylow:
Lets add the regions in the AWS C++ SDK list first. We can always add more later.

znation:
Fixes #57.

Issue #58, opened 2017-12-11
Export text classifier to Core ML


emannuelOC:
I would like to export the Sentence Classifier model to Core ML. Right now I'm getting the `'SentenceClassifier' object has no attribute 'export_coreml'` error.
Is there some way to export it?

srikris:
Unfortunately, there isn't at the moment but this one is easy to add. We will do it for the next release and add a user guide step for exporting this to Core ML. Thanks for your patience.

znation:
Fixed with #99.

PR #99, opened 2017-12-14
Sentence classifier export to coreML

alonpal:
Added Core ML support for the sentence classifier, including unittests and user guide.
All tests pass.

Requested in https://github.com/apple/turicreate/issues/58.


src/unity/python/turicreate/test/test_sentence_classifier.py
```diff
@@ -85,6 +88,25 @@ def test_evaluate(self):
         """
         self.model.evaluate(self.docs)
 
+    def test_export_coreml(self):
+        filename = tempfile.mkstemp('bingo.mlmodel')[1]
+        self.model.export_coreml(filename)
+
+    @unittest.skipIf(sys.platform != 'darwin', 'Only supported on Mac')
```
znation: This isn't quite the right check (since predictions are only supported on macOS 10.13+).

You may want to put your Core ML export test in https://github.com/apple/turicreate/blob/master/src/unity/python/turicreate/test/test_coreml_export.py to share the common code there for evaluating turicreate predictions against Core ML. If you'd prefer to keep it here, the skipIf needs to check macOS version.

znation: I see there is an `if` statement within for the macOS version, but I think on prior macOS versions we'd want this test to report as `skipped` rather than `passed`.

userguide/text_classifier/introduction.md
```diff
@@ -71,3 +71,37 @@ classifier model using:
...
+##### Deploying to Core ML
+
+You can use the trained sentence classifier in Xcode by exporting it to Core ML. Exporting is done by:
+
+```python
+model.export_coreml('MySentenceClassifier.mlmodel')
+```
+
+Dragging the saved model into Xcode and inspecting it looks like the following:
+<img src="images/sentence_calssifier_model.png"></img>
```
znation: Typo in filename: should be `sentence_classifier_model.png`.

userguide/text_classifier/introduction.md
```diff
@@ -71,3 +71,37 @@ classifier model using:
...
+
+To use the model in Swift we can use the code below. Note that the model expects a bag-of-words representation of the input text.
+
+```swift
+let bagOfWords = bow(text: text)
```
znation: I think we should use [NSLinguisticTagger](https://developer.apple.com/documentation/foundation/nslinguistictagger/tokenizing_natural_language_text) in the example for tokenizing strings into bag-of-words.

src/unity/python/turicreate/test/test_sentence_classifier.py
```diff
@@ -85,6 +88,23 @@ def test_evaluate(self):
         """
         self.model.evaluate(self.docs)
 
+    def test_export_coreml(self):
+        filename = tempfile.mkstemp('bingo.mlmodel')[1]
+        self.model.export_coreml(filename)
+
+    @unittest.skipIf(_mac_ver() < (10, 13), 'Only supported on Mac')
```
znation: Please change this message to `'Only supported on macOS 10.13+'`.

znation:
I messed up the merge; these changes are now in master, but GitHub doesn't show it as such on this PR because the commit hashes changed (I rebased), so this PR now shows as 0 files changed. Closing as merged.

PR #110, opened 2017-12-16
WIP: Load models in Python 3

gustavla:
This enables loading models in Python 3 (#2). Saving models already work.

The issue is that when models are loaded, all strings in the state are `bytes` in Python 3. However, the code expects all these strings to be `str` (unicode). This PR simply converts all the `bytes` to `str`. However, this would break if there were genuine bytestrings that do represent ascii text. It doesn't seem to be case anywhere (since I pass all relevant tests) and any future such bytestrings should probably be stored as `array` objects anyway. If a non-ascii string is discovered, with high probability `decode` would raise an exception.

However, this is still not a very satisfactory solution, so I have marked it work-in-progress to open it up for discussion. This PR serves as a starting point to demonstrate a pragmatic solution. If the code that actually bridges C++ and Python would natively produce unicode objects in Python 3, that would arguably be a more elegant solution.

### DIFF
--- a/src/unity/python/turicreate/toolkits/_model.py
+++ b/src/unity/python/turicreate/toolkits/_model.py
@@ -26,6 +26,35 @@ import six as _six
 
 MODEL_NAME_MAP = {}
 
+if _six.PY3:
+    def _state_str_conversion(data):
+        """
+        Conversion of data from type bytes to str (unicode). If data is a
+        container (dict, list, tuple) it applies itself recursively.
+        """
+        if isinstance(data, dict):
+            new_dict = {}
+            for k, v in data.items():
+                if isinstance(k, bytes):
+                    new_k = k.decode('utf-8')
+                else:
+                    new_k = k
+                new_dict[new_k] = _state_str_conversion(v)
+            return new_dict
+        elif isinstance(data, bytes):
+            return data.decode('utf-8')
+        elif isinstance(data, (list, tuple)):
+            new_list = []
+            for l in data:
+                new_list.append(_state_str_conversion(l))
+            return data.__class__(new_list)
+        else:
+            return data
+else:
+    def _state_str_conversion(data):
+        return data
+
+
 def load_model(location):
     """
     Load any Turi Create model that was previously saved.
@@ -69,6 +98,7 @@ def load_model(location):
 
     _internal_url = _make_internal_url(location)
     saved_state = glconnect.get_unity().load_model(_internal_url)
+    saved_state = _state_str_conversion(saved_state)
     if saved_state['archive_version'] == 1:
         cls = MODEL_NAME_MAP[saved_state['model_name']]
         if 'model' in saved_state:


znation:
@gustavla Hmm. I think we have lots of options here to resolve this issue (correct me if any of these won't work):

1. Assume strings contain `utf-8` encoded text when necessary to make an assumption for Python 3, and error out otherwise. Python 2 would still continue to work with arbitrary byte strings, but the models and data produced may not be loadable in Python 3. (I think that's what's in this PR?)
    * Pros: non-intrusive code change, no UX change to existing scenarios.
    * Cons: Turi Create in Python 3 would not quite be compatible with Python 2 in terms of behavior/data/models.

2. Require strings to contain `utf-8` encoded text, and validate this in both Python 2 and Python 3 to prevent incompatibilities between the two. Note: this would potentially break loading existing SFrames or models from 4.0.
    * Pros: full compatibility between Python 2 and Python 3.
    * Cons: breaking compatibility with 4.0, no support for arbitrary byte strings or other encodings.

3. Keep the same semantics in Python 3 as in Python 2, which implies we would only ever use the `bytes` type in both. (Currently, we use the `str` type in both, but this has different semantics in Python 2 and Python 3; as I understand `bytes` has the same semantics in both Python 2 and 3 as our current semantics, and is what `str` maps to in Python 2).
    * Pros: full compatibility between Python 2 and Python 3, full compatibility with 4.0.
    * Cons: not very Pythonic (in Python 3 especially). Users expect the `str`/`unicode` type to work in Python 3 (and probably the `unicode` type in 2 as well).

4. Add a new `text` type in SFrame that implies utf-8 encoding and uses the `unicode` Python type; this is basically option 2 + option 3. We could potentially rename the current `str` type to `bytes` as well to remove confusion.
    * Pros: full compatibility between Python 2 and 3, full compatibility with 4.0.
    * Cons: two string types. This could cause user confusion (in the same way that Python having two string types does 😄 ). However, I think the biggest source of confusion is the word `str` and could be resolved by just calling the types `bytes` and `unicode`/`text`/etc.

gustavla:
Thanks @znation, that is really useful overview! I think we need to distinguish model save/load and SFrame save/load though, at least from an implementation point of view. Right now, SFrame in Python 3 automatically decodes strings as UTF-8 (likewise in Python 2 when displaying strings). This does not happen automatically for the model states for some reason, which is exactly what this PR does manually. If the model states behaved like SFrames, this PR would not be needed at all.

As for the SFrame, its behavior is essentially Option 1 right now. This mostly works except for two main issues (and a separate bug that I just opened an issue for in #157):
- In Python 2 only bytes indexing is supported and in Python 3 only unicode indexing is supported
- Python 2 and 3 both support saving byte strings that are not valid UTF-8. In Python 2 this is mostly fine (printing fails with a missing symbol), while in Python 3 this is a fatal error.

That we are allowing non-ascii/non-utf8 bytes to go into an SFrame is not great, since these should probably more appropriately use the `array` type in the first place. There is an option between 1 and 2 where we keep the Python 2 behavior but introduce deprecation warnings (that we could wait to enforce strictly until a major version bump). Apart from the breaking change, the fact that Python 2 and 3 only support indexing with their respective `str` is a bit strict, but it is consistent with Python 3 behavior in general:

```python
>>> {b'foo': 100}[u'foo']
```
This works in Python 2, and it doesn't matter what combination of  `b` and `u` you have. The same exact discussion applies to regular values and string equality checks (`b'foo' == u'foo'`). However, if your key is a unicode with non-ascii characters, then you can't index it with a byte string (even if it's utf-8 encoded). In Python 3, indexing across bytes and unicode is simply forbidden altogether even if you only have ascii in your strings. What makes Turi Create's behavior a bit weird is that whatever you put into the SFrame, you always end up with UTF-8-encoded bytes in Python 2, and unicode in Python 3. This can lead to unintuitive indexing or quality checks where the value you just put in does not seem to match what you get out:

Currently in Python 2:
```python
>>> key = u'\u00aa'
>>> tc.SArray([{key: 100}])[0][key]
KeyError
>>> tc.SArray([{key: 100}])[0][key.encode('utf-8')]
100
>>> tc.SArray([key])[0] == key
__main__:1: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
False
>>> tc.SArray([key])[0] == key.encode('utf-8')
True
```
Currently in Python 3:
```python
>>> key = b'foo'
>>> tc.SArray([{key: 100}])[0][key]
KeyError
>>> tc.SArray([{key: 100}])[0][key.decode('utf-8')]
100
>>> tc.SArray([key])[0] == key
False
>>> tc.SArray([key])[0] == key.decode('utf-8')
True
```
I purposely made this example using an SArray, since that means the `KeyError` or the equality check is happening from a native Python data type and not directly from an SFrame/SArray. Of course, for things like SFrame column indexing, we could theoretically support dual bytes/unicode indexing and do the encode/decode implicitly. However, we could not extend this kind of support to the example above or two equality checks. That is, under Option 1 or 2. If we introduce a separate string type, as Option 4 proposes, then we could avoid this type of inconsistent behavior.

I'm not sure Option 4 as you describe it is even possible. The reason is because we already support unicode input in Python 2 and it has the implicit UTF-8 encoding behavior as outlined above. So, if we keep Python 2 behavior exactly (assuming compatibility with 4.0), the new text type would only ever be used in Python 3 when a unicode object is used. This would be really strange in many ways, especially since Python 2 would then be able to open and load these as unicode, but if the information is re-saved it would again go in as a UTF-8-encoded bytes. It would also be strange to have two unicode-supporting string types.

For Option 4 to really make sense it would have to be more like Python 3, where the Turi Create bytes/unicode would share semantics with Python's bytes/unicode types (and UTF-8 is never implied. although it can be an implementation detail for the unicode type). However, that would be a pretty major breaking change to the current behavior. Probably much more disruptive than the breaking change in Option 2.

I don't think Option 3 offers an acceptable UX in Python 3.

znation:
Thanks for the analysis @gustavla! I like your suggestion of Option 2 eventually, but with Option 1 and deprecation warnings in the meantime.

ylow:
I might suggest an Option 5:

- Use str in Python 3. Assume utf-8 decode everywhere. However, on failure to decode, fallback to cp1252. (optionally we could raise a one-time warning when this happens).

cp1252 maps 8-bit ascii to first 256 unicode code points hence allowing str to behave exactly as expected even in the event SFrames contain binary data. This should allow us to maintain broadest user code compatibility between Python 2 and Python 3 (except for certain edge cases where I actually have binary data, which coincidentally decodes as utf-8).

Ideally SFrames should contain only utf-8 encoded strings and we can enforce that later in several places (in all ingestion points, or during SFrame encoding).

gustavla:
I think @ylow's Option 5 is a winner! I also agree that in the future we should consider strictly enforcing UTF-8 when creating an SFrame.

gustavla:
Leaving the broader discussion and coming back to what this PR tries to do. I have put up an alternative PR (#161) that does this conversion a bit earlier, using the same mechanism that SFrame uses. This will ensure more consistent behavior. It is not quite as clean as it could be due to some legacy stuff, but I still prefer it over this solution. I'm closing this PR for now in favor of #161.

PR #113, opened 2017-12-17
Improving the Activity Classifier's unit testing

igiloh:
The current unit test for the AC's sequence iterator only covered the `prep_data` step, and only a few of the possible edge cases.
I've re-designed the unit test to:
1. Cover more edge cases in the data preparation step
2. Test the outputs of the iterator's `next()` method - there are several non trivial conversions in the process.

### DIFF
--- a/src/unity/python/turicreate/test/test_sframe_sequence_iterator.py
+++ b/src/unity/python/turicreate/test/test_sframe_sequence_iterator.py
@@ -8,6 +8,7 @@ from __future__ import division as _
 from __future__ import absolute_import as _
 import unittest
 import turicreate as tc
+import numpy as np
 from turicreate.toolkits.activity_classifier import _sframe_sequence_iterator as sframe_sequence_iterator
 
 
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/test/test_sframe_sequence_iterator.py
```diff
@@ -59,3 +82,131 @@ def test_ceil_dev(self):
...
+        # The iterator length should be in the same length as the expected_batches list.
+        # We expect a "StopIteration" exception to be raised on the next call to next()
+        stop_iteration_raised = False
+        try:
+            seq_iter.next()
+        except StopIteration:
+            stop_iteration_raised = True
+        self.assertTrue(stop_iteration_raised , "Error - the iterator is longer than expected")
+
+    @classmethod
```
alonpal: `classmethod` should receive `cls` instead of `self`. 
On the other hand, `_prepare_expected_chunked_dataset`, `_prepare_expected_batches`, and `setUp` should not be `classmethod`.  The setup method should be `setUp` which is called before every test, or `setUpClass` which is a `classmethod` invoked once for the entire test class.

src/unity/python/turicreate/test/test_sframe_sequence_iterator.py
```diff
@@ -59,3 +82,131 @@ def test_ceil_dev(self):
...
+        self.assertTrue(seq_iter.batch_size == batch_size)
+        self.assertTrue(seq_iter.num_batches == expected_num_batches)
+
+        for batch_num ,expected_batch in enumerate(expected_batches):
+            observed_batch = seq_iter.next()
+
+            np.testing.assert_array_equal(expected_batch['features'] , observed_batch.data[0].asnumpy(),
+                                          "Error - features in batch %d" % batch_num)
+
+            if (use_taget):
```
alonpal: Skip the parentheses (also typo): `if use_target:`

src/unity/python/turicreate/test/test_sframe_sequence_iterator.py
```diff
@@ -59,3 +82,131 @@ def test_ceil_dev(self):
...
+                np.testing.assert_array_equal(expected_batch['target'] , observed_batch.label[0].asnumpy(),
+                                              "Error - target in batch %d" % batch_num)
+
+                np.testing.assert_array_equal(expected_batch['weights'] , observed_batch.label[1].asnumpy(),
+                                              "Error - weights in batch %d" % batch_num)
+            else:
+                self.assertTrue(observed_batch.label is None)
+
+            if (use_pad and "padding" in expected_batch):
```
alonpal: Skip the parentheses

src/unity/python/turicreate/test/test_sframe_sequence_iterator.py
igiloh: Fixed

PR #122, opened 2017-12-19
Consolidate source drops

### LINKED ISSUE #87: Consolidate external source drops
We have source drops of external projects in several places:

```
src/external
src/nnvm
subtree/xgboost
deps/src
```

We should consolidate these into a single external directory, and make it clear that these are external source drops and generally shouldn't be modified except to take new source drops of these projects.

znation:
Fixes #87.


igiloh:
Isn't the xgboost subtree an actual git subtree\submodule?
I think (not really sure here) that it's customary to keep these as we did - in a directory called `<repo_root/subtree>`, isn't it?

znation:
@igiloh It used to be a subtree (prior to the initial commit in this repo). It's now a plain old source drop, mislabeled as subtree.

src/external/xgboost/CMakeLists.txt
```diff
@@ -4,11 +4,11 @@ make_library(xgboost
   SOURCES
     src/gbm/gbm.cpp
     src/tree/updater.cpp
+    src/io/dmlc_simple.cpp
     src/io/io.cpp
     subtree/rabit/src/engine_empty.cc
   REQUIRES
     libjson
-    dmlc
```
TobyRoseman: If xgboost doesn't need dmlc, do we still need it at all?

znation: I can't find where it's building the `dmlc` library - nothing else in the CMakeLists seems to produce it. But if you can find it let's remove it.

Issue #17, opened 2017-12-09
CUDA 9 and cuDNN 7 support


linthieda:
Since CUDA 9 and cuDNN 7 has already released, and there are multiple API changes for these low-level libraries. Something like `cudnnGetConvolutionForwardAlgorithm_v7()` may have better performance than older version. We should adapt to these as well.

gustavla:
All CUDA code runs through MXNet, which is a deep learning framework. If MXNet supports CUDA 9, then so can we. The problem is that they have not put up a wheel for mxnet-cu90==0.11.0, which is the only version we currently support (there are breaking changes in 0.12).

We hope to extend our MXNet version support soon, in which case we should also get CUDA 9 support for free. I'm not sure what cuDNN they currently support.

srikris:
It looks like [mxnet 0.12](https://cwiki.apache.org/confluence/display/MXNET/MXNet+0.12.0+Release+Notes) supports CUDA 7 and 9 so we should try and depend on that version in the next release (of course assuming all our tests pass and the model quality stays the same).

Thanks for the feedback @linthieda!

helloniklas:
@gustavla @srikris So I managed to compile from source and run MXNET 1.0.1 on GPU on my MacBook Pro with a Sonnett eGPU and NVIDIA 1080 card.

Since NVIDIA's macOS 10.13.2 driver is only supporting CUDA 9 and not CUDA 8, CUDA 9 is essential for this. I'm also running cuDNN 7 with this setup.

With Turi being an Apple thing and supporting CoreML export for Xcode presumably a lot of users will be on Mac. Would be great if next version will use MXNET 1.0.0 as you will then get GPU support for Mac with NVIDIA cards.

gustavla:
@helloniklas We are close to both MXNet 0.12 and 1.0 support. However, the 1.0.0 version causes the object detector to crash (this is fixed in their 1.0.1 nightlies), so we will have to wait until 1.0.1 has been released before we can extend official support to 1.*. Hopefully they release it before our next version so that we can get it in. Thanks for your patience!

gustavla:
All changes to support all newer versions of MXNet (which include CUDA 9 from 0.12.1) have been merged in (#129, #169). Unless you do a from source installation, this will all be available in the next release.

PR #129, opened 2017-12-20
Improve mxnet support for activity classifier save/load

gustavla:
This PR updates the activity classifier (AC) save/load to work with more versions of mxnet. There are still some issues with export to Core ML for newer versions of mxnet, so this PR alone does not expand support yet (#17).

The issue with the AC is that it saves and loads the network graph using mxnet json files. MXNet is pretty good at backward compatibility, but not forward compatibility. If we support more than one version of mxnet at a time, this creates a problem for us where we can't even be same-version compatible:

1. User 1 (turicreate==4.0.0 and mxnet==0.12.1) saves a model.
2. User 2 (turicreate==4.0.0 and mxnet==0.11.0) loads the model.

This won't work due to forward incompatibility in mxnet. The solution is to avoid saving and loading the graph and simply building it up and copying over the weights. This has much better forward compatibility.

# Support (old and new)

Let's look at current support and support after this PR. I'll show my testing as matrices where:

* rows: model saved in
* columns: model loaded in

Classifier (IC)/Similarity (IS)/Detector (OD): (all work on all combinations of save/load, although currently you get warnings if you do not use MXNet 0.11.0. This PR will eliminate those warnings. This broad support is thanks to the same changes that I'm making to the AC in this PR)

TC/MXNet|4/0.11.0|4/0.12.1|4/1post1
---|:-:|:-:|:-:
v4/0.11.0|:white_check_mark:|:white_check_mark:|:white_check_mark:
v4/0.12.1|:white_check_mark:|:white_check_mark:|:white_check_mark:
v4/1post1|:white_check_mark:|:white_check_mark:|:white_check_mark:

Activity Classifier (AC): (since this PR changes the AC saver/loader, I tested a bunch of combinations)

TC/MXNet|4/0.11.0|4/0.12.1|4/1post1|PR/0.11.0|PR/0.12.1|PR/1post1
---|:-:|:-:|:-:|:-:|:-:|:-:
v4/0.11.0|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:
v4/0.12.1|:no_entry_sign:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:
v4/1post1|:no_entry_sign:|:no_entry_sign:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:
PR/0.11.0|:warning:|:warning:|:warning:|:white_check_mark:|:white_check_mark:|:white_check_mark:
PR/0.12.1|:warning:|:warning:|:warning:|:white_check_mark:|:white_check_mark:|:white_check_mark:
PR/1post1|:warning:|:warning:|:warning:|:white_check_mark:|:white_check_mark:|:white_check_mark:

:white_check_mark: Works
:no_entry_sign: Does not work (ugly error)
:warning: Does not work (fails gracefully)

"PR" refers to the Turi Create model as defined by this PR's commit. "1post1" is short for 1.0.0.post1 (1.0.0 segfaults the object detector, and this seems to have been resolved in the post1 version).

**Top-left**: This is the status quo
**Right half**: Backwards and same-version compatible
**Bottom-left**: Forward incompatible (with respect to TC version). See note at the bottom.

The "graceful" failure in 4.0 actually says "Corrupted model. Cannot load a model with this version." for OD/AC, and for IC/IS it does not even check the version! This PR in an isolated commit also improves this and makes the message friendlier and tells the user to upgrade Turi Create. Unfortunately, whenever we upgrade the file format for IC/IS, it will fail very ungracefully on 4.0.

## Why not be forward compatible?

We could make newer models load in 4.0 as well. However, that is a commitment to write the backward migration to mxnet for all its future versions. For instance, in mxnet 5.0, we would still need to write json graphs that look like 0.11.0. It's better to break this compatibility now, since we would probably break it eventually. At least going forward, we have much better chances of being forward compatible (in mxnet version) for the AC, just like it turned out we are for IC/IS/OD.

### DIFF
--- a/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -206,12 +206,15 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
     # Train the model
     log = _fit_model(loss_model, data_iter, valid_iter,
                      max_iterations, num_gpus, verbose)
+
+    # Set up prediction model
     pred_model.bind(data_shapes=data_iter.provide_data, label_shapes=None,
-                    for_training=False, shared_module=loss_model)
+                    for_training=False)
+    arg_params, aux_params = loss_model.get_params()
+    pred_model.init_params(arg_params=arg_params, aux_params=aux_params)
 
     # Save the model
     state = {
-        '_loss_model': loss_model,
         '_pred_model': pred_model,
         'verbose': verbose,
         'training_time': _time.time() - start_time,
@@ -258,14 +261,13 @@ class ActivityClassifier(_CustomModel):
     This model should not be constructed directly.
     """
 
-    _PYTHON_ACTIVITY_CLASSIFIER_VERSION = 1
+    _PYTHON_ACTIVITY_CLASSIFIER_VERSION = 2
 
     def __init__(self, state):
         self.__proxy__ = _PythonProxy(state)
 
     def _get_native_state(self):
         state = self.__proxy__.get_state()
-        state['_loss_model'] = _mxnet_utils.get_mxnet_state(state['_loss_model'])
         state['_pred_model'] = _mxnet_utils.get_mxnet_state(state['_pred_model'])
         return state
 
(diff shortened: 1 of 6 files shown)


src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -281,12 +282,27 @@ def _load_version(cls, state, version):
             ('weights', (state['_recalibrated_batch_size'], state['_predictions_in_chunk'], 1))
         ]
 
+        from ._model_architecture import _define_model
+        import mxnet as _mx
         context = _mxnet_utils.get_mxnet_context(max_devices=state['num_sessions'])
-        state['_loss_model'] = _mxnet_utils.load_mxnet_model_from_state(
-            state['_loss_model'], data, labels, None, context)
```
igiloh: Doesn't this mean we're no longer **backward** compatible?
If someone saved a model using version 1, the weights are now saved only in the loss model, and therefore when   we later in lines 301-303 load them from `state['_pred_model']` they would be all zeros, won't they?

I can understand not being forward compatible (saved in new version should not load in old version). But backwards compatibility is important.
We could check for `if cls._PYTHON_ACTIVITY_CLASSIFIER_VERSION == 1 or '_loss_model' in state` then extract the params from loss model. Else extract from pred model.
Right?

igiloh:
Please see my comment about backward compatibility. Otherwise LGTM.
I would still wait for @alonpal's review as well, as he's more familiar with the MXnet arch in the AC.

src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
gustavla: Thanks for the review! In the current v4, there is no weight sharing when it gets saved to file. All weights are saved twice. Looking at the actualy saved files, a model saved with v4 takes 4 MB while a model saved with v4+ takes 2 MB. Therefore, there is no problem for v4+ to simply ignore half of those weights and load the model entirely from the `pred_model`.

Also, regarding backward compatibility. Every cell in the 6x6 matrix I showed in the original post is the result of an **actual test** and not just my hopes (I wanted to be very thorough!), so I have tested and verified full backward compatibility.

gustavla:
Sounds good, I will wait for @alonpal's review. Thanks!

PR #130, opened 2017-12-20
Fix compilation issue with Aws::String

gustavla:


src/fileio/s3_api.cpp
```diff
@@ -414,7 +414,7 @@ std::string delete_object_impl(s3url parsed_url,
     clientConfiguration.proxyHost = proxy.c_str();
     clientConfiguration.requestTimeoutMs = 5 * 60000;
     clientConfiguration.connectTimeoutMs = 20000;
-    clientConfiguration.region = fileio::get_region_name_from_endpoint(clientConfiguration.endpointOverride.c_str());
+    clientConfiguration.region = fileio::get_region_name_from_endpoint(clientConfiguration.endpointOverride.c_str()).c_str();
```
ylow: In all cases, can you assign to a std::string, then take c_str() of that?
i.e.
```
    auto region_name = fileio::get_region_name_from_endpoint(...)
    clientConfiguration.region = region_name.c_str()
```

Since get_region_name_from_endpoint returns a copy not a reference.

gustavla:
Tested on Ubuntu 16.04 and macOS 10.13.

PR #131, opened 2017-12-20
Add `set -e` to configure to make it fail early

gustavla:
It's quite a common situation that the python installation fails in `./configure` (often due to local experimental changes, but sometimes due to external issues). Right now, if this happens `./configure` returns 0 (no error) and the errors are not immediately visible since there is so much printed after them. As a result, I proceed with `make` and I discover that I have a broken python installation much later (usually around 97% build progress).

I keep adding this flag locally, so I think this will be useful for everyone.

### DIFF
--- a/configure
+++ b/configure
@@ -6,6 +6,7 @@ DEBUG_DIR=debug
 TURI_HOME=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 DEPS_PREFIX=$PWD/deps/local
 
+set -e
 
 function print_help {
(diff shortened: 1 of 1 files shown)


gustavla:
One of the typical situations when this happens to me is if I accidentally call `./configure` in a terminal that has sourced `scripts/python_env.sh`. That should probably be addressed separately, but at least this PR will catch it early along with other Python installation issues.

Another typical case is when a pip install fails (for instance due to incompatibility of a single package) and pip rolls everything back and you are left with an empty python installation.

PR #156, opened 2018-01-02
License year

benhc123:


TobyRoseman:
@benhc123 - Thanks for the pull request. However this change has already been made.

Issue #154, opened 2018-01-02
Unable to run c++ unitests


mordoron:
After installing and building, per the build.md file. Running:
```
cd debug/test
make
ctest 
```

Yields:
*********************************
No test configuration file found!
*********************************
Usage

  ctest [options]

running ```ctest .```
Yields:
Test project /Users/mordoron/GitHub/turicreate/debug/test
No tests were found!!!

Details:
MacOS High Sierra
version 10.13.2

TobyRoseman:
@mordoron - thanks for reporting this issue. I've put up a pull request which fixes this issue for me. It would be helpful if you could verify the fix too.

PR #158, opened 2018-01-03
Fix cpp unit tests

TobyRoseman:
Fix for #154.


BUILD.md
```diff
@@ -72,6 +72,6 @@ From the repo root:
 
     cd debug/test
     make
-    ctest
+    ctest .
```
mordoron: Actually, after the enable_testing() change ``ctest`` works as well as ``ctest .``

PR #159, opened 2018-01-03
Added GUI Functionality for Visualization

abhishekpratapa:
- Modified `Save` and `Save Vega` to `Export as PNG` and `Export as Vega JSON`
- Added a prompt to prevent accidentally overwriting existing files in both `Save Vega`, and `Save`
- Disabled `Save` and `Save Vega` for explore
- Got `Print` working in Visualization
- Added Comma Separators to Summary Stats
- Box and Whiskers, single point for one Datum, fixed by changing the median bar color


src/visualization/Turi Create Visualization/src/VegaContainer.swift
```diff
@@ -271,6 +287,22 @@ class VegaContainer: NSObject, WKScriptMessageHandler {
             }
         }
     }
+    
+    public func get_image(completion: @escaping (NSImage) -> Void) {
+        let jsString = String(format: "export_png();");
```
znation: This could just be `let jsString = "export_png();"`, right?

src/visualization/Turi Create Visualization/src/VegaContainer.swift
```diff
@@ -271,6 +287,22 @@ class VegaContainer: NSObject, WKScriptMessageHandler {
...
+        self.view.evaluateJavaScript(jsString, completionHandler: { (value , err) in
+            
+            if(err != nil){
+                return
+            }
+            
+            let s = String(describing: value!)
+            let dataDecoded = Data(base64Encoded: s, options: .ignoreUnknownCharacters)!
```
znation: Is it a good idea to ignore unknown characters here? Shouldn't we present an error to the user instead? I would think if there are any unknown characters, then something went terribly wrong, and the image probably won't look right.

src/visualization/Turi Create Visualization/src/VegaContainer.swift
abhishekpratapa: Yes, your absolutely correct. I changed it.

src/visualization/Turi Create Visualization/src/VegaContainer.swift
abhishekpratapa: Not entirely sure about this one, though I'll admit its not officially from Apple:

All usage of decoding base64 strings into data, on various swift projects use one of the two options. I'll have to investigate further to figure out how to deal with this.

- options: NSDataBase64DecodingOptions.fromRaw(0)
- options: .ignoreUnknownCharacters

znation: I think we should go with `NSDataBase64DecodingOptions.fromRaw(0)` - when I've seen that pattern before, it essentially means "no options" but is needed to satisfy the type system (because options is a value type and can't be null).

abhishekpratapa: Done

abhishekpratapa:
Updated GUI Functionality for Visualization

PR #161, opened 2018-01-04
Fixes model loading in Python 3

gustavla:
This is an alternative to #110. This PR decodes the model state to unicode strings in Python 3 in the Cython bridge instead of as an afterthought from Python.

However, legacy models need to be decoded and unpickled using the old method, so this PR first peeks at the `archive_version` and then uses old behavior if it's 0 and new behavior if it's 1. Note, `disable_cpp_str_decode()` doesn't change anything in Python 2, however it is important in Python 3 so that we don't get a decode error when reading legacy models. This PR doesn't actually offer support of loading legacy models in Python 3 (this is hard since it involves unpickling Python 2), but it does offer a more user-friendly error message. This is why it's important to avoid the decoding error happening. An alternative to peeking at the archive version would be to simply catch the decoding errors and re-throw a friendly message, but I think assuming decoding errors are always caused by legacy models could mask unrelated future errors.

All 4.0 models are of course loadable in both Python 2 and 3 with this PR.

As a bonus, I also added a nicer error for `archive_version > 1` to be more future proof. Right now, it actually tries to do the legacy loading for newer version numbers.

### DIFF
--- a/src/unity/python/turicreate/cython/cy_variant.pyx
+++ b/src/unity/python/turicreate/cython/cy_variant.pyx
@@ -694,19 +694,6 @@ cdef variant_type from_value(object v) except *:
 
 ################################################################################
 
-cdef read_archive_version(variant_map_type& d):
-    """
-    Reads only the 'archive_version' from a model state.
-    """
-    cdef variant_map_type_iterator it = d.begin()
-    ret = 0
-    while (it != d.end()):
-        if cpp_to_str(deref(it).first) == 'archive_version':
-            ret = to_value(deref(it).second)
-            break
-        inc(it)
-    return ret
-
 
 cdef dict to_dict(variant_map_type& d):
     """
--- a/src/unity/python/turicreate/cython/cy_variant.pyx
+++ b/src/unity/python/turicreate/cython/cy_variant.pyx
@@ -694,6 +694,20 @@ cdef variant_type from_value(object v) except *:
 
 ################################################################################
 
+cdef read_archive_version(variant_map_type& d):
+    """
+    Reads only the 'archive_version' from a model state.
+    """
+    cdef variant_map_type_iterator it = d.begin()
+    ret = 0
+    while (it != d.end()):
+        if cpp_to_str(deref(it).first) == 'archive_version':
+            ret = to_value(deref(it).second)
+            break
+        inc(it)
+    return ret
+
+
 cdef dict to_dict(variant_map_type& d):
     """
     Converts a variant map type to a python dictionary.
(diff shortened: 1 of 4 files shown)


src/unity/python/turicreate/cython/cy_variant.pyx
```diff
@@ -694,6 +694,20 @@ cdef variant_type from_value(object v) except *:
 
 ################################################################################
 
+cdef read_archive_version(variant_map_type& d):
+    """
+    Reads only the 'archive_version' from a model state.
+    """
+    cdef variant_map_type_iterator it = d.begin()
+    ret = 0
```
TobyRoseman: would `-1` make more sense here?

gustavla: I wasn't sure how far the convention of storing `archive_version` dates back. If there are old models that does not store it, but can still be loaded by the legacy loader, we could be unnecessarily cutting off support.

However, if it does date back I like this idea. The change needs to be coupled with handling and raising a `ToolkitError` in `_model.py` though, so I'll add that too.

src/unity/python/turicreate/cython/cy_variant.pyx
```diff
@@ -694,6 +694,20 @@ cdef variant_type from_value(object v) except *:
 
 ################################################################################
 
+cdef read_archive_version(variant_map_type& d):
```
ylow: Move this function into cy_unity.pyx It doesn't quite belong in cy_variant.

src/unity/python/turicreate/cython/cy_variant.pyx
gustavla: @ylow What about this? Do you think we should assume 0 if not encoded?

src/unity/python/turicreate/cython/cy_variant.pyx
gustavla: Will do!

src/unity/python/turicreate/cython/cy_variant.pyx
ylow: 0 is fine.

gustavla:
@ylow @TobyRoseman I moved `read_archive_version` and made sure it compiles and passes relevant tests. I pushed but must have merged in a bunch of other commits by accidents. I fixed by force pushing after a rebase, so both commit shas updated and reviews were reset. You'll have to give it a quick scan again and approve, sorry about that!

Let's first decide whether or no 0 or -1 should be the default (@TobyRoseman pointed this out).

ylow:
0 should be fine. Though -1 might be clearer. I don't anticipate this to be an issue. archive_version has been there from the beginning.

gustavla:
OK! I changed the default to -1 and added an error message for it. I also made byte-loading default for that too so decode errors won't trigger before the friendlier error message.

PR #163, opened 2018-01-04
WIP: Added some more GUI fixes

abhishekpratapa:
## Finished
- [x] Fixed: Passing empty labels into SArray

   Test
``` python
import turicreate as tc 
array = tc.SArray(range(1, 20000))  
array.show(title = "my title", ylabel = "", xlabel = "")
```

- [x] Visualization crashes if `inf` exists in the data
- [x] Streamlined CSS to point to parent element for text color (Assigning color selection exclusively in the body element for the content)
- [x] Fixed inconsistent padding on the left and right for Summary view by creating a Vega Variable to control the left padding

    Test
``` python
import turicreate as tc 
array = tc.SArray(range(1, 20000)) 
frame = tc.SFrame([array])
frame.show()
```

### DIFF
--- a/src/unity/lib/visualization/histogram.cpp
+++ b/src/unity/lib/visualization/histogram.cpp
@@ -192,13 +192,10 @@ void histogram_result::add_element_simple(const flexible_type& value) {
     return;
   }
 
+  // ignore nan values
+  // ignore inf values
   if (value.get_type() == flex_type_enum::FLOAT &&
-      std::isinf(value.get<flex_float>())) {
-    return;
-  }
-
-  if (value.get_type() == flex_type_enum::INTEGER &&
-      std::isinf(value.get<flex_int>())) {
+      !std::isfinite(value.get<flex_float>())) {
     return;
   }
 
@@ -208,13 +205,6 @@ void histogram_result::add_element_simple(const flexible_type& value) {
 
   // resize bins if needed
   this->rescale(this->min, this->max);
-
-  // ignore nan values
-  if (value.get_type() == flex_type_enum::FLOAT &&
-      std::isnan(value.get<flex_float>())) {
-    return;
-  }
-
 
   // update count in bin
   size_t bin = get_bin_idx(value, this->scale_min, this->scale_max);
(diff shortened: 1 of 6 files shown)


src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -2860,6 +2860,14 @@ def show(self, title=None, xlabel=None, ylabel=None):
         import os
         (tcviz_dir, _) = os.path.split(os.path.dirname(__file__))
         path_to_client = os.path.join(tcviz_dir, 'Turi Create Visualization.app', 'Contents', 'MacOS', 'Turi Create Visualization')
+
+        if title is "":
```
znation: I think you may need this logic in [show.py](https://github.com/apple/turicreate/blob/master/src/unity/python/turicreate/visualization/show.py#L72) as well?

src/unity/lib/visualization/histogram.cpp
```diff
@@ -188,11 +188,21 @@ void histogram_result::add_element_simple(const flexible_type& value) {
...
    */
 
-  // ignore undefined values
+
   if (value.get_type() == flex_type_enum::UNDEFINED) {
     return;
   }
 
+  if (value.get_type() == flex_type_enum::FLOAT &&
+      std::isinf(value.get<flex_float>())) {
```
znation: Might want to use [isfinite](http://en.cppreference.com/w/cpp/numeric/math/isfinite) to cover NaN as well.

src/unity/lib/visualization/histogram.cpp
```diff
@@ -188,11 +188,21 @@ void histogram_result::add_element_simple(const flexible_type& value) {
...
+    return;
+  }
+
+  if (value.get_type() == flex_type_enum::INTEGER &&
+      std::isinf(value.get<flex_int>())) {
```
znation: Integers can't be inf or nan (only float). I think you can remove this check.

src/unity/lib/visualization/histogram.cpp
```diff
@@ -222,13 +233,16 @@ void histogram::init(const gl_sarray& source) {
   size_t input_size = m_source.size();
   if (input_size >= 2 &&
       m_source[0].get_type() != flex_type_enum::UNDEFINED &&
-      m_source[1].get_type() != flex_type_enum::UNDEFINED) {
+      m_source[1].get_type() != flex_type_enum::UNDEFINED &&
+      !std::isinf(m_source[0].get<flex_float>()) &&
```
znation: Same here; you might want to use isfinite instead of isinf.

src/unity/lib/visualization/histogram.cpp
abhishekpratapa: Updated

src/unity/lib/visualization/histogram.cpp
abhishekpratapa: Done!

src/unity/python/turicreate/data_structures/sarray.py
abhishekpratapa: Updated with requested changes

znation: Actually I think you want to use `==` instead of `is` for string comparison; see [this post](https://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -69,6 +69,9 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
     (tcviz_dir, _) = os.path.split(os.path.dirname(__file__))
     path_to_client = os.path.join(tcviz_dir, 'Turi Create Visualization.app', 'Contents', 'MacOS', 'Turi Create Visualization')
 
+    if title is "":
```
znation: Same here; string comparison should be `==`.

PR #164, opened 2018-01-04
Add official support for MXNet 0.12 and unofficial support for MXNet 1.X

gustavla:
With this PR, we have support for pretty much all versions of MXNet from 0.11.

I updated the support in `setup.py` to `mxnet >= 0.11, mxnet < 1.0.0`. This PR makes 1.0.0 theoretically work, but that version causes a segfault in the object detector. This segfault seems to be fixed in every subsequent nightly release, so hopefully it will not be an issue in 1.0.1. When that version is released, we can extend official support for 1.X and remove or update the `mxnet < 1.0.0`.

### DIFF
--- a/src/unity/python/setup.py
+++ b/src/unity/python/setup.py
@@ -166,7 +166,7 @@ if __name__ == '__main__':
             "decorator >= 4.0.9",
             "prettytable == 0.7.2",
             "requests >= 2.9.1",
-            "mxnet == 0.11",
+            "mxnet >= 0.11, mxnet < 1.0.0",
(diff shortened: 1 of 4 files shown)


gustavla:
I just pushed a commit that:
- Updates the MXNet version check so it does not complain for 0.12 and designates "0.12.1" as the currently recommended version.
- Updated mxnet-cu80 to recommend mxnet-cu90 and adding a note that it can be adjusted if your CUDA version is different.

With these changes, this PR fully rolls out official 0.12.1 support.

PR #166, opened 2018-01-05
Fix SArray.show with undefined categorical

znation:
This would only repro on an SArray with categorical values with at least
one category that has the same count as the (hidden) "undefined" category.
Even though this doesn't repro on many real datasets, the minimal repro is:

```python
import turicreate as tc
sa = tc.SArray(['x', None])
sa.show()
```

Despite ignoring undefined values lower down in this function, the
item_frequency.cpp vega_column_data function attempts to sort all
keys/values for item frequency. If there is a null key with the same
count as another value, it sorts by key, and then throws when comparing
undefined with a string value.


znation:
@abhishekpratapa @srikris Please review, thanks!

abhishekpratapa:
It works for me, I tested it.

PR #171, opened 2018-01-10
Flexible types conversion system rewrite. 

hoytak:
Rewrote the flexible type converter logic to be much more readible and
flexible.

The logic uses a different metaprogramming method documented in
flexible_type_converter.hpp.  This gives much better control over
error messages and better flexibility in adding additional types.

It should be fully backwards compatible with the previous version.


src/flexible_type/flexible_type_converter.hpp
```diff
@@ -10,657 +10,899 @@
...
-    else if (val.get_type() == flex_type_enum::VECTOR) {
-      flex_vec f = val.to<flex_vec>();
-      flex_list ret;
-      ret.resize(f.size());
-      for (size_t i = 0;i < f.size(); ++i) ret[i] = f[i];
-      return ret;
+template <> struct ft_converter<CVTR__FLOATING_POINT> {
+
+  template <typename T> static constexpr bool matches() {
+    return std::is_floating_point<T>::value;
```
ylow: Why is this and Integer case separate? Why not merge as an std::is_arithmetic test?

src/flexible_type/flexible_type_converter.hpp
```diff
@@ -10,657 +10,899 @@
...
+
+// Dictionaries -- sequences convertible to an exact match.
+static constexpr int CVTR__FLEX_DICT_EXACT                = 8;
+static constexpr int CVTR__FLEX_DICT_CONVERTIBLE_SEQUENCE = 9;
+static constexpr int CVTR__FLEX_DICT_CONVERTIBLE_MAPS     = 10;
+
+// Lists -- any other sequences that cannot be converted into a
+// dictionary or vector but can be converted into a list of flexible
+// type.
+static constexpr int CVTR__FLEX_LIST_EXACT                = 11;
```
ylow: Can all the EXACT cases be merged? Previously is_flexible_type_member<T>::value is used.

src/flexible_type/flexible_type_converter.hpp
```diff
@@ -10,657 +10,899 @@
...
-  static constexpr bool value = true;
-  std::vector<T> get(const flexible_type& val) {
-
-    if (val.get_type() != flex_type_enum::VECTOR) {
-      std::string errormsg = 
-          std::string("Expecting an array of numbers, But we got a ") 
-          + flex_type_enum_to_name(val.get_type());
-      throw(errormsg);
-    }
+template <> struct ft_converter<CVTR__FLEX_VEC_EXACT> {
```
ylow: This file should ideally be sorted in the same order as the conversion test sequence.

src/flexible_type/flexible_type.hpp
```diff
@@ -281,7 +282,14 @@ class flexible_type {
    */
   flexible_type(flexible_type& other) noexcept;
 
+  template <size_t N>
+  flexible_type(const char v[N]):flexible_type() {
```
TobyRoseman: might be nice to add the `void` return type to these functions.

src/flexible_type/flexible_type_converter.hpp
```diff
@@ -10,657 +10,899 @@
...
-  typedef typename 
-      boost::mpl::count<transformed_sequence, 
-                        std::integral_constant<bool, true>>::type num_good;
- 
-  // true if all args are convertible.  
-  static constexpr bool value = (num_good::value == sizeof...(Args));
-};
 
+// The obvious case of another flexible type.
+static constexpr int CVTR__FLEXIBLE_TYPE                  = 1;
```
TobyRoseman: Is there a reason these all can't be an enum?

src/flexible_type/flexible_type.hpp
ylow: This is a constructor.

src/flexible_type/flexible_type_converter.hpp
```diff
@@ -10,657 +10,899 @@
...
-    if (val.get_type() != flex_type_enum::INTEGER && 
-        val.get_type() != flex_type_enum::FLOAT) {
-      std::string errormsg = 
-          std::string("Expecting a numeric type, But we got a ") 
-          + flex_type_enum_to_name(val.get_type());
-      throw(errormsg);
+template <> struct ft_converter<CVTR__FLEX_VEC_CONVERTIBLE_SEQUENCE> {
+
+  template <typename V> static constexpr bool matches() {
+    return (is_sequence_container<V>::value
```
ylow: Previous version of this captured ```is_arithmetic<typename first_nested_type<V>>``` here to provide a fast-ish path for ```std::vector<int>``` etc. The current path for this will go through CVTR__FLEX_LIST_CONVERTIBLE_SEQUENCE which the current implementation will incur a conversion to flexible_type for each element.

src/flexible_type/flexible_type.hpp
TobyRoseman: make sense

src/flexible_type/flexible_type_converter.hpp
hoytak: Yes.  I wasn't aware it wasn't.  Will do.

src/flexible_type/flexible_type_converter.hpp
hoytak: Good point.  I can add that.

src/flexible_type/flexible_type_converter.hpp
hoytak: Yes, they can be merged.  I can do that easily.  Will do.

hoytak:
All tests pass with https://github.com/apple/turicreate/pull/178 on both mac and linux except for one python test on distances that does not raise an exception on bad input when it should.  Investigating.

hoytak:
Okay, the test failure is really a question about correct behavior.  Suppose we do this: 

flexible_type f = 1.0; 
flex_string s = f;  // Error or cast? 

With this PR, this goes through the "to<flex_string>()" operator, which does not error out in this case.  

Previous behavior was a mix of to<...> and get<...> depending on the types.  

Should this implicitly cause a cast?  

If we go with "no" universally, it's not backwards compatible -- double and int are cast to each other. 
if we defer to the to<...> operator, then it's more liberal than current. 

I'm inclined to have it be strict with get<...>, except for numerics. 

-- Hoyt

ylow:
@hoytak I am inclined to have it be strict too.

ylow:
Will that change be in this PR?

ylow:
Actually, I take that back. I am confused. The current behavior for implicit cast to flex_string is via `to<flex_string>`?
```
inline flexible_type::operator flex_string() const {
  return to<flex_string>();
}
```

I recall flexible_type is always implicit castable to string (for printability) and many things depend on this behavior.

hoytak:
The current behavior is exactly what you mentioned; everything is implicitly castable to string.  And I agree with you -- I think it should be either strict or follow the semantics of to<...>(), not somewhere in between. 

The weird behavior in question is due to the following use case.  The Levenstein distance is exposed directly through the extensions mechanism as 

`double levenshtein(const std::string& a, std::string& b);`

The test that's failing simply calls this function with a list and a dict types and asserts that an exception is raised -- i.e. the previous behavior here raised an error; the new mechanism does not.  

I suspect that what changed is the converter did actually change behavior in this, but the implicit casting behavior went through the flex_list converter method and showed up only here.  Thus the old converter didn't match the behavior of the class and now it does.

src/flexible_type/flexible_type_converter.hpp
```diff
@@ -128,9 +128,54 @@ template <> struct ft_converter<CVTR__FLEXIBLE_TYPE_EXACT> {
     dest = src;
   }
  
+  static void get(flex_float& dest, const flexible_type& src) {
```
znation: What's with these static void methods? I don't see any calls to them in this commit (and as of now what I'm looking at is the latest commit in this PR). Were calls to these methods going to come later?

hoytak: No, they are being used by the template calling classes at the end of this file.  

The way this works is that it tests every one of these classes in order by calling the matches() static method, and if it is true, it calls the get / set methods as appropriate. 

Because of how the templating works, none of these classes are ever instantiated, hence the static method.  Also, the destination and source of the conversions need to be overloaded by type, they are arguments, hence the void.

src/flexible_type/flexible_type_converter.hpp
```diff
@@ -10,657 +10,843 @@
...
+
+  flexible_type_internals::ft_resolver<flexible_type_internals::NUM_CONVERTER_STRUCTS-1>::get(t, f);
+};
+
+template <typename T> GL_HOT_INLINE_FLATTEN
+static void convert_to_flexible_type(flexible_type& f, T&& t) {
+  static_assert(is_flexible_type_convertible<T>::value, "Type not convertable to flexible_type.");
+
+  flexible_type_internals::ft_resolver<flexible_type_internals::NUM_CONVERTER_STRUCTS-1>::set(f, std::forward<T>(t));
+};
```
hoytak: @znation -- set is being called here.

src/flexible_type/flexible_type_converter.hpp
```diff
@@ -10,657 +10,843 @@
...
 
-/**
- * all_flexible_type_convertible<Args...>::value is true if every type in 
- * Args can be converted to and from a flexible_type via 
- * flexible_type_converter<T>.
+template <typename T> GL_HOT_INLINE_FLATTEN
+static void convert_from_flexible_type(T& t, const flexible_type& f) {
+  static_assert(is_flexible_type_convertible<T>::value, "Type not convertable from flexible_type.");
+
+  flexible_type_internals::ft_resolver<flexible_type_internals::NUM_CONVERTER_STRUCTS-1>::get(t, f);
```
hoytak: @znation get is called here.

znation: Seems a bit too magical to me, but I'll go with it. 👍

PR #176, opened 2018-01-13
Added patch to gracefully handle pyfpe not being present.

hoytak:
On ubuntu 16.04, the default environment does not create a pyfpe.h file in the installation process.  Thus we should simply warn out if it's not present.

### DIFF
--- a/scripts/install_python_toolchain.sh
+++ b/scripts/install_python_toolchain.sh
@@ -89,5 +89,6 @@ for f in `ls ../../env/bin`; do
 done
 popd
 
-linux_patch_sigfpe_handler
+# This is not always present and so it may fail.
+linux_patch_sigfpe_handler || echo "WARNING: Error patching pyfpe.h file."
 


znation:
@hoytak I think this actually is an error -- but it's not being reported correctly. I've seen this error at build time when I didn't have `libpython2.7-dev` apt-get installed (having it installed globally seems to change the behavior of `virtualenv`). I suspect we do actually need that apt-get install, for reasons other than this one header file (and we probably do actually need this file?).

hoytak:
@znation That makes sense.  The issue then seems to be that the install_python_toolchain.sh script does not check to make sure python2.7-dev is present, and then seems to fail silently when it is not installed. 

I believe the CMake stage of the build process checks for python-dev not being present, but this is run after this patch is applied.  I think the right thing to do would be to move this patching to a stage in the CMake build process immediately after checking for python headers.

znation:
@hoytak Interesting; I think your explanation is correct. Moving the patching to a stage in the CMake build makes sense to me. Alternately, would it be easier to check for python2.7-dev in `./configure`? I think either fix would be fine.

ylow:
It is tricky to move this to a CMake build stage since it is an in-place file modification. Doesn't quite work with Make build rules. In theory this can be done at configure time, but similar caveats apply. 

Probably the simplest solution is for install_python_toolchain.sh script could probably just look for the existence of Python.h in the expected location and fail early with a good error. In the meantime shall I close this PR and you submit a new one? Or do you want to update this PR?

znation:
I like @ylow's suggestion of checking for Python.h early in install_python_toolchain.sh.

znation:
@hoytak Can we close this PR? Given the discussion I don't think the change here is the right fix.

PR #178, opened 2018-01-13
Fix for not thread safe TS_ASSERT in parallel block.

hoytak:
Turns out that the the TS_ASSERT_* macros are not threadsafe on linux (due to the underlying boost calls not being threadsafe).  This causes sporadic failures in some C++ tests.  The immediate workaround is to replace all these statements by builtin ASSERT statements, which raise an exception.

### DIFF
--- a/test/unity/toolkits/sparse_similarity/test_brute_force_all_pairs.cxx
+++ b/test/unity/toolkits/sparse_similarity/test_brute_force_all_pairs.cxx
@@ -134,7 +134,7 @@ void run_test(const std::vector<std::vector<std::pair<size_t, T> > >& data_1,
 
     size_t idx = ref_idx * m + query_idx;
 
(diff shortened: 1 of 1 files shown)


znation:
I've never seen `ASSERT_FALSE` - checking [cppreference](http://en.cppreference.com/mwiki/index.php?title=Special%3ASearch&search=ASSERT_FALSE) it doesn't appear to be built in. Where does it come from?

If you want to use a builtin, you could `#include <cassert>` and use `assert(xyz == false)`.

ylow:
@zach ASSERT_FALSE is our macros. It is preferred to the casserts.

znation:
@ylow Oh, that makes sense. I've only seen the `DASSERT_` variety used before; now I get why there's a `D` in front. 😄

PR #180, opened 2018-01-14
Allow reading of empty CSV files

znation:
It's a bit unintuitive to error out on `turicreate.SFrame.read_csv` on
an empty file with the message:

```
RuntimeError: No files corresponding to the specified path
```

Instead, let's return an empty SFrame. Removing some of the
empty-checking code seems to work and gives an empty SFrame with no
errors.


igiloh:
What if the user has accidentally given the wrong path?
You would want him to get an error as early as possible, making it easy for him to catch the mistake and fix it. Returning an empty SFrame would result in a much less intuitive failure further down the pipeline.

znation:
@igiloh I realize the description of the change may not have been clear. This behavior change is only intended for empty files that exist (which previously gave an incorrect error message), not for nonexistent files. With this PR, the behavior is now:

```python
tc.SFrame.read_csv('empty.csv')
```
```
Columns:
	None

Rows: 0

Data:
	[]
```

```python
tc.SFrame.read_csv('nonexistent.csv')
```
```
RuntimeError: No files corresponding to the specified path (/Users/zach/temp3/nonexistent.csv).
```

Does that address your concern?

igiloh:
@znation yes, thank you!

src/sframe/parallel_csv_parser.cpp
```diff
@@ -990,22 +989,6 @@ std::map<std::string, std::shared_ptr<sarray<flexible_type>>> parse_csvs_to_sfra
   
   for (auto p : file_and_status) {
     if (p.second == file_status::REGULAR_FILE) {
-      // throw away empty files
-      try {
-        general_ifstream fin(p.first);
```
ylow: We do want to keep this try catch.

This was added a long time for the (apparently not uncommon) case where you are reading a collection of CSV files (`tc.SFrame.read_csv('blah/*.csv')`), where each CSV file has a header, AND some files are 0 bytes long. Not skipping the 0 byte files cause the CSV header parse to fail resulting in an unexpected parse failure. 

We could catch the parse failure, but it can be hard to differentiate between a malformed CSV file (in which case we should fail), and this very particular use case (in which case we should just skip and continue). Hence it was easier to just check for the 0 byte case specifically before hand.

znation: I'm very surprised to hear this use case isn't uncommon... I thought about it when modifying the code here, and it seemed like a use case we wouldn't necessarily want to support. I'll put back the try/catch and implement my pass-through fix only for collections of all zero-byte files.

znation:
@ylow Feedback addressed; please re-review.

src/sframe/parallel_csv_parser.cpp
```diff
@@ -1017,6 +1018,13 @@ std::map<std::string, std::shared_ptr<sarray<flexible_type>>> parse_csvs_to_sfra
   file_and_status.clear(); // don't need these anymore
   
   // ensure that we actually found some valid files
+  if (files.empty() && !empty_files.empty()) {
+    // If all we got is empty files, let's take them --
+    // we'll produce an empty SFrame.
+    // But if we have any non-empty files, we should omit the
+    // empty ones so the column schema doesn't conflict.
+    files = empty_files;
```
ylow: Or perhaps just exit here, and put back the lines below.
```
frame.open_for_write({},{},frame_sidx_file);
frame.close();
return {};
```

znation: Not sure if that would have quite the intended behavior -- I'm not sure what you mean by put back the lines below (I didn't remove those lines?). The way it's written currently, everything goes through the same path whether or not the files are empty (diverging only on whether they're all-empty or all-non-empty), and everything seems to work in both cases. I tested the following combinations:

* One empty file (results in empty SFrame with this change).
* One non-empty file (results in non-empty SFrame before and after).
* Two empty files (results in empty SFrame with this change).
* Two non-empty files (results in non-empty SFrame before and after).
* A mix of empty and non-empty files (with this change, ignores the empty files, same as before).

ylow: Sorry, what I wrote was ambiguous.  Might this be a simpler alternative?

- Replacing this section with this:
```
frame.open_for_write({},{},frame_sidx_file);
frame.close();
return {};
```
 - restore the deletions at 1029-1031.

znation: Oh, I see. Sure, I'll make that change.

znation:
@ylow Feedback addressed again; please re-review.

PR #181, opened 2018-01-14
Fix JSON export of array dtype

znation:
This fixes the formatting of the `flex_vec` (array) dtype in CSV export
from SFrame.

Still quite a few bugs lurking in the JSON export -- I also tested
inf/nan handling and it's broken. Leaving that commented out in the test
for now. I suspect other things are broken too (namely datetime or Image
types - not sure how to reasonably JSON export those).

We might want to consider extensions._json for JSON export.

### DIFF
--- a/src/sframe/csv_writer.cpp
+++ b/src/sframe/csv_writer.cpp
@@ -25,8 +25,15 @@ void csv_writer::csv_print_internal(std::string& out, const flexible_type& val) 
       out += std::string(val);
       break;
     case flex_type_enum::DATETIME:
-    case flex_type_enum::VECTOR:
       out += std::string(val);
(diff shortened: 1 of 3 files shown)


znation:
@hoytak I'm not sure what you mean by previous difference. If you think changing the `flex_vec::to<std::string>` behavior is the better fix I'm happy to do that instead; I figured just changing the CSV writer was the more scoped fix. (Also: most of the types already have their own CSV writer output behavior, alongside the change I just made here; they don't use the `to` method as is. Should they?)

ylow:
Yes. The csv localized fix is better. The flexible_type has never promised JSON compatibility and it is not about to sign up for that.

PR #186, opened 2018-01-14
Generate compile_commands.json

znation:
This file is used with tooling like IDEs to get autocompletion and other
features that need to be aware of compile flags used on each source
file. Seems to affect build performance minimally, and only takes a few
MB of disk space when populated. I think it makes sense to turn it on
globally so that it's there when tools/IDEs need it.


igiloh:
My IDE (CLion) seems to be able to build the project without this file, directly from the Cmake files.
I think we should define some target IDEs to test this on, and verify if the `compile_command` is indeed required for them, and whether it affects build time (from scratch).
Maybe we could ask the community which IDEs are important, as well as ask for assistance in testing it?

znation:
@igiloh I think I misspoke with IDEs - this change is intended for editors that don't do their own build, and don't have proper completion without the aid of external plugins. Those plugins typically depend on something like `compile_commands.json`.

I tested it with the [YouCompleteMe](http://valloric.github.io/YouCompleteMe/#c-family-semantic-completion) vim plugin and it works -- I get true semantic completion instead of the normal vim "tag"-based completion.

A few other tools that I haven't tried, but claim to be compatible with this `compile_commands.json` file:

* [linter-gcc](https://atom.io/packages/linter-gcc) for Atom
* [vscode-cpptools](https://github.com/Microsoft/vscode-cpptools/blob/master/Documentation/Getting%20started.md#how-to-specify-the-include-paths) for Visual Studio Code
* [flycheck-cmake](https://github.com/xwl/flycheck-cmake) for Emacs
* [auto-complete-clang](https://github.com/xwl/auto-complete-clang) for Emacs

I also measured the perf of the build on each branch, over 5 runs each, with the command:
```bash
rm -rf debug/ release/ && ./configure && pushd debug/src/unity && time make -j8 && popd
```

In my 5 runs, there was no difference between the wall clock time of the build on `master` and this PR branch. (In fact, this branch was 2s faster on average, but with a stdev of around 5s.) See [compile_commands perf impact.txt](https://github.com/apple/turicreate/files/1633054/compile_commands.perf.impact.txt).

PR #188, opened 2018-01-17
Fix C++ unit test build

znation:
Apparently this broke with #179 -- it seems that change introduced
a new static library `visualization` with a mutual dependency on
`unity_core`. For some reason this works when building `unity_core`
itself, but breaks for any consumer of `unity_core` as a static
library (namely the C++ unit tests that depend on it implicitly).


znation:
Nevermind, this isn't the right fix. Looks like we're just missing a dependency in the `visualization/CMakeLists.txt`.

PR #193, opened 2018-01-18
Fix TS_ASSERTS to be threadsafe

ylow:
An annoying side effect is that TS_ASSERTS are now *really* slow.
In particular for some tests that loop TS_ASSERTS millions of times,
this will degrade test performance significantly. Those tests may have
to change from:

  TS_ASSERT(something);

to

  if (something) {
    TS_ASSERT(something);
  }

It will also be problematic if TS_ASSERT blocks, in particular, in
tests of threading primitives. For instance this is not a good idea.

  TS_ASSERT(wait_for_thread());

instead:

  int ret = wait_for_thread();
  TS_ASSERT(ret);

### DIFF
--- a/src/util/test_macros.hpp
+++ b/src/util/test_macros.hpp
@@ -5,16 +5,20 @@
  */
 #ifndef TEST_MACROS_HPP
 #define TEST_MACROS_HPP
+#include <mutex>
+static std::recursive_mutex __b_lock__;
+#define _TS_ADD_LOCK_GUARD std::lock_guard<std::recursive_mutex> __guard_b__(__b_lock__)
 
-#define TS_ASSERT(x)                     BOOST_CHECK(x);
-#define TS_ASSERT_EQUALS(x,y)            BOOST_TEST((x) == (y));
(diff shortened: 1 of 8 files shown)


src/flexible_type/flexible_type.hpp
```diff
@@ -445,6 +445,33 @@ class flexible_type {
   template <typename T>
   const T& get() const;
 
+
+  /**
```
znation: This part doesn't look related to making TS_ASSERT thread safe. Should there be two separate commits here?

ylow: Ah yes. indeed. This also fixes a bunch of other test failures. I can rename the PR. Or I can submit that as a different PR. What do you prefer?

ylow: I think it I will split into a different PR.

ylow:
PR has been split.

hoytak:
Why don't we just format all our ASSERT_* statements so they can be picked up by the rest of our test system, and just use those here?

ylow:
ASSERTs just raise exceptions. TS_ASSERTS bind into the test harness system (boost test in this case)

PR #197, opened 2018-01-18
Make explicit unsafe use of flexible_type

ylow:
Adds 2 new methods to flexible_type

  reinterpret_get<T>
  reinterpret_mutable_get<T>

These two functions behave identically to get<T> and mutable_get<T> but
bypasses type checking. They are used in limited situations where it is
useful to reinterpret between integer and floating points types and to
make it clear when this is happening.

This also fixes the test failures introduced in #177

### DIFF
--- a/src/flexible_type/flexible_type.hpp
+++ b/src/flexible_type/flexible_type.hpp
@@ -463,6 +463,33 @@ class flexible_type {
   template <typename T>
   const T& get() const;
 
+
+  /**
+   * Gets a modifiable type unchecked modifiable reference to the value stored 
+   * inside the flexible_type.
+   *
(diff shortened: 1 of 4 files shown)


ylow:
I would prefer to add functions which try to have similar semantics as the existing methods. The current methods try to separate the internal representation from the user facing methods. So I prefer not to have a method like get_raw_data_ptr() that leaks the internal representation. 

Basically, if we let the internal representation change drastically, we can still implement reinterpret_get so that no other code breaks. But on the other hand if we provide get_raw_data_ptr(), we will need to audit every use of it to make sure that no one else is making bad assumption.

PR #199, opened 2018-01-18
Fix README.md dead links

emannuelOC:
Some links in the README.md file were dead since the name of the files they were pointing to had changed.


TobyRoseman:
@emannuelOC - Thanks for fixing this.

PR #200, opened 2018-01-18
Fix loose flex type casting in histogram

znation:
Now that flex_int can't implicitly cast to flex_float, we need to use
the `to` method instead of `get`.


hoytak:
This would not have done what you thought it should previously.

znation:
@hoytak Could you elaborate? It definitely _was_ doing what I wanted previously... and I hope it's now doing that with `to`. The goal of these checks are:

* Assume the value is either int, float, or undefined.
* If the value is undefined, skip to the next case.
* If the value is inf/nan, skip to the next case.

hoytak:
@znation -- apologies; a previous comment didn't end up posting correctly.  Basically, under the old version, the actual bits of an integer would have been interpreted as those of a double.  If the integer sequence happened to be NaN in a double, e.g. 0x7fc00000, then it would have triggered this.  For this test, it's best to cast it to a double first.

PR #215, opened 2018-01-24
Added guard against memory leak if start_server is called multiple ti…

hoytak:
…mes.

### DIFF
--- a/src/unity/server/unity_server_control.cpp
+++ b/src/unity/server/unity_server_control.cpp
@@ -34,6 +34,10 @@ EXPORT void start_server(const unity_server_options& server_options) {
 
   unity_server_initializer server_initializer;
 
+  if(SERVER) { 
+    stop_server(); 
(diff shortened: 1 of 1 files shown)


src/unity/server/unity_server_control.cpp
```diff
@@ -34,6 +34,10 @@ EXPORT void start_server(const unity_server_options& server_options) {
 
   unity_server_initializer server_initializer;
 
+  if(SERVER) {
```
TobyRoseman: Rather than stopping and starting the server in this case, would it be better to only start the server if it isn't already started, i.e. change the next six lines to something thing like this:
```
if(!SERVER) { 
    SERVER = new unity_server(server_options);
    SERVER->start(server_initializer);
}

hoytak: No, as the all of the options may change.

PR #218, opened 2018-01-25
Fix AC loss func

igiloh:
According to [MXne API doc](https://mxnet.incubator.apache.org/api/python/gluon/loss.html?highlight=softmaxcrossentropyloss#mxnet.gluon.loss.SoftmaxCrossEntropyLoss), the cross entropy loss is averaged out (and not summed) over all axes but the  batch axis. This means that our current loss is off by a factor of `seq_len`.
Thanks for @glrasson for noticing this.

We've also spotted a doc mistake in the shapes before calling the last dense layer.

### DIFF
--- a/src/unity/python/turicreate/toolkits/activity_classifier/_model_architecture.py
+++ b/src/unity/python/turicreate/toolkits/activity_classifier/_model_architecture.py
@@ -34,7 +34,7 @@ def _define_model(features, target_map, pred_win, seq_len, context):
                        seq_len=seq_len, dropout=0.2)
 
     # Dense
-    dense = _mx.sym.Reshape(data=lstm, shape=(-1, _net_params['lstm_h']))  # NTC to NxTC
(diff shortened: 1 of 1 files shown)


gustavla:
The final training loss gets added to the model state, right? That means we won't be able to compare between version 4.0 and subsequent versions.

By the way, as a validation, have you checked that the loss starts from log C now, where C is the number of target?

alonpal:
The calculated loss is actually meant to be the "mean loss per sequence" and not "mean loss across timestamps” although in my experience they are quite similar for activity classification.

In order to calculate the mean loss across timestamps, notice how `loss_per_seq` is weighted, so the correct calculation would be: `loss = _mx.sym.sum(loss_per_seq * seq_sum_weights) / (_mx.sym.sum(seq_sum_weights) + 1e-5)`

The same applies for the accuracy calculation.

PR #221, opened 2018-01-27
Add NDArray Type

ylow:
This somewhat large commit adds an NDArray type to SFrame.

ndarray<T> type
---------------
In src/flexible_type/ndarray.hpp

ndarray<T> implements a standard ndarray layout comprising of
 - a vector of elements (really a shared_ptr over elements)
 - a vector of shape
 - a vector of strides
 - an 'start' offset integer

Array indexing is computed as follows:

  ndarray[i,j,k] = elements[start + i * stride[0] + j * stride[1] + k * stride[2]]

Note the stride are not based on the element size. i.e. if we have a simple
1-D array, stride[0] is always 1. (as opposed to sizeof(T)).

The NDArray's default construction layout is "C" ordering: the stride
array is non-increasing. That is, for a simple 3D ndarray[i,j,k], elements
ndarray[i,j,k], ndarray[i,j,k+1] are adjacent.

However, due to to explicit representation of offset, shape and stride, almost
arbitrary layouts and slices can be represented by the ndarray.

The current implementation of the ndarray enforces that all mutation
operations require elems to be unique (elems.use_count() == 1). Hence views
are not modifiable; modifying a view automtacally incurs a data copy. This
might be relaxed in the future (perhaps with a inherited view<T> type) but
this assumption allows for simpler memory management right now.

NDArray operations are right now not particularly efficient. There are no
fast paths that avoid recomputing the array indexing for every element.
For instance, an operator+= on two NDArrays of the same 2D shape, the code is
roughly as efficient as:

    for (i = 0; i < shape[0]; ++i) {
      for (j = 0; j < shape[1]; ++j) {
        left_index = i * left.stride[0] + j * left.stride[1]
        right_index = i * right.stride[0] + j * right.stride[1]
        left.elements[left_index] += right.elements[right_index]
      }
    }

When really if both arrays have the same stride and shape, we could:

    for (i = 0;i < shape[0] * shape[1]; ++i) {
      left.elements[i] += right.elements[i]
    }

flexible_type
-------------

ndarray<double> is typedefed to flex_nd_vec (flex_type_enum::ND_VECTOR) which
is a new addition to the flexible_type union. The following conversion paths
are currently implemented.
 - image <--> ndarray
 - array <--> ndarray
 - list --> ndarray
 - ndarray --> str

To support some of these better, some image operations
(encode_image, decode_image) have been moved around a bit.

Operators +,-,*,/ are implemented only against nd_vector, int, and float.

SFrame
------
Once the ndarray is supported in the flexible_type supporting it in SFrame
requires a surprisingly little amount of work. Encoders and Decoders for the
new type have been implemented such that an image represented as an ndarray
will generally not require more storage than the raw image equivalent.

The VECTOR SUM and AVG groupby aggregators have been expanded to support
ND_VECTOR which allows groupby() and sum() operators to automatically take
advantage of them.

Four unnecessary std::moves were also removed to remove clang warnings.

Python SFrame
-------------
The ndarray is exposed to Python as the type "np.ndarray". This forces
numpy to be a *required* dependency. Only *exactly* np.ndarray types will be
converted to ndarray. No other cast paths are permitted.

Conversion from np.ndarray to flexible_type[containing an ndarray] always
requires a copy.

Conversions from flexible_type[containing an ndarray] to np.ndarray does not
require a copy relying on Python's buffer protocol to expose the ndarray<double>
memory layout to Python.

What works on ndarray columns:
 - Most numeric ops
 - sum()
 - SFrame groupby sum/avg on ndarray columns

Also SArray / SFrame indexing cache size had to be tweaked and reduced to
better support large ndarrays.

Backward Compatibility
----------------------
Since this new datatype allows all np.ndarray to be maintained as np.ndarray,
this breaks backwards compatibility for operations of the form:

    a = [np.array([1,2,3]),np.array([2,3,4])]
    tc.SArray(a)
    a.dtype == array.array # previous
    a.dtype == np.ndarray  # now

There were a number of tests which assume the previous behavior. All have been
fixed.

Previously, due to implicit np.ndarray --> array.array casting, it might have
been possible to call model prediction using ndarrays. Now, since ndarray
are not (yet) supported by any toolkit, this will likely result in an error.

Known Issues
------------
 - The ndarray type is not supported by any of the toolkits. All toolkits
 need to be upgraded to be aware of this type.
 - show() does not support ndarray.
 - explore() works, but how it is displayed may not be ideal.

### DIFF
--- a/src/flexible_type/flexible_type.hpp
+++ b/src/flexible_type/flexible_type.hpp
@@ -47,7 +47,7 @@ namespace turi {
  *  - \ref flex_int
  *  - \ref flex_float
  *  - \ref flex_string
- *  - \ref flex_vec
+ *  - \ref flex_nd_vec
  *  - \ref flex_list
  *  - \ref flex_dict
  *  - \ref flex_image 
@@ -741,6 +741,11 @@ class flexible_type {
   operator flex_vec() const;
 
   /**
+   * Implicit cast to ndarray<double>
+   */
+  operator flex_nd_vec() const;
+
+  /**
    * Implicit cast to vector<flexible_type>
    */
   operator flex_list() const;
@@ -1227,6 +1232,7 @@ class flexible_type {
     flex_float dblval;
     std::pair<atomic<size_t>, flex_string>* strval;
     std::pair<atomic<size_t>, flex_vec>* vecval;
+    std::pair<atomic<size_t>, flex_nd_vec>* ndvecval;
     std::pair<atomic<size_t>, flex_list>* recval;
     std::pair<atomic<size_t>, flex_dict>* dictval;
     std::pair<atomic<size_t>, flex_image>* imgval;
@@ -1258,6 +1264,16 @@ class flexible_type {
          val.vecval = new std::pair<atomic<size_t>, flex_vec>(*(val.vecval));
          val.vecval->first.value = 1;
          decref(prev, flex_type_enum::VECTOR);
+       }
+       break;
+     case flex_type_enum::ND_VECTOR:
+       if (val.ndvecval->first.value == 1) return;
+       else {
+         union_type prev;
+         prev = val;
+         val.ndvecval = new std::pair<atomic<size_t>, flex_nd_vec>(*(val.ndvecval));
+         val.ndvecval->first.value = 1;
+         decref(prev, flex_type_enum::ND_VECTOR);
        }
        break;
      case flex_type_enum::LIST:
@@ -1310,6 +1326,12 @@ class flexible_type {
          v.vecval = NULL;
        }
        break;
+     case flex_type_enum::ND_VECTOR:
+       if (v.ndvecval->first.dec() == 0) {
+         delete v.ndvecval;
+         v.ndvecval = NULL;
+       }
+       break;
      case flex_type_enum::LIST:
        if (v.recval->first.dec() == 0) {
          delete v.recval;
@@ -1341,6 +1363,9 @@ class flexible_type {
        break;
      case flex_type_enum::VECTOR:
        v.vecval->first.inc();
+       break;
+     case flex_type_enum::ND_VECTOR:
+       v.ndvecval->first.inc();
        break;
      case flex_type_enum::LIST:
        v.recval->first.inc();
@@ -1531,6 +1556,22 @@ template <>
 inline FLEX_ALWAYS_INLINE const flex_vec& flexible_type::get<flex_vec>() const {
   DFLEX_TYPE_ASSERT(get_type() == flex_type_enum::VECTOR);
   return val.vecval->second;
+}
+
+
+// ND_VECTOR
+template <>
+inline FLEX_ALWAYS_INLINE flex_nd_vec& flexible_type::mutable_get<flex_nd_vec>() {
+  DFLEX_TYPE_ASSERT(get_type() == flex_type_enum::ND_VECTOR);
+  ensure_unique();
+  return val.ndvecval->second;
+}
+
+
+template <>
+inline FLEX_ALWAYS_INLINE const flex_nd_vec& flexible_type::get<flex_nd_vec>() const {
+  DFLEX_TYPE_ASSERT(get_type() == flex_type_enum::ND_VECTOR);
+  return val.ndvecval->second;
 }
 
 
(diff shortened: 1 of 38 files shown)


src/flexible_type/flexible_type.hpp
```diff
@@ -1260,6 +1266,16 @@ class flexible_type {
...
        }
        break;
+     case flex_type_enum::ND_VECTOR:
+       if (val.ndvecval->first.value == 1) return;
+       else {
+         union_type prev;
+         prev = val;
+         val.ndvecval = new std::pair<atomic<size_t>, flex_nd_vec>(*(val.ndvecval));
+         val.ndvecval->first.value = 1;
+         decref(prev, flex_type_enum::VECTOR);
```
hoytak: Is this supposed to be ::VECTOR here?

src/flexible_type/ndarray.hpp
```diff
@@ -0,0 +1,715 @@
...
+  */
+  bool has_canonical_stride() const {
+    if (m_stride.size() == 0) return true;
+    if (m_stride[0] == 0) return false;
+    for (size_t i = 1; i < m_stride.size(); ++i) {
+      if (m_stride[i] == 0 || m_stride[i - 1] < m_stride[i]) {
+        return false;
+      }
+    }
+    return true;
```
hoytak: Need to test that final stride equals 1?

src/flexible_type/ndarray.hpp
```diff
@@ -0,0 +1,715 @@
...
+    do {
+      T& v = (*this)[fast_index(idx)];
+      v = -v;
+    } while(increment_index(idx));
+    return *this;
+  }
+
+
+  /// element-wise equality. The other array must have the same shape.
+  bool operator==(const ndarray<T>& other) const {
```
hoytak: Add test for &other == this ?

src/flexible_type/flexible_type.hpp
ylow: Yes. Good catch.

src/flexible_type/ndarray.hpp
ylow: Yeah. This just tests for C-ordering. is_full() will test if it is complete.

src/flexible_type/ndarray.hpp
ylow: good catch.

test/flexible_type/flexible_datatype.cxx
```diff
@@ -1,25 +1,3 @@
-/*
```
TobyRoseman: We should probably replace this with our new header.

test/flexible_type/ndarray_test.cxx
```diff
@@ -0,0 +1,572 @@
+#define BOOST_TEST_MODULE
```
TobyRoseman: Also probably want a header here too.

test/flexible_type/flexible_datatype.cxx
ylow: done

test/flexible_type/ndarray_test.cxx
ylow: done

src/unity/python/turicreate/cython/cy_flexible_type.pyx
```diff
@@ -1553,6 +1695,23 @@ cdef inline array.array pyvec_from_flex_vec(const flex_vec& fv):
     array.extend_buffer(ret, <char*>fv.data(), n)
     return ret
 
+@cython.boundscheck(False)
+cdef inline object pyndarray_from_flex_nd_vec(const flex_nd_vec& fv):
+    wrapper = NDArrayWrapper()
+    wrapper.initialize(&fv)
+    return np.asarray(wrapper)
+    # old unused code path
```
hoytak: Dead code removal?

src/unity/python/turicreate/cython/cy_flexible_type.pyx
```diff
@@ -534,10 +593,6 @@ cdef __fill_numpy_info():
     for c in <str>(__typecodes.get("Character", "") + "SU"):
         _dtype_to_flex_enum_lookup[<int>(ord(c))] = STRING
 
-cdef type np_ndarray
```
hoytak: The reason this was left here is for speed.  import numpy as np goes through python's module and dict lookup system, this is a local C variable.  Thus if we refer to np_ndarray, it avoids the dictionary lookup performed to get the type value.

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4084,10 +4086,10 @@ def groupby(self, key_column_names, operations, *args):
                   val = operation[key]
                   if type(val) is tuple:
                     (op, column) = val
-                    if (op == '__builtin__avg__' and self[column[0]].dtype is array.array):
+                    if (op == '__builtin__avg__' and self[column[0]].dtype in [array.array, numpy.ndarray]):
```
hoytak: use np_ndarray here ?

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -2012,6 +2012,8 @@ def __get_pretty_tables__(self, wrap_text=False, max_row_width=80,
         def _value_to_str(value):
             if (type(value) is array.array):
                 return str(list(value))
+            elif (type(value) is numpy.ndarray):
```
hoytak: use np_ndarray here.

Issue #153, opened 2018-01-02
Build Prerequisites are Missing for Build on Mac


mordoron:
I experience two failures building turicreate from source on Mac High-Sierra:

1. ```bin/sh -c \"/Users/mordoron/GitHub/turicreate/debug/src/visualization/Turi\Create\Visualization.build/Debug/Tui\ Create\ Visualization.build/ScriptFC1A878A1FB14D2900A67DAD.sh\"/Users/mordoron/GitHub/turicreate/debug/src/visualization/Turi Create Visualization.build/Debug/Turi Create Visualization.build/ScriptFC1A878A1FB14D2900A67DAD.sh: line 3: npm: command not found/Users/mordoron/GitHub/turicreate/debug/src/visualization/Turi CreateVisualization.build/Debug/Turi Create Visualization.build/ScriptFC1A878A1FB14D2900A67DAD.sh: line 4: npm: command not found Command /bin/sh failed with exit code 127```
This is due to npm not being installed on my Mac. 
We may want to add this as a prerequisite to the build.md file.

2. ```/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libunity_coreml_model_export.a(NeuralNetwork.cpp.o) has no symbols```
This was fixed by updating Xcode.app.
Again we may want to add a prerequisite with Xcode version.

znation:
We depend on [node](https://nodejs.org/en/) for visualization (typically the latest official LTS version should work). We should check for it in `./configure` and error out, and we should update the build instructions to reflect that we depend on it.

As for the second point - @mordoron, what Xcode version gives this issue, and what version fixed it? I'm not sure we have been rigorous about testing the build with different Xcode versions so it's not surprising it might fail with something earlier than we've tried.

mordoron:
Thanks for looking into this.
My current Xcode version (on which the build succeeds) is: Version 9.2 (9C40b)
I am not 100% sure as for the version on which it failed, but you can see below the update (from my app-store record) which seems to have fixed the issue:


Hope this helps...

 Mor

znation:
Fixed with #230.

PR #230, opened 2018-02-01
Note nodejs/npm as a build dependency on macOS

znation:
* Update BUILD.md to accurately describe pre-existing dependencies vs.
  supplied dependencies.
* Error out in CMakeLists.txt if npm is missing, and
  TC_BUILD_VISUALIZATION_CLIENT is enabled.


znation:
Fixes #153.

PR #237, opened 2018-02-02
Fixed an issue with export to coreml.

### LINKED ISSUE #55: Support/examples for using TuriCreate with pandas?
The example in the README shows how to use Turi Create with an SFrame. However, [SFrame has not had a release since July 2016](https://pypi.python.org/pypi/SFrame).

pandas is pretty much the de-facto DataFrame library in the Python data science ecosystem. It is under [active development](https://pypi.python.org/pypi/SFrame), has a [substantial community](https://github.com/pandas-dev/pandas/stargazers) (with [tens of thousands of StackOverflow questions](https://stackoverflow.com/questions/tagged/pandas)), and provides [tools for many common steps in data analysis](http://pandas.pydata.org/pandas-docs/stable/basics.html).

Please consider aligning with the general momentum in the Python data science community by providing examples of using pandas with Turi Create.

srikris:
Fixes #55 

@alonpal @hoytak Please review.

### DIFF
--- a/src/unity/python/turicreate/test/test_coreml_export.py
+++ b/src/unity/python/turicreate/test/test_coreml_export.py
@@ -68,8 +68,8 @@ class CoreMLExportTest(unittest.TestCase):
         # Arrange
         model = self.model
         test_data = self.sf[:]
-        #test_data['cat_column'] = 'new_cat'
-        #test_data['dict_column'] = [{'new_cat': 1} for i in range(len(test_data))]
+        test_data['cat_column'] = 'new_cat'
+        test_data['dict_column'] = [{'new_cat': 1} for i in range(len(test_data))]
 
         # Assert
         model.predict(test_data)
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/test/test_coreml_export.py
```diff
@@ -22,6 +22,26 @@
...
+        'a': [{'x': 1}, {'y': 1}],
+        'b': [0, 1],
+    })
+    model = tc.logistic_classifier.create(data, target='b', features=['a'],
+            verbose=False)
+
+    # Test set with a new class for categorical variables
+    test = tc.SFrame({'a': [{'z': 1}]})
+    model.predict(test)
+    model.export_coreml('LogisticClassifier.mlmodel')
```
gustavla: It looks like this file is not cleaned up. You can use the `tempfile` package.

gustavla: Also, why is this test not inside the test class?

srikris:
@gustavla Address your comments.

PR #238, opened 2018-02-03
Raise an error on Type Mismatch in Flexible Type Converter

ylow:
flexible_type.get<T>() only fails on debug mode on a type mismatch. We need to explicitly check it here.


znation:
Oh! Good catch @ylow. Looks good to me.

PR #242, opened 2018-02-05
Image similarity export to CoreML

srikris:
- Adds CoreML export to Image similarity

Requires:
- Test cases
- Documentation
- Extract reference data from the model

### DIFF
--- a/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
+++ b/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
@@ -476,7 +476,7 @@ class ImageSimilarityModel(_CustomModel):
         input_name = self.feature_extractor.data_layer
         output_name = 'distance'
         input_features = [(input_name, _datatypes.Array(*(self.input_image_shape)))]
-        output_features = [(output_name, _datatypes.Array(*(num_examples, )))]
+        output_features = [(output_name, _datatypes.Array(num_examples))]
 
         # Create a neural network
         builder = _neural_network.NeuralNetworkBuilder(
@@ -507,9 +507,8 @@ class ImageSimilarityModel(_CustomModel):
                                   input_channels=embedding_size, output_channels=num_examples,
                                   input_name=feature_layer, output_name='v^2-2vu')
 
-        builder.add_elementwise('element_wise-u^2', mode='MULTIPLY',
-                                input_names=[feature_layer, feature_layer],
-                                output_name='element_wise-u^2')
+        builder.add_unary('element_wise-u^2', mode='power', alpha=2,
+                          input_name=feature_layer, output_name='element_wise-u^2')
 
         # Produce a vector of length num_examples with all values equal to u^2
         builder.add_inner_product('u^2', W=_np.ones((embedding_size, num_examples)),
--- a/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
+++ b/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
@@ -409,3 +409,92 @@ class ImageSimilarityModel(_CustomModel):
         +----------+----------+----------------+------+
         """
         return self.similarity_model.similarity_graph(k, radius, include_self_edges, output_type, verbose)
+
+    def export_coreml(self, filename, reference_data):
+        """
+        Save the model in Core ML format.
+
+        See Also
+        --------
+        save
+
+        Examples
+        --------
+        >>> model.export_coreml('myModel.mlmodel')
+        """
+        import numpy as _np
+        import coremltools as _cmt
+        from coremltools.models import datatypes as _datatypes, neural_network as _neural_network
+        from .._mxnet_to_coreml import _mxnet_converter
+        from turicreate.toolkits import _coreml_utils
+
+        input_name = self.feature_extractor.data_layer
+        output_name = 'distance'
+
+        output_size = reference_data.num_rows()
+
+        input_features = [(input_name, _datatypes.Array(*(self.input_image_shape)))]
+        output_features = [(output_name, _datatypes.Array(*(output_size, )))]
+
+        builder = _neural_network.NeuralNetworkBuilder(
+            input_features, output_features, mode='classifier')
+
+        # Convert the feature exctraction network
+        mx_feature_extractor = self.feature_extractor._get_mx_module(
+            self.feature_extractor.ptModel.mxmodel,
+            self.feature_extractor.data_layer,
+            self.feature_extractor.feature_layer,
+            self.feature_extractor.context,
+            self.input_image_shape
+        )
+
+        batch_input_shape = (1, ) + self.input_image_shape
+        _mxnet_converter.convert(mx_feature_extractor, mode=None,
+                                 input_shape={input_name: batch_input_shape},
+                                 builder=builder, verbose=False)
+
+        # Calculate the euclidean distance between the newly extracted query features
+        # and each extracted reference feature.
+        # Calculation of sqrt(v^2 - 2vu + u^2) ensues.
+        W = self.feature_extractor.extract_features(reference_data, self.feature).to_numpy()
+        b = (W * W).sum(axis=1)
+        embedding_size = W.shape[1]
+
+        feature_layer = self.feature_extractor.feature_layer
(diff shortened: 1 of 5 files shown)


src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -409,3 +409,92 @@ def similarity_graph(self, k=5, radius=None, include_self_edges=False,
         +----------+----------+----------------+------+
         """
         return self.similarity_model.similarity_graph(k, radius, include_self_edges, output_type, verbose)
+
+    def export_coreml(self, filename, reference_data):
+        """
+        Save the model in Core ML format.
+
```
TobyRoseman: Is client code needed to sort the output and pick the top k? If so, it's probably worth a mention here.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -409,3 +409,92 @@ def similarity_graph(self, k=5, radius=None, include_self_edges=False,
...
+        feature_layer = self.feature_extractor.feature_layer
+        builder.add_inner_product('v^2-2vu', W=-2 * W, b=b, has_bias=True,
+                                  input_channels=embedding_size, output_channels=W.shape[0],
+                                  input_name=feature_layer, output_name='v^2-2vu')
+
+        builder.add_elementwise('element_wise-u2', mode='MULTIPLY',
+                                input_names=[feature_layer, feature_layer],
+                                output_name='element_wise-u2')
+
+        builder.add_inner_product('u2', W=_np.ones((embedding_size, output_size)),
```
TobyRoseman: should this be called `u^2` rather than `u2`?

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -409,3 +409,92 @@ def similarity_graph(self, k=5, radius=None, include_self_edges=False,
...
+        batch_input_shape = (1, ) + self.input_image_shape
+        _mxnet_converter.convert(mx_feature_extractor, mode=None,
+                                 input_shape={input_name: batch_input_shape},
+                                 builder=builder, verbose=False)
+
+        # Calculate the euclidean distance between the newly extracted query features
+        # and each extracted reference feature.
+        # Calculation of sqrt(v^2 - 2vu + u^2) ensues.
+        W = self.feature_extractor.extract_features(reference_data, self.feature).to_numpy()
+        b = (W * W).sum(axis=1)
```
TobyRoseman: `v_squared` I think is a better variable name than `b`, but perhaps I don't understand what's going on.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -409,3 +409,92 @@ def similarity_graph(self, k=5, radius=None, include_self_edges=False,
...
+            self.feature_extractor.context,
+            self.input_image_shape
+        )
+
+        batch_input_shape = (1, ) + self.input_image_shape
+        _mxnet_converter.convert(mx_feature_extractor, mode=None,
+                                 input_shape={input_name: batch_input_shape},
+                                 builder=builder, verbose=False)
+
+        # Calculate the euclidean distance between the newly extracted query features
```
TobyRoseman: I'm having a really difficult time understanding the code that ensues. Any additional comments you can leave here would be very helpful.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -409,3 +409,135 @@ def similarity_graph(self, k=5, radius=None, include_self_edges=False,
...
+        # Get the reference data from the model
+        proxy = self.similarity_model.__proxy__
+        reference_data = _np.array(_tc.extensions._nearest_neighbors._nn_get_reference_data(proxy))
+        num_examples, embedding_size = reference_data.shape
+
+        # Get the input and output names
+        input_name = self.feature_extractor.data_layer
+        output_name = 'distance'
+        input_features = [(input_name, _datatypes.Array(*(self.input_image_shape)))]
+        output_features = [(output_name, _datatypes.Array(*(num_examples, )))]
```
gustavla: Why not just `_datatypes.Array(num_examples)`?

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -409,3 +409,135 @@ def similarity_graph(self, k=5, radius=None, include_self_edges=False,
...
+        v_squared = (V * V).sum(axis=1)
+
+        feature_layer = self.feature_extractor.feature_layer
+        builder.add_inner_product('v^2-2vu', W=-2 * V, b=v_squared, has_bias=True,
+                                  input_channels=embedding_size, output_channels=num_examples,
+                                  input_name=feature_layer, output_name='v^2-2vu')
+
+        builder.add_elementwise('element_wise-u^2', mode='MULTIPLY',
+                                input_names=[feature_layer, feature_layer],
+                                output_name='element_wise-u^2')
```
gustavla: Can also be done with `add_unary` and `alpha=2` to avoid an unnecessary edge in the graph.

gustavla:
Looks great so far, nice work!

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -409,3 +409,134 @@ def similarity_graph(self, k=5, radius=None, include_self_edges=False,
...
+            self.feature_extractor.feature_layer,
+            self.feature_extractor.context,
+            self.input_image_shape
+        )
+        batch_input_shape = (1, ) + self.input_image_shape
+        _mxnet_converter.convert(mx_feature_extractor, mode=None,
+                                 input_shape={input_name: batch_input_shape},
+                                 builder=builder, verbose=False)
+
+        # To add the nearest neighbors model we add calculation of the euclidean
```
TobyRoseman: Great! This code is much easier to follow.

vpndemand:
From iOS App, need to give one person image and model need to find the similar person images.I trained some human images in Image similarity and  exported to  coreML model.but it gives only output (distances between the input and reference images-Type(MultiArray (Double 20503) and it is not giving similar images.

TobyRoseman:
@vpndemand - please see [our user guide about exporting a image similarity model to Core ML](https://apple.github.io/turicreate/docs/userguide/image_similarity/export-coreml.html). You need to sort the images by distances and pick the `k` images with the smallest values. There is example code for that [in the user guide](https://apple.github.io/turicreate/docs/userguide/image_similarity/export-coreml.html).

znation:
@TobyRoseman Should we consider using `MLCustomModel` to get the images encoded into the model itself, for an (optional) exported model interface that results in image output directly from the model? Or does that not quite fit the user scenario here?

TobyRoseman:
@vpndemand - there is currently no way to get images from the Core ML model. You will need to store the images used to create the model in your app. The Core ML model just returns the distances from your input image to all of the images used to create the model. The n-th index in the array returned by the Core ML Model corresponds to the distance from your input image to the n-th image passed to `tc.image_similarity.create(...)`. 

Please create an issue, if you have further questions.

@znation - I think using an `MLCustomModel` here is a great idea!

znation:
I opened https://github.com/apple/turicreate/issues/724 as a feature request to track returning images from the Core ML model directly.

@vpndemand Feel free to add any comments there if there is something you're looking for specifically; in the meantime, you should be able to use the index of each probability returned in Core ML to look up the corresponding image. You will need to copy the images (with indices) into your app, such that (for example) for index 3 into the probabilities returned by the Core ML model, you would return the image at index 3 in the training SFrame.

vpndemand:
@TobyRoseman @znation Thanks guys for the support.Also i've done Image classification for binary Images. (data = tc.image_analysis.load_images('PetImages', with_path=True))
How to do the same thing for multiple classification to load images from different folders for each sets.

znation:
Hi @vpndemand - since this is a pull request thread (that has already been merged a while back), please open new issues for each issue you are encountering separately, so we can separate out these discussions. I think you should open two new issues: one for your question about multi-class classification of images in folders, and one for your issue with black and white images. We'll address those separately, thanks!

vpndemand:
@znation Im doing thanks

PR #244, opened 2018-02-06
Build + test fixes to prepare for 4.1 release

znation:
* Remove unused spacy dependency in scenario tests
* Bump coremltools dependency to 0.8
* Use assertAlmostEqual to get better failure msg in test_perplexity
* Get scenario tests ready for Python 3 execution
* Expect wheels on Linux for scenario and unit tests
* Make wheels on Linux
* Rename make_egg -> make_wheel
* Bump version number to 4.1a1
* Use VIRTUALENV env variable in run_scenario_tests

### DIFF
--- a/src/unity/python/dummy.c
+++ b/src/unity/python/dummy.c
@@ -1,0 +1,8 @@
+#include <stdio.h>
+
+int main()
+{
+      return 0;
+}
+
+void initdummy() {}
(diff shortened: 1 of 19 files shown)


src/unity/python/dummy.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
```
TobyRoseman: Why is this file needed?

znation: Probably isn't. I'll remove.

PR #245, opened 2018-02-06
Fixes a doc bug in groupby

### LINKED ISSUE #120: SFrame groupby should document that None values are skipped
Feedback from #116 suggests that we should clarify that all None values are skipped by the aggregators.

srikris:
Added some more details to fix #120

### DIFF
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -3824,6 +3824,16 @@ class SFrame(object):
         --------
         aggregate
 
+        Notes
+        -----
+        * Numeric aggregator (such as sum, mean, stdev etc.) follow the skip
+        None policy i.e they will omit all missing values from the aggregation.
+        As an example, `sum([None, 5, 10]) = 15` because the `None` values is
+        skipped.
+        * Aggregators have a default value when no values (after skipping all
+        `None` values) are present. For example, `sum([]) = 0`, `mean([]) =
+        None`, and `count([]) = 0`.
+
         Examples
         --------
         Suppose we have an SFrame with movie ratings by many users.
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3824,6 +3824,16 @@ def groupby(self, key_column_names, operations, *args):
         --------
         aggregate
 
+        Notes
+        -----
+        * Numeric aggregator (such as sum, mean, stdev etc.) follow the skip
```
znation: `s/Numeric aggregator/Numeric aggregators/`

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3824,6 +3824,16 @@ def groupby(self, key_column_names, operations, *args):
...
+        None policy i.e they will omit all missing values from the aggregation.
+        As an example, `sum([None, 5, 10]) = 15` because the `None` values is
```
znation:
```
s/`None` values/`None` value/
```

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3824,6 +3824,16 @@ def groupby(self, key_column_names, operations, *args):
...
+        skipped.
+        * Aggregators have a default value when no values (after skipping all
```
znation: What are the default values for the other aggregators? I think it would be useful to list them comprehensively here.

srikris: Fixed.

src/unity/python/turicreate/data_structures/sframe.py
srikris: Fixed.

src/unity/python/turicreate/data_structures/sframe.py
srikris: Fixed.

PR #247, opened 2018-02-07
Fix some Unicode string cases in Python 2

### LINKED ISSUE #157: Unable to create SFrame with non-ascii column name in Python 2
I have been investigating unicode support in Python 2 and 3 as part of #110 and SFrame currently has a mostly consistent behavior, except for one thing that stands out as a bug: Creation of non-ascii column names in Python 2.

Python 2 supports putting both `bytes` (`str` in Python 2) and `unicode` into the SFrame and SArray. It is mostly accepted and printed as a UTF-8-encoded string and returned as UTF-8-encoded bytes objects (`str` in Python 2). However, it fails when it is a column name and entered as a dictionary through the constructor:

```python
>>> tc.SFrame({u'\u00aa': [1, 2, 3]})
  File ".../turicreate/data_structures/sframe.py", line 833, in __init__
    raise ValueError('Unknown input type: ' + format)
  File ".../turicreate/cython/context.py", line 49, in __exit__
    raise exc_type(exc_value)
TypeError: function takes exactly 5 arguments (1 given)
```

It's the same result for unicode string as for a UTF-8-encoded bytes string. It also fails for a bytes object that is not valid UTF-8 (e.g. `b'\xaa'`). It is only a problem when using the `SFrame` constructor, and there are several ways to get a non-ascii string in there:

```python
>>> sf = tc.SFrame()
>>> sf[u'\u00aa'.encode('utf-8')] = [1, 2, 3]
```

This error does not happen on Python 3 and the resulting SFrame can be saved from Python 3 and correctly loaded into Python 2.

znation:
* Assumes utf-8 everywhere (Python 3 already makes this assumption).
* Allow unicode column accessors in SFrame.
* Allow unicode column name in construction of SFrame.
* Allow unicode column_name_prefix in str_to_datetime and unpack.

Fixes #157

### DIFF
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -751,6 +751,8 @@ class SFrame(object):
         else:
             self.__proxy__ = UnitySFrameProxy()
             _format = None
+            if six.PY2 and isinstance(data, unicode):
+                data = data.encode('utf-8')
             if (format == 'auto'):
                 if (HAS_PANDAS and isinstance(data, pandas.DataFrame)):
                     _format = 'dataframe'
@@ -3080,16 +3082,11 @@ class SFrame(object):
         """
         if not _is_non_string_iterable(column_names):
             raise TypeError("column_names must be an iterable")
-        if not (all([isinstance(x, str) or isinstance(x, type) or isinstance(x, bytes)
+        if not (all([isinstance(x, six.string_types) or isinstance(x, type) or isinstance(x, bytes)
                      for x in column_names])):
-            raise TypeError("Invalid key type: must be str, bytes or type")
+            raise TypeError("Invalid key type: must be str, unicode, bytes or type")
 
-        column_names_set = set(self.column_names())
-        # quick validation to make sure all selected string columns exist
-        requested_str_columns = [s for s in column_names if isinstance(s, str)]
-        for i in requested_str_columns:
-            if i not in column_names_set:
-                raise RuntimeError("Column name " +  i + " does not exist")
+        requested_str_columns = [s for s in column_names if isinstance(s, six.string_types)]
 
         # Make sure there are no duplicates keys
         from collections import Counter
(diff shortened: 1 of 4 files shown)


src/unity/python/turicreate/test/test_unicode_strings.py
```diff
@@ -0,0 +1,45 @@
...
+from __future__ import absolute_import as _
+
+import six
+import unittest
+import turicreate as tc
+
+class UnicodeStringTest(unittest.TestCase):
+
+    def test_unicode_column_accessor(self):
+        import turicreate as tc
```
TobyRoseman: You can remove this line.

src/unity/python/turicreate/test/test_unicode_strings.py
```diff
@@ -0,0 +1,45 @@
...
+        sf = tc.SFrame({'a': range(100)})
+        self.assertEqual(sf[u'a'][0], sf['a'][0])
+
+    def test_unicode_unpack_prefix(self):
+        import turicreate as tc
```
TobyRoseman: And this line

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3520,7 +3517,9 @@ def __getitem__(self, key):
         """
         if type(key) is SArray:
             return self._row_selector(key)
-        elif type(key) is str:
+        elif isinstance(key, six.string_types):
```
gustavla: Doesn't this also allow `bytes` access in Python 3? I am fine with dual bytes/unicode indexing in Python 2 since they coerce implicitly. However, in Python 3 (where they don't, i.e. `b'foo' != u'foo'`), dual indexing can cause confusion and would be inconsistent with Python 3 semantics.

Consider this situation in Python 3:

```python
sf = tc.SFrame({'foo': [{'bar': 100}]})
sf['foo'][0]  # returns Python dictionary
sf['foo'][0]['bar']  # Returns 100

sf[b'foo']  # This would work if we allow bytes indexing
sf[0][b'foo']  # We could allow this too if extended to SArray
sf[0][b'foo'][b'bar']  # This will never work, since we can't control the semantics of native Python types
```
I think it's better to let SFrame/SArray follow Python 3 semantics and not mix.

gustavla: Never mind, I misremembered what `six.string_types` gives you. This gives the desired semantics already.

PR #253, opened 2018-02-08
Image similarity export user guide

alonpal:


userguide/image_similarity/README.md
```diff
@@ -111,6 +111,68 @@ similar_images = similarity_graph.edges
...
+	do {
+		let model = try VNCoreMLModel(for: MyImageSimilarityModel().model)
+		
+		let request = VNCoreMLRequest(model: model, completionHandler: { [weak self] request, error in
+		    self?.processQuery(for: request, error: error)
+		})
+		request.imageCropAndScaleOption = .centerCrop
+		return request
+	} catch {
+            fatalError("Failed to load Vision ML model: \(error)")
```
TobyRoseman: This is error message that is prepended may not be the best. There are other thing that can fail in the code block besides loading the model, right?

userguide/image_similarity/README.md
```diff
@@ -111,6 +111,68 @@ similar_images = similarity_graph.edges
 [91440 rows x 4 columns]
 ```
 
+##### Deploying to Core ML
+
+With the Core ML framework, you can use the model to
```
TobyRoseman: How about this for the first sentence? `Using the Core ML Framework, you can retrieve the distance between a query image and all images in the reference data.`

userguide/image_similarity/README.md
```diff
@@ -111,6 +111,68 @@ similar_images = similarity_graph.edges
...
+   }
+}()
+```
+
+Notice that the image similarity model in Swift returns the distances between the query image and all images in the reference dataset. In order to retrieve the ids of the most similar images in the reference dataset, we must perform an argsort operation on the returned distances. We can do this by modifying the [Core ML sample application
+](https://developer.apple.com/documentation/vision/classifying_images_with_vision_and_core_ml) as such:
+
+```swift
+func processQuery(for request: VNRequest, error: Error?) {
+        DispatchQueue.main.async {
```
znation: All of the code inside the function appears to be overly indented. The final closing brace should have 0 spaces in front of it.

userguide/image_similarity/README.md
```diff
@@ -111,6 +111,68 @@ similar_images = similarity_graph.edges
...
+		})
+		request.imageCropAndScaleOption = .centerCrop
+		return request
+	} catch {
+            fatalError("Failed to load Vision ML model: \(error)")
+   }
+}()
+```
+
+Notice that the image similarity model in Swift returns the distances between the query image and all images in the reference dataset. In order to retrieve the ids of the most similar images in the reference dataset, we must perform an argsort operation on the returned distances. We can do this by modifying the [Core ML sample application
```
TobyRoseman: I'm not sure I've ever heard of an `argsort operation` before. I would just say `In order to retrieve the most similar images in the reference dataset, must sort by distance.`

znation: Instead of `in Swift` let's say `in Core ML`. The Core ML framework can be used from multiple languages including Swift.

Also, what is an `argsort` operation? I've never heard of this before. Is there a definition we can link to somewhere?

userguide/image_similarity/README.md
```diff
@@ -111,6 +111,68 @@ similar_images = similarity_graph.edges
...
+](https://developer.apple.com/documentation/vision/classifying_images_with_vision_and_core_ml) as such:
+
+```swift
+func processQuery(for request: VNRequest, error: Error?) {
```
TobyRoseman: On the python side, there is a `k` values (which defaults to 5). It only returns the top k results. I might be good to add that here as well.

userguide/image_similarity/README.md
alonpal: Removed and made consistent with all other image toolkits.

userguide/image_similarity/README.md
alonpal: I guess [argsort](https://docs.scipy.org/doc/numpy/reference/generated/numpy.argsort.html) isn't as common as I thought.

znation:
@TobyRoseman I signed off on this after changes from @alonpal. Does it look good to you as well?

TobyRoseman:
Looks great to me.

PR #258, opened 2018-02-08
Make sure distributions contain README.rst

gustavla:
Right now, the README.rst does not show up on PyPI. This is addressed in
two ways (not sure if both are necessary, but it seems to be common
practice to do both):

- Add to manifest file
- Add to `long_description` in setup.py

### DIFF
--- a/src/unity/python/setup.py
+++ b/src/unity/python/setup.py
@@ -171,7 +171,7 @@ if __name__ == '__main__':
                      "*.demo", "*.demo.*", "demo.*", "demo", "*.demo", "*.demo.*", "demo.*", "demo"]),
         url='https://github.com/apple/turicreate',
         license='LICENSE.txt',
-        description='Turi Create enables developers and data scientists to apply machine learning to build state of the art data products.',
+        description='Turi Create simplifies the development of custom machine learning models.',
         long_description=long_description,
         classifiers=classifiers,
         install_requires=[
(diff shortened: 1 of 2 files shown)


src/unity/python/setup.py
```diff
@@ -169,6 +172,7 @@ def run(self):
         url='https://github.com/apple/turicreate',
         license='LICENSE.txt',
         description='Turi Create enables developers and data scientists to apply machine learning to build state of the art data products.',
```
znation: While we're at it, let's update `description` too. I'm not sure `data products` is the right wording for what we are enabling. Maybe let's take the first sentence or two from the README.

gustavla: Good idea! If we use both sentences, is "in your app" too vague? It might make sense to identify it as particularly useful for deployment on Apple devices.

znation: Maybe let's just go with the same short description that we use on the GitHub repo: "Turi Create simplifies the development of custom machine learning models."

znation:
This seems to have broken in Python 3:
```
+ VERSION_NUMBER=4.1.1
+ /builds/turi/turicreate-build/scripts/../deps/env/bin/python setup.py bdist_wheel
  File "setup.py", line 127, in <module>
    long_description = f.read()
  File "/builds/turi/turicreate-build/scripts/../deps/env/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 106: ordinal not in range(128)
```

gustavla:
Sigh. Apparently that README has unicode apostrophes. Easiest fix is to just switch to ascii apostrophes. Not sure why this is failing inside `read` though.

PR #276, opened 2018-02-13
Temporary speed up large strings in categorical spec

abhishekpratapa:
Added a temporary fix of truncation of large strings in categorical spec for speed up in categorical visualization.


src/unity/lib/visualization/item_frequency.cpp
```diff
@@ -105,7 +106,13 @@ std::string item_frequency_result::vega_column_data(bool sframe) const {
     size_t count = pair.second.get<flex_int>();
 
     ss << "{\"label\": ";
-    ss << escape_string(value);
+
+    if(value.length() >= 200){
+      ss << escape_string(value).substr(0, 199) << i << "\"";
```
znation: What's with the `<< i << "\""`?

znation: I think you should do `escape_string(value.substr(0,199))` instead.

abhishekpratapa: ```<< i << "\""``` is for, theoretically say that we have a really long string vector like this:

```
[
    "hellowhatsupmarie" : 400,
    "hellowhatsupadam" : 300,
    "hellowhatsuptom" : 200
] 
```
And we truncate everything to "hellowhats" because that's our theoretical cutoff, then:

```
[
    "hellowhats" : 400,
    "hellowhats" : 300,
    "hellowhats" : 200
] 
```

would be our output and technically we are 2x-ing the theoretical string shown to vega, in this instance only "hello..." would be shown in most cases. To differentiate between duplicated string values caused by truncation I added the `<< i << "\"";` to take care of this situation

znation: In that case, I would do `escape_string(value.substr(0,199) + std::to_string(i))` to get the quotes/escaping from `escape_string`. You shouldn't need to manually add a quote mark onto the end.

PR #277, opened 2018-02-13
Allow CUDA GPU to be used if available

### LINKED ISSUE #260: Document which platforms turicreate is able to support GPU acceleration.
It's helpful that `turicreate` communicates at runtime that ["Using GPUs is currently not supported on Mac"][runtime error]. However, runtime feedback can be a little late in a setup process (e.g. after working through how to install MXNet from source with CUDA support on a Mac).

Could this limitation be mentioned in the [GPU documentation]? It may help some Mac enthusiasts with NVIDIA cards find this out early and avoid going down the road of trying to setup GPU support for  `turicreate` on their Mac.

[runtime error]: https://github.com/apple/turicreate/blob/v4.1/src/unity/python/turicreate/toolkits/_mxnet_utils.py#L85
[GPU documentation]: https://github.com/apple/turicreate/blob/master/README.md#gpu-support

gustavla:
Right now, a Mac with a CUDA card is unnecessarily blocked.

Fixes #260

### DIFF
--- a/src/unity/python/turicreate/toolkits/_mxnet_utils.py
+++ b/src/unity/python/turicreate/toolkits/_mxnet_utils.py
@@ -80,8 +80,9 @@ def get_num_gpus_in_use(max_devices=None):
     return len(gpu_ctx)
 
 def assert_valid_num_gpus():
+    from turicreate.util import _CUDA_GPU_IDS
     num_gpus = _tc_config.get_num_gpus()
-    if _sys.platform == 'darwin' and num_gpus > 0:
+    if not _CUDA_GPU_IDS and _sys.platform == 'darwin' and num_gpus > 0:
         raise _ToolkitError('Using GPUs is currently not supported on Mac')
     _numeric_param_check_range('num_gpus', num_gpus, -1, _six.MAXSIZE)
 


src/unity/python/turicreate/toolkits/_mxnet_utils.py
```diff
@@ -80,8 +80,9 @@ def get_num_gpus_in_use(max_devices=None):
     return len(gpu_ctx)
 
 def assert_valid_num_gpus():
+    from turicreate.util import _CUDA_GPU_IDS
     num_gpus = _tc_config.get_num_gpus()
-    if _sys.platform == 'darwin' and num_gpus > 0:
+    if not _CUDA_GPU_IDS and _sys.platform == 'darwin' and num_gpus > 0:
         raise _ToolkitError('Using GPUs is currently not supported on Mac')
```
igiloh: You should probably refine the error message as well

gustavla: I thought about it. However, since you only see this message if you don't have a CUDA card, I kept it as is so as not to complicate it for the average user.

PR #279, opened 2018-02-14
Escaped the BOMs during a read_csv.

### LINKED ISSUE #227: csv header parsing problem with BOM UTF8
I'm using Python 2.7.13 with turicreate version 4.0 shown by "pip show turicreate".

To recreate the bug, 
1. Write character 'a' to "buggy.csv" in UTF-8 encoding.

2. Run the following code:
`import turicreate as tc
data1 = tc.SFrame('buggy.csv')
print data1['a']
`

Expected output:
[]

Actual output:
RuntimeError: Column name a does not exist.

I used a debugger and found that the corresponding column in data1 for 'a' is '\xef\xbb\xbfa', which means that the BOM is incorrectly parsed as header.

srikris:
Fixes #227


src/unity/python/turicreate/test/test_sframe.py
```diff
@@ -204,11 +204,29 @@ def test_auto_parse_csv(self):
...
+                content = f.read()
+            with open(csvfile.name, 'w') as f:
+                f.write(codecs.BOM_UTF8)
+                f.write(content)
 
+            sf = SFrame.read_csv(csvfile.name, header=True)
             self.assertEqual(sf.dtype, [float, int, str])
             self.__test_equal(sf, df)
 
+    def test_csv_with_BOM(self):
```
znation: It looks like the test names got switched; `test_csv_with_BOM` tests without a BOM, and `test_auto_parse_csv` tests with a BOM now. Was that intended?

srikris: Oh. That was a mistake. Fixing

src/sframe/parallel_csv_parser.cpp
```diff
@@ -28,6 +28,20 @@ namespace turi {
...
+ * Escape BOMs
+ */
+void skip_BOM(general_ifstream& fin) {
+   char fChar, sChar, tChar;
+   fChar = fin.get();
+   sChar = fin.get();
+   tChar = fin.get();
+   bool isBOM = ((fChar == (char)0xEF) && (sChar == (char)0xBB) && (tChar == (char)0xBF));
+   if (!isBOM) {
+       fin.seekg(0);
```
srikris: Apparently, I can't do this operation on a `tar.gz` file.

znation: Can you do something like seekg(-3) to go back 3 chars?

ylow: Hmm... the general_fstream does not support seek for compressed files. There is an fstream put back mechanism.

ylow: This might work http://www.cplusplus.com/reference/istream/istream/putback/

PR #290, opened 2018-02-19
Fix validation-set statistics saved to linear/logistic regression models

nickjong:
The changes for https://github.com/apple/turicreate/pull/274 were incomplete. They added validation-set fields, but populated them with recomputed training-set stats. The optimizer interface must explicitly support computing validation-set stats.

### DIFF
--- a/src/unity/toolkits/supervised_learning/linear_regression_opt_interface.cpp
+++ b/src/unity/toolkits/supervised_learning/linear_regression_opt_interface.cpp
@@ -236,10 +236,10 @@ void linear_regression_opt_interface::compute_first_order_statistics(const
 /**
  * Compute the second order statistics
 */
-void linear_regression_opt_interface::compute_second_order_statistics(const
-    DenseVector& point, DenseMatrix& hessian, DenseVector& gradient, double&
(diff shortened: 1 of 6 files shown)


src/unity/toolkits/supervised_learning/linear_regression_opt_interface.cpp
```diff
@@ -236,10 +236,10 @@ void linear_regression_opt_interface::compute_first_order_statistics(const
 /**
  * Compute the second order statistics
 */
-void linear_regression_opt_interface::compute_second_order_statistics(const
-    DenseVector& point, DenseMatrix& hessian, DenseVector& gradient, double&
-    function_value) {
-  
+void linear_regression_opt_interface::compute_second_order_statistics(
+    const ml_data& data, const DenseVector& point, DenseMatrix& hessian,
```
nickjong: Note that this `data` parameter shadows the class's `data` member. That's why there are no changes to the function itself. (We could change the parameter name, but honestly I think the class data members should be renamed to begin with m_ or perhaps end with _. But I didn't want to do that in this PR.)

PR #293, opened 2018-02-20
Single dictionary support for OD.

### LINKED ISSUE #87: Consolidate external source drops

srikris:
Fixes #87. Object detectors can now take in an annotation column that's just a single dictionary (if its all just 1 objects)

### DIFF
--- a/src/unity/python/turicreate/test/test_object_detector.py
+++ b/src/unity/python/turicreate/test/test_object_detector.py
@@ -185,7 +185,6 @@ class ObjectDetectorTest(unittest.TestCase):
         annotated_img = tc.object_detector.util.draw_bounding_boxes(sf_copy[self.feature],
                 sf_copy[self.annotations])
 
-
     def test_invalid_num_gpus(self):
         num_gpus = tc.config.get_num_gpus()
         tc.config.set_num_gpus(-2)
--- a/src/unity/python/turicreate/test/test_object_detector.py
+++ b/src/unity/python/turicreate/test/test_object_detector.py
@@ -171,6 +171,21 @@ class ObjectDetectorTest(unittest.TestCase):
     def test_create_with_empty_dataset(self):
         tc.object_detector.create(self.sf[:0])
 
+    def test_dict_annotations(self):
+        sf_copy = self.sf[:]
+        sf_copy[self.annotations] = sf_copy[self.annotations].apply(lambda x: x[0] if len(x) > 0 else None)
+        dict_model = tc.object_detector.create(sf_copy,
+                          feature=self.feature,
+                          annotations=self.annotations,
+                          max_iterations=1,
+                          model=self.pre_trained_model)
+
+        pred = dict_model.predict(sf_copy)
+        metrics = dict_model.evaluate(sf_copy)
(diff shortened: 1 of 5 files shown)


src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
```diff
@@ -166,6 +170,9 @@ def is_valid(ann):
                     return ok_optional
 
                 # Unchanged boxes, for evaluation
+                if not hasattr(label, '__iter__'):
```
TobyRoseman: Probably want to remove this before you check it in.

src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
```diff
@@ -149,6 +149,10 @@ def _next(self):
             oshape = orig_image.shape
             if self.load_labels:
                 label = row[self.annotations_column]
+                if label == None:
+                    label = []
+                if type(label) == dict:
```
gustavla: How about an `elif` here? Whenever I see if statements like this I assume both can occur and waste a few cycles realizing they can't.

src/unity/python/turicreate/toolkits/object_detector/util/_visualization.py
```diff
@@ -120,6 +120,10 @@ def draw_bounding_boxes(images, annotations, confidence_threshold=0):
     def draw_single_image(row):
         image = row['image']
         anns = row['annotations']
+        if anns == None:
+            anns = []
+        if type(anns) == dict:
+            anns = [anns]
```
gustavla: Same here, `elif`

src/unity/python/turicreate/test/test_object_detector.py
```diff
@@ -171,6 +171,21 @@ def test_create_with_invalid_annotations_not_dict(self):
...
+                          annotations=self.annotations,
+                          max_iterations=1,
+                          model=self.pre_trained_model)
+
+        pred = dict_model.predict(sf_copy)
+        metrics = dict_model.evaluate(sf_copy)
+        annotated_img = tc.object_detector.util.draw_bounding_boxes(sf_copy[self.feature],
+                sf_copy[self.annotations])
+
+
```
gustavla: Remove one of these empty lines. /The PEP8 Police

src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
srikris: Done

src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
srikris: Done.

src/unity/python/turicreate/toolkits/object_detector/util/_visualization.py
srikris: Done

PR #294, opened 2018-02-20
Viz/plot implementation

abhishekpratapa:
- Added functionality for plot object in TuriCreate
- Added Plot object to show methods of TuriCreate


src/unity/lib/visualization/CMakeLists.txt
```diff
@@ -14,10 +14,13 @@ make_library(
     process_wrapper.cpp
     scatter.cpp
     show.cpp
+    summary_view.cpp
     thread.cpp
     transformation.cpp
     vega_data.cpp
     vega_spec.cpp
+
```
znation: Remove this blank line.

src/unity/lib/visualization/CMakeLists.txt
```diff
@@ -14,10 +14,13 @@ make_library(
...
+    plot/plot.cpp
```
znation: Since there isn't anything else in the `plot/` directory, we probably don't need a `plot` directory yet. I'd move `plot.[h|c]pp` up into the parent directory.

src/unity/lib/visualization/plot/plot.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <unity/lib/visualization/transformation.hpp>
+#include <unity/lib/visualization/vega_data.hpp>
+#include <unity/lib/visualization/histogram.hpp>
+#include <unity/lib/visualization/item_frequency.hpp>
+#include <unity/lib/visualization/summary_view.hpp>
+
+namespace turi{
+  namespace visualization{
+    void Plot::show() {
+      using namespace turi;
```
znation: You don't need `using` for namespaces that you're already in. You can remove this and the next line.

src/unity/lib/visualization/plot/plot.cpp
```diff
@@ -0,0 +1,58 @@
...
+          if (self->transformer->eof()) {
+             break;
+          }
+        }
+      });
+    }
+
+    void Plot::materialize() {
+      do {
+        transformer.get()->get()->vega_column_data();
```
znation: You should be able to just do `transformer->get()`. The first get is to resolve the shared pointer to a raw pointer (which arrow syntax does automatically), and the vega_column_data is unnecessary here, I think.

src/unity/lib/visualization/plot/plot.hpp
```diff
@@ -0,0 +1,42 @@
...
+                                              vega_spec(vega_spec),
+                                              path_to_client(path_to_client),
+                                              size_array(size_array),
+                                              transformer(transformer){}
+
+        void show();
+        void materialize();
+        std::string get_spec();
+
+        std::string vega_spec;
```
znation: Class members should be prefixed with `m_` to denote that they are members, since their usage is otherwise syntactically indistinct from scope variables. That makes the code in the `.cpp` more readable.

src/unity/lib/visualization/plot/plot.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace turi {
+  namespace visualization {
+
+    class EXPORT Plot: public toolkit_class_base{
+      public:
+
+        Plot(){};
+
+        Plot(const std::string path_to_client, const std::string vega_spec, std::shared_ptr<transformation_base> transformer, double size_array):
```
znation: For readability, please put each argument on its own line.

znation: And you should probably take a const reference to std::strings, not just a const string.

src/unity/lib/visualization/summary_view.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  for (const auto& transformer : m_transformers) {
+    all_rows_processed += transformer->get_rows_processed();
+  }
+
+  return flex_int(all_rows_processed);
+}
+
+size_t summary_view_transformation::get_batch_size() const {
+  // all have the same batch size and same number of rows processed.
```
znation: I don't think this comment is still true.

src/unity/lib/visualization/summary_view.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+bool summary_view_transformation::eof() const {
+  // all have the same batch size and same number of rows processed.
+  // we also guaranteed in the constructor that there is at least 1 transformer.
+  // we need only ask the 1 we know of whether it's eof.
+  return m_transformers[(m_transformers.size()-1)]->eof();
+}
+
+flex_int summary_view_transformation::get_rows_processed() const {
+  // all have the same batch size and same number of rows processed.
```
znation: I don't think this comment is still true.

src/unity/lib/visualization/summary_view.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  if(index >= column_names.size()){
+    index = 0;
+  }
+
+  return fused_out;
+}
+
+bool summary_view_transformation::eof() const {
+  // all have the same batch size and same number of rows processed.
```
znation: I don't think this comment is still true.

src/unity/lib/visualization/summary_view.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+    default:
+      throw std::runtime_error("Unexpected dtype. SFrame plot expects int, float or str dtypes.");
+  }
+  return ss.str();
+}
+
+summary_view_transformation::summary_view_transformation(const std::vector<std::shared_ptr<transformation_base>> transformers)
+  : m_transformers(transformers) {
+    // check some assumptions that make fuse work
```
znation: Let's make the terminology consistent here - it doesn't make sense that to get a summary view transformation, you "fuse" a "collection" of transformers. (Though that is what this code evolved from.)

Maybe we should have a summary_view_transformation take a collection of transformers in its constructor, instead of putting a method `fuse` on the collection itself?

src/unity/lib/visualization/summary_view.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "transformation.hpp"
+
+namespace turi {
+namespace visualization {
+
+class summary_view_transformation_output : public transformation_output {
+  private:
+    std::vector<std::shared_ptr<transformation_output>> m_outputs;
+  public:
+    std::vector<std::string> column_names;
```
znation: Please prefix class member names with `m_`. It's also generally not good practice to reveal data members directly from classes like that, but this code is fairly internal so it's fine for now.

src/unity/lib/visualization/summary_view.hpp
```diff
@@ -0,0 +1,47 @@
...
+    virtual std::string vega_column_data(bool sframe = false) const override;
+};
+
+class summary_view_transformation : public transformation_base {
+  private:
+    std::vector<std::shared_ptr<transformation_base>> m_transformers;
+    size_t index = 0;
+
+  public:
+    std::vector<std::string> column_names;
```
znation: Please name all class members with `m_`.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -0,0 +1,29 @@
+import time
+import json
+
+class Plot(object):
+    def __init__(self):
+        self.plot_object = tc.extensions._Plot()
```
znation: Where does `tc` come from here? Doesn't it need to be imported in this file?

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -0,0 +1,29 @@
+import time
+import json
+
+class Plot(object):
```
znation: I was imagining that this class would inherit from `tc.extensions._Plot`, instead of containing it as a member, but there may be reasons for that not to make sense. If it works better as a member that's fine.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -0,0 +1,29 @@
...
+        self.plot_object = tc.extensions._Plot()
+
+    def show(self):
+        self.plot_object.show()
+
+    def save_vega(self, filepath, include_data=True):
+        file_contents = self.plot_object.get_vega(include_data)
+
+        with open(filepath, 'w') as fp:
+            json.dump(file_contents, fp)
```
znation: Doesn't `get_vega` return an `std::string`? In that case you should just use `fp.write(file_contents)`.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def get_data(self, filepath, include_data=True):
+        return self.plot_object.get_data()
```
znation: Here, because `get_data` returns `std::string`, you should use `json.loads(...)`.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -0,0 +1,29 @@
+import time
```
znation: These imports will get exported publicly from this file - prefix them with underscores like `import time as _time` to prevent that.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -0,0 +1,29 @@
...
+    def get_vega(self, include_data=True):
+        return self.plot_object.get_vega(include_data)
+
+    def _repr_javascript_(self):
+        from IPython.core.display import display, HTML, clear_output, update_display
+
+        vega_spec = self.get_vega()["vega_spec"]
+
+        vega_html = '<html lang="en"><head><script src="https://cdnjs.cloudflare.com/ajax/libs/vega/3.0.8/vega.js"></script><script src="https://cdnjs.cloudflare.com/ajax/libs/vega-embed/3.0.0-rc7/vega-embed.js"></script></head><body><div id="vis"></div><script>var vega_json = '+json.dumps(json.dumps(vega_spec)).replace("'", "&apos;")+'; var vega_json_parsed = JSON.parse(vega_json); vegaEmbed("#vis", vega_json_parsed);</script></body></html>'
+        display(HTML('<html><body><iframe style="border:0;margin:0" width="'+str(width)+'" height="'+str(height)+'" srcdoc='+"'"+vega_html+"'"+' src="demo_iframe_srcdoc.htm"><p>Your browser does not support iframes.</p></iframe></body></html>'));
```
znation: Unnecessary trailing semicolon.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -0,0 +1,29 @@
...
+            json.dump(file_contents, fp)
+
+    def get_data(self, filepath, include_data=True):
+        return self.plot_object.get_data()
+
+    def get_vega(self, include_data=True):
+        return self.plot_object.get_vega(include_data)
+
+    def _repr_javascript_(self):
+        from IPython.core.display import display, HTML, clear_output, update_display
```
znation: The imports for `clear_output` and `update_display` don't seem to be used.

src/unity/python/turicreate/visualization/show.py
```diff
@@ -77,7 +77,7 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
 
     import turicreate as tc
 
-    path_to_client = get_client_app_path()
+    path_to_client = _get_client_app_path()
```
znation: Not sure when this broke; good catch.

src/unity/lib/visualization/plot/plot.cpp
abhishekpratapa: noted

src/unity/lib/visualization/CMakeLists.txt
abhishekpratapa: got it

src/unity/lib/visualization/CMakeLists.txt
abhishekpratapa: removed

src/unity/lib/visualization/plot/plot.cpp
abhishekpratapa: changed

src/unity/lib/visualization/plot/plot.hpp
abhishekpratapa: changed

src/unity/lib/visualization/summary_view.cpp
abhishekpratapa: changed comment to make sense

src/unity/lib/visualization/summary_view.cpp
abhishekpratapa: it isn't changed comment to reflect this

src/unity/lib/visualization/summary_view.cpp
abhishekpratapa: changed comment

src/unity/lib/visualization/summary_view.cpp
abhishekpratapa: Potentially, I'll change that in another PR if that's alright, this one is getting big

src/unity/lib/visualization/summary_view.hpp
abhishekpratapa: prefixed

src/unity/lib/visualization/summary_view.hpp
abhishekpratapa: prefixed

src/unity/python/turicreate/visualization/plot.py
abhishekpratapa: I changed the way we access Plot. It was getting tricky to integrate with the existing cython bindings we had for SFrame and SArray. If you think there's a better way of doing it. I'd be more than happy to change it!

src/unity/python/turicreate/visualization/plot.py
abhishekpratapa: get_vega returns a json string since, it's easier to include the data on the Python side, rather than shoving the data into the vega_spec on the C++ side. So I parse the vega, parse the data and stitch the two together on the front end.

src/unity/python/turicreate/visualization/plot.py
abhishekpratapa: Changed code

src/unity/python/turicreate/visualization/plot.py
abhishekpratapa: Noted changed the public imports to _imports

src/unity/python/turicreate/visualization/plot.py
abhishekpratapa: removed

src/unity/python/turicreate/visualization/plot.py
abhishekpratapa: removed imports

src/unity/python/turicreate/visualization/show.py
abhishekpratapa: thanks! but now fixed, yeay!

src/unity/lib/variant.cpp
```diff
@@ -10,6 +10,8 @@
 #include <unity/lib/api/unity_sarray_interface.hpp>
 #include <unity/lib/api/unity_graph_interface.hpp>
 #include <unity/lib/api/function_closure_info.hpp>
+#include <unity/lib/api/plot_interface.hpp>
```
znation: Making `Plot` part of `variant_type` seems like overkill. I thought we planned to use the toolkit SDK to expose the Plot object - what changed? Does the old-style Cython interface not allow returning types exposed through the toolkit SDK?

src/unity/lib/visualization/transformation.cpp
```diff
@@ -4,70 +4,15 @@
...
-size_t fused_transformation::get_batch_size() const {
-  // all have the same batch size and same number of rows processed.
-  // we also guaranteed in the constructor that there is at least 1 transformer.
-  // we need only ask the 1 we know of for rows_processed.
-  return m_transformers[0]->get_batch_size();
-}
-
-std::shared_ptr<fused_transformation> transformation_collection::fuse() {
-  return std::make_shared<fused_transformation>(*this);
+std::shared_ptr<summary_view_transformation> transformation_collection::fuse(std::vector<std::string> column_names, std::vector<flex_type_enum> column_types, size_t size) {
```
znation: Once again, I don't think we should have a `fuse` method on transformation_collection - rather, the constructor to `summary_view_transformation` should directly take in whatever parameters it needs. The caller should specify that they are constructing a `summary_view_transformation`, not fusing a transformation collection.

src/unity/lib/visualization/summary_view.cpp
```diff
@@ -54,12 +54,7 @@ std::shared_ptr<transformation_output> summary_view_transformation::get() {
   std::shared_ptr<transformation_output> output =  m_transformers[m_index] -> get();
   ret.push_back(output);
 
-  auto fused_out = std::make_shared<summary_view_transformation_output>(ret);
-
-  fused_out->m_column_names = m_column_names;
-  fused_out->m_column_types = m_column_types;
-  fused_out->m_size = m_size;
-  fused_out->m_index = m_index;
+  auto fused_out = std::make_shared<summary_view_transformation_output>(summary_view_transformation_output(ret, m_column_names, m_column_types, m_size, m_index));
```
znation: You don't need to call the `summary_view_transformation_output` constructor explicitly - that will make a copy and then invoke the copy constructor in make_shared. You should be able to just pass the parameters directly to make_shared.

PR #296, opened 2018-02-20
Propose fix typo

jeis2497052:
OK to fix this ?


TobyRoseman:
Thanks for the fix @jeis2497052

PR #297, opened 2018-02-20
Converts some ASSERT_MSG to more appropriately use log_and_throw

TobyRoseman:
Also some general clean ups.

### DIFF
--- a/src/sframe/sframe.hpp
+++ b/src/sframe/sframe.hpp
@@ -360,7 +360,9 @@ class sframe : public swriter_base<sframe_output_iterator> {
    * column id is out of range.
    */
   inline std::string column_name(size_t i) const {
-    ASSERT_MSG(i < index_info.column_names.size(), "Index out of range!");
+    if(i >= index_info.column_names.size()) {
+      log_and_throw("Index out of range!");
+    }
 
     return index_info.column_names[i];
   }
@@ -371,11 +373,15 @@ class sframe : public swriter_base<sframe_output_iterator> {
    */
   inline flex_type_enum column_type(size_t i) const {
     if (writing) {
-      ASSERT_MSG(i < group_writer->get_index_info().columns.size(), "Index out of range!");
+      if(i >= group_writer->get_index_info().columns.size()) {
+        log_and_throw("Index out of range!");
+      }
       return (flex_type_enum)
           (atoi(group_writer->get_index_info().columns[i].metadata["__type__"].c_str()));
     } else {
-      ASSERT_MSG(i < columns.size(), "Index out of range!");
+      if(i >= columns.size()) {
+        log_and_throw("Index out of range!");
+      }
       return columns[i]->get_type();
     }
   }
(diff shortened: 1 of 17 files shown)


src/sframe/sframe.hpp
```diff
@@ -360,7 +360,9 @@ class sframe : public swriter_base<sframe_output_iterator> {
    * column id is out of range.
    */
   inline std::string column_name(size_t i) const {
-    ASSERT_MSG(i < index_info.column_names.size(), "Index out of range!");
+    if(i >= index_info.column_names.size()) {
+      log_and_throw("Index out of range!");
```
hoytak: Since these error messages get propagated all the way to the user, and they may not have context for how it's called, we should add  a bit more description.  Maybe "Column index out of range."

src/unity/python/turicreate/connect/main.py
```diff
@@ -16,15 +16,12 @@
 import decorator
 import logging
 
-""" The module level logger object """
-__LOGGER__ = logging.getLogger(__name__)
```
hoytak: I don't think we should delete this one.  It could be helpful for interactive debugging.

src/unity/python/turicreate/connect/main.py
```diff
@@ -16,15 +16,12 @@
...
-
 """ Global variables """
 __UNITY_GLOBAL_PROXY__ = None
 __SERVER__ = None
 
-ENGINE_START_ERROR_MESSAGE = 'Cannot connect to SFrame engine. ' + \
-    'If you believe this to be a bug, check https://github.com/turi-code/SFrame/issues for known issues.'
+ENGINE_START_ERROR_MESSAGE = 'Cannot connect to turicreate engine. ' + \
```
hoytak: I don't think this is relevant any more. .  We can delete it.

src/unity/python/turicreate/connect/main.py
```diff
@@ -39,15 +36,18 @@ def __catch_and_log__(func, *args, **kargs):
 
 @__catch_and_log__
 def launch(server_log=None):
-    server = None
+    global __UNITY_GLOBAL_PROXY__
+    global __SERVER__
+
+    from ..extensions import _publish
```
hoytak: This should be after the set_log_progess and sever start, as importing this triggers a lot of logged code execution.

src/unity/python/turicreate/__init__.py
```diff
@@ -25,7 +25,7 @@
 from .data_structures.sarray_builder import SArrayBuilder
 from .data_structures.sframe_builder import SFrameBuilder
 
-from turicreate.data_structures.sgraph import load_sgraph, load_graph
```
hoytak: We need to be really careful deleting these.  The main reason for the deprecation of load_graph is that it's required to load some saved models.  Before submitting this PR, we should test that all our load model tests pass.

src/sframe/sframe.hpp
```diff
@@ -371,11 +373,15 @@ class sframe : public swriter_base<sframe_output_iterator> {
    */
   inline flex_type_enum column_type(size_t i) const {
     if (writing) {
-      ASSERT_MSG(i < group_writer->get_index_info().columns.size(), "Index out of range!");
+      if(i >= group_writer->get_index_info().columns.size()) {
+        log_and_throw("Index out of range!");
```
hoytak: Column index

TobyRoseman:
Based on suggestion by @hoytak, I'm closing this PR. I will submit seperate PRs for the clean ups and the ASSERT_MSG conversions.

Issue #302, opened 2018-02-24
Font used in Linux visualization looks very different than on macOS


znation:
See screenshot:
![screen shot 2018-02-23 at 4 37 28 pm](https://user-images.githubusercontent.com/5673251/36623276-117def4e-18b8-11e8-899d-7cb95d7e80fa.png)

We should probably pick a font that looks as close as possible.

Issue #301, opened 2018-02-24
Double scrollbars on Linux visualization


znation:
Repro steps:
```python
import turicreate as tc
sf = tc.SFrame.read_csv('https://docs-assets.developer.apple.com/turicreate/datasets/tc-clang-format-results.csv')
sf.show()
```

Expected:

One scrollbar on the right.

Actual:

Two scrollbars on the right. The inner one seems disabled. See screenshot.
![screen shot 2018-02-23 at 4 37 28 pm](https://user-images.githubusercontent.com/5673251/36623271-f9a7dc90-18b7-11e8-9d6e-9a286044223d.png)

PR #309, opened 2018-02-26
Added fallback font TuriCreate and fixed scroll bar issue in linux

abhishekpratapa:
- added fallback font for linux
- fixed scroll bar issue in linux


znation:
Fixes #302 
Fixes #301

PR #314, opened 2018-03-01
Updated Plot Api

abhishekpratapa:
- Made Plot api plumbing for the show method
- Updated .show() method on Jupyter Notebook
- Updated Tooltips on Jupyter Notebook
- Created doc string for public methods of Plot Object

### DIFF
--- a/src/unity/python/turicreate/visualization/plot.py
+++ b/src/unity/python/turicreate/visualization/plot.py
@@ -8,11 +8,11 @@ import json as _json
 
 class Plot(object):
     """
-    An immutable, object representative of a visualization.
+    An immutable object representation of a visualization.
 
     Notes
     -----
-    - A plot object is returned through a proxy via the SFrame's or SArray's .plot() method
+    - A plot object is returned via the SFrame's or SArray's .plot() method
 
     Examples
     --------
@@ -44,7 +44,7 @@ class Plot(object):
         Notes
         -----
         - In a Jupyter Notebook the .show() method displays an inline plot
-        - In any other environment, the .show() method displays a native GUI
+        - In any other environment, the .show() method launches a native GUI and displays the plot there
 
         Examples
         --------
@@ -62,7 +62,7 @@ class Plot(object):
                 display = True
             else:
                 pass
-        except:
+        except NameError:
             pass
         finally:
             if not display:
--- a/src/unity/python/turicreate/visualization/plot.py
+++ b/src/unity/python/turicreate/visualization/plot.py
@@ -60,14 +60,11 @@ class Plot(object):
             if(get_ipython().__class__.__name__ == "ZMQInteractiveShell"):
                 self._repr_javascript_()
                 display = True
-            else:
-                pass
         except NameError:
             pass
         finally:
             if not display:
                 import sys
-                import os
                 if sys.platform != 'darwin' and sys.platform != 'linux2':
                      raise NotImplementedError('Visualization is currently supported only on macOS and Linux.')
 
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/visualization/plot.py
```diff
@@ -7,22 +7,99 @@
 import json as _json
 
 class Plot(object):
+    """
+    An immutable, object representative of a visualization.
```
znation: s/representative/representation

znation: And you don't need a comma between immutable and object.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -7,22 +7,99 @@
...
 
+    Notes
+    -----
+    - A plot object is returned through a proxy via the SFrame's or SArray's .plot() method
```
znation: The `through a proxy` part doesn't seem relevant to users. You can just say a plot object is returned from the `plot` method on Turi Create data structures.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -7,22 +7,99 @@
...
-        import os
-        if sys.platform != 'darwin' and sys.platform != 'linux2':
-             raise NotImplementedError('Visualization is currently supported only on macOS and Linux.')
+        """
+        A method for displaying the Plot object
+
+        Notes
+        -----
+        - In a Jupyter Notebook the .show() method displays an inline plot
+        - In any other environment, the .show() method displays a native GUI
```
znation: How about `launches a native GUI and displays the plot there`?

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -7,22 +7,99 @@
...
-    def save_vega(self, filepath, include_data=True):
+        """
+        display = False
+        try:
+            if(get_ipython().__class__.__name__ == "ZMQInteractiveShell"):
+                self._repr_javascript_()
+                display = True
+            else:
+                pass
+        except:
```
znation: It would be better to catch the specific exception that happens when we're not in a Jupyter Notebook (and allow other exceptions to get raised from here).

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -7,22 +7,99 @@
...
+        >>> plt.show()
 
-    def save_vega(self, filepath, include_data=True):
+        """
+        display = False
+        try:
+            if(get_ipython().__class__.__name__ == "ZMQInteractiveShell"):
+                self._repr_javascript_()
+                display = True
+            else:
```
TobyRoseman: This else-block seems to have no purpose. So you probably want to delete it.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -7,22 +7,99 @@
...
+                pass
+        except NameError:
+            pass
```
TobyRoseman: Why is it ok to ignore this error?

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -7,22 +7,99 @@
...
+        finally:
+            if not display:
+                import sys
+                import os
```
TobyRoseman: Doesn't look like you're using this import.

src/unity/python/turicreate/visualization/plot.py
abhishekpratapa: Using exceptions for control flow

TobyRoseman: Sounds reasonable (I think). Please add a comment.

Issue #291, opened 2018-02-19
make_wheel.sh does not run cpp unit tests and error out on python unit tests


TobyRoseman:
This is with Python 2.7 on Linux.

Here is the relevant output:
```
================= Done Build Source ================


+ [[ -z '' ]]
+ [[ -z '' ]]
+ cpp_test
+ echo -e '\n\n\n================= Running Unit Test ================\n\n\n'



================= Running Unit Test ================



+ cd /home/toby/Documents/turicreate/scripts/../release
+ push_ld_library_path
+ OLD_LIBRARY_PATH=
+ export LD_LIBRARY_PATH=/home/toby/Documents/turicreate/scripts/../deps/local/lib:/home/toby/Documents/turicreate/scripts/../deps/local/lib64:
+ LD_LIBRARY_PATH=/home/toby/Documents/turicreate/scripts/../deps/local/lib:/home/toby/Documents/turicreate/scripts/../deps/local/lib64:
+ /home/toby/Documents/turicreate/scripts/../scripts/run_cpp_tests.py -j 1
Found 155 tests.
Found 0 files in cache.
Hashed 155 files in 165.550021887 seconds.
Ready to test 155 files.
No tests were found!!!
Test project /home/toby/Documents/turicreate/release

+ pop_ld_library_path
+ export LD_LIBRARY_PATH=
+ LD_LIBRARY_PATH=
+ unit_test
+ echo -e '\n\n\n================= Running Unit Test ================\n\n\n'



================= Running Unit Test ================



+ cd /home/toby/Documents/turicreate/scripts/..
+ scripts/run_python_test.sh release
+ [[ -z release ]]
+ BUILD_TYPE=release
+++ dirname scripts/run_python_test.sh
++ cd scripts
++ pwd
+ SCRIPT_DIR=/home/toby/Documents/turicreate/scripts
+ WORKSPACE=/home/toby/Documents/turicreate/scripts/..
+ unset PYTHONPATH
+ cd /home/toby/Documents/turicreate/scripts/..
+ scripts/install_python_toolchain.sh
++ set -e
++ [[ -z '' ]]
++ VIRTUALENV=virtualenv
++ PIP=pip3
++ PYTHON_FULL_NAME=m
++ [[ '' == \p\y\t\h\o\n\2\.\7 ]]
++ virtualenv /home/toby/Documents/turicreate/deps/env
New python executable in /home/toby/Documents/turicreate/deps/env/bin/python
[...]

PR #315, opened 2018-03-01
Remove unused stuff and make_wheel.py

TobyRoseman:
The last commit address #291 
The second and third commit were previously included and approved as part of #297

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -1,4 +1,4 @@
-coremltools
+coremltools==0.8
 scipy==0.19.1
 numpy==1.11.1
 cython==0.24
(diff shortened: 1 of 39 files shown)


scripts/requirements.txt
```diff
@@ -1,3 +1,4 @@
+coremltools
```
znation: I think we are trying not to duplicate requirements between `setup.py` and `requirements.txt` -- and coremltools is already in setup.py. Why is it needed here? (If all setup.py requirements are needed here, we should copy them all over.)

TobyRoseman: All the requirements from `setup.py` except this one are already in this file.

znation: I see. In that case, should we pin a version here so it doesn't drift from the setup.py version?

Issue #326, opened 2018-03-02
Missing value in Jupiter notebook visualization causes it to open in a new window and not inline


srikris:
```
import turicreate as tc
import numpy

a = test_str(100, 10)

def add_value(sa, value = None, num_values = 0):
    if num_values > 0:
        sa = sa.append(tc.SArray([value] * num_values, sa.dtype))
    return sa

def generate_str(size = 1000, num_none = 0, prefix = "foo_"):
    sa = tc.SArray(numpy.random.randint(3, 20, size = size))
    sa = add_value(sa, None, num_none)
    sa = prefix + sa.astype(str)
    return sa
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-9-003118562ce8> in <module>()
----> 1 a = test_str(100, 10)

/Volumes/Data/srikris/workspace/test/viz/viz_test.py in test_str(size, num_none, prefix)
     80     sa = generate_str(size, num_none, prefix)
     81     print("Showing")
---> 82     sa.show()
     83     return sa
     84 

/Volumes/Data/srikris/workspace/turicreate/debug/src/unity/python/turicreate/data_structures/sarray.py in show(self, title, xlabel, ylabel)
   2899         returned_plot = self.__plot(title, xlabel, ylabel)
   2900 
-> 2901         returned_plot.show()
   2902 
   2903     def __plot(self, title=None, xlabel=None, ylabel=None):

/Volumes/Data/srikris/workspace/turicreate/debug/src/unity/python/turicreate/visualization/plot.py in show(self)
     59         try:
     60             if(get_ipython().__class__.__name__ == "ZMQInteractiveShell"):
---> 61                 self._repr_javascript_()
     62                 display = True
     63         except NameError:

/Volumes/Data/srikris/workspace/turicreate/debug/src/unity/python/turicreate/visualization/plot.py in _repr_javascript_(self)
    129         from IPython.core.display import display, HTML
    130 
--> 131         vega_spec = self._get_vega(True)["vega_spec"]
    132 
    133         vega_html = '<html lang="en"> \

/Volumes/Data/srikris/workspace/turicreate/debug/src/unity/python/turicreate/visualization/plot.py in _get_vega(self, include_data)
    117         if(include_data):
    118             spec = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_spec'}))
[...]

PR #333, opened 2018-03-03
Updated tc.show to work in Jupyter Notebook

abhishekpratapa:
- updated tc.show to work in Jupyter Notebook


abhishekpratapa:
Fixed #326

PR #334, opened 2018-03-03
Fixed build failure in Python 3.6

srikris:


src/unity/python/turicreate/test/test_sframe.py
```diff
@@ -208,7 +208,7 @@ def test_auto_parse_csv_with_bom(self):
             with open(csvfile.name) as f:
                 content = f.read()
             with open(csvfile.name, 'w') as f:
-                f.write(codecs.BOM_UTF8)
+                f.write(str(codecs.BOM_UTF8))
```
gustavla: The BOM is not a string and should never be treated as such. This is not just by principle, but if you run `str(codecs.BOM_UTF8)` in Python 3, you'll notice this won't work.

I think this whole section needs to be done as:

```python
with open(csvfile.name, 'rb') as f:
    content = f.read()
with open(csvfile.name, 'wb') as f:
    f.write(codecs.BOM_UTF8)
    f.write(content)
```

srikris: You are right. Fixed.

PR #349, opened 2018-03-07
Implemented exact sample and random_split

### LINKED ISSUE #347: Exact random split fraction sizes
Right now, a random split of an SFrame is a coin toss for each sample, with the probability defined by `fraction`. However, if the dataset has a known size, it would be nice if the `fraction` actually corresponded to the final fraction (and not just in expectation).

```python
>>> sf = tc.SFrame({'x': range(1000)})
>>> train, test = sf1.random_split(0.5)
>>> len(train)
476
>>> len(test)
524
```

ylow:
Implemented the following APIs:

    SFrame.random_split(..., exact=True)
    SFrame.sample(..., exact=True)
    SArray.sample(..., eaxct=True)

The inexact sample method simply does:

    threshold = 2^64 * fraction
    sample = sf[[0...size].hash(seed) < threshold]

The exact sample method does the same, but has to do a bit more work:

    random_numbers = [0...size].hash(seed)
    threshold = random_numbers.sort()[fraction * size]
    sample = sf[random_numbers < threshold]

Really, we only need a partial sort instead of a sort, but we do not
have a convenient implementation of partial sort to use. So this
implementation currently is quite expensive.

Fixes #347


src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -2594,20 +2594,24 @@ def flat_map(self, column_names, fn, column_types='auto', seed=None):
...
         Sample a fraction of the current SFrame's rows.
 
         Parameters
         ----------
         fraction : float
-            Approximate fraction of the rows to fetch. Must be between 0 and 1.
-            The number of rows returned is approximately the fraction times the
-            number of rows.
+            Fraction of the rows to fetch. Must be between 0 and 1.
+            if exact is False (default), the number of rows returned is
```
gustavla: if -> If (capital "I")

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -2637,25 +2641,33 @@ def sample(self, fraction, seed=None):
...
+        An exact fraction partition can be optionally obtained by setting 
+        exact=True.
 
         Parameters
         ----------
         fraction : float
-            Approximate fraction of the rows to fetch for the first returned
-            SFrame. Must be between 0 and 1.
+            Fraction of the rows to fetch. Must be between 0 and 1.
+            if exact is False (default), the number of rows returned is
```
gustavla: Same here

PR #354, opened 2018-03-08
Fix ./configure on Python 2.7

znation:
Don't rely on PY_MAJOR_VERSION env variable (this is not set
consistently). Instead, we can check the PYTHON_VERSION variable, which
is set by `./configure` before `install_python_toolchain` runs.

### DIFF
--- a/scripts/install_python_toolchain.sh
+++ b/scripts/install_python_toolchain.sh
@@ -11,7 +11,7 @@ PIP=pip3
 # TODO - not sure why 'm' is necessary here (and not in 2.7)
 # note that PYTHON_VERSION includes the word "python", like "python2.7" or "python3.6"
 PYTHON_FULL_NAME=${PYTHON_VERSION}m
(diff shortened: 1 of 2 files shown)


TobyRoseman:
So it looks like this change will prevent the `make_wheel.sh` script from working. If someone runs `make_wheel.sh` without first having he python tool chain installed (i.e. with out running `configure` first), it will not work.

We should probably just have one, either `PY_MAJOR_VERSION` or `PYTHON_VERSION`, not both. That is what I would recommend for this pull request.

Long term we should probably think about how to simply things with respect to building. It's a little too complicated in my opinion.

znation:
Thanks @TobyRoseman, I agree with everything you said. I will rework this fix taking `make_wheel.sh` into account, with a longer term plan to simplify.

znation:
@TobyRoseman Please re-review. PY_MAJOR_VERSION has been eliminated.

PR #366, opened 2018-03-13
Add a setup.py for a source distribution

### LINKED ISSUE #27: Prevent installing on systems that are not x86_64
We don't have any checks for x86_64. We should have our wheels restrict to only the platforms that we support. Ideally with names like this: numpy-1.13.3-cp35-cp35m-manylinux1_x86_64.whl

znation:
We will use this one for erroring out on unsupported platforms, by
uploading it as an egg (source distribution). If a user's pip install
falls back to the egg (no compatible wheel) they will see this error
message.

* Fixes #27

### DIFF
--- a/src/unsupported_python/setup.py
+++ b/src/unsupported_python/setup.py
@@ -38,7 +38,6 @@ class InstallEngine(install):
         sys.exit(1)
 
 if __name__ == '__main__':
-    from distutils.util import get_platform
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Environment :: Console",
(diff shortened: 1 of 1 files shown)


gustavla:
Is the plan to add a wheel with this for the current 4.2 and not for any future releases? That makes sense to me, just double-checking that plans.

znation:
@gustavla Yes, that was my plan.

src/unsupported_python/setup.py
```diff
@@ -0,0 +1,73 @@
...
+        * Outdated pip version (try `pip install -U pip`).
+
+        ==================================================================================
+
+        """)
+        sys.stderr.write(msg)
+        sys.exit(1)
+
+if __name__ == '__main__':
+    from distutils.util import get_platform
```
gustavla: Why is this imported?

znation: It doesn't need to be. Good catch.

Issue #342, opened 2018-03-06
Single example predict methods for Vision toolkits


srikris:
- Image classification
- Object detection
- Image similarity

These 3 above should double down on a single example predict call

PR #372, opened 2018-03-13
Single-sample input gives single-sample results

gustavla:
- Adds single-sample input to object detector
- Changes single-sample input for classification to unpack the return

Addresses #342.

Previous classifier behavior:
```python
>>> cc.predict(image)
dtype: str
Rows: 1
['bike']
```
New behavior:
```python
>>> cc.predict(image)
'bike'
```
Since this breaks current behavior, this should not be merged yet.

### DIFF
--- a/src/unity/python/turicreate/test/test_object_detector.py
+++ b/src/unity/python/turicreate/test/test_object_detector.py
@@ -210,6 +210,18 @@ class ObjectDetectorTest(unittest.TestCase):
         pred0 = self.model.predict(sf[:0])
         self.assertEqual(len(pred0), 0)
 
+    def test_single_image(self):
+        # Predict should work on a single image and product a list of dictionaries
+        # (we set confidene threshold to 0 to ensure predictions are returned)
+        pred = self.model.predict(self.sf[self.feature][0], confidence_threshold=0)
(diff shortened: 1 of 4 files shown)


src/unity/python/turicreate/test/test_object_detector.py
```diff
@@ -220,6 +220,18 @@ def test_predict(self):
         pred0 = self.model.predict(sf[:0])
         self.assertEqual(len(pred0), 0)
 
+    def test_single_image(self):
```
TobyRoseman: Could you also update `test_single_image` in `test_image_classifier.py`?

gustavla: Good idea. I added some checks of the return type.

I also pushed some updates to the docstrings to reflect these changes.

znation:
@gustavla Please rebase against latest master, thanks!

znation:
@gustavla Should this change still go in? It needs to be rebased.

gustavla:
It is now rebased with some fixes to broken tests.

gustavla:
I also ran all relevant tests (OD/IC/IS) and they pass.

PR #373, opened 2018-03-13
Avoiding default atof behavior for integers and floats

shantanuchhabra:
Redirected from unity_sarray to go to flex_int and flex_float and added logic in flexible_type_detail to deal with invalid strings like "2hello", "INFiltrate", etc. Built and tested locally.

### DIFF
--- a/src/flexible_type/flexible_type_detail.hpp
+++ b/src/flexible_type/flexible_type_detail.hpp
@@ -540,7 +540,7 @@ struct get_int_visitor {
     long long int converted = std::strtoll(t.c_str(), &end, 10); 
     if (*end != '\0') {
       // was not at end of element so throw error
-      throw std::runtime_error("Invalid conversion: " + t + "cannot be interpreted as an integer");
+      throw std::runtime_error("Invalid conversion: " + t + " cannot be interpreted as an integer");
     } else {
       return converted; 
     }
@@ -566,7 +566,7 @@ struct get_float_visitor {
     double converted = std::strtod(t.c_str(), &end); 
     if (*end != '\0') {
       // was not at end of element so throw error
-      throw std::runtime_error("Invalid conversion: " + t + "cannot be interpreted as a float");
+      throw std::runtime_error("Invalid conversion: " + t + " cannot be interpreted as a float");
     } else {
       return (float)converted; 
     }
--- a/src/flexible_type/flexible_type_detail.hpp
+++ b/src/flexible_type/flexible_type_detail.hpp
@@ -535,7 +535,16 @@ struct get_int_visitor {
     return dt.posix_timestamp();
   }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(flex_float i) const {return i; }
-  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { return std::atoll(t.c_str()); }
+  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { 
+    char *end;
+    long long int converted = std::strtoll(t.c_str(), &end, 10); 
+    if (**end != '\0') {
+      // was not at end of element so throw error
+      throw std::runtime_error("Invalid conversion: String contains more characters than just float");
+    } else {
+      return converted; 
+    }
+  }
 };
 
 /**
@@ -552,7 +561,16 @@ struct get_float_visitor {
   }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(flex_int i) const {return i; }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(flex_float i) const {return i; }
-  inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(const flex_string& t) const { return std::atof(t.c_str()); }
+  inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(const flex_string& t) const { 
+    char *end;
+    double converted = std::strtod(t.c_str(), &end); 
+    if (**end != '\0') {
+      // was not at end of element so throw error
+      throw std::runtime_error("Invalid conversion: String contains more characters than just float");
+    } else {
+      return (float)converted; 
+    }
+  }
 };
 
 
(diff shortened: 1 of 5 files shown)


src/flexible_type/flexible_type_detail.hpp
```diff
@@ -535,7 +535,16 @@ struct get_int_visitor {
     return dt.posix_timestamp();
   }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(flex_float i) const {return i; }
-  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { return std::atoll(t.c_str()); }
+  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { 
+    char *end;
+    long long int converted = std::strtoll(t.c_str(), &end, 10); 
+    if (*end != '\0') {
+      // was not at end of element so throw error
+      throw std::runtime_error("Invalid conversion: String contains more characters than just integer");
```
znation: Let's put the value into the error message, so the user has context on why it failed:
```cpp
"Invalid conversion: \"" + t + "\" cannot be interpreted as an integer."
```

Same for the float conversion below.

shantanuchhabra: That's a great idea, on it!

TobyRoseman:
@shantanuchhabra - a unit test would be great. I would recommend creating a simple unit test. Verify that it fails without your change. Then verify that it passes with your change.

shantanuchhabra:
Changed the error message and also added a unit test!

src/flexible_type/flexible_type_detail.hpp
```diff
@@ -535,7 +535,16 @@ struct get_int_visitor {
     return dt.posix_timestamp();
   }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(flex_float i) const {return i; }
-  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { return std::atoll(t.c_str()); }
+  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { 
+    char *end;
+    long long int converted = std::strtoll(t.c_str(), &end, 10); 
+    if (*end != '\0') {
+      // was not at end of element so throw error
+      throw std::runtime_error("Invalid conversion: " + t + "cannot be interpreted as an integer");
```
TobyRoseman: probably want a space before "cannot"

src/flexible_type/flexible_type_detail.hpp
```diff
@@ -552,7 +561,16 @@ struct get_float_visitor {
   }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(flex_int i) const {return i; }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(flex_float i) const {return i; }
-  inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(const flex_string& t) const { return std::atof(t.c_str()); }
+  inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(const flex_string& t) const { 
+    char *end;
+    double converted = std::strtod(t.c_str(), &end); 
+    if (*end != '\0') {
+      // was not at end of element so throw error
+      throw std::runtime_error("Invalid conversion: " + t + "cannot be interpreted as a float");
```
TobyRoseman: same as previous comment

src/flexible_type/flexible_type_detail.hpp
```diff
@@ -535,7 +535,16 @@ struct get_int_visitor {
     return dt.posix_timestamp();
   }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(flex_float i) const {return i; }
-  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { return std::atoll(t.c_str()); }
+  inline FLEX_ALWAYS_INLINE_FLATTEN flex_int operator()(const flex_string& t) const { 
+    char *end;
+    long long int converted = std::strtoll(t.c_str(), &end, 10); 
+    if (*end != '\0') {
```
ylow: I rather this not use (*end) == '\0' as a comparison. The null character could occur anywhere.
Use string.data(), and check for position of the pointer instead of comparing the value at the pointer.

src/flexible_type/flexible_type_detail.hpp
```diff
@@ -552,7 +561,16 @@ struct get_float_visitor {
   }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(flex_int i) const {return i; }
   inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(flex_float i) const {return i; }
-  inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(const flex_string& t) const { return std::atof(t.c_str()); }
+  inline FLEX_ALWAYS_INLINE_FLATTEN flex_float operator()(const flex_string& t) const { 
+    char *end;
+    double converted = std::strtod(t.c_str(), &end);
```
ylow: Ditto

src/unity/lib/unity_sarray.cpp
```diff
@@ -1300,9 +1300,9 @@ std::shared_ptr<unity_sarray_base> unity_sarray::lazy_astype(flex_type_enum dtyp
       flexible_type ret;
       try {
         if (dtype == flex_type_enum::INTEGER) {
-          ret = std::stoll(f.get<flex_string>());
+          f.to<flex_int>();
```
ylow: to<...> returns. It does not cast inplace. See the flexible_type documentation.

ylow:
Hi, sorry. My review came late. Can we revert?

src/unity/lib/unity_sarray.cpp
znation: Of course - not sure how I missed that. How does this pass unit tests?

znation:
@shantanuchhabra Let's address @ylow's comments. You can either start fresh (I'll git revert this PR change) or you can put up a 2nd PR to address the comments - let me know which you'd prefer.

shantanuchhabra:
I'll create a second PR!

shantanuchhabra:
Please see #376 :)

Issue #355, opened 2018-03-09
NDArray: Problem creating SFrame when OWNDATA is False


gustavla:
I can't seem to put an ndarray into an SFrame where the `OWNDATA` flag is `False`:

```ipython
In [58]: x = np.array([1, 2, 3])

In [59]: x.flags
Out[59]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

In [60]: tc.SFrame({'x': [x]})
Out[60]: 
Columns:
	x	ndarray

Rows: 1

Data:
+------------+
|     x      |
+------------+
| [1. 2. 3.] |
+------------+
[1 rows x 1 columns]

In [61]: y = x[:]

In [62]: y.flags
Out[62]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

In [63]: tc.SFrame({'y': [y]})
[...]
TypeError: Could not convert python object with type ndarray('int64') to flexible_type.
```

ylow:
I think this issue is with type interpretation inconsistencies when OWNDATA is false. The following works:
```
x = np.array([1, 2, 3],dtype='d')
y = x[:]
tc.SFrame({'y': [y]})
```

gustavla:
Actually, that example also fails, but I realized now it only happens on Python 3.

gustavla:
Actually, #385 fixes this issue.

PR #386, opened 2018-03-21
Avoid ndarray to fix fatal error in activity classifier

gustavla:
This only affects Python 3. This causes failure in many of the activity classifier tests.


gustavla:
Actually, this is the issue that is causing this: https://github.com/apple/turicreate/issues/355

gustavla:
... which is fixed by #385, so once that is merged this fix is no longer needed. Closing this.

PR #387, opened 2018-03-21
Added support for file:// as valid url protocol.

hoytak:
Currently, file:// is not recognized as a valid protocol for opening file urls, even though it's widely available elsewhere.

### DIFF
--- a/src/fileio/union_fstream.cpp
+++ b/src/fileio/union_fstream.cpp
@@ -82,6 +82,11 @@ union_fstream::union_fstream(std::string url,
       original_input_stream_handle = std::static_pointer_cast<std::istream>(s3stream);
     }
   } else {
+    // Remove the preceeding file:// if it's a local URL.
+    if(boost::algorithm::starts_with(url, "file://")) {
+      url = url.substr(7);
+    }
+
     // must be local file
     if (is_output_stream) {
       // Output stream must be a local openable file.
--- a/test/fileio/general_fstream_test.cxx
+++ b/test/fileio/general_fstream_test.cxx
@@ -40,6 +40,13 @@ struct general_fstream_test {
       TS_ASSERT_EQUALS(helper_test_seek(fname), 0);
     }
 
+    void test_local_url_with_prefix() {
+      std::string fname = "file://" + std::string(tmpname);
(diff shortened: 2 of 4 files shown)


src/fileio/union_fstream.cpp
```diff
@@ -82,6 +82,11 @@ union_fstream::union_fstream(std::string url,
       original_input_stream_handle = std::static_pointer_cast<std::istream>(s3stream);
     }
   } else {
+    // Remove the preceeding file:// if it's a local URL.
+    if(boost::algorithm::starts_with(url, "file://")) {
```
znation: Protocols are case insensitive according to the [spec](https://tools.ietf.org/html/rfc3986#section-3.1). It looks like the case insensitivity is handled properly in the get_protocol function above - but here, I think we should probably match all case variations of `file://`.

test/fileio/general_fstream_test.cxx
```diff
@@ -40,6 +40,13 @@ struct general_fstream_test {
       TS_ASSERT_EQUALS(helper_test_seek(fname), 0);
     }
 
+    void test_local_url_with_prefix() {
+      std::string fname = "file://" + std::string(tmpname);
```
znation: I would add a unit test case for a mixed-case variant of `file://` as well.

src/fileio/union_fstream.cpp
hoytak: Hmmm.  Interesting.  This brings up another point, though, as none of the other code paths (e.g. s3 or http) use type insensitive checking.  I'd be happy to add it here, though, but we should probably make sure that HTTP and others work in a separate issue.

znation: 👍

hoytak: So, I actually took a stab at fixing this, and we always assume that protocols are lower case throughout the code in fileio/*.  It's quite a bit of work to fix this.  I vote we get this in now, but then file an issue to fix the case insensitive protocols globally.

znation: Thanks for investigating. That sounds like the right approach to me.

Issue #380, opened 2018-03-16
Remove dependency on pillow (PIL)


gustavla:
Pillow is one of Turi Create's hard dependencies. However, it is only used in two situations:

- For unit tests to generate random `tc.Image` images. If we made it easy to take a numpy array and turn it into a `tc.Image`, this need for PIL would go away.
- When calling `image.pixel_data`. Since we have our own bindings to decoding libraries, we should just use them instead of using PIL here.

These two cases can easily be removed. The second also leads to the weird situation where sometimes decoding is done with PIL and sometimes with our own decoder:

```python
# Decodes using PIL
x = image.pixel_data

# Decodes using Turi Create (although the decoded image still passes through PIL)
tc.image_analysis.resize(image, image.width, image.height, image.channels, 
                         decode=True).pixel_data
```

Making this more consistent is partly motivated by #375 and #77, since one solution there is to convert the <8 bit depths to 8 bit in the decoding process. However, this can easily expose the fact that there are two underlying decoding methods, when inconsistencies start appearing.

Note, the main motivation is not to get rid of a dependency (that's just a bonus) and this does not preclude future usages of PIL in Turi Create, if that's ever useful. However, in those cases, it is probably better to make pillow an optional dependency.

TobyRoseman:
@gustavla - this sounds great. 

Regarding the first situation - having an easy way to turn a numpy array into a `tc.Image` would be useful. However it's easier to just make pillow a dependency for running our units test rather than a dependency for Turi Create. You should just need to add it to [this file](https://github.com/apple/turicreate/blob/15ef2d84ee554e446a2a1cc1057d2360672520fc/scripts/requirements.txt)

gustavla:
@TobyRosemen Good point! I'll start with that and then we can do numpy -> `tc.Image` separately.

PR #390, opened 2018-03-21
Do not use PIL for decoding images for pixel_data

gustavla:
This partly addresses #380. It does not remove pillow as a dependency (that can be considered separately), but does remove PIL as a means to decode images.

Images are exported into numpy arrays in two flavors:

- In format CHW in floats (for feeding to mxnet)
- in format HWC in unsigned chars (for `pixel_data`)

I have consolidated both these extraction needs in `copy_image_to_memory`.

### API Break
Also, this is technically a breaking API change. The `pixel_data` used to be able to fall back to returning `array.array` in the case numpy is not installed. I removed this since numpy is a hard dependency of turicreate, so this is only breaking API if used in a non-standard way (by delibrately uninstalling numpy). @znation What do you think? Should I mark this PR as api-breaking? Or, should I extend the implementation to also support `array.array`?

### DIFF
--- a/src/unity/extensions/additional_sframe_utilities.cpp
+++ b/src/unity/extensions/additional_sframe_utilities.cpp
@@ -16,6 +16,55 @@
 using namespace turi;
 
 
+template <typename T>
+void copy_image_to_memory(const image_type& img, T *outptr,
+                          const std::vector<size_t>& outstrides,
+                          bool channel_last) {
+  ASSERT_EQ(outstrides.size(), 3);
+  size_t index_h, index_w, index_c;
+  if (channel_last) {
+    // Format: HWC
+    index_h = 0;
+    index_w = 1;
+    index_c = 2;
+  } else {
+    // Format: CHW
+    index_c = 0;
+    index_h = 1;
+    index_w = 2;
+  }
+
+  // Decode if needed
+  if (!img.is_decoded()) {
+    char* buf = NULL;
+    size_t length = 0;
+    if (img.m_format == Format::JPG) {
+      decode_jpeg((const char*)img.get_image_data(), img.m_image_data_size, &buf, length);
+    } else if (img.m_format == Format::PNG) {
+      decode_png((const char*)img.get_image_data(), img.m_image_data_size, &buf, length);
+    }
+    size_t cnt = 0;
+    for (size_t i = 0; i < img.m_height; ++i) {
+      for (size_t j = 0; j < img.m_width; ++j) {
+        for (size_t k = 0; k < img.m_channels; ++k) {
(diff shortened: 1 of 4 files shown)


znation:
@gustavla Considering numpy is already required, I wouldn't consider this an API-breaking change. I'm not sure the original logic made sense, but I don't expect anyone would uninstall numpy after installing Turi Create -- and if they did, I don't think we are guaranteeing that will continue to work.

src/unity/extensions/additional_sframe_utilities.cpp
```diff
@@ -16,6 +16,55 @@
 using namespace turi;
 
 
+template <typename T>
+void copy_image_to_memory(const image_type& img, T *outptr,
+                          const std::vector<size_t>& outstrides,
```
znation: The name `outstrides` is a bit confusing - it at first suggests to me that it's an out parameter (like `outptr`). But looking at the code, it looks like an in parameter, that is meant to be used with the out parameter `outptr`. What about calling it something like `strides_for_outptr` instead?

src/unity/extensions/additional_sframe_utilities.cpp
```diff
@@ -16,6 +16,55 @@
...
+    index_w = 2;
+  }
+
+  // Decode if needed
+  if (!img.is_decoded()) {
+    char* buf = NULL;
+    size_t length = 0;
+    if (img.m_format == Format::JPG) {
+      decode_jpeg((const char*)img.get_image_data(), img.m_image_data_size, &buf, length);
+    } else if (img.m_format == Format::PNG) {
```
TobyRoseman: It seems like later you assume `buf` is not null. So this should be an `else` rather than a `else if`?

Perhaps the test here should just be an assert?

src/unity/extensions/additional_sframe_utilities.cpp
gustavla: Yes, it's the strides associated with `outptr`. These names are taken from `sframe_load_to_numpy` in the same file, so let's change them together if we do. In that case, any opinion on `field_length` too? Would simply `strides` be clear?

src/unity/extensions/additional_sframe_utilities.cpp
gustavla: Good point. The else should be unreachable (although that can easily change with the introduction of a new image type). I'll add an else-assert here.

gustavla: Done.

gustavla:
I noticed that grayscale images were returned as `(H, W, 1)` and not `(H, W)` as the current turicreate behavior. This also broke `_to_pil_image` in that case. I have fixed this and made sure we have test coverage for that.

Issue #77, opened 2017-12-12
ToolkitError: Unsupported PNG bit depth: 4


cor-rine:
While creating graph lab images - PNG images are not currently supported by the ToolKit, resulting in the following error to be thrown:
`ToolkitError: Unsupported PNG bit depth: 4`

Complete error details:

```
---------------------------------------------------------------------------
ToolkitError                              Traceback (most recent call last)
/Users/corrinetoracchio/Projects/corSheetMusic/turi/read-timed-notes.py in <module>()
     11 
     12                 for imageFile in data:
---> 13                          gl.Image(imageFile.replace('\n', ''))

/Users/corrinetoracchio/anaconda/envs/gl-env/lib/python2.7/site-packages/graphlab/data_structures/image.pyc in __init__(self, path, format, **_Image__internal_kw_args)
     70             from ..util import _make_internal_url
     71             from .. import extensions as _extensions
---> 72             img = _extensions.load_image(_make_internal_url(path), format)
     73             for key, value in list(img.__dict__.items()):
     74                 setattr(self, key, value)

/Users/corrinetoracchio/anaconda/envs/gl-env/lib/python2.7/site-packages/graphlab/extensions.pyc in <lambda>(*args, **kwargs)
    166 
    167 def _make_injected_function(fn, arguments):
--> 168     return lambda *args, **kwargs: _run_toolkit_function(fn, arguments, args, kwargs)
    169 
    170 def _class_instance_from_name(class_name, *arg, **kwarg):

/Users/corrinetoracchio/anaconda/envs/gl-env/lib/python2.7/site-packages/graphlab/extensions.pyc in _run_toolkit_function(fnname, arguments, args, kwargs)
    155     if ret[0] != True:
    156         if len(ret[1]) > 0:
--> 157             raise _ToolkitError(ret[1])
    158         else:
    159             raise _ToolkitError("Toolkit failed with unknown error")

ToolkitError: Unsupported PNG bit depth: 4

znation:
@cor-rine PNG images should be supported with other bit depths (like 8 bits per channel). We'll keep this issue open to track adding support for 4-bit images. In the meantime, as a workaround you could use another tool like ImageMagick to convert the bit depth to 24-bit (8 bits per channel, 3 channels) or 32-bit (8 bits per channel, 4 channels) prior to loading the images into Turi Create. Hope that helps!

cor-rine:
[...]

PR #393, opened 2018-03-22
Support PNG images with non-8-bit depth

### LINKED ISSUE #375: Unsupported PNG bit depth: 1
Another PNG loading issue:

```python
>>> tc.Image('woodenspoonbinary1.png')
turicreate.toolkits._main.ToolkitError: Unsupported PNG bit depth: 1
```
More info on this file:
```
woodenspoonbinary1.png PNG 2048x1536 2048x1536+0+0 8-bit Gray 2c 5687B 0.000u 0:00.000
```
This is a palette file (just like addressed in #278), but the difference is that it is a grayscale palette and not an RGB palette.

This example image can be found [here](http://homepages.inf.ed.ac.uk/rbf/UTENSILS/).

gustavla:
This adds support for 1, 2, and 4 by decoding them into 8-bit.  It adds support for 16-bit, by extracting only the most significant byte. This incurs loss of precision, so not ideal, but I think probably better than not supporting it at all.

Adds unit tests with references images and color correctness checks.

Fixes #375, #77.


nickjong:
Nice!

PR #408, opened 2018-03-27
Gitignore cleanup

znation:
Removing lots of `.gitignore` contents that are actually shadowing intentionally tracked files in our repo.


znation:
@TobyRoseman Those Xcode files are meant to be gitignored (so I don't want to remove them from our gitignore). In theory projects should never require xcuserdata - they should use [shared schemes](https://stackoverflow.com/a/13954760) instead if needed by CI systems for building. If we hit build issues I'll move these to a shared scheme (but otherwise I think they aren't relevant for us).

PR #409, opened 2018-03-27
WIP: Refactoring code, Table View Functionality

abhishekpratapa:
- Refactoring Code
- Adding Table View Functionality


src/unity/lib/unity_sframe.cpp
```diff
@@ -1950,6 +1950,186 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
       }
     };
 
+    auto getAccordian = [self, &reader, &ew, &column_names, &empty_tz](std::string column_name, size_t index) {
```
znation: Accordion (with an o). And I'm not sure I'd want to use the name of the specific UI control here - can we come up with a name that represents the information being passed through, rather than the way it's displayed to the user? How about `getFlexibleTypeValue`?

src/unity/lib/unity_sframe.cpp
```diff
@@ -1950,6 +1950,186 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
...
+
+        flexible_type value = gl_sa[index];
+
+        switch (value.get_type()) {
+          case flex_type_enum::UNDEFINED:
+            break;
+          case flex_type_enum::FLOAT:
+            {
+              std::stringstream ss;
+              ss << "{\"modal_spec\": {\"index\": " << index << ", \"column\":" << turi::visualization::extra_label_escape(column_name);
```
znation: Similarly, I don't think we should call this `modal_spec` - the UI may evolve over time and may or may not be modal. I think all of the metadata here could be top-level within this dictionary, as:
```
{
    "index": ...,
    "column": ...,
    "type": ...,
    "data": ...
}
```

src/unity/lib/unity_sframe.cpp
```diff
@@ -2016,27 +2196,39 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
...
+      MethodType response;
+
       auto sa = gl_sarray(std::vector<flexible_type>(1, input)).astype(flex_type_enum::DICT);
       flex_dict dict = sa[0].get<flex_dict>();
       for (const auto& pair : dict) {
         const auto& key = pair.first.get<flex_string>();
         const auto& value = pair.second;
         if (key == "method") {
-          DASSERT_EQ(value.get<flex_string>(), "get_rows");
+          if(value.get<flex_string>() == "get_rows"){
```
znation: Interesting that method name and parameters are parsed in two different places. It may simplify the logic for sharing parameters between different methods, but looks a bit confusing (as it implies all methods can take all parameters). For now it's fine (since we only have two methods) but we may want to revisit this later to associate specific parameters with specific methods.

src/visualization/Turi Create Visualization/src/Pipe.swift
```diff
@@ -46,6 +46,9 @@ class Pipe {
 
         #endif
         
+        process_data(data: Debugging.BoxesAndWhiskers.spec)
```
znation: Please remove this.

src/visualization/Turi Create Visualization/src/js/lib/elements/Plot/Plot.css
```diff
@@ -0,0 +1,19 @@
+.vega_container{
```
znation: You might want to use a vendor prefix (`tc_`) on any CSS we write; otherwise, it's possible something in Vega itself (or any of our other JS dependencies) could pick up and accidentally use our CSS. (Or, if it's intentional that we're using an element ID or class name from a dependent library, please add a comment in the CSS describing which library it's intended for.)

src/visualization/Turi Create Visualization/src/js/lib/elements/Plot/Plot.css
```diff
@@ -0,0 +1,19 @@
...
+    
+}
```
znation: An empty CSS block will do nothing; I think you can remove it.

src/visualization/Turi Create Visualization/src/js/lib/elements/Plot/Plot.js
```diff
@@ -0,0 +1,166 @@
+import React, { Component } from 'react';
+import vegaEmbed from 'vega-embed';
+import {vega as vegaTooltip} from'vega-tooltip';
+import './Plot.css';
+
+var vega = require('vega');
+
+window.vega_type = vega;
```
znation: Why is this called `vega_type`? And why does it need to be global?

src/visualization/Turi Create Visualization/src/js/lib/elements/Plot/Plot.js
```diff
@@ -0,0 +1,166 @@
...
+    componentDidMount(){
+        this.addSpec(this.props.vega_spec);
+    }
+
+    checkLoadedFlag(data, $this) {
+        if($this.vegaLoading) {
+            window.setTimeout($this.checkLoadedFlag, 100, data, $this);
+        } else {
+            if($this.vegaLoading){
+                debugger;
```
znation: Remove these `debugger` calls? They may get annoying when debugging unrelated issues (if they are supposed to happen frequently).

src/visualization/Turi Create Visualization/src/js/lib/elements/Plot/Plot.js
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            var newData = data.values.slice();
+
+            var changeSet = vega.changeset();
+
+            if(data.progress != null){
+                var progress = parseFloat(data.progress);
+                if(parseInt(progress*100, 10) >= 100){
+                    document.getElementById('progress_bar').style.width = parseInt(progress*100)+"%";
```
znation: This line seems repeated in the if/else cases. You can probably move it up above the `if` so it isn't repeated.

abhishekpratapa:
- newer pull request

PR #414, opened 2018-03-30
Add data preparation step to object detection user guide

### LINKED ISSUE #187: How to download the dataset cats-dogs.sframe?
I want to sue the example, but I don't find the way to download the datasets. Thanks for your suggestion!

gustavla:
- Adds new data preparation page with reference to IG02 dataset
- Changes some examples in the main page to reference this dataset (replacing `cats-dogs.sframe`)

This closes #187.

This dataset requires the next release to properly load the images, so we should probably hold on merging this until right before the release (for people browsing the user guide on GitHub).


userguide/object_detection/data-preparation.md
```diff
@@ -0,0 +1,134 @@
...
+
+# Load all images in random order
+raw_sf = tc.image_analysis.load_images(ig02_path, recursive=True,
+                                       random_order=True)
+
+# Split file names so that we can determine what kind of image each row is
+# E.g. bike_005.mask.0.png -> ['bike_005', 'mask']
+info = raw_sf['path'].apply(lambda path: os.path.basename(path).split('.')[:2])
+
+# Rename as 'name', 'type', 'mask_index'
```
TobyRoseman: I don't think you want to mention 'mask_index' here. Also I would probably say "Rename the columns to" rather than "Rename as".

userguide/object_detection/data-preparation.md
```diff
@@ -0,0 +1,134 @@
...
+This dataset contains masks marking the exact pixels belonging to each object
+instance. Each image may have several masks to indicate the presence of
+multiple objects. Red pixels indicate object presence and green indicate
+presence but with [occlusion](https://en.wikipedia.org/wiki/Occultation). For
+object detection, we will convert this richly annotated data to bounding boxes.
+Here is a snippet to turn this into an SFrame:
+
+```python
+import turicreate as tc
+import numpy as np
```
TobyRoseman: Looks like you're not using `glob`. Also you're only using `numpy` from your helper function, so I would just move the import to your helper function.

userguide/object_detection/data-preparation.md
```diff
@@ -0,0 +1,134 @@
...
+    x0, x1 = np.where(mask.max(0))[0][[0, -1]]
+    y0, y1 = np.where(mask.max(1))[0][[0, -1]]
+
+    return {'x': (x0 + x1) / 2, 'width': (x1 - x0),
+            'y': (y0 + y1) / 2, 'height': (y1 - y0)}
+
+# Convert masks to bounding boxes (drop masks that did not contain bounding box)
+sf_masks['coordinates'] = sf_masks['image'].apply(mask_to_bbox_coordinates)
+
+# There can be empty masks (which returns None), so let's sort them out
```
TobyRoseman: Maybe "so let's get rid of those" rather than "so let's sort them out"

userguide/object_detection/data-preparation.md
gustavla: Very good catch! `mask_index` is old stuff, indeed. I agree with your other suggestion too.

userguide/object_detection/data-preparation.md
gustavla: :+1:

userguide/object_detection/data-preparation.md
gustavla: Agreed.

All changes have been pushed. Thanks!

userguide/object_detection/data-preparation.md
```diff
@@ -0,0 +1,133 @@
...
+
+In this section, we will show you how to download a publicly available dataset
+and load it into an SFrame. This will allow you to try out the Object Detection
+toolkit for yourself. To get this dataset into the format expected by our
+toolkit, we will rely on many useful SFrame functions.
+
+*Note: Please make sure that you have Turi Create 4.3 or above for these steps*
+
+The dataset that we will use is [INRIA Annotations for Graz-02
+(IG02)](https://lear.inrialpes.fr/people/marszalek/data/ig02/). Our goal is to
```
afranklin: We just need a reference to the dataset disclaimer, after the period on this line. Thanks!
`[<sup>1</sup>](../datasets.md)`

gustavla: Good idea. I changed it to use markdown footnotes, which looks good.

gustavla: At least through gitbook - GitHub viewing is not as nice, but functionally OK.

PR #415, opened 2018-03-31
Plot API now has Independent Plots

shantanuchhabra:
Added code for Plot API. Built and tested.

Usage:

```
turicreate.visualization.histogram(sa)
turicreate.visualization.item_frequency(sa)
turicreate.visualization.columnwise_summary(sf)
turicreate.visualization.box_plot(sa1, sa2)
turicreate.visualization.heatmap(sa1, sa2)
turicreate.visualization.scatter(sa1, sa2)
turicreate.visualization.categorical_heatmap(sa1, sa2)
```


src/unity/lib/unity_sarray.cpp
```diff
@@ -2836,79 +2836,17 @@ std::shared_ptr<model_base> unity_sarray::plot(const std::string& path_to_client
     log_and_throw("Nothing to show; SArray is empty.");
   }
 
-  std::shared_ptr<unity_sarray> self = \
-             std::make_shared<unity_sarray>(*this);
+  std::shared_ptr<unity_sarray> self = std::make_shared<unity_sarray>(*this);
+  gl_sarray sa = gl_sarray(self);
```
znation: I think you can do `gl_sarray sa(self)` to use copy constructor instead of copy assign operator (style preference, I think -- I think any efficiency difference is likely to get optimized away).

PR #431, opened 2018-04-03
Code for saving as PNG and SVG

shantanuchhabra:
Added code to support saving of plots in PNG and SVG formats. Assumes that the user already has node.


src/unity/python/turicreate/visualization/plot.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import absolute_import as _
 import logging as _logging
 import json as _json
+import os
```
TobyRoseman: Don't pollute the namespace. It should be `import os as _os`

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -126,33 +127,44 @@ def save(self, filepath, include_data=True):
...
-
-        if(include_data):
-            data = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_data'}))["data_spec"]
-
-            for x in range(0, len(spec["vega_spec"]["data"])):
-                if(spec["vega_spec"]["data"][x]["name"] == "source_2"):
-                    spec["vega_spec"]["data"][x] = data
-                    break;
+        if type(filepath) != str:
+            raise ValueError("File path provided is not a string")
```
TobyRoseman: I would make it `filepath provided is not a string`. So it matches the parameter name.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -126,33 +127,44 @@ def save(self, filepath, include_data=True):
...
 
     def _get_vega(self, include_data=True):
         if(include_data):
-            spec = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_spec'}))
+            spec = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_spec'}))["vega_spec"]
             data = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_data'}))["data_spec"]
-            for x in range(0, len(spec["vega_spec"]["data"])):
-                if(spec["vega_spec"]["data"][x]["name"] == "source_2"):
-                    spec["vega_spec"]["data"][x] = data
+            for x in range(0, len(spec["data"])):
```
TobyRoseman: you don't need the first parameter to `range`.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -126,33 +127,44 @@ def save(self, filepath, include_data=True):
...
+            spec = self._get_vega(include_data = include_data)
+            with open(filepath, 'w') as fp:
+                _json.dump(spec, fp)
+        elif filepath.endswith(".png") or filepath.endswith(".svg"):
+            # save as png/svg, but json first
+            spec = self._get_vega(include_data = True)
+            extension = filepath[-3:]
+            with open("temp_file.vg.json", 'w') as fp:
+                _json.dump(spec, fp)
+            dirname = os.path.dirname(__file__)
```
TobyRoseman: I don't think it's safe to assume that the directory where turicreate is installed will be writable. 

I think you should use the `tempfile` module:
https://docs.python.org/2/library/tempfile.html

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -126,33 +127,44 @@ def save(self, filepath, include_data=True):
...
     def _get_vega(self, include_data=True):
         if(include_data):
-            spec = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_spec'}))
+            spec = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_spec'}))["vega_spec"]
             data = _json.loads(self.__proxy__.get('call_function', {'__function_name__': 'get_data'}))["data_spec"]
-            for x in range(0, len(spec["vega_spec"]["data"])):
-                if(spec["vega_spec"]["data"][x]["name"] == "source_2"):
-                    spec["vega_spec"]["data"][x] = data
+            for x in range(0, len(spec["data"])):
+                if(spec["data"][x]["name"] == "source_2"):
```
TobyRoseman: what is the significance of "source_2"?

shantanuchhabra: @abhishekpratapa ?

znation: `source_2` seems to be the default name for a dataset within Vega (if you don't otherwise give it a name) - so we've adopted it by convention for our dataset. We should probably rename it to something clearer (since we don't need to use the default for this), although that change would need to touch Python and JavaScript code and is orthogonal to this PR, so it could come later as a cleanup PR. For now let's keep `source_2`.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -139,16 +140,19 @@ def save(self, filepath, include_data=True):
...
-            absolute_path_to_vg2png_vg2svg = os.path.join(dirname, relative_path_to_vg2png_vg2svg)
-            os.system("node " + absolute_path_to_vg2png_vg2svg + " temp_file.vg.json " + filepath)
+            dirname = _os.path.dirname(__file__)
+            relative_path_to_vg2png_vg2svg = "../node_modules/vega/bin/vg2" + extension
+            absolute_path_to_vg2png_vg2svg = _os.path.join(
+                dirname, relative_path_to_vg2png_vg2svg)
+            _os.system("node " + absolute_path_to_vg2png_vg2svg + " " + temp_file_path + " " + filepath)
             # delete temp file that user didn't ask for
-            os.system("rm temp_file.vg.json") 
+            _os.system("rm " + temp_file_path)
```
znation: You can use [os.unlink](https://docs.python.org/2/library/os.html#os.unlink) instead.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -139,16 +140,19 @@ def save(self, filepath, include_data=True):
...
                 _json.dump(spec, fp)
-            dirname = os.path.dirname(__file__)
-            relative_path_to_vg2png_vg2svg = "../../../../../../node_modules/vega/bin/vg2" + extension
-            absolute_path_to_vg2png_vg2svg = os.path.join(dirname, relative_path_to_vg2png_vg2svg)
-            os.system("node " + absolute_path_to_vg2png_vg2svg + " temp_file.vg.json " + filepath)
+            dirname = _os.path.dirname(__file__)
+            relative_path_to_vg2png_vg2svg = "../node_modules/vega/bin/vg2" + extension
+            absolute_path_to_vg2png_vg2svg = _os.path.join(
+                dirname, relative_path_to_vg2png_vg2svg)
+            _os.system("node " + absolute_path_to_vg2png_vg2svg + " " + temp_file_path + " " + filepath)
```
znation: Before we call into an external `node` binary, we should verify that one exists in the user's path. If this call is going to fail, we should instead present a nice error message to the user telling them to install node (and linking to the Node download page).

znation: That can come in a future PR.

src/visualization/Turi Create Visualization/src/js/package.json
```diff
@@ -15,9 +15,11 @@
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react": "^6.24.1",
+    "vega": "^3.2.1",
```
znation: You can remove this line.

src/visualization/Turi Create Visualization/src/js/package.json
```diff
@@ -15,9 +15,11 @@
...
     "webpack": "^3.8.1",
     "webpack-dev-server": "^2.9.4"
   },
+  "postinstall": "cd node_modules/canvas && node-gyp rebuild",
```
znation: You can remove this line.

Issue #455, opened 2018-04-09
Plot.save fails mysteriously if node is not present


znation:
Repro steps:

1. Remove node from the machine (if present). I renamed `/usr/local/bin/node` to `node_bak`.
2. Run the following Python code:

```python
import turicreate as tc
sf = tc.SFrame.read_csv('https://docs-assets.developer.apple.com/turicreate/datasets/tc-clang-format-results.csv')
plt = tc.visualization.columnwise_summary(sf)
plt.save('/Users/zach/Desktop/temp.svg')
```

Expected:
A friendly error message about requiring Node.js (with a link to download).

Actual:

```
sh: node: command not found
```

Additionally, no file is written at the location specified.

shantanuchhabra:
Thanks a ton for pointing this out!

PR #499, opened 2018-04-10
Fix for Issue #455

shantanuchhabra:
Added a helpful message for the user in case they don't have node installed. Changed the implementation a little to support that.


barchart.json
```diff
@@ -0,0 +1,95 @@
+{
```
znation: Did you mean to commit this file? Or was it intended only for local testing?

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -126,33 +139,61 @@ def save(self, filepath, include_data=True):
...
+            dirname = _os.path.dirname(__file__)
+            relative_path_to_vg2png_vg2svg = "../vg2" + extension
+            absolute_path_to_vg2png_vg2svg = _os.path.join(dirname, 
+                relative_path_to_vg2png_vg2svg)
+            (exitcode, stdout, stderr) = _run_cmdline("node " + 
+                absolute_path_to_vg2png_vg2svg + " " 
+                + temp_file_path + " " + filepath)
+            if exitcode == NODE_NOT_FOUND_ERROR_CODE:
+                # user doesn't have node installed
+                raise NameError("Node.JS not found. Please download " +
```
znation: Should be `Node.js` (lowercase js).

znation: And we should explain that saving as png and jpg (specifically) requires Node.js - otherwise it's implied that you need it to use Turi Create in general (which is not the case).

src/unity/python/turicreate/CMakeLists.txt
```diff
@@ -49,6 +49,15 @@ set_property(DIRECTORY PROPERTY INSTALLATION_EXTENSIONS "${INSTALLATION_EXTENSIO
 set_property(DIRECTORY PROPERTY INSTALLATION_BINARY_TARGETS "${INSTALLATION_BINARY_TARGETS}")
 set_property(DIRECTORY PROPERTY INSTALLATION_BINARY_FILES "${INSTALLATION_BINARY_FILES}")
 
+add_custom_target(
```
znation: I thought we already had this custom target. How did this work prior to this PR?

src/unity/python/turicreate/CMakeLists.txt
```diff
@@ -49,6 +49,15 @@ set_property(DIRECTORY PROPERTY INSTALLATION_EXTENSIONS "${INSTALLATION_EXTENSIO
...
+  visualization_vega ALL
+  DEPENDS visualization_client
```
znation: Does this actually depend on visualization_client? If we are committing the files being copied directly into the repo, it seems like it doesn't.

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -5,9 +5,20 @@
 import json as _json
 import os as _os
 from tempfile import mkstemp as _mkstemp
+from subprocess import Popen as _Popen 
+from subprocess import PIPE as _PIPE
 
 _target = 'auto'
 
+NODE_NOT_FOUND_ERROR_CODE = 127
```
TobyRoseman: Unless you want to publicly export this from this file, you should start the name with an underscore.

shantanuchhabra: Oh yeah good catch, thanks!!

shantanuchhabra:
Resolved the merge issue so the changes are much more succinct. @znation

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -139,21 +150,35 @@ def save(self, filepath, include_data=True):
...
-                dirname, relative_path_to_vg2png_vg2svg)
-            _os.system("node " + absolute_path_to_vg2png_vg2svg + " " + temp_file_path + " " + filepath)
+            absolute_path_to_vg2png_vg2svg = _os.path.join(dirname, 
+                relative_path_to_vg2png_vg2svg)
+            (exitcode, stdout, stderr) = _run_cmdline("node " + 
+                absolute_path_to_vg2png_vg2svg + " " 
+                + temp_file_path + " " + filepath)
+            if exitcode == _NODE_NOT_FOUND_ERROR_CODE:
+                # user doesn't have node installed
+                raise NameError("Node.js not found. Saving as PNG and SVG" +
```
znation: Looking again at this, I don't think `NameError` is appropriate for this use -- it's supposed to mean an undeclared identifier (local or global variable) is being referenced. Maybe this should just be `RuntimeError`.

Issue #442, opened 2018-04-09
ND-Array Table


abhishekpratapa:
<img width="932" alt="screen shot 2018-04-09 at 1 24 51 pm" src="https://user-images.githubusercontent.com/4616962/38521068-6c93eae6-3bf9-11e8-8773-b9cf8dc9495d.png">

ND-Array: size not scoped to 190px

- Same as not having accordion implemented

PR #500, opened 2018-04-10
Added ND-array support

abhishekpratapa:
- Added ND-Array type

### DIFF
--- a/src/unity/lib/unity_sframe.cpp
+++ b/src/unity/lib/unity_sframe.cpp
@@ -1842,6 +1842,15 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
               case flex_type_enum::LIST:
                 ss << value.to<std::string>();
                 break;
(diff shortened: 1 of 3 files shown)


abhishekpratapa:
fixed #442

src/unity/lib/unity_sframe.cpp
```diff
@@ -2004,35 +2013,9 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
             }
             break;
           case flex_type_enum::LIST:
-            {
```
znation: What's the deal with all this code being removed? It wasn't needed? Or was it causing bug(s)?

(If the former -- maybe we shouldn't change it now?)
(If the latter -- what bugs was it causing? Should it go in as a separate PR?)

abhishekpratapa: Duplicated code!

Issue #440, opened 2018-04-09
List of Lists Crash on Accordion 


abhishekpratapa:
- List of Lists crashes on Accordion.
- Error: Invalid JSON

PR #503, opened 2018-04-10
Added recursive functionality to list view

abhishekpratapa:
- Added recursive functionality to lists.


abhishekpratapa:
fixed #440

src/unity/lib/unity_sframe.cpp
```diff
@@ -1840,8 +1934,11 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
                 }
                 break;
               case flex_type_enum::LIST:
-                ss << value.to<std::string>();
-                break;
+                {
```
znation: Shouldn't this whole section (switch/case over type, with stringification) get refactored into `listHelper`? Seems like almost identical logic.

src/unity/lib/unity_sframe.cpp
```diff
@@ -1673,7 +1673,101 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
       ew << ss.str();
     }
 
-    auto getRows = [self, &reader, &ew, &column_names, &empty_tz, &image_queue](size_t start, size_t end) {
+    std::function<std::string(flexible_type&)> listHelper;
+    listHelper = [&listHelper, &empty_tz](flexible_type& value) {
```
znation: Can we come up with a better name than `listHelper`? How about `escapeForTable`? (Is this only for the accordion, only for inline display in the table, or both? Let's pick a name that communicates the answer to that question...)

abhishekpratapa: Just for the inline display in the table. How's `escapeForInlineTable`?

znation: I find the word `inline` confusing. I think `escapeForTable` is fine.

Issue #450, opened 2018-04-09
Plot APIs should ValueError on incorrect input type


znation:
The Plot APIs under `tc.visualization` expect certain data types, and when the wrong types are passed in, should give a friendly error message to the user explaining the mismatch. Instead, the API currently errors low-level in C++ and doesn't tell the user why it failed:

```python
import turicreate as tc
sf = tc.SFrame.read_csv('https://docs-assets.developer.apple.com/turicreate/datasets/tc-clang-format-results.csv')
plt = tc.visualization.box_plot(sf, sf)
```

```python
---------------------------------------------------------------------------
ToolkitError                              Traceback (most recent call last)
<ipython-input-9-97dd2288d4ba> in <module>()
----> 1 plt = tc.visualization.box_plot(sf, sf)

/Users/zach/turicreate/debug/src/unity/python/turicreate/visualization/show.pyc in box_plot(x, y, xlabel, ylabel, title)
    121     title = _get_title(title)
    122     plt_ref = tc.extensions.plot_boxes_and_whiskers(path_to_client, x, y, 
--> 123       xlabel, ylabel, title)
    124     return Plot(plt_ref)
    125 

/Users/zach/turicreate/debug/src/unity/python/turicreate/extensions.pyc in <lambda>(*args, **kwargs)
    171 
    172 def _make_injected_function(fn, arguments):
--> 173     return lambda *args, **kwargs: _run_toolkit_function(fn, arguments, args, kwargs)
    174 
    175 def _class_instance_from_name(class_name, *arg, **kwarg):

/Users/zach/turicreate/debug/src/unity/python/turicreate/extensions.pyc in _run_toolkit_function(fnname, arguments, args, kwargs)
    160     if not ret[0]:
    161         if len(ret[1]) > 0:
--> 162             raise _ToolkitError(ret[1])
    163         else:
    164             raise _ToolkitError("Toolkit failed with unknown error")

ToolkitError: boost::bad_get: failed value get using boost::get
```

PR #504, opened 2018-04-10
Bugfix for issue #445 and issue #450

shantanuchhabra:


src/unity/python/turicreate/visualization/plot.py
```diff
@@ -107,13 +107,18 @@ def save(self, filepath, include_data=True):
...
         Parameters
         ----------
+        filepath: string
+            The destination filepath where the plot object must be saved as.
+            The extension of this filepath determines what format the plot will
+            be saved as. Currently supported formats are JSON, PNG, and SVG.
         include_data : bool, optional
             If True, save's the Plot in a vega spec with the data spec
-            included.
+            included. This parameter is not important when saving as a PNG/SVG
```
znation: This is really weird. I think we should get rid of this parameter entirely, instead. (When would the user want to save without data??).

src/unity/python/turicreate/visualization/plot.py
```diff
@@ -107,13 +107,18 @@ def save(self, filepath, include_data=True):
...
 
         Notes
         -----
-        - The save method saves the Plot object in a vega json format
+        - The save method saves the Plot object in a vega json format, or a
```
znation: I think this could be worded more clearly - but is redundant with the parameters/description above anyway. I'd remove this entire `Notes` section.

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
     Plot(plt_ref).show()
 
 def scatter(x, y, xlabel="X", ylabel="Y", title=None):
+    """
+    Plots the data in `x` on the X axis and the data in `y` on the Y axis
+    in a 2d scatterplot, and returns the resulting Plot object.
```
znation: Consistency: we usually use two words (`scatter plot`).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
+    
+    The function supports SArrays of dtypes: int, float.
+
+    Parameters
+    ----------
+    x : SArray
+      The data to plot on the X axis of the scatter plot. 
+      Must be numeric (int/float)
```
znation: Please put a period at the end (complete sentences encouraged but not required).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
+    ylabel : str (optional)
+      The text label for the Y axis. Defaults to "Y".
+    title : str (optional)
+      The title of the plot. Defaults to None. If the value is None, the title
+      will be "<xlabel> vs. <ylabel>". Otherwise, the string passed in as the
+      title will be used as the plot title.
+
+    Return Value
+    ------------
+    An object of Plot that is the scatter plot
```
znation: This could be worded more clearly. See [activity classifier](https://github.com/apple/turicreate/blob/master/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py#L90) for an example of a return value documented (with name, type, and description). Please follow that format/syntax - I'm not sure "Return Value" as written here will work in our HTML template.

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
+
+    Examples
+    --------
+    Make a scatter plot.
+
+    >>> x = turicreate.SArray([1,2,3,4,5])
+    >>> y = x * 2
+    >>> scplt = turicreate.visualization.scatter(x, y)
+    """
+    if (type(x) != tc.data_structures.sarray.SArray or
```
znation: Use `isinstance`, in case users (or we) ever subclass `SArray`.

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
+    xlabel : str (optional)
+      The text label for the X axis. Defaults to "X".
+    ylabel : str (optional)
+      The text label for the Y axis. Defaults to "Y".
+    title : str (optional)
+      The title of the plot. Defaults to None. If the value is None, the title
+      will be "<xlabel> vs. <ylabel>". Otherwise, the string passed in as the
+      title will be used as the plot title.
+
+    Return Value
```
znation: Same comment as for scatter plot (should be `Returns`, with name/type/description in special syntax).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
```
znation: Same comment as for scatter plot (should be Returns, with name/type/description in special syntax).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
```
znation: Same comment as for scatter plot (should be Returns, with name/type/description in special syntax).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
+    and returns the resulting Plot object.
+    
+    The function supports SFrames.
+
+    Parameters
+    ----------
+    sf : SFrame
+      The data to get a columnwise summary for
+    
+    Return Value
```
znation: Same comment as for scatter plot (should be Returns, with name/type/description in special syntax).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
+    """
+    Plots a columnwise summary of the sframe provided as input, 
+    and returns the resulting Plot object.
+    
+    The function supports SFrames.
+
+    Parameters
+    ----------
+    sf : SFrame
+      The data to get a columnwise summary for
```
znation: Please use complete sentences with punctuation.

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
+    xlabel : str (optional)
+      The text label for the X axis. Defaults to "Values".
+    ylabel : str (optional)
+      The text label for the Y axis. Defaults to "Count".
+    title : str (optional)
+      The title of the plot. Defaults to None. If the value is None, the title
+      will be "<xlabel> vs. <ylabel>". Otherwise, the string passed in as the
+      title will be used as the plot title.
+    
+    Return Value
```
znation: Same comment as for scatter plot (should be Returns, with name/type/description in special syntax).

src/unity/python/turicreate/visualization/show.py
```diff
@@ -96,46 +96,302 @@ def show(x, y, xlabel="X", ylabel="Y", title=None):
...
```
znation: Same comment as for scatter plot (should be Returns, with name/type/description in special syntax).

src/unity/python/turicreate/visualization/plot.py
shantanuchhabra: Done!

src/unity/python/turicreate/visualization/plot.py
shantanuchhabra: As discussed in person, removing the `include_data` flag from `save` but keeping it in the `_get_vega` and `_get_data` functions.

Issue #451, opened 2018-04-09
API Docs for Visualization incomplete


shantanuchhabra:
https://apple.github.io/turicreate/docs/api/turicreate.visualization.html

shantanuchhabra:
I think this should also be fixed by #504 because inline docs feed into HTML, right @znation ?

znation:
@shantanuchhabra You'll also need to modify the [rst file](https://github.com/apple/turicreate/blob/master/src/unity/python/doc/source/turicreate.visualization.rst) that renders the link above.

PR #507, opened 2018-04-10
Fix for #451

shantanuchhabra:
Added support in the RST file to render documentation for Plot API on the API docs here: https://apple.github.io/turicreate/docs/api/turicreate.visualization.html


src/unity/python/doc/source/turicreate.visualization.rst
```diff
@@ -4,14 +4,24 @@ Visualization
...
   :nosignatures:
 
  show
  SArray.show
+ SArray.plot
+ SArray.explore
  SFrame.show
+ SFrame.plot
  SFrame.explore
+ visualization.scatter
```
znation: I'm not sure how these links will render or what they will link to. We should try rendering this page as HTML and see what it looks like.

You can run `make html` from `src/unity/python/doc` (maybe `source` in there, I don't remember), while in the development python env, and see how it looks.

PR #508, opened 2018-04-11
Bugfix for 505

shantanuchhabra:
Added content for the GitBook docs at https://apple.github.io/turicreate/docs/userguide/vis/


userguide/vis/README.md
```diff
@@ -38,6 +38,14 @@ There are three primary visualization methods in Turi Create:
   and
   [SArray.explore](https://apple.github.io/turicreate/docs/api/generated/turicreate.SArray.explore.html).
 
+Turi Create also enables you to create independent plots including 
+scatter plots ([visualization.scatter](https://apple.github.io/turicreate/docs/api/generated/turicreate.visualization.scatter.html)),
```
znation: I think we could just link the words "scatter plots" directly to the API docs for scatter, and the same for the rest of them (we don't need to put the API names here - it seems too verbose).

Issue #379, opened 2018-03-16
No Top Level load_sarray


TobyRoseman:
We have `turicreate.load_model`, `turicreate.load_sframe` and `turicreate.load_sgraph`. However there is no `turicreate.load_sarray`. For consistency we should add one.

karthikdash:
Can I work on this?

TobyRoseman:
@karthikdash - absolutely, go right ahead. I would be happy to code review it when you are done, or answer any questions you might have along the way.

PR #516, opened 2018-04-12
Added Top Level load_sarray (#379)

karthikdash:
In reference to #379


src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -54,6 +54,36 @@ def _create_sequential_sarray(size, start=0, reverse=False):
...
+    actually a directory, created when the SArray is saved.
+
+    Paramaters
+    ----------
+    filename : string
+        Location of the file to load. Can be a local path or a remote URL.
+
+    Returns
+    -------
+    out : SFrame
```
TobyRoseman: This should be an `SArray`

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -54,6 +54,36 @@ def _create_sequential_sarray(size, start=0, reverse=False):
...
+
+    See Also
+    --------
+    SArray.save
+
+    Examples
+    --------
+    >>> sa = turicreate.SArray(data=[1,2,3,4,5], dtype=int)
```
TobyRoseman: I don't think having `dtype=int` is adding anything here. I would remove it.

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -54,6 +54,36 @@ def _create_sequential_sarray(size, start=0, reverse=False):
...
+    >>> sa.save('my_sarray')
```
TobyRoseman: I think this would make it clearer: `'my_sarray'` -> `./my_sarray`.

Same for the next line.

src/unity/python/turicreate/data_structures/sarray.py
znation: Should be `SArray` here.

src/unity/python/turicreate/test/test_sarray.py
```diff
@@ -9,6 +9,7 @@
 from __future__ import division as _
 from __future__ import absolute_import as _
 from ..data_structures.sarray import SArray
+from .. import load_sarray
```
znation: This should be imported directly from SArray, with `from ..data_structures.sarray import load_sarray`.

TobyRoseman:
Thanks @karthikdash Great job.

PR #524, opened 2018-04-12
Fix build for 4.3

znation:
For some reason, the test build is failing on Linux/release. This seems
to fix it. I'm not sure why.


ylow:
which unit tests fail?

znation:
@ylow No tests fail; the build fails (for most C++ tests, seemingly) inside `release/test`. They all have the same linker errors, about missing symbols that seem unrelated to visualization (like cppipc), and the linker command appears to provide `.a` files that actually do contain those symbols, so it's a mystery to me why it's erroring.

znation:
I'll merge this now, and post-4.3 we can figure out the root cause and hopefully back out this hacky fix.

PR #525, opened 2018-04-12
Last minute fixes

shantanuchhabra:
Added vg2png, vg2svg, and vega_3.2.1.js to setup.py


src/unity/python/setup.py
```diff
@@ -176,7 +179,10 @@ def run(self):
 		     'Turi Create Visualization/natives_blob.bin',
 		     'Turi Create Visualization/libcef.so',
 		     'Turi Create Visualization/v8_context_snapshot.bin',
-		     'Turi Create Visualization/cef_extensions.pak'
+		     'Turi Create Visualization/cef_extensions.pak',
+             'visualization/vega_3.2.1.js',
```
znation: This doesn't need to be included twice - the whole file list is applied for all builds (only the comments in each section indicate mac/linux). You can make a 3rd section for `# Plot.save dependencies`.

PR #526, opened 2018-04-12
Removed Canvas Dependecy

romieklund:
- removed canvas dependency

### DIFF
--- a/src/unity/python/turicreate/visualization/_plot.py
+++ b/src/unity/python/turicreate/visualization/_plot.py
@@ -9,8 +9,10 @@ from subprocess import PIPE as _PIPE
 
 _target = 'auto'
 
+_SUCCESS = 0
+_CANVAS_PREBUILT_NOT_FOUND_ERROR = 1
 _NODE_NOT_FOUND_ERROR_CODE = 127
-_CANVAS_PREBUILT_NON_EXISTENT = 1
+_PERMISSION_DENIED_ERROR_CODE = 243
 
 def _run_cmdline(command):
     # runs a shell command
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/visualization/_plot.py
```diff
@@ -159,24 +160,36 @@ def save(self, filepath):
...
                 relative_path_to_vg2png_vg2svg)
-            (exitcode, stdout, stderr) = _run_cmdline("node " + 
-                absolute_path_to_vg2png_vg2svg + " " 
+            (exitcode, stdout, stderr) = _run_cmdline("node " +
+                absolute_path_to_vg2png_vg2svg + " "
                 + temp_file_path + " " + filepath)
+
+            print(exitcode)
+            print(stdout)
+            print(stderr)
```
shantanuchhabra: Please remove print statements

src/unity/python/turicreate/visualization/_plot.py
```diff
@@ -159,24 +160,36 @@ def save(self, filepath):
...
+
             if exitcode == _NODE_NOT_FOUND_ERROR_CODE:
                 # user doesn't have node installed
                 raise RuntimeError("Node.js not found. Saving as PNG and SVG" +
                     " requires Node.js, please download and install Node.js " +
                     "from here and try again: https://nodejs.org/en/download/")
+
+            if exitcode == _CANVAS_PREBUILT_NON_EXISTENT:
+                raise RuntimeError("Canvas-Prebuilt not found. Saving as PNG and SVG" +
```
znation: Let's keep to the naming convention of the package itself (all lowercase, `canvas-prebuilt`).

src/unity/python/turicreate/visualization/vg2png
```diff
@@ -47,5 +48,5 @@ function render(spec) {
     .initialize()
     .toCanvas()
     .then(function(canvas) { writePNG(canvas, outputFile); })
-    .catch(function(err) { console.error(err); });
-}
\ No newline at end of file
+    .catch(function(err) { process.exit(_CANVAS_PREBUILT_NON_EXISTENT); });
```
znation: Let's keep the `console.error` (before the process.exit). Ideally the user should be able to see the whole error message, in case it's something weird and isn't the missing canvas-prebuilt. (This will catch all possible errors, so it's possible it won't be because of that.)

znation:
Is this ready to merge?

Issue #519, opened 2018-04-12
Terminal hangs when training 10 classes of 60 images


apoorvbagla:
Successfully trained 5 classes butTerminal hangs when training 10 classes of 60 images

TobyRoseman:
@apoorvbagla - What toolkit are you using, image classification, image similarity, object detection, or something else? How much RAM do you have?

apoorvbagla:
Im using **image classification** tool. Im running the python code on MacOS High Sierra with **8GB RAM**

This is the code Im running for preprocessing the data to create SFrame data which works fine for any size of data
 
import turicreate as tc
from turicreate import SFrame
from turicreate import SArray
import pandas as pd
import os

pathToImages = 'Images'
tempObjectLabel = []
files = os.listdir(pathToImages)
for file in files:
	tempObjectLabel.append(file)

objectLabel = tempObjectLabel[1:]

# Load images
data = tc.image_analysis.load_images(pathToImages, with_path=True)
# data = data.add_row_number('row')
#csv = pd.read_csv('annotations.csv')

temp = []

for j, item in enumerate(data):
    temp.append(os.path.split(item['path'])[1])


label = []

for j, item in enumerate(data):
	a = os.path.split(item['path'])[0]
	b = os.path.split(a)[1]
	if b not in objectLabel:
		label.append('misc')
		continue
	for i in temp:
		if i == str(os.path.split(item['path'])[1]):
			a = os.path.split(item['path'])[0]
			b = os.path.split(a)[1]
			print(b)
			label.append(b)
			break

data['label'] = SArray(data=label, dtype=str)

# the data is in no particular order, so we have to loop it to match
# we also have the 'misc' images, which won't have an annotation, to skip

# Save the data for future use
data.save('training.sframe')

#data['image_with_ground_truth'] = tc.object_detector.util.draw_bounding_boxes(data["image"], data["annotations"])

# Explore interactively
data.explore()

This is the code to train the model which hangs in case of around 10 classes and gives Kill 9 error on large number of classes.

import turicreate as tc
data = tc.SFrame('training.sframe')

train_data, test_data = data.random_split(0.8)

model = tc.image_classifier.create(train_data, target='label', max_iterations=2000)

model.save('mymodel.model')

model.export_coreml('MyCustomModel.mlmodel')

TobyRoseman:
@apoorvbagla - Thanks for sharing the code. I'm sorry you are having this issue. It is likely the same problem as #419

[...]

Issue #419, opened 2018-04-01
Large number of classes/large labels cause too much RAM usage with image classification


PietroMessineo:
Hi guys,

Finally I came out with a solution to my problem. Turi is working finally great but there is just one problem: if you will rename a folder with slash such as "example/human" and you will recall it in python as 'example:human' Turi will start to have problem during the exportation of the model and instead to spend several seconds on a base of 200 pics and 21 classes it will stay for 2 hours.

I don't know what can be the problem but try to fix it. Have a good easter 👍

znation:
@PietroMessineo Can you please provide some Python code that will reproduce this issue? (Even if you are unable to provide the data, we can probably recreate similar enough data, given a description of the columns + types). Thanks!

PietroMessineo:
Hi @znation, 

Actually I'm using three different python, but I'm encountering the problem during the export, below the code: 

`import turicreate as tc

**#Load the data**
data = tc.SFrame('alfa.sframe')

**# Make a train-test split**
train_data, test_data = data.random_split(0.8)

tc.config.set_runtime_config('TURI_FILEIO_MAXIMUM_CACHE_CAPACITY', 2*1024*1024*1024)

**# Create a model**
model = tc.image_classifier.create(train_data, target='label', model='squeezenet_v1.1', max_iterations=50)

**# Save predictions to an SFrame (class and corresponding class-probabilities)**
predictions = model.classify(test_data)

**# Evaluate the model and save the results into a dictionary**
results = model.evaluate(test_data)
print "Accuracy         : %s" % results['accuracy']
print "Confusion Matrix : \n%s" % results['confusion_matrix']

**# Save the model for later usage in Turi Create**
model.save('Alfa2.model')
`

znation:
@PietroMessineo How does the folder with a slash in the name relate to the Python code you provided? I don't see where that code uses a folder with a slash in the name.

PietroMessineo:
Hi @znation, in macOS the slash is recognize as semicolon so the code is as follow: 

`import turicreate as tc

**# load data**
image_data = tc.image_analysis.load_images('alfaset', with_path=True)

labels = ['A:a', 'B:b', 'C:c', 'D:d', 'E:e', 'F:f', 'G:g', 'H:h','I:i','J:j','K:k','L:l','M:m','N:n','O:o','P:p','Q:q','R:r','S:s','T:t','U:u','V:v','W:w','X:x','Y:y','Z:z']

def get_label(path, labels=labels):
[...]

PR #531, opened 2018-04-14
Never Use Validation Set for Image Classification

TobyRoseman:
Fix for #519 and #419.

If the total number of images in the training set is at least 100, then by default logistic regression will create a validation set. Having a validation set causes second order statistics to be calculated at the end of training. This calculate is very costly when there are a large number of features. The fix is easy, just don't use a validation set.

### DIFF
--- a/src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
+++ b/src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
@@ -382,7 +382,7 @@ class LogisticClassifier(_Classifier):
 
     """
     def __init__(self, model_proxy):
-        '''__init__(self)'''
+
         self.__proxy__ = model_proxy
         self.__name__ = self.__class__._native_name()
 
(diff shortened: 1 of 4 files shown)


srikris:
I'm not sure this is the right fix. The second order information is being computed but never used. This is an O(N * F * F) operation where F is the number of features (which is around 1K).

```
// Compute validation-set stats.
if (lr_interface->num_validation_examples() > 0) {
         // Recycle lvalues from stats to use as out parameters here, now that we're
        // otherwise done reading from stats.
        lr_interface->compute_validation_second_order_statistics(
             stats.solution, stats.hessian, stats.gradient, stats.func_value);
             state["validation_loss"] =  stats.func_value;
}
```

This is an extremely expensive operation and I'm not sure why we are doing it at the end of training. The right fix should be to replace that function with

```
 lr_interface->compute_validation_first_order_statistics
```

If that function doesn't exist, we should add it since the code that follows doesn't actually need to compute the Hessian.

@znation This seems like a fix worth a point release.

srikris:
Seems like it was introduced in https://github.com/apple/turicreate/commit/285cb3a1f65919b29afd5c34a8df0ea163677259

srikris:
Please refer to #532 which keeps the validation tracking but fixes the underlying issue.

PR #547, opened 2018-04-22
Cleanups and removal of unused code

TobyRoseman:


scripts/cmake_setup.sh
```diff
@@ -48,5 +48,5 @@ make -j4
 make install
 set +e
 
-exit 0
+return 0
```
znation: I think this one should still be `exit` since it's a top level thing and not within a function? See https://stackoverflow.com/questions/4419952/difference-between-return-and-exit-in-bash-functions

scripts/collect_archive.sh
```diff
@@ -12,7 +12,7 @@ done
 # http://stackoverflow.com/questions/8063228/how-do-i-check-if-a-variable-exists-in-a-list-in-bash
 if [[ !("${bin_list[*]}" =~ (^|[[:space:]])"$executable_to_run"($|[[:space:]])) ]]; then
   $@
-  exit
+  return
```
znation: Same comment as above.

scripts/cmake_setup.sh
```diff
@@ -9,7 +9,7 @@ set -e
 function print_help {
   echo "Usage: ./cmake_setup.sh <cmake_source_directory> <install_prefix>"
   echo
-  exit 1
+  return 1
```
znation: Will this change the behavior of the script? I think it may have been intended to early-exit when `print_help` is called.

scripts/dump_library_list.sh
```diff
@@ -12,7 +12,7 @@ done
 # http://stackoverflow.com/questions/8063228/how-do-i-check-if-a-variable-exists-in-a-list-in-bash
 if [[ !("${bin_list[*]}" =~ (^|[[:space:]])"$executable_to_run"($|[[:space:]])) ]]; then
   $@
-  exit
+  return
```
znation: Same comment as above.

scripts/make_pydocs.sh
```diff
@@ -20,7 +20,7 @@ print_help() {
   echo
   echo "Produce an sphinx docs at pydocs/" 
   echo "Example: ./make_pydocs.sh"
-  exit 1
+  return 1
```
znation: Will this change the behavior of the script? I think it may have been intended to early-exit when `print_help` is called.

scripts/make_wheel.sh
```diff
@@ -45,7 +45,7 @@ print_help() {
   echo
   echo "Produce a local wheel and skip test and doc generation"
   echo "Example: ./make_wheel.sh --skip_test"
-  exit 1
+  return 1
```
znation: Will this change the behavior of the script? I think it may have been intended to early-exit when `print_help` is called.

scripts/make_wheel.sh
```diff
@@ -120,7 +120,7 @@ build_source() {
 
   if [[ $(($python27 + $python35 + $python36)) -gt 1 ]]; then
     echo "Two or more versions of Python specified. Pick one."
-    exit 1
+    return 1
```
znation: Will this change the behavior of the script? I think it may have been intended to early-exit when this error occurs.

scripts/python_env.sh
```diff
@@ -3,12 +3,20 @@
...
+# Check build type parameter
 if [[ -z $1 ]]; then
         echo "build type must be specified. "
         echo "Usage: $0 [debug | release ]"
-        return
+        return 1
 fi
+BUILD_TYPE=$1
+BUILD_TYPE=`echo $BUILD_TYPE | tr "[A-Z]" "[a-z]"` # convert to lower case
+if [ $BUILD_TYPE != "debug" ] && [ $BUILD_TYPE != "release" ]; then
```
znation: I think you'll want quotes on the left hand side of each string comparison as well, like `"$BUILD_TYPE" != "debug"`, in case the $BUILD_TYPE variable is set to empty string (which will cause a syntax error in the absence of quotes).

scripts/run_cpp_tests.py
```diff
@@ -69,7 +69,7 @@
   cmd = 'ctest --output-on-failure -j {0} --schedule-random -R "({1})"'.format(args.j, '|'.join(runtests))
   if args.dry_run:
     print('Dry run requested! Proposed ctest command:', cmd)
-    exit()
+    return
```
znation: Seems like this would change the behavior of the script (it was intended to early-exit here).

scripts/run_cpp_tests.py
```diff
@@ -91,4 +91,4 @@
   # Save to cache
   pickle.dump(cache, open(args.cache_file, "wb" ))
 
-  exit(exit_code)
+  return exit_code
```
znation: Is `return` the correct way to exit Python with an exit code? I thought it was [sys.exit](https://stackoverflow.com/questions/285289/exit-codes-in-python).

src/unity/python/turicreate/_sys_util.py
```diff
@@ -64,7 +63,8 @@ def make_unity_server_env():
 
     # Set mxnet envvars
     if 'MXNET_CPU_WORKER_NTHREADS' not in env:
-        num_workers = min(2, int(env.get('OMP_NUM_THREADS', _sys_info.NUM_CPUS)))
+        from multiprocessing import cpu_count
```
znation: Was the `minipsutil` thing plumbed through to `sys_info` before (is that where `NUM_CPUS` was coming from)? If so, is there any chance that `multiprocessing.cpu_count` would give a different number of CPUs? Or was the now-defunct server-client architecture the only reason we had the CPU counting logic on the C++ side before?

TobyRoseman: Yes, `minisputil` was plumbed through to `sys_info`.

If I recall correctly, `minisputil` is remnant of jobs and task. 

I think this should be safe; I would trust something in the python standard library to give the right answer, more than I would trust the code in `minisputil`. It's only anecdotal evidence but they give the same answer on machine.

scripts/python_env.sh
TobyRoseman: Fixed. I didn't know that would be a syntax error. Thanks.

TobyRoseman:
@znation - Thanks for the review. You're right the changing of `exit` to `return` is not what should be done. I have removed that commit. Thanks for catching this error.

PR #584, opened 2018-05-11
Fix recommender model load from 4.0

znation:
Not sure how this happened, but model load for recommender broke between
4.0 and 4.1. This seems to fix it, but the root cause is still unclear.

Models saved in 4.0 use vector<vector<double>> for new_user_seed_items.
Models saved in 4.1+ use vector<pair<size_t, double>>.

Possibly something to do with flexible_type automatic conversion?


znation:
After chatting with @hoytak in person, I plan to flip the try/catch blocks so the newer-model case doesn't take a perf hit on exception handling.

znation:
Changed got merged with #623.

PR #589, opened 2018-05-14
Ternary Logic & and | operators

ylow:
The current scheme of missing value propagation can lead to very
unexpected results. For instance,

    a = SArray([0,1,None])
    res = a[(a > 0) | (a == None)]

While I expect res to contain [1,None], this is currently not the case.
It will only return [1]. This is because "a > 0" propagates
None, and | propagates None.

This PR implements instead the three-value logic described in
https://en.wikipedia.org/wiki/Three-valued_logic#Kleene_and_Priest_logics

Essentially, instead of naive propagation of missing values, & and |
should following these truth tables, where '-' is "unknown".

    and | F - T
    -----------
      F | F F F
      - | F - -
      T | F - T

    or | F - T
    ----------
     F | F - T
     - | - - T
     T | T T T

With these truth tables, the above code now works as expected.

This is a fundamental behavioral change and has a chance of breaking
people's code. On the other hand it is likely that if people's code
changes behavior with this PR, it is likely that their code is currently
not doing what they expect.


znation:
@ylow since this does potentially break existing code, I'm tagging it as `api-breaking`. Let's keep it open for now and we'll merge at the next opportunity for API-breaking changes.

CMakeLists.txt
```diff
@@ -159,7 +159,7 @@ message(WARNING "OpenMP Libraries were not found")
...
+set(WERROR_FLAGS "-Werror -Wno-error=tautological-undefined-compare -Wno-error=reorder -Wno-error=exceptions -Wno-error=switch -Wno-error=sometimes-uninitialized -Wno-error=unused-lambda-capture -Wno-error=missing-braces -Wno-error=absolute-value -Wno-error=potentially-evaluated-expression -Wno-error=null-arithmetic -Wno-error=format -Wno-error=pessimizing-move -Wno-error=comment -Wno-error=main -Wno-error=constant-conversion -Wno-error=deprecated-declarations -Wno-error=return-type -Wno-error=inconsistent-missing-override -Wno-error=overloaded-virtual -Wno-error=unused-private-field -Wno-error=unused-variable -Wno-error=unused-local-typedef -Wno-error=shift-count-overflow")
```
znation: Why do we want to allow shift-count-overflow to not be an error (generally)? If it's for a specific place in the code, let's use `#pragma clang diagnostic` to turn it off for that line.

PR #597, opened 2018-05-19
Append fixes

ylow:
Implemented compaction heuristics for sframe/sarray append so that
recursive appends, or continuous appends of small sframe/sarrays do not
result in poorly organized SFrames.

The basic mechanism of the compaction process is two methods, a
sarray_fast_compact and a sarray_compact method. sarray_fast_compact
will look for contiguous segments with less than
FAST_COMPACT_BLOCKS_IN_SMALL_SEGMENT (8 by default) number of blocks,
and compact them into a single segment.

The sarray_compact method will try to perform a fast compaction, and if
unable to bring it below a threshold (SFRAME_COMPACTION_THRESHOLD =
256), will perform a full rewrite of the entire sarray.

SFrame compact is similar.

In addition, a block writing option ("disable_padding") is pushed
all the way through to sframe/sarray. This option is enabled when
small sframe/sarrays (things which fit in memory) are written.

### DIFF
--- a/src/sframe/sframe_compact.cpp
+++ b/src/sframe/sframe_compact.cpp
@@ -1,0 +1,59 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+#include <sframe/sframe.hpp>
+#include <sframe/sframe_compact.hpp>
+namespace turi {
+
+bool sframe_fast_compact(const sframe& sf) {
+  bool ret = false;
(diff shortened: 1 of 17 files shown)


src/sframe/sframe_compact.cpp
```diff
@@ -0,0 +1,59 @@
...
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+#include <sframe/sframe.hpp>
+#include <sframe/sframe_compact.hpp>
+namespace turi {
+
+bool sframe_fast_compact(const sframe& sf) {
+  bool ret = false;
+  for (size_t i = 0;i < sf.num_columns(); ++i) {
```
hoytak: Could this loop be done in parallel?

Issue #633, opened 2018-06-07
Text classification has wrong instruction


avlaskin:
1. We trained a text classification model using:
`model = tc.sentence_classifier.create(train, target_key, features=predictors)`

2. At the end of the trining we got a message:
`
TERMINATED: Iteration limit reached.
This model may not be optimal. To improve it, consider increasing `max_iterations`.
`
3. We tried to retrain the model using following parameters:
`model = tc.sentence_classifier.create(train, target_key, features=predictors, max_iterations=20)`
got an error:
`TypeError: create() got an unexpected keyword argument 'max_iterations'`

Essentially the instruction is wrong. Version of Turi is the following:
turicreate-4.3.2-cp36-cp36m-macosx_10_12_intel.macosx_10_12_x86_64.whl

srikris:
Thanks for catching the issue. We will add this parameter to the sentence classifier.

znation:
Fixed with #637.

PR #637, opened 2018-06-07
Add fix for issue #633

avlaskin:
This small PR will add a parameter _max_iterations_ to sentence classifier which is built on top of logistic classifier. Former has this optional parameter in create function. 

There is one more consideration: 
LogisticClassifier has a parameter description as following

> max_iterations : **float**, optional

which we believe is a mistake as it is generally should be **int** number.


srikris:
Thanks for taking a stab at this!

znation:
Thanks for the fix @avlaskin!

PR #640, opened 2018-06-08
Build from source issues on macOS (isysroot, tcmps)

gustavla:
- Explicitly set isysroot on macOS
- More explicitly find and add all frameworks used in tcmps
- More explicitly find and add foundation for `platform`

These are the changes that gave me a successful build, but I do not guarantee it is the minimal set of changes. I added the isysroot last and maybe that would have fixed that tcmps was finding the wrong SDK in the first place.  I'm not sure if the platform change is related to my issues at all. I left it all in since it's probably good to be more explicit about library discovery and this version is so far the only one I can say with certainty builds to completion.


src/platform/config/CMakeLists.txt
```diff
@@ -8,8 +8,20 @@ project(config)
 
 if(APPLE)
 
-  find_library(foundation_config NAMES Foundation REQUIRED)
-  message("Foundation found at ${foundation_config}.")
+    #find_library(foundation_config NAMES Foundation REQUIRED)
+    #message("Foundation found at ${foundation_config}.")
+
+  # We need to get the proper SDK paths in the build
+  execute_process(
```
znation: This is repetitive with logic in the top level CMakeLists.txt. We should keep the SDK path lookup in a single place, and use a variable to refer to it. We also should keep macOS/iOS framework search paths consistent (such that if we're doing an iOS build, it should use iOS SDK path everywhere).

CMakeLists.txt
```diff
@@ -111,6 +111,13 @@ if(TC_BUILD_IOS)
   )
   SET(CMAKE_OSX_SYSROOT "${IOS_ISYSROOT}")
   SET(COMPILER_FLAGS "${COMPILER_FLAGS} -Wno-c++11-narrowing -isysroot ${IOS_ISYSROOT}")
+elseif(APPLE)
+  EXECUTE_PROCESS(
```
znation: We should probably refactor this whole block to share logic with the `xcrun` command around line 735.

CMakeLists.txt
```diff
@@ -111,6 +111,13 @@ if(TC_BUILD_IOS)
...
+    COMMAND xcrun --sdk macosx --show-sdk-path
+    OUTPUT_VARIABLE MACOS_ISYSROOT
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+  SET(CMAKE_OSX_SYSROOT "${MACOS_ISYSROOT}")
```
znation: [This post](https://stackoverflow.com/a/10170214) suggests that just setting `CMAKE_OSX_SYSROOT` should be sufficient to get correct framework search paths. With this fix, I think the `find_library` commands elsewhere may be able to be simplified to use default paths.

src/unity/toolkits/tcmps/CMakeLists.txt
```diff
@@ -1,6 +1,34 @@
 project(unity_toolkits)
 
 if(APPLE AND HAS_MPS AND NOT TC_BUILD_IOS)
+    # We need to get the proper SDK paths in the build
+    execute_process(
```
znation: Same comment; duplicated with logic above. Let's see if we can get a single top level CMAKE_OSX_SYSROOT to result in the correct paths for everything else. The `find_library` calls are probably still needed, but using default paths should do the right thing now that that variable is set.

PR #657, opened 2018-06-12
Single images should work with `load_images` now

shantanuchhabra:


src/unity/lib/image_util.cpp
```diff
@@ -230,13 +235,41 @@ std::vector<std::string> get_directory_files(std::string url, bool recursive) {
...
  * Construct an sframe of flex_images, with url pointing to directory where images reside. 
  */
 std::shared_ptr<unity_sframe> load_images(std::string url, std::string format, bool with_path, bool recursive,
                                           bool ignore_failure, bool random_order) {
     log_func_entry();
-    std::vector<std::string> all_files = get_directory_files(url, recursive);
+
+    std::vector<std::string> all_files;
+    if (has_ending(url, std::string(".jpg")) 
+      || has_ending(url, std::string(".png"))) {
```
TobyRoseman: we also support ".jpeg".

src/unity/lib/image_util.cpp
```diff
@@ -12,6 +12,11 @@
 #include <sframe_query_engine/util/aggregates.hpp>
 #include <unity/lib/toolkit_function_macros.hpp>
 
+#define PATH_IS_ABSOLUTE(s) ((s.length() > 1) && (s[0] == '/'))
+#ifdef _WIN32
+#include <windows.h>
+#define PATH_IS_ABSOLUTE(s) (!PathIsRelative(s))
```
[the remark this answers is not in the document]
znation: According to the [API docs](https://msdn.microsoft.com/en-us/library/windows/desktop/bb773660(v=vs.85).aspx), we should be including at least one more header here. Though maybe it works without that for some reason.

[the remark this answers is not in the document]
znation: We should probably just use [boost filesystem](https://www.boost.org/doc/libs/1_51_0/libs/filesystem/doc/reference.html#path-query) `is_absolute` to do this and not worry about cross-platform code here.

src/unity/lib/image_util.cpp
```diff
@@ -230,13 +235,41 @@ std::vector<std::string> get_directory_files(std::string url, bool recursive) {
   return ret;
 }
 
+
+bool has_ending (std::string fullString, std::string ending) {
```
znation: To scope this function to just this `.cpp` file (internal linkage), mark it as `static`.

src/unity/lib/image_util.cpp
```diff
@@ -230,13 +235,41 @@ std::vector<std::string> get_directory_files(std::string url, bool recursive) {
...
+    if (fullString.length() >= ending.length()) {
```
znation: I think with [rfind](http://en.cppreference.com/w/cpp/string/basic_string/rfind) the body of this function can be simplified to:

```
return fullString.rfind(ending) == fullString.length() - ending.length();
```

src/unity/lib/image_util.cpp
znation: And the case-sensitive variants of them. You may want to make `has_ending` case insensitive always (though the name should probably change at that point), or make these comparisons case insensitive (if you think `has_ending` needs to be case sensitive generally).

TobyRoseman:
I would really like to see a unit test here. Should be simple, you could add something like:

```python
sf5 = image_analysis.load_images(image_url_dir, "auto", False, False)
self.assertEqual(sf5.num_columns(), 1)
self.assertEqual(sf5.num_rows(), 1)
```

Around line 158 of `src/unity/python/turicreate/test/test_image_type.py`.

PR #658, opened 2018-06-12
Added userguide docs for Custom Recommender Model for iOS 

shantanuchhabra:


userguide/SUMMARY.md
```diff
@@ -16,6 +16,7 @@ This is the User Guide for Turi Create
     * [Recommender systems](recommender/README.md)
         * [Using models](recommender/using-trained-models.md)
         * [Choosing a model](recommender/choosing-a-model.md)
+        * [Custom Recommender Model for iOS Applications](recommender/custom-model-ios.md)
```
srikris: Simply call this export to CoreML (like all other user guide chapters)

userguide/recommender/README.md
```diff
@@ -93,4 +93,6 @@ The following sections provide more information about the recommender model:
 - [Choosing a model](choosing-a-model.md)
   * data you might encounter (implicit or explicit)
   * types of models worth considering (item-based similarity, factorization-based models, and so on).
+- [Custom Recommender Model for iOS Applications](custom-model-ios.md)
```
srikris: Same as above

userguide/recommender/custom-model-ios.md
```diff
@@ -0,0 +1,50 @@
+# Custom Recommender Model for iOS Applications
+
+The Turi Create Recommender Model is also available for use in your iOS/macOS
```
srikris: This is only available on iOS 12+ and macOS 10.14+ . Please make that clear.

userguide/SUMMARY.md
znation: The chapter name is a bit verbose. How about "Deployment to Core ML", to match the other toolkits that have a section like this?

userguide/recommender/README.md
znation: Same comment as above.

userguide/recommender/custom-model-ios.md
```diff
@@ -0,0 +1,50 @@
+# Custom Recommender Model for iOS Applications
```
znation: Same comment as above.

userguide/recommender/custom-model-ios.md
```diff
@@ -0,0 +1,50 @@
...
+
+The Turi Create Recommender Model is also available for use in your iOS/macOS
+Applications as a custom model via exporting to Core ML. After creating your model in
```
znation: Applications shouldn't be capitalized? And I think it should be "apps".

userguide/recommender/custom-model-ios.md
```diff
@@ -0,0 +1,50 @@
...
+Turi Create, you need to export your model to Core ML by leveraging the simple
```
znation: Simplify the wording here. "After creating your model in Turi Create, save it in Core ML format by calling `export_coreml` as follows".

userguide/recommender/custom-model-ios.md
znation: Note first and foremost that this is a beta feature (ask Aaron for wording to match the other places we call out beta-only features). For now, the user guide is dual-purpose, shared across 4.x and 5.x.

userguide/recommender/custom-model-ios.md
```diff
@@ -0,0 +1,50 @@
...
+
+```python
+# assume my_recommender is the trained Turi Create Recommender Model
+my_recommender.export_coreml("MyRecommender.mlmodel")
+```
+
+After you drag and drop the exported Core ML model in your iOS Application, it will  look something like this:
+![Core ML Model Screenshot in Xcode](xcode-shot.png)
+
+As you can see above, we need to give the model two inputs, `interactions` and `k`.
```
srikris: You are missing the instructions to include the binary into the app and where to get it from and all the instructions on how to link the binary to your app (i.e adding the dylib into a new phase etc.)

userguide/recommender/custom-model-ios.md
```diff
@@ -0,0 +1,50 @@
...
+```swift
+let model = MyRecommender();
+```
+
+* We now construct the two inputs we need to give to our model to get back some
+predictions.
+
+```swift
+let interactions = NSMutableDictionary();
+interactions[242] = 1.0; // this signifies that the user interacted with item 242.
```
znation: You may want to give an example with explicit ratings (I think we got that working? If not don't worry about it).

userguide/recommender/custom-model-ios.md
```diff
@@ -0,0 +1,50 @@
...
+```swift
+guard let modelOutput = try? model.prediction(input: modelInput) else {
+    fatalError("Unexpected runtime error.");
+}
+print(modelOutput.probabilities);
+print(modelOutput.recommendations);
+
+```
+
+In your app, you would probably have more complex inputs to the model and also want to
```
znation: I'm not sure what this paragraph means. Hopefully the inputs to the model, and outputs from the model, stay consistent regardless of what the rest of the app is doing (this gives me the impression the model itself changes). If what you are trying to imply is that this code sample doesn't comprise an entire app, and the apps that use recommender models can vary across lots of different use cases, we may want to give some examples of what those use cases are.

userguide/SUMMARY.md
znation: Note: they are called "Deployment to Core ML"

userguide/recommender/coreml-deployment.md
```diff
@@ -0,0 +1,70 @@
...
+
+```python
+# assume my_recommender is the trained Turi Create Recommender Model
+my_recommender.export_coreml("MyRecommender.mlmodel")
+```
+
+After you drag and drop the exported Core ML model in your iOS app, it will look something like this:
+![Core ML Model Screenshot in Xcode](xcode-shot.png)
+
+Before you can go ahead and use the recommender model in your app, you must
```
znation: Not the right way for most people to get this file. Please point people to the [GitHub releases page](https://github.com/apple/turicreate/releases), where we provide these dylibs starting with 5.0b1.

userguide/recommender/coreml-deployment.md
```diff
@@ -18,15 +18,10 @@ my_recommender.export_coreml("MyRecommender.mlmodel")
...
-create the `targets/libTuriCreate.dylib` file from the root directory of the 
-Turi Create repository by running the following command from your terminal:
-
-```bash
-./build_capi.sh -i -j 8 -r
-```
-
-After the C API build completes, you can drag and drop the `libTuriCreate.dylib` file from the `targets` directory of the root Turi Create repository into your Xcode project. 
+Before you can go ahead and use the recommender model in your app, you must
+drag and drop the `libRecommender-arm64.dylib` available on the
```
znation: Maybe clarify which library is for which type of app? For a mac app, or for iOS Simulator, the x86_64 one would be used. For an iOS app, the arm64 one would be used.

userguide/recommender/coreml-deployment.md
```diff
@@ -0,0 +1,69 @@
...
+# assume my_recommender is the trained Turi Create Recommender Model
+my_recommender.export_coreml("MyRecommender.mlmodel")
+```
+
+After you drag and drop the exported Core ML model in your iOS app, it will look something like this:
+![Core ML Model Screenshot in Xcode](xcode-shot.png)
+
+Before you can go ahead and use the recommender model in your app, you must
+drag and drop the relevant `.dylib` into your app. If you are writing an app 
+for a Mac or an iOS simulator, you would want `libRecommender-x86_64.dylib`.
```
srikris: Per my earlier comment. You didn't mention how to exactly do this step. You have to drag and drop it and then make sure its in the copy phases. Attach a screenshot of that.

PR #662, opened 2018-06-14
add missing step to fix type error in userguide/text/analysis.md

zmcartor:
Within this section, following the sample code produces an error because `SArray.dict_trim_by_values` expects the `dtype` to be `dict`, not `str`.

This change adds a step to first call `turicreate.text_analytics.count_words` to convert the `SFrame dtype=str` into a ` dtype=dict` before proceeding with the `trim_by_keys` and `stopwords` example.


TobyRoseman:
Thanks @zmcartor

PR #676, opened 2018-06-14
Docfix 452

shantanuchhabra:
Just more description for TC viz


shantanuchhabra:
Extraneous changes made their way in here.

Issue #542, opened 2018-04-18
Link to recommender userguide docs is incorrect


shantanuchhabra:
In the screenshot attached, the link highlighted by "recommender systems chapter of the User Guide" goes to https://apple.github.io/turicreate/docs/userguide/recommender/README.html which gives a 404. Instead, it should go to https://apple.github.io/turicreate/docs/userguide/recommender/

<img width="925" alt="screen shot 2018-04-17 at 10 37 28 pm" src="https://user-images.githubusercontent.com/14062641/38913517-fb393f20-428f-11e8-9baa-6ad00bb3a97b.png">

PR #678, opened 2018-06-15
Fix #452

### LINKED ISSUE #452: Description of help(tc.visualization) is incomplete
Need to add more information in the header of __init__.py in the visualization directory.

Also, investigate if it's possible to add the exposed functions to the "Package Contents" section

shantanuchhabra:


src/unity/python/turicreate/visualization/__init__.py
```diff
@@ -3,6 +3,121 @@
...
+# 	provided in `turicreate.visualization.set_target` (defaults to 'auto').
+# 
+# ====
+# save
+# ====
+#
+# 	A method for saving the Plot object in a vega representation
+#
+#   Parameters
+#	----------
```
srikris: This won't render correctly in Sphinx. Can you check how this will render in bumpy docs.

src/unity/python/turicreate/visualization/__init__.py
```diff
@@ -3,6 +3,121 @@
...
+#   >>> plt.save('vega_spec.json')
+#
+#   We can also save the vega representation of the plot without data:
+#   >>> plt.save('vega_spec.json', False)
+#
+#   We can save the plot as a PNG/SVG using:
+#   >>> plt.save('test.png')
+#   >>> plt.save('test.svg')
+#
+# ====
```
srikris: This won't render correctly at all. Can you check how this renders.

src/unity/python/turicreate/visualization/__init__.py
```diff
@@ -3,6 +3,121 @@
...
+#   Notes
+#   -----
+#   - A plot object is returned via the SFrame's or SArray's .plot() method
+#
+#   Examples
+#   --------
+#   Suppose 'plt' is an Plot Object
+#
+#   We can view it using:
+#   >>> plt.show()
```
srikris: This won't render as code. Can you check

src/unity/python/turicreate/visualization/__init__.py
shantanuchhabra: Where would these render though? There are separate, correctly formatted docs for the API docs, next to each individual function. This commit was to make the response to `help(tc.visualization)` more useful (from the command line)

srikris:
These are also rendered by our API Docs.

src/unity/python/turicreate/visualization/__init__.py
znation: The help text and API docs should be the same (and come from the same place).

src/unity/python/turicreate/visualization/__init__.py
```diff
@@ -3,6 +3,121 @@
 #
 # Use of this source code is governed by a BSD-3-clause license that can
 # be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+#
+# Turi Create Visualization allows you to create Plot objects and 
+# manipulate them using various functionalities, including
```
znation: Remove "using various functionalities" - this sounds like fluff.

src/unity/python/turicreate/visualization/__init__.py
```diff
@@ -3,6 +3,121 @@
...
+# saving them (.save) and viewing them (.show). Moreover, Turi Create 
+# Visualization also allows the user to call `show` on data sources like
```
znation: Be consistent - are we calling the audience "you" or "the user"? I think the former.

src/unity/python/turicreate/visualization/__init__.py
znation: I think you'll want to put the help/API text in triple quotes at the top of the file (perhaps above the license, if that's what is needed to get it to show up correctly).

shantanuchhabra:
This PR should fix both #452 and #542

shantanuchhabra:
Please see #710

Issue #340, opened 2018-03-06
how to save image_data with bounding boxes as '.jpg'  in object_detection task


erinliu0428:
In the object_detection task, we use
`data['image_with_ground_truth'] =  tc.object_detector.util.draw_bounding_boxes(data['image'], data['annotations'])`
to create a  turicreate.Image  type data which contains image  data with bounding boxes. So  how to save the turicreate.Image  type  data  as  '.jpg' or '.png' ?  The file '~/turicreate/data_structures/image.py '   doesn't  have a function like this

erinliu0428:
I know already.We can use
`img = data['image_with_ground_truth']
 img._to_pil_image.save('test.jpg')`

gustavla:
Hi @erinliu0428! This is definitely missing from `tc.Image` (along with other basic functionality that we hope to add). I think what you are describing works, but it is not part of the official API, so it could break at any new version without warning. A more API-stable approach would be 

```python
from PIL import Image
Image.fromarray(img.pixel_data).save('test.jpg')
```

I will leave this open and mark it as a feature request, since there should be an even simpler way.

erinliu0428:
@gustavla Thank you for your attention !
And I have another basic need.After a object_detection model created, I want to predict some new pics. I need the new pics are parted from a video directly without be saved in a path.So if there is a way we can load video directly, or after loading video and parting images  by other toolkits  convert this type of image to tc.Image?   By the way,I'm curious about how tc.Image._image_data is generated. I thought if two pics are same size ,the shape of ._image_data  should be the same .But it turns out wrong.

srikris:
That's actually a great feature request. I filed it at #342 and #343 to track this.

shantanuchhabra:
Fixed by #709

baizhenmao95:
@znation Thank you very much for your reply in #88 for "save the images at full size"; but when I tried the code:
`from PIL import Image
Image.fromarray(img.pixel_data).save('test.jpg')`
There was an AttributeError: 'SArray' object has no attribute 'pixel_data'
I also tried the code:
`img = data['image_with_ground_truth']
img._to_pil_image.save('test.jpg')`
but there was also the AttributeError: 'SArray' object has no attribute '_to_pil_image'

Is there any other things that I should do in advance?
Thank you!

znation:
[...]

PR #707, opened 2018-06-18
Feature Request #340

shantanuchhabra:
Don't let the branch name confuse you, this is a PR to add the first feature request mentioned in #340


src/unity/python/turicreate/data_structures/image.py
```diff
@@ -194,6 +194,10 @@ def _to_pil_image(self):
         from PIL import Image as _PIL_Image
         return _PIL_Image.fromarray(self.pixel_data)
 
+    def save(self, filename):
+        from PIL import Image as _PIL_Image
```
znation: To avoid duplicating the above code, call `self._to_pil_image().save(filename)`.

znation:
Your commit history has lots of unrelated commits in it.

src/unity/python/turicreate/data_structures/image.py
shantanuchhabra: I was considering that but I saw @gustavla 's comment at #340 where he said that it might break without warning at any new version and a more API-stable approach would be to do:
```
from PIL import Image
Image.fromarray(img.pixel_data).save('test.jpg')
```

shantanuchhabra:
Fixed the appearance of unrelated commits and opened a new PR #709

PR #709, opened 2018-06-18
Feature Request #340 completed

shantanuchhabra:
added save functionality to the src/unity/python/turicreate/data_structures/image.py Image class

### DIFF
--- a/src/unity/python/turicreate/data_structures/image.py
+++ b/src/unity/python/turicreate/data_structures/image.py
@@ -194,6 +194,10 @@ class Image(object):
         from PIL import Image as _PIL_Image
         return _PIL_Image.fromarray(self.pixel_data)
 
+    def save(self, filename):
+        from PIL import Image as _PIL_Image
+        _PIL_Image.fromarray(self.pixel_data).save(filename)
+
     def show(self):
         """
         Displays the image. Requires PIL/Pillow.


src/unity/python/turicreate/data_structures/image.py
```diff
@@ -194,6 +194,10 @@ def _to_pil_image(self):
         from PIL import Image as _PIL_Image
         return _PIL_Image.fromarray(self.pixel_data)
 
+    def save(self, filename):
+        from PIL import Image as _PIL_Image
```
znation: Same comment as last time - please use `_to_pil_image` from within this function, so we are sharing code wherever possible (i.e. in the future, the way we convert to a PIL image may change, and if we have to change it in two places the logic may diverge).

shantanuchhabra: Yep, thanks! Done.

gustavla:
A bit late here since this merged, but this is what I wrote on the closed PR:

Turi Create links directly with libpng and libjpeg, and we recently changed `pixel_data` to rely on this instead of PIL. In the spirit of avoiding unnecessary PIL dependence, I think we should use our own direct linkage here as well.

We do have `write_image` in `src/image/io.cpp` to do this, so we just need to expose it appropriate (look at `_extensions.load_image` for instance). Note, there is also a potential for avoiding a decode/encode, if the requested file format is the same as the internal representation inside the SFrame.

Also, should format only be implicit in the filename? What should happen if `filename` does not end with `.png`/`.jpg`/`.jpeg`?

gustavla:
Also, this function should have a proper docstring.

znation:
Good comments @gustavla, and thanks for opening another issue to track that!

PR #710, opened 2018-06-18
Bugfix for #452 and #542

shantanuchhabra:
Picking up from #678 but with a cleaner commit history :-)

### DIFF
--- a/src/unity/python/turicreate/visualization/__init__.py
+++ b/src/unity/python/turicreate/visualization/__init__.py
@@ -4,111 +4,35 @@
 # Use of this source code is governed by a BSD-3-clause license that can
 # be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
 """
-Turi Create Visualization allows you to create, view, and save Plot objects.
+Turi Create Visualization allows you to create, view, and save Plot objects. 
+The Plot class is used to create an immutable object representation of a 
+visualization and has the following methods:
+- `show`: A method for displaying the Plot object.
+- `save`: A method for saving the Plot object in a vega representation.
 
-class Plot
+Turi Create Visualization also allows you to call `show` 
+on data sources like SFrames and SArrays:
 
-    An immutable object representation of a visualization.
+- `show`: Plots the data in `x` on the X axis and the data in `y` on the Y axis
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/visualization/__init__.py
```diff
@@ -4,111 +4,35 @@
...
+
+- scatter(x, y, xlabel, ylabel, title)
+- categorical_heatmap(x, y, xlabel, ylabel, title)
+- heatmap(x, y, xlabel, ylabel, title)
+- box_plot(x, y, xlabel, ylabel, title)
+- columnwise_summary(sf)
+- histogram(sa, xlabel, ylabel, title)
+- item_frequency(sa, xlabel, ylabel, title)
+
+To know more, please see the API docs at https://apple.github.io/turicreate/docs/api/
```
znation: I would remove this line, or rephrase it to reference the docstrings of the referenced classes/methods. Since the user is already in the UX flow of looking up docstrings on Python objects, they'll probably want to stay in that flow as they explore. How about "For more detailed information, please refer to the docstrings for each of those classes/methods."

src/unity/python/turicreate/visualization/__init__.py
```diff
@@ -4,111 +4,35 @@
...
+Turi Create Visualization allows you to create, view, and save Plot objects. 
+The Plot class is used to create an immutable object representation of a 
+visualization and has the following methods:
+- `show`: A method for displaying the Plot object.
+- `save`: A method for saving the Plot object in a vega representation.
+
+Turi Create Visualization also allows you to call `show` 
+on data sources like SFrames and SArrays:
+
+- `show`: Plots the data in `x` on the X axis and the data in `y` on the Y axis
```
znation: This description of `show` (looks to be copy/pasted from `tc.show`) doesn't fit the paragraph above (about calling it on SFrames/SArrays). I would rephrase the paragraph above, and this detailed method description, into an overall description of what the `show` method implies when it exists on a data structure or namespace. Maybe something like: "The show method, available on data structures in Turi Create as well as at the top level in the namespace, streamlines the process of creating and subsequently calling `show` on an automatically-selected Plot object."

PR #713, opened 2018-06-18
Softened metadata update requirements for CoreML models.

hoytak:
Currently, the add_metadata field for the coreml model export did not
allow updating of metadat fields without having some baseline fields
filled in.  This is problematic when additional fields had to be updated
later.

### DIFF
--- a/src/unity/toolkits/coreml_export/coreml_export_utils.cpp
+++ b/src/unity/toolkits/coreml_export/coreml_export_utils.cpp
@@ -15,21 +15,33 @@ namespace turi {
 
 void add_metadata(CoreML::Specification::Model& model_spec,
                   const std::map<std::string, flexible_type>& context) {
-  DASSERT_TRUE(context.count("short_description"));
-  DASSERT_TRUE(context.count("version"));
-  DASSERT_TRUE(context.count("class"));
-  DASSERT_TRUE(!context.count("model_fields") || context.at("model_fields").get_type() == flex_type_enum::DICT);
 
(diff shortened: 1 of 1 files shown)


src/unity/toolkits/coreml_export/coreml_export_utils.cpp
```diff
@@ -15,21 +15,33 @@ namespace turi {
...
+    }
+
+    if (context.count("license")) {
+      metadata->set_license(context.at("license").to<flex_string>());
+    }
+
+    if (context.count("user_defined")) {
+      DASSERT_TRUE(context.at("user_defined").get_type() == flex_type_enum::DICT);
+
+      auto user_defined = metadata->mutable_userdefined();
```
znation: I have an ask, as part of this change: can we also set a user defined field named `Turi Create Version` to the current version of TC (regardless of whether the caller passed in `user_defined`)? That would help in determining whether a model came from TC. Right now we only (sometimes?) write that into the description field, but users can easily and intentionally overwrite that and then we won't know if the model came from TC, or what version it came from.

znation: CC @abhishekpratapa

hoytak: I'm fine with this, but let's make it part of another change instead of this PR.

hoytak: Opened a new issue to track that.

znation: Sounds good, thanks!

Issue #687, opened 2018-06-15
Documentation link for visualization.set_target results in a 404


afranklin:
Observed Behavior:
1. Go to this page: https://apple.github.io/turicreate/docs/userguide/vis/
2. Scroll to the Show section and click the link for turicreate.visualization.set_target
3. Result:  404
https://apple.github.io/turicreate/docs/api/generated/turicreate.visualization.set_target.html

Issue #686, opened 2018-06-15
typo in linear-regression docs


afranklin:
Go to:
https://apple.github.io/turicreate/docs/userguide/supervised-learning/linear-regression.html

Look at the paragraph that describes how to access the coefficients

Note that it describes the previous version of Turi Create 

It should say:

selected_words_model.coefficients

PR #717, opened 2018-06-19
Misc doc fixes

afranklin:
Simple fixes for the following issues:

- https://github.com/apple/turicreate/issues/687
- https://github.com/apple/turicreate/issues/686
- https://github.com/apple/turicreate/issues/542

### DIFF
--- a/userguide/supervised-learning/linear-regression.md
+++ b/userguide/supervised-learning/linear-regression.md
@@ -115,8 +115,8 @@ shows the largest positive and negative coefficients. After a model is
 created, we can access the coefficients as follows:
 
 ```python
-coefs = model['coefficients']
-print(coefs)
+coefs = model.coefficients
+coefs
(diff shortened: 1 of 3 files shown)


userguide/supervised-learning/linear-regression.md
```diff
@@ -115,8 +115,8 @@ shows the largest positive and negative coefficients. After a model is
 created, we can access the coefficients as follows:
 
 ```python
-coefs = model['coefficients']
-print(coefs)
+coefs = model.coefficients
+coefs
```
TobyRoseman: I think you probably still want the `print` and the parentheses. Without that, the below output will not be displayed unless you're in interactive mode of python/ipython.

afranklin: Thanks - added back the print() statement

PR #719, opened 2018-06-19
Image similarity vision feature print screen

TobyRoseman:
Fixes unreliable Image Similarity unit test by using a better comparison metric.
Adds vision feature print screen to image similarity (#705)

### DIFF
--- a/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
+++ b/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
@@ -53,8 +53,8 @@ def create(dataset, label = None, feature = None, model = 'resnet-50', verbose =
            - "squeezenet_v1.1" : Uses a pretrained squeezenet model.
 
            - "VisionFeaturePrint_Screen": Uses an OS internal feature extractor.
-                                          Only on available on iOS,tvOS 12.0+,
-                                          macOS 10.14+.
+                                          Only on available on iOS 12.0+,
+                                          macOS 10.14+ and tvOS 12.0+.
 
         Models are downloaded from the internet if not available locally. Once
         downloaded, the models are cached for future use.
@@ -215,7 +215,7 @@ class ImageSimilarityModel(_CustomModel):
         from turicreate.toolkits.nearest_neighbors import NearestNeighborsModel
         state['similarity_model'] = NearestNeighborsModel(state['similarity_model'])
         if state['model'] == "VisionFeaturePrint_Screen" and _mac_ver() < (10,14):
-            raise ToolkitError("Can not load model on this operating system. This model uses VisionFeaturePrint_Screen, " \
+            raise ToolkitError("Can not load model on this operating system. This model uses VisionFeaturePrint_Screen, "
                                "which is only supported on macOS 10.14 and higher.")
         state['feature_extractor'] = _image_feature_extractor._create_feature_extractor(state['model'])
         state['input_image_shape'] = tuple([int(i) for i in state['input_image_shape']])
@@ -541,6 +541,7 @@ class ImageSimilarityModel(_CustomModel):
             # neural network.
             BGR_VALUE = _cmt.proto.FeatureTypes_pb2.ImageFeatureType.ColorSpace.Value('BGR')
             DOUBLE_ARRAY_VALUE = _cmt.proto.FeatureTypes_pb2.ArrayFeatureType.ArrayDataType.Value('DOUBLE')
+            INPUT_IMAGE_SHAPE = 299
 
             top_spec = _cmt.proto.Model_pb2.Model()
             top_spec.specificationVersion = 3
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -48,8 +49,13 @@ def create(dataset, label = None, feature = None, model = 'resnet-50', verbose =
         Uses a pretrained model to bootstrap an image similarity model
 
            - "resnet-50" : Uses a pretrained resnet model.
+
            - "squeezenet_v1.1" : Uses a pretrained squeezenet model.
 
+           - "VisionFeaturePrint_Screen": Uses an OS internal feature extractor.
+                                          Only on available on iOS,tvOS 12.0+,
```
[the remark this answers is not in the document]
TobyRoseman: I believe the tvOS version is also 12.0+

Would the following be more clear? `Only on available on iOS 12.0+, macOS 10.14+, and tvOS 12.0+.`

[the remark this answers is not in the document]
znation: Oh, you're right, it is. Sorry I assumed the earlier one was a typo. I do think it is more clear the way you're suggesting.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -201,6 +214,9 @@ def _load_version(cls, state, version):
         _tkutl._model_version_check(version, cls._PYTHON_IMAGE_SIMILARITY_VERSION)
         from turicreate.toolkits.nearest_neighbors import NearestNeighborsModel
         state['similarity_model'] = NearestNeighborsModel(state['similarity_model'])
+        if state['model'] == "VisionFeaturePrint_Screen" and _mac_ver() < (10,14):
+            raise ToolkitError("Can not load model on this operating system. This model uses VisionFeaturePrint_Screen, " \
```
gustavla: Nit: Note that `\` is not necessary inside parens.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -491,40 +507,84 @@ def export_coreml(self, filename):
...
+            DOUBLE_ARRAY_VALUE = _cmt.proto.FeatureTypes_pb2.ArrayFeatureType.ArrayDataType.Value('DOUBLE')
+
+            top_spec = _cmt.proto.Model_pb2.Model()
+            top_spec.specificationVersion = 3
+            desc = top_spec.description
+
+            input = desc.input.add()
+            input.name = self.feature
+            input.type.imageType.width = 299
+            input.type.imageType.height = 299
```
gustavla: Do we need to repeat 299 here? We already stored this inside `input_image_shape`.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
gustavla: No need to change, just an FYI.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -491,40 +507,84 @@ def export_coreml(self, filename):
...
+            pipeline = top_spec.pipeline
+            scene_print = pipeline.models.add()
+            scene_print.specificationVersion = 3
+            scene_print.visionFeaturePrint.scene.version = 1
+
+            input = scene_print.description.input.add()
+            input.name = self.feature
+            input.type.imageType.width = 299
+            input.type.imageType.height = 299
+            input.type.imageType.colorSpace = BGR_VALUE
```
gustavla: Do all image feature extractors in TC use BGR? I'm looking inside `mlmodel_image_feature_extractor.mm` and that seems to be the case. So are all the MXNet models also trained on BGR?

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
TobyRoseman: Interesting. I didn't know that. I updated here and in the image classifier.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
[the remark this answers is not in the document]
TobyRoseman: No worries. The wording is confusing. I've updated it here and in the image classifier.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
TobyRoseman: Ok. Done.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
TobyRoseman: I'm not sure. I believe the all the Core ML models used by image classification/similarity use BGR. I believe the MxNet models use RGB.

Issue #456, opened 2018-04-09
Initial window size for canvas should be at least as big as the plot that appears


shantanuchhabra:
I tried to make a heatmap and the plot was bigger than the initial canvas size. I would expect the initial canvas to resize based on how big of a plot it's about to render. This is on macOS High Sierra, here's the repro:

```
x = tc.SArray(range(1,3000000))
y = tc.SArray(range(5,3000004))
heat = tc.visualization.heatmap(x,y)
heat.show()
```

PR #722, opened 2018-06-19
Added Resize Functionality to Plots

abhishekpratapa:
- added resize functionality
- limit to 80% of the screen's width and height


abhishekpratapa:
Fixes #456

znation:
@abhishekpratapa This adds a lot of complexity to the code, and I'm not sure it's the right approach. Have we considered making sure all of the Vega specs will fit within the default window size, vs. resizing the window to fit our Vega specs? I think the former may be a better fix, since it's both simpler, and will result in a consistent UX (consistent window size regardless of data structure passed in). Users can always resize/zoom on their own if they want; this issue is about the initial design bug where the Vega spec mismatches the window size (and we chose both of those to begin with).

abhishekpratapa:
Feature request

PR #723, opened 2018-06-19
Added a check for macOS < 10.14 in OD NMS

### LINKED ISSUE #632: Object detection should give a better error message when it fails on <10.14
A macOS version check seems to missing in the detector's `export_coreml` that will make sure it isn't trying to save a NMS model with macOS older than 10.14.

We should error out and offer the following suggestions:
- Upgrade to 10.14
-   Add `include_non_maximum_suppression=False`

We could also automatically fall back to the latter and write a warning, but I don't think I like that option as much.

shantanuchhabra:
Fixes #632


shantanuchhabra:
Addressed by #778

PR #731, opened 2018-06-20
Propagate -mmacosx-version-min=10.12 to external dependencies

### LINKED ISSUE #729: Warnings due to mismatched deployment target when linking with dependencies built without -mmacosx-version-min
We need to propagate -mmacosx-version-min=10.12 to all the external dependencies we build.

nickjong:
I almost got this right the last time. In the macOS case, I had erred on the side of continuing to use CMAKE_OSX_DEPLOYMENT_TARGET, but we need to modify COMPILER_FLAGS to propagate to the external CMake invocations.

Fixes #729

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,8 +84,9 @@ if(APPLE)
   else()
     SET(TC_BASE_SDK macosx)
 
-    # Sets -mmacosx-version-min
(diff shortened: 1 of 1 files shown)


nickjong:
(The only other documented purpose of CMAKE_OSX_DEPLOYMENT_TARGET is to allow CMake to infer a value for CMAKE_OSX_SYSROOT, but we set that manually.)

PR #772, opened 2018-06-26
Added a features parameter to the internal calls of supervised learning.

hoytak:


hoytak:
@znation -- There is some logic in python that should be moved to C++ now that this is here, but this is needed for calling through the C-API layer.

PR #774, opened 2018-06-26
Fixed seeds for internal validation set splitting.

hoytak:


src/unity/toolkits/supervised_learning/automatic_model_creation.cpp
```diff
@@ -67,7 +67,7 @@ std::shared_ptr<supervised_learning_model_base> create_automatic_classifier_mode
 
   // If no validation set and enough training data, create a validation set.
   if(validation_data.empty() && data.size() >= 100) {
-    std::pair<gl_sframe, gl_sframe>  split = data.random_split(.95);
+    std::pair<gl_sframe, gl_sframe>  split = data.random_split(.95, 0);
```
znation: Not sure fixed seed is a good idea here. When splitting, isn't it better to have some real randomness? Fixed seed implies a fixed split for any given dataset, which while giving reproducible results, could give worse-than-average results consistently for some datasets.

srikris: We should not fix the seed. We should make it a parameter in the Python API.

```
seed = None (default, current behavior)
seed = 0 (user can force the seed to be fixed)
```

srikris:
Dup of #774

znation:
@srikris I think you closed this PR as a dupe of itself?

srikris:
@znation Actually this is a dup of #869. That's what I meant to dup it against. Could you verify and close this PR.

znation:
Agreed, this is a dupe of #869.

PR #775, opened 2018-06-26
* Resolve all warnings and remove -Werror exceptions

### LINKED ISSUE #459: warning: control may reach end of non-void function [-Wreturn-type]
We cannot enable -Werror in part because the compiler believes that some functions don't return in all cases.

/Users/jong/Source/turicreate/src/fileio/general_fstream_source.cpp:73:1: warning: 
      control may reach end of non-void function [-Wreturn-type]

At least some cases are related to assert macros that wrap a throw statement in do {} while (0);

### LINKED ISSUE #460: warning: unused variable 'size_list' [-Wunused-variable]
We cannot enable -Werror in part because our report contains some unused variables, e.g.

/Users/jong/Source/turicreate/src/unity/lib/visualization/boxes_and_whiskers.cpp:32:10: warning: 
      unused variable 'size_list' [-Wunused-variable]
  size_t size_list = grouped.size();
         ^

### LINKED ISSUE #462: warning: deprecated declarations
We cannot enable -Werror in part because our code uses deprecated functions, e.g.

/Users/jong/Source/turicreate/test/fileio/general_fstream_test.cxx:25:16: warning: 
      'tmpnam' is deprecated: This function is provided for compatibility
      reasons only. Due to security concerns inherent in the design of
      tmpnam(3), it is highly recommended that you use mkstemp(3) instead.
      [-Wdeprecated-declarations]
      tmpname (tmpnam(NULL)) {
               ^

We should remove our usage of deprecated functions and enable -Werror=deprecated-declarations until we can enable -Werror globally

### LINKED ISSUE #465: warning: implicit conversion of constant (literal)
We cannot enable -Werror in part because of some numeric literals that must be implicitly converted to the constant that they initialize, e.g.

/Users/jong/Source/turicreate/test/fileio/general_fstream_test.cxx:130:20: warning: 
      implicit conversion from 'int' to 'std::__1::basic_string<char,
      std::__1::char_traits<char>, std::__1::allocator<char> >::value_type'
      (aka 'char') changes value from 255 to -1 [-Wconstant-conversion]
        s[2 * i] = 255;
                 ~ ^~~

We should manually convert such literals and adopt -Werror=constant-conversion until we can adopt -Werror globally

### LINKED ISSUE #466: warning:  bool literal returned from 'main' [-Wmain]
We cannot enable -Werror in part because we sometimes "return false" inside main, e.g.
/Users/jong/Source/turicreate/test/fileio/fs.cpp:203:7: warning: bool literal
      returned from 'main' [-Wmain]
      return false;
      ^      ~~~~~

We should return an int instead and build with -Werror=main until we can adopt -Werror globally

### LINKED ISSUE #467: warning: '/*' within block comment [-Wcomment]
We cannot enable -Werror in part because some C-style comments include '/*', e.g.

```
/Users/jong/Source/turicreate/src/unity/lib/visualization/vega_spec.cpp:60:64: warning: 
      '/*' within block comment [-Wcomment]
 * Prepares a raw JSON format string (from one of the vega_spec/*.json files)
                                                               ^
```

We should change the comment style or use pragmas to avoid these warnings, and adopt -Werror=comment until we can adopt -Werror globally

### LINKED ISSUE #469: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
My favorite warning! We cannot adopt -Werror in part because we use std::move in some instances where it's not only redundant but precludes some optimizations, e.g.

```
/Users/jong/Source/turicreate/test/generics/gl_string.cxx:363:11: warning: 
      moving a temporary object prevents copy elision [-Wpessimizing-move]
      v = std::move(gl_string(ev.begin(), ev.end()));
          ^
/Users/jong/Source/turicreate/test/generics/gl_string.cxx:363:11: note: remove
      std::move call here
      v = std::move(gl_string(ev.begin(), ev.end()));
          ^~~~~~~~~~                               ~
```

We should remove these unhelpful std::move invocations and adopt -Werror=pessimizing-move until we can adopt -Werror globally

### LINKED ISSUE #470: warning: format specifies type 'long' but the argument has type 'turi::flex_int' (aka 'long long') [-Wformat]
We cannot adopt -Werror in part because of mismatches between some format specifiers and the values being formatted, e.g.

```
/Users/jong/Source/turicreate/src/sframe/sframe_io.cpp:31:68: warning: format
      specifies type 'long' but the argument has type 'turi::flex_int'
      (aka 'long long') [-Wformat]
  ...= snprintf(cur, space_remaining, "%ld", val.get<flex_int>());
                                       ~~~   ^~~~~~~~~~~~~~~~~~~
                                       %lld
```

We should resolve these issues and adopt -Werror=format until we can adopt -Werror globally

### LINKED ISSUE #471: warning: comparison between NULL and non-pointer ('unsigned char' and NULL) [-Wnull-arithmetic]
We cannot adopt -Werror in part because we sometimes use NULL for things other than testing against pointers, e.g.

```
/Users/jong/Source/turicreate/src/unity/toolkits/coreml_export/xgboost_exporter.cpp:25:53: warning: 
      comparison between NULL and non-pointer ('unsigned char' and NULL)
      [-Wnull-arithmetic]
    is_little_endian = ((unsigned char*)(&test))[0] != NULL;
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~
```

We should probably use 0 in these cases. Where NULL is desired, we probably want nullptr anyway. Then we can adopt -Werror=null-arithmetic, pending global adoption of -Werror

### LINKED ISSUE #472: warning: expression with side effects will be evaluated despite being used as an operand to 'typeid' [-Wpotentially-evaluated-expression]
We cannot adopt -Werror in part because we're called typeid on non-trivial expressions, e.g.
```
/Users/jong/Source/turicreate/test/sframe/sframe_test.cxx:1519:25: warning: 
      expression with side effects will be evaluated despite being used as an
      operand to 'typeid' [-Wpotentially-evaluated-expression] 
       TS_ASSERT(typeid(*parallel_vals[i]) == typeid(*val));
                        ^
```
We should either avoid this usage or use pragmas to explicitly mark this as safe, if applicable. Then we can adopt -Werror=potentially-evaluated-expression or -Werror as appropriate

### LINKED ISSUE #474: warning: taking the absolute value of unsigned type 'bool' has no effect [-Wabsolute-value]
We cannot adopt -Werror in part because we sometimes compute the absolute value of a boolean expression, e.g.
```
/Users/jong/Source/turicreate/test/optimization/optimization_interface_test.cxx:371:17: warning: 
      taking the absolute value of unsigned type 'bool' has no effect
      [-Wabsolute-value]
      TS_ASSERT(abs(stats.residual < 1e-5));
                ^
```
We should remove the unnecessary abs() calls and adopt -Werror=absolute-value or -Werror, as appropriate

### LINKED ISSUE #477: warning: suggest braces around initialization of subobject [-Wmissing-braces]
We cannot adopt -Werror in part because we can sometimes be more explicit when initializing objects, e.g.
```
/Users/jong/Source/turicreate/test/sframe_query_engine/optimizations.cxx:41:23: warning: 
      suggest braces around initialization of subobject [-Wmissing-braces]
      history.insert({n.v[0], n.v[3]});
                      ^~~~~~~~~~~~~~
                      {             }
```
We should adopt the suggested fixes or silence the warnings, so that we can adopt -Werror=missing-braces or -Werror, as appropriate

### LINKED ISSUE #479: warning: lambda capture 'reader' is not used [-Wunused-lambda-capture]
We cannot adopt -Werror in part because some lambdas capture variables without using them, e.g.
```
/Users/jong/Source/turicreate/src/unity/lib/unity_sframe.cpp:1868:33: warning: 
      lambda capture 'reader' is not used [-Wunused-lambda-capture]
    auto getAccordion = [self, &reader, &ew, &column_names, &empty_tz](s...
                                ^
```
We should prune unused captures from all lambdas and then adopt -Werror=unused-lambda-capture or -Werror, as appropriate

### LINKED ISSUE #481: warning: variable 'none_action' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
We cannot adopt -Werror in part because some code paths return uninitialized values, e.g.
```
/Users/jong/Source/turicreate/src/unity/toolkits/ml_data_2/ml_data_missing_values.cpp:24:14: warning: 
      variable 'none_action' is used uninitialized whenever 'if' condition is
      false [-Wsometimes-uninitialized]
  } else if (missing_value_action_str == "impute") {
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
We should ensure that these values are always initialized or silence these warnings via pragmas if safe. Then we can adopt -Werror=sometimes-uninitialized or -Werror, as appropriate

### LINKED ISSUE #483: warning: enumeration value 'ND_VECTOR' not handled in switch [-Wswitch]
We cannot adopt -Werror in part because some switch statements do not handle every case, e.g.
```
/Users/jong/Source/turicreate/src/unity/toolkits/feature_engineering/dict_transform_utils.cpp:117:12: warning: 
      enumeration value 'ND_VECTOR' not handled in switch [-Wswitch]
  switch(v.get_type()) {
           ^
```
These warnings may indicate actual user-facing bugs, since ND_VECTOR was only recently added. We should ensure that all case statements handle all enum values, then adopt -Werror=switch or -Werror as appropriate

nickjong:
This fixes #460, fixes #479, fixes #462, fixes #459, fixes #483, fixes #467, fixes #470, fixes #481, fixes #477, fixes #465, fixes #466, fixes #469, fixes #472, fixes #474, and fixes #471.

* Simplify fix for Linux C++ test build (#595).

This patch resolves a TODO item in commit 818ead2 (Fix Linux C++ test build).
It appears that the undefined symbol errors of the following form:

> undefined reference to `boost::test_tools::tt_detail::print_log_value<decltype(nullptr)>::operator()(std::ostream&, decltype(nullptr))'

were due to the fact that GCC does not automatically emit such overloads unless
the relevant files are compiled with -std=c++11 or greater. (In this case, the
relevant files happen to be external Boost source files.) This patch adds the
-std=c++11 flag to the corresponding cmake build command, which allows us to
avoid defining the function overload ourselves.

See also:

https://svn.boost.org/trac10/ticket/12969

* Fix transposed memset arguments.

A call to memset() in vector_internals had its arguments apparently transposed,
with the length argument coming second instead of third. This patch swaps the
arguments to the correct order.

* Fix compiler warnings (division by zero).

When constructing floating-point infinity values, divide by floating-point zero
rather than integer zero, in order to avoid compiler warnings.

* Fix missing return statements/unreachable code.

- We also add a macro ASSERT_UNREACHABLE(), which is roughly equivalent to
  "assert(false)" but is also able to leverage the existing error-reporting
  and stack-trace machinery.

* Fix function return type (broadcast_queue.hpp).

* Fix signed/unsigned comparisons.

- In some cases involving ordering comparisons, it is important to ensure that
  the operands are within range of the signed type actually used to perform the
  comparison. For this purpose, this patch adds a family of functions called
  "truncate_check" to perform these range checks at runtime. (Of course, if we
  can ensure statically that a given range check is satisfied, then we may
  remove the invocation of truncate_check() and avoid the runtime overhead.)

* Fix unused variables (#460, #479).

* Fix strict-aliasing issues.

* Use boost::filesystem for temporary paths (#462).

- One of the C-API unit tests (capi_models.cxxtest) was attempting to write to a
  path of the form "/var/tmpFILE" instead of "/var/tmp/FILE", which was causing
  the test to fail due to permissions issues on Linux. This patch replaces the
  temporary path with an implementation based on boost::filesystem.

- We also switch to an auto-generated temporary path rather than a hard-coded
  path in the capi_sframe unit test, in order to avoid errors if the current
  user does not have permission to write the path in question.

- Finally, we replace a use of the deprecated function tmpnam(), addressing a
  compiler warning (#462).

* Implement several new features in run_cpp_tests.py.

- Make caching behavior optional, so that we can avoid hashing many large files
  when this is not necessary.

- Add an option to disable random shuffling of the individual tests (i.e., the
  "--schedule-random" flag in ctest). This makes it much easier to compare
  results across multiple runs.

- Add an option to disable expensive tests (i.e., those that typically take
  longer than a few seconds); this is useful to get a quick overall read on
  correctness.

- Display ctest output incrementally via subprocess.Popen, rather than saving it
  all until the end of the run.

* Address compiler warnings for protobuf includes.

- Some auto-generated protobuf source files cause compiler warnings (e.g.,
  -Wunused-function). Existing sources that include these files guard them with
  clang pragmas to disable the warnings, but GCC does not recognize the
  pragmas, causing additional compiler warnings. Instead, we tell CMake to add
  GCC-specific flags to disable the warnings in the command-line invocations
  for the relevant source files.

* Fix deprecation warning for Boost header (template_arity_spec.hpp).

* Prevent excessive debugging output from Boost build ("common.copy", "common.mkdir", etc.).

* Fix CMake script to restrict C++-specific compiler flag to C++ source files.

Builds of external source files were triggering compiler warnings due to the
presence of the "-fvisibility-inlines-hidden" flag, which is only valid for C++
source (not plain C). This patch restricts this flag to C++ source files only.

* Add missing cases in conditionals/switch-case blocks (#459, #483).

* Fix order of operations (xor vs. equals).

* Fix warnings regarding C++ comment syntax: avoid backslash and '/*' tokens (#467).

* Fix minor format-string issues (#470).

- snprintf() expects the format specifier "%lld" to refer to a value of type
  "long long", not flex_int; we assert statically that flex_int is synonymous
  with "long", and add a static cast.

- Use of the null terminator "\0" in the format string for sscanf() elicits a
  GCC compiler warning; we remove this character.

* Fix unused variable in test/random/test_alias.cpp.

The variable "k" is assigned to, but not subsequently used, in the
run_alias_benchmark() function. Rather than remove it, we use it as a dummy
accumulator, in order to preserve the existing sequence of random number
generation calls.

* Fix variable shadowing issue in sparse_similarity/item_processing.hpp.

The variable "row_idx", referring to an atomic counter shared between threads,
was incorrectly shadowed by another local variable of the same name. This patch
renames the outer-scope variable, and updates one of the corresponding
references in the inner scope.

* Fix return type (void functions incorrectly marked as returning bool).

* Fix potentially uninitialized values (#481).

* Factor out utility function "ceil_divide" (also fixes -Wunused-function warnings).

* Add missing braces in std::array initialization (#477).

Extra braces are required in C++11; see:
https://en.cppreference.com/w/cpp/container/array

* Add missing virtual method override annotations.

* Fix implicit conversion of constant (#465).

* Fix boolean value returned from main() (#466).

* Fix std::move invocations on temporary variables (#469).

- We still make std::move calls in the case of the gl_string unit test, to
  attempt to preserve the existing test semantics.

* Fix potential side-effects in typeid() call (#472).

* Fix abs() call to refer to intended operand instead of comparison result (#474).

- Previously the code was testing "abs(stats.residual < 1e-5)" instead of
  "abs(stats.residual) < 1e-5".

* Add local directives to disable warnings on overloaded virtual methods (-Woverloaded-virtual).

- In some cases, we have a base class method overridden by a derived class
  method of the same name, but different type signature. This causes clang to
  generate compiler warnings (-Woverloaded-virtual). Ideally, we would simply
  rename the methods in the derived class, but this may require other API
  changes. For now, we instead add preprocessor directives to prevent these
  warnings at a few key sites.

* Remove test of tc_ft_type_enum against NULL (#471).

- Previously we ran CHECK_NOT_NULL on a value of type tc_ft_type_enum. This is
  likely not intended, since this is not a pointer type, and in fact may have
  incorrectly resulted in errors thrown upon receiving the value FT_TYPE_INTEGER
  (= 0). This patch removes that check.

* Fix uninitialized value in TF-IDF computation.

* Disable -Wsign-compare warning for auto-generated Cython source files.

* Disable some warnings for selected external source files (as well as CoreML-related internal files).

* Fix uninitialized variable in unity_sframe.

* Add #ifdef guards around compiler pragma directives to prevent GCC warnings.

* Fix memory-clearing behavior in flexible_type.

- In order to clear a flexible_type value, in addition to setting val.intval
  (the first 8 bytes) and val.stored_type (the last 4 bytes), we must also set
  val.dtval.m_microsecond (the middle 4 bytes). Previously, these 4 bytes may
  have been uninitialized on some uses, causing GCC compiler warnings.

* Add missing argument in image similarity GPU test case.

* Explicitly specify Python version for virtualenv.

- This seems to be necessary in some environments when both Python 2.x and
  Python 3.x are installed on macOS.

* Add missing API_AVAILABLE annotations for macOS.

- This fixes several compiler warnings about specific APIs only being available
  from macOS 10.13 (or 10.14) onwards.

* Enable -Werror on Linux (and in release builds).

* Remove global -Werror exceptions.


nickjong:
Sigh, for the record, everything in this PR was authored by Joe Zimmerman, but our GitHub settings forced a merge commit listing me as the author. Sorry, Joe!

PR #776, opened 2018-06-26
Removes Fiber

### LINKED ISSUE #444: Remove unused fiber
The fiber under `src/platform/fiber/` is no longer being used. We should remove it.

TobyRoseman:
Fixes #444


hoytak:
Let's get Yucheng to approve this before we merge -- when we did the initial pass to clean up all these old libraries, he wanted to leave it in but I forget the reason.

Issue #701, opened 2018-06-16
Error reading protobuf spec. validator error: The .mlmodel supplied is of version 3, intended for a newer version of Xcode. This version of Xcode supports model version 2 or earlier.".


amiraeitan:
I'm getting the above error when trying to run:
`model.export_coreml('MyCustomObjectDetector.mlmodel')`

```
model.export_coreml('MyCustomObjectDetector.mlmodel')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/amir/anaconda/envs/CVision/lib/python2.7/site-packages/turicreate/toolkits/object_detector/object_detector.py", line 1477, in export_coreml
    mlmodel = coremltools.models.MLModel(model)
  File "/Users/amir/anaconda/envs/CVision/lib/python2.7/site-packages/coremltools/models/model.py", line 209, in __init__
    self.__proxy__ = _get_proxy_from_spec(filename)
  File "/Users/amir/anaconda/envs/CVision/lib/python2.7/site-packages/coremltools/models/model.py", line 103, in _get_proxy_from_spec
    return _MLModelProxy(filename)
```

I'm on coremltools 2.0b1
pip 9.0.1
turicreate 5.0b1
python 2.7.13
Xcode 10.0 beta
macOS High Sierra 10.13.4

Not sure what to do, I wish the error would be more descriptive about what one should do in order to resolve this.

znation:
@amiraeitan This message means that Turi Create tried to save a model, while running on macOS 10.13, that isn't actually compatible with macOS 10.13. This seems like a bug, assuming we should be able to export object detector models that run on 10.13.

@gustavla Is it expected that a 5.0b1 object detector model, when exported to Core ML, requires Xcode 10+ (macOS 10.14+)? That's the behavior we're seeing here. We should make sure the spec version is the minimum needed to execute the model (in this case probably 1 or 2?).

znation:
@amiraeitan In the meantime, you can work around this bug by setting `include_non_maximum_suppression` to `False`. See https://apple.github.io/turicreate/docs/api/generated/turicreate.object_detector.ObjectDetector.export_coreml.html#turicreate.object_detector.ObjectDetector.export_coreml

gustavla:
@znation I'm a bit uncertain about this now. Linux doesn't have a runtime support for the Core ML model, but it can still export the file itself. Shouldn't this also be true for macOS 10.13 then?

znation:
[...]

PR #778, opened 2018-06-26
Turi Create fix for Issue #701

shantanuchhabra:
- Avoiding loading the Core ML model using the coremtools.models.MLModel call
- Still need to create a PR on the `coremltools` side

### DIFF
--- a/src/unity/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/unity/python/turicreate/toolkits/object_detector/object_detector.py
@@ -1533,11 +1533,8 @@ class ObjectDetector(_CustomModel):
             u'Boxes \xd7 Class confidence (see user-defined metadata "classes")'
         model.description.output[1].shortDescription = \
             u'Boxes \xd7 [x, y, width, height] (relative to image size)'
-        from turicreate import __version__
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1517,35 +1517,40 @@ def export_coreml(self, filename,
...
+            'include_non_maximum_suppression': str(
+                include_non_maximum_suppression),
+            'non_maximum_suppression_threshold': str(
+                iou_threshold),
+            'confidence_threshold': str(confidence_threshold),
+            'iou_threshold': str(iou_threshold),
+            'feature': self.feature,
+            'annotations': self.annotations,
+            'classes': ','.join(self.classes)
+        }
```
gustavla: I suggest you create another function in `_coreml_utils` that takes the same arguments as `_set_model_metadata`, except without the `mlmodel`. That function can return a dictionary with the meta-data. Now, replace most of the body of `_set_model_metadata` with that function and call it from here as well.

In this solution, we now have two places that handle `turicreate_version`, `type`, and `version`, and it will be harder to keep them consistent.

PR #779, opened 2018-06-27
Fix release build break on 10.13 / Xcode 9

znation:
With macOS 10.13 and Xcode 9, I am now getting a build break on latest
master, due to the `-Werror` change that went in. I'm not sure why this
doesn't repro on macOS 10.14 beta / Xcode 10 beta, but the Cython
generated code on my machine gives warnings (now treated as error). This
change suppresses the specific warning in the compilation of the
generated code.


nickjong:
Strange, I also encountered this last night on my 10.14 machine, although I have no idea how the `-Werror` change built in CI. And to sanity-check that commit after rebasing I only did a debug build, sigh.

PR #784, opened 2018-06-28
Allowed CMAKE_OSX_DEPLOYMENT_TARGET to be set externally.

hoytak:


nickjong:
Thinking about this some more, I'm not sure why we need this. The deployment target for our Objective-C libraries should not need to vary any more than the deployment target of the frameworks that we depend on, right? For example, CoreML presumably has a deployment target of 10.13, but binaries with any deployment target can link against it. (As long as they use runtime checks to avoid using it on pre-10.13 systems.)

znation:
@hoytak Can you address @nickjong's comment? This change looks fine to me, but it sounds like it may not be necessary.

srikris:
@hoytak @nickjong Can we close on this. Are we taking this change or not?

hoytak:
Let me investigate more on this -- I believe we can drop it.

Issue #789, opened 2018-06-29
Default value used for TURI_CACHE_FILE_LOCATIONS if that path does not exist


TobyRoseman:
```bash
$ # First, show default value
$ python -c "import turicreate as tc;
print(tc.config.get_runtime_config()['TURI_CACHE_FILE_LOCATIONS'])"

/var/folders/g_/_qq7342d05qdqmylcrsvnr9w0000gn/T/

$ # Set value 
$ export TURI_CACHE_FILE_LOCATIONS=/tmp/will_exist_later

$ python -c "import turicreate as tc;
print(tc.config.get_runtime_config()['TURI_CACHE_FILE_LOCATIONS'])"

/var/folders/g_/_qq7342d05qdqmylcrsvnr9w0000gn/T/

$ mkdir /tmp/will_exist_later

$ python -c "import turicreate as tc;
print(tc.config.get_runtime_config()['TURI_CACHE_FILE_LOCATIONS'])"

/tmp/will_exist_later
```

It's strange that it silently reverts to the default value if the given path does not exist. I would expect it to either try to create the path or give an error.

dhivyaaxim:
@TobyRoseman  @znation  Please assign this ticket to me as I would like to try fixing it and contribute. Thanks in advance!

PR #790, opened 2018-06-29
Better error message for unwritable cache path

### LINKED ISSUE #286: IOError: [Errno 13] Permission denied: '/var/tmp/model_cache/resnet-50-symbol.json'
**Error**
Done.
Downloading https://docs-assets.developer.apple.com/turicreate/models/resnet-50-symbol.json
  File "classifier.py", line 9, in <module>
    model = turi.image_classifier.create(trainingBuffers, target="chairType", model="resnet-50")
  File "/Library/Python/2.7/site-packages/turicreate/toolkits/image_classifier/image_classifier.py", line 116, in create
    ptModel = _pre_trained_models.MODELS[model]()
  File "/Library/Python/2.7/site-packages/turicreate/toolkits/_pre_trained_models.py", line 98, in __init__
    ], path)
  File "/Library/Python/2.7/site-packages/turicreate/toolkits/_pre_trained_models.py", line 52, in _download_and_checksum_files
    with open(fn, 'wb') as f:
IOError: [Errno 13] Permission denied: '/var/tmp/model_cache/resnet-50-symbol.json'

TobyRoseman:
Fixes #286 

Give a useful error message to the user when we can't write to the cache path. Also only create the model caches directory in one place.

### DIFF
--- a/src/unity/python/turicreate/toolkits/_pre_trained_models.py
+++ b/src/unity/python/turicreate/toolkits/_pre_trained_models.py
@@ -15,14 +15,24 @@ from six.moves.urllib import parse as _urlparse
 
 MODELS_URL_ROOT = 'https://docs-assets.developer.apple.com/turicreate/models/'
 
+
 def _get_model_cache_dir():
-    tmp_dir = _tc.config.get_runtime_config()['TURI_CACHE_FILE_LOCATIONS']
-    return _os.path.join(tmp_dir, 'model_cache')
+    cache_dir = _tc.config.get_runtime_config()['TURI_CACHE_FILE_LOCATIONS']
+    download_path = _os.path.join(cache_dir, 'model_cache')
+
+    if not _os.path.exists(download_path):
+        try:
+            _os.makedirs(download_path)
+        except:
+            raise RuntimeError("Could not write to the turicreate file cache, which is currently set to \"{cache_dir}\".\n"
+                               "To continue you must update this location to a writable path by calling:\n"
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/toolkits/_pre_trained_models.py
```diff
@@ -15,14 +15,24 @@
...
-    if not _os.path.exists(dirname):
-        _os.makedirs(dirname)
+    if not _os.path.exists(download_path):
+        try:
+            _os.makedirs(download_path)
+        except:
+            raise RuntimeError("Could not write to the turicreate file cache, which is currently set to \"{cache_dir}\".\n"
+                               "To continue you must update this location to a writable path by calling:\n"
+                               "\ttc.config.set_runtime_config(\'TURI_CACHE_FILE_LOCATIONS\', <path>)\n"
+                               "Where <path> is a writable file path that exists.".format(cache_dir=cache_dir))
```
gustavla: It's also possible to set the env var with this name too. Is this worth mentioning briefly?

gustavla: Also "... that exists" is technically not true, since we attempt to create all the paths. As long we have permission to do so. It's a subtle distinction that is not that important though, and I agree that it's better for users that hit this issue to manually confirm that there is a directory there.

TobyRoseman: Thanks for the review. 

I don't believe it's worth mentioning the environment variable. This error message is already long enough.

Surprisingly the "that exists" part is accurate. See #789 which I discovered while fixing this bug.

PR #794, opened 2018-06-29
Clean up TCMPS symbols

nickjong:
Our libtcmps.dylib really shouldn’t export pure C functions with names like “Init” and “Load”, which are bound to collide someday with someone who links against this library. Instead, we should add a prefix to all C functions and Objective-C classes. Going with “TCMPS” as the best fit for existing names in this code base. While we’re at it, the C++ symbols are moved into a new turi::mps namespace, instead of being declared at the top level.

Also did some miscellaneous cleanup in the top-level C interface for TCMPS. All of these changes should be behavior-preserving, although previously some Objective-C code was running outside of autorelease pools (that we know will drain).

### DIFF
--- a/src/unity/toolkits/tcmps/mps_graph_trainer.h
+++ b/src/unity/toolkits/tcmps/mps_graph_trainer.h
@@ -1,30 +1,11 @@
 #ifndef GRAPH_C_API_H_
 #define GRAPH_C_API_H_
 
-#import "mps_cnnmodule.h"
-#import "mps_graph_cnnmodule.h"
 #import "mps_trainer.h"
-#import <exception>
-
-#define EXPORT __attribute__((visibility("default")))
-
-/*! \brief  macro to guard beginning and end section of all functions */
-#define API_BEGIN() try {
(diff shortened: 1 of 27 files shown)


src/unity/toolkits/tcmps/mps_graph_trainer.h
```diff
@@ -1,63 +1,44 @@
...
-
-#define EXPORT __attribute__((visibility("default")))
-
-/*! \brief  macro to guard beginning and end section of all functions */
-#define API_BEGIN() try {
-//\brief every function starts with API_BEGIN();
-//     and finishes with API_END() or API_END_HANDLE_ERROR
-#define API_END()                                                              \
-  }                                                                            \
-  catch (...) {                                                                \
```
igiloh: Do we know how TC will behave now if there's an exception in MPS (or tcmps) code?  
Is it failing somewhat gracefully?

nickjong: I don't see anything on the Python side that inspects the (status code) return values from TCMPS.... I filed #801

PR #800, opened 2018-07-02
Resolve #614 - a few bug fixes in AC train-test split 

### LINKED ISSUE #614: Error when using `validation_set='auto'` in Activity classifier
Creating an activity classifier model with `validation_set='auto'` (which is the default) give an error:
```
  File "/Users/ilaigiloh/anaconda2/envs/TC4/lib/python2.7/site-packages/turicreate/toolkits/activity_classifier/_activity_classifier.py", line 190, in create
    validation_set = validation_set.filter_by(target_map.keys(), target)
  File "/Users/ilaigiloh/anaconda2/envs/TC4/lib/python2.7/site-packages/turicreate/data_structures/sframe.py", line 4371, in filter_by
    column_name + "' in SFrame.")
TypeError: Type of given values does not match type of column 'activity' in SFrame.
```
This will only happen if the dataset has more than 100 different sessions. For smaller datasets we currently don't create a validation set at all (by design).

igiloh:
1. Solved two bugs in the way the train-validation split was used in the Activity Classifier (causing `create()` to crash when `validation_set='auto'`).
2. Improved the `random_split_by_session() ` mechanism. The current implementation might have broken line order - which is not allowed for the AC. The new mechanism is optimized for the typical case of 100-1000 training sessions.
3. Added explicit unit test that covers the auto validation set in all relevant cases.

### DIFF
--- a/src/unity/python/turicreate/test/test_activity_classifier.py
+++ b/src/unity/python/turicreate/test/test_activity_classifier.py
@@ -19,14 +19,15 @@ import pytest
 from turicreate.toolkits._internal_utils import _mac_ver
 
 
-def _load_data(self):
+def _load_data(self, num_examples = 1000, num_features = 3, max_num_sessions = 4,
+               randomize_num_sessions = True, num_labels = 9, prediction_window = 5):
     random.seed(42)
 
-    self.num_examples = 1000
-    self.num_features = 3
-    self.num_sessions = random.randint(1, 4)
-    self.num_labels = 9
-    self.prediction_window = 5
+    self.num_examples = num_examples
+    self.num_features = num_features
+    self.num_sessions = random.randint(1, max_num_sessions) if randomize_num_sessions else max_num_sessions
+    self.num_labels = num_labels
+    self.prediction_window = prediction_window
 
     self.features = ['X1-r', 'X2-r', 'X3-r']
     self.target = 'activity_label'
@@ -36,9 +37,8 @@ def _load_data(self):
     random_labels = [random.randint(0, self.num_labels - 1) for i in range(self.num_examples)]
 
     self.data = tc.util.generate_random_sframe(column_codes='r' * self.num_features, num_rows=self.num_examples)
-    self.data['session_id'] = random_session_ids
+    self.data[self.session_id] = random_session_ids
     self.data[self.target] = random_labels
-
 
 class ActivityClassifierCreateStressTests(unittest.TestCase):
     @classmethod
@@ -89,6 +89,74 @@ class ActivityClassifierCreateStressTests(unittest.TestCase):
                                               session_id=self.session_id,
                                               prediction_window=self.prediction_window,
                                               validation_set=None)
+
+
+class ActivityClassifierAutoValdSetTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        self.fraction = 0.9
+        self.seed = 42
+
+    def _create_auto_validation_set(self):
+        model = tc.activity_classifier.create(self.data,
+                            features=self.features ,
+                            target=self.target,
+                            session_id=self.session_id,
+                            prediction_window=self.prediction_window,
+                            validation_set='auto')
+        predictions = model.predict(self.data)
+
+    def _test_random_split_by_session(self, num_sessions, is_small= False):
+
+        train, valid = tc.activity_classifier.util.random_split_by_session(self.data, self.session_id, self.fraction, self.seed)
+
+        train_num_sessions = len(train[self.session_id].unique())
+        train_frac = float(train_num_sessions / num_sessions)
+        expected_frac = 1.0 if is_small else self.fraction
+
+
+        self.assertAlmostEqual(train_frac, expected_frac, places=1,
+                               msg= "Got {} train sessions out of {}, which is {:.3f}, and not the expected {}".format(
+                               train_num_sessions, num_sessions, train_frac, expected_frac))
+
+        if is_small:
+            self.assertTrue(valid is None, "For {} sessions - expected behaviour is no validation set.".format(num_sessions))
+            return
+
+        valid_num_sessions = len(valid[self.session_id].unique())
+        valid_frac = float(valid_num_sessions / num_sessions)
+        expected_valid_frac = 1.0 - self.fraction
+
+        self.assertAlmostEqual(valid_frac, expected_valid_frac, places=1,
+                               msg= "Got {} train sessions out of {}, which is {:.3f}, and not the expected {}".format(
+                               valid_num_sessions, num_sessions, valid_frac, expected_valid_frac))
+
+        train_sessions_set = set(train[self.session_id].unique())
+        valid_sessions_set = set(valid[self.session_id].unique())
+
+        self.assertTrue(train_sessions_set.isdisjoint(valid_sessions_set),
+                        "After train-test split, the train and validation sets should not include the same sessions")
+
+    def test_create_auto_validation_set_small(self):
+        num_sessions = tc.activity_classifier.util._MIN_NUM_SESSIONS_FOR_SPLIT / 2
+        _load_data(self, max_num_sessions=num_sessions, randomize_num_sessions=False)
+
+        self._create_auto_validation_set()
+        self._test_random_split_by_session(num_sessions, is_small=True)
+
+    def test_create_auto_validation_set_typical(self):
+        num_sessions = tc.activity_classifier.util._MAX_SESSIONS_TO_USE_IS_IN / 4
+        _load_data(self, num_examples=10000, max_num_sessions=num_sessions, randomize_num_sessions=False)
+
+        self._create_auto_validation_set()
+        self._test_random_split_by_session(num_sessions)
+
+    def test_create_auto_validation_set_large(self):
+        num_sessions = tc.activity_classifier.util._MAX_SESSIONS_TO_USE_IS_IN + 200
+        _load_data(self, num_examples=10000, max_num_sessions=num_sessions, randomize_num_sessions=False)
+
+        self._create_auto_validation_set()
+        self._test_random_split_by_session(num_sessions)
 
 class ActivityClassifierTest(unittest.TestCase):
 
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/toolkits/activity_classifier/util.py
```diff
@@ -68,7 +70,43 @@ def random_split_by_session(dataset, session_id, fraction=0.9, seed=None):
...
     chosen, not_chosen = unique_sessions.random_split(fraction, seed)
-    train = dataset.filter_by(chosen['session'], session_id)
-    valid = dataset.filter_by(not_chosen['session'], session_id)
-    return train, valid
+
+    def split_using_is_in(data, session_id, not_chosen):
+        not_chosen_filter = data[session_id].is_in(not_chosen['session'])
+
+        train = data[1 - not_chosen_filter]
+        valid = data[not_chosen_filter]
```
nickjong: All of this logic makes sense, but it's frustrating, since random_split works by generating a random Boolean value for each row and then performing a filter internally, before returning the two resulting SArrays. We're having to re-implement so much of this operation, and the is_in case requires a linear search through `not_chosen` for each row, and the filter case requires sorting.

Could we somehow generate a random boolean for each row that yields the same result for each session ID? In principle, all it would take is a suitable hash function:

```
chosen_filter = data[session_id].apply(lambda x: hash_fn(x + seed) < fraction * MAX_HASH_FN_OUTPUT)
train = data[chosen_filter]
valid = data[1 - chosen_filter]
```

The total operation should then be linear, at face value. All that is required is a suitable hash function. For `hash_fn(x)` it may even suffice to re-initialize a RNG with `x` as the seed, perform one draw from uniform(0,1) and compare against `fraction`.

If there's something I'm missing, or if Python doesn't really make it easy enough to do the above, I could always investigate this approach as a followup someday. How long does this approach take in practice?

src/unity/python/turicreate/test/test_activity_classifier.py
```diff
@@ -17,28 +17,59 @@
...
+        raise ValueError("Can't divide {} lines into {} sessions.".format(num_examples, num_sessions))
+
+    min_lines_per_session = int(0.85 * examples_per_session)
+    max_lines_per_session = int(1.15 * examples_per_session)
+
+    lines_in_each_session = [random.randint(min_lines_per_session, max_lines_per_session) for i in range(num_sessions)]
+    lines_in_each_session[-1] += num_examples - sum(lines_in_each_session)
+
+    if lines_in_each_session[-1] < 0:
+        raise ValueError()
```
nickjong: This doesn't guard against `lines_in_each_session[-1] == 0`

And the probability of an error increases as the number of sessions increases.

It might be more robust just to normalize the sampled `lines_in_each_session`, by multiplying each value by `num_examples / sum(lines_in_each_session)`, followed by the existing fudge factor (in line 62) to deal with rounding.

(Interestingly, this sampling procedure would begin to resemble how you sample multinomial parameters from a Dirichlet distribution: https://en.wikipedia.org/wiki/Dirichlet_distribution#Random_number_generation)

src/unity/python/turicreate/toolkits/activity_classifier/util.py
igiloh: Hmm - interesting. This sounds like a good idea.  
My only concern is that it only guarantees that the sessions are actually divided according to the divided fraction in expectancy. For a small number of sessions (which is the typical case) - the actual fraction might be quite different.  
I'm not sure how accurate our current `random_split` mechanism is (which we use to divide the sessions). Maybe @ylow has some more information.

I have a bolt task where I've compared the `is_in` and the `filter` implementations for various dataset sizes (number of sessions). I'll run it again using your idea - and compare run time and fraction accuracy.

src/unity/python/turicreate/toolkits/activity_classifier/util.py
```diff
@@ -68,7 +70,43 @@ def random_split_by_session(dataset, session_id, fraction=0.9, seed=None):
...
+        valid = valid.sort(id_column)
+        train = train.remove_column(id_column)
+        valid = valid.remove_column(id_column)
+
+        return train, valid
+
+    # For a small number of unique sessions, creating a filter using SFrame.is_in() would
+    # be efficient - since filtering with a binary SArray maintains order, and no sorting
+    # is required.
+    if len(unique_sessions) < _MAX_SESSIONS_TO_USE_IS_IN:
```
igiloh: For now - I've set some arbitrary hard limit, which I've found empirically.  
Giving it some more thought - the `is_in` solution takes o(k * n), where `k` is the number of sessions, and the filter+sort solution takes o(nlogn).   
Maybe I can improve this condition by setting `if number_of_sessions < log(len(dataset))`?  

@nickjong - what do you think? (that's all assuming I won't go with your suggested `apply` method).

src/unity/python/turicreate/test/test_activity_classifier.py
igiloh: I've known my way of guaranteeing `sum(lines_in_each_session) == num_examples` is off right when I've written it - but couldn't think of a better method. Yours is the better (and probably the correct) method. Thanks!

I'll read about the Dirichlet sampling tomorrow, when my mind is still fresh. Right now the very mention of it brings back haunting memories from horrible college courses... 😄

src/unity/python/turicreate/toolkits/activity_classifier/util.py
nickjong: Ah, good point about the variance in the split. Looking at the implementation: [unity_sframe::random_split](https://github.com/apple/turicreate/blob/e43e2133d725cf8cbdb3d92f8d4d46facb58a3aa/src/unity/lib/unity_sframe.cpp#L1071) calls into [unity_sarray::make_uniform_boolean_array](https://github.com/apple/turicreate/blob/e43e2133d725cf8cbdb3d92f8d4d46facb58a3aa/src/unity/lib/unity_sarray.cpp#L1690), which uses essentially the idea above.

Interestingly, the C++ version has an `exact` parameter that causes it split as close as possible to the desired proportion. But that implementation also uses sorting. I would think there's a more efficient way to choose a random k-subset of n elements. I wonder if we could just use [numpy.random.choice](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.choice.html). Something like:
```
validation_session_ids = numpy.random.choice(num_sessions, fraction*num_sessions, replace=False)
# Use some numpy magic I don't know to convert to an array of length num_sessions with 1 in each row indicated by validation_session_ids
# Convert to filter SArray
```

src/unity/python/turicreate/test/test_activity_classifier.py
nickjong: Eh, the link I posted is of mostly academic interest. The sampling procedure boils down to generating `num_sessions` draws from a random distribution and then normalizing to obtain a random list of `num_sessions` probabilities that sum to 1.

The Dirichlet connection comes in that you can view `lines_in_each_session` as a multinomial random variable that we could sample e.g. using https://en.wikipedia.org/wiki/Multinomial_distribution#Sampling_from_a_multinomial_distribution. But we don't really care about `lines_in_each_session` actually having a multinomial distribution. Sampling a Dirichlet is a convenient way of giving us something similar, literally choosing a random multinomial distribution (represented by its parameters).

Maybe the real question is to what degree the test even relies on `lines_in_each_session` being random....

src/unity/python/turicreate/toolkits/activity_classifier/util.py
nickjong: I think what you have is fine. There's only so far we can get with asymptotic analysis, and if this continues to be a problem, I suspect it's a better investment of our time to find a way to make the O(n) approach work.

src/unity/python/turicreate/toolkits/activity_classifier/util.py
igiloh: Good idea - but it goes beyond the scope of this PR.  
Let's open a separate issue for that?

igiloh: BTW - session_id may be a string - so using `random.seed(x) ; random.uniform(0,1)` as the "hash" function won't work. At least not out of the box without pre-processing (in the string case).  
I'll try to see if I can find some more suitable hash func.

src/unity/python/turicreate/test/test_activity_classifier.py
igiloh: None of our tests are truly random at the moment.  
They all set the seed to `42` (always!), and most don't even use randomness.  

Nevertheless - I do believe that we'll add proper pseudo-randomness to the tests someday - so it's better to write new tests with that mindset.

igiloh:
I've changed the `random_split_by_session()` implementation to the O(n) one suggested by @nickjong.  
The split accuracy doesn't seem be much affected - but the speedup is tremendous. Thanks Nick!

I've also changed the unit tests accordingly - there's no need to test a very large case, if the implementation is similar for all data sizes.

@nickjong - can you please give this PR a fresh review?

igiloh:
@nickjong - let's open a separate issue for your suggestion regarding `random_split` [improvements](https://github.com/apple/turicreate/pull/800#discussion_r199876248).

src/unity/python/turicreate/toolkits/activity_classifier/util.py
```diff
@@ -68,7 +70,28 @@ def random_split_by_session(dataset, session_id, fraction=0.9, seed=None):
...
+        seed = long(time.time() * 256)
+
+    # Create a random binary filter (boolean SArray), using the same probability across all lines
+    # that belong to the same session. In expectancy - the desired fraction of the sessions will
+    # go to the training set.
+    # Since boolean filters preserve order - there is no need to re-sort the lines within each session.
+    def random_session_pick(session_id):
+        # If we will use only the session_id as the seed - the split will be constant for the
+        # same dataset across different runs, which is of course undesired
+        _random.seed(hash(session_id) + seed)
```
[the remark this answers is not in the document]
igiloh: Sounds right. Will do.

igiloh:
@nickjong - I've handled the last two issues you've mentioned.  
Can you please look at the last commit quickly before I merge this PR?

Thanks!

nickjong:
Great, looks good to me! Let's go ahead and merge!

PR #806, opened 2018-07-02
Reduce Object Detector batch size to 16 for Mac GPUs with less than 4 GB RAM

### LINKED ISSUE #629: loss is nan when using beta to train detector
Hi,
I’m using the beta (turiCreate 5) to train an image detector. The code I’m using to train the detector is the same that I used to train a detector using the last version of Turi Create (which went fine). Now, I’m getting nan in the loss category as it’s training with my mac’s GPU. Is this a bug or should I adjust my code to fix this?
Thanks!

nickjong:
To avoid memory allocation failures that currently manifest as NAN output, object_detector.create() should query the memory resources of the Metal device and adjust the batch size if necessary.

We should also clean up our code for reasoning about Metal devices. At present, the first time we use Metal on any given thread, we query the device list and save it for future reference. This approach doesn't do the right thing if the device list changes (due to eGPUs), and it performs redundant work across threads. This PR takes a few steps towards supporting this use case more robustly.

Fixes #629


src/unity/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -265,11 +270,16 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
...
+    if batch_size < 1:
+        batch_size = 32  # Default if not user-specified
+    num_mxnet_gpus = _mxnet_utils.get_num_gpus_in_use(max_devices=batch_size)
+    batch_size_each = batch_size // max(num_mxnet_gpus, 1)
     # Note, this may slightly alter the batch size to fit evenly on the GPUs
     batch_size = max(num_mxnet_gpus, 1) * batch_size_each
     use_mps = _use_mps() and num_mxnet_gpus == 0
+    if use_mps and _mps_device_memory_limit() < 4 * 1024 * 1024 * 1024:
+        # Reduce batch size for GPUs with less than 4GB RAM
+        batch_size = 16
```
nickjong: At face value, we may also need to reset `batch_size_each`, but AFAICT none of the references to this variable below are needed (in the `if use_mps` case).

What does the net.forward() call in line 422 do? Also, it looks like `ymap_shape` and `loss` are only used by the MXNet path. If so, moving those assignments to that path might make things a bit more clear

src/unity/toolkits/tcmps/mps_device_manager.m
```diff
@@ -0,0 +1,96 @@
...
+  return self;
+}
+
+- (void)dealloc {
+  MTLRemoveDeviceObserver(_deviceObserver);
+}
+
+- (BOOL)shouldPreferDevice:(id <MTLDevice>)candidate overDevice:(nullable id <MTLDevice>)current {
+  // Something is better than nothing.
+  if (!current) return YES;
```
igiloh: Not sure.  
Gustav's whole approach (based on some input from MPS, if I recall correctly) was that a low power GPU (which is usually just the onboard one) would perform slowly than the CPU.

nickjong: Well, FWIW, this code merely preserves the behavior defined in the original mps_dev.mm. Note that in practice MPS usage is gated on the Python side by `has_fast_mps_support`, which only returns true if there's a "high-power" Metal device.

src/unity/python/turicreate/toolkits/object_detector/object_detector.py
gustavla: Yes, the `batch_size_each` should only be relevant for multi-GPU. Although it's used to set `ymap_shape`, so for single-GPU we should just make that `batch_size_each == batch_size`. As you point out, I don't see a usage of `ymap_shape` in the MPS path right now, but it could be inviting future bugs to leave `batch_size` and `batch_size_each` inconsistent.

The `net.forward` should really have a comment explaining it. So, `net` here is an MXNet model with the right initial parameters that should be copied into MPS. However, the parameters may not have been initialized yet (since this is deferred until the first forward pass in MXNet). The call to `forward` forces this initialization.

Of course, all this is a bit of a hack. Ideally we should not be relying on MXNet at all to initialize the MPS network. However, getting to that point will take a bit of refactoring.

gustavla: Also, we should probably print the batch size if `verbose` is True, just like we show what `max_iterations` was automatically set to (as well as what computational resource was used).

PR #810, opened 2018-07-03
Made error reporting in unity sframe append more robust.

hoytak:


src/unity/lib/unity_sframe.cpp
```diff
@@ -844,20 +844,50 @@ std::shared_ptr<unity_sframe_base> unity_sframe::append(
...
+                          other_column_names.begin(), other_column_names.end(),
+                          std::inserter(in_this, in_this.begin()));
+
+      std::set_difference(other_column_names.begin(), other_column_names.end(),
+                          column_names.begin(), column_names.end(),
+                          std::inserter(in_other, in_other.begin()));
+
+        std::ostringstream ss; 
+          ss << "Error: Attempting to append SFrame with columns "
+            << in_other << " to SFrame with columns " << in_this << ".";
```
znation: Do we want to use the set difference for `in_this`? I would expect as a user to get the full list of columns of the SFrame being appended to (including the intersection with `in_other`).

znation: Or maybe simplify it to the same message style as for types below (just show the names of columns in the `other_sframe` that aren't valid in `this` one)?

hoytak: Fair point; this just reports the mismatch.   What about this?  "Columns [....] in base SFrame not found in appending SFrame."  ?

znation: That sounds good.

hoytak:
@znation -- could you review the new changes? 

Thanks!

PR #814, opened 2018-07-03
Added code to keep track of size of SArray, fixes #758

### LINKED ISSUE #758: read_csv prints load file summary but read_json does not.
Please fix in C++ layer, below unity_sarray.

shantanuchhabra:
@hoytak and I discussed that it'd be best to print out the size of the SArray parsed in the C++ layer as load file summary, so added code to do that.

### DIFF
--- a/src/unity/lib/unity_sarray.cpp
+++ b/src/unity/lib/unity_sarray.cpp
@@ -266,16 +266,22 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
           continue;
         }
 
+        size_t num_elems_parsed = 0;
         bool has_non_dict_elements = false;
         for (const auto& element : parse_result.first.get<flex_list>()) {
           if (element.get_type() == flex_type_enum::DICT ||
               element.get_type() == flex_type_enum::UNDEFINED) {
             (*output) = element;
             ++output;
+            ++num_elems_parsed;
           } else {
             has_non_dict_elements = true;
           }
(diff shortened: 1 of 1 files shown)


src/unity/lib/unity_sarray.cpp
```diff
@@ -266,17 +266,23 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
...
             (*output) = element;
             ++output;
+            ++num_elems_parsed;
           } else {
             has_non_dict_elements = true;
           }
         }
 
+        logstream(LOG_PROGRESS) << "Successfully parsed an SArray of size "
+                                << num_elems_parsed << " from the JSON file, "
```
znation: Grammar: I don't think this should have a comma.

src/unity/lib/unity_sarray.cpp
```diff
@@ -266,17 +266,23 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
...
               element.get_type() == flex_type_enum::UNDEFINED) {
             (*output) = element;
             ++output;
+            ++num_elems_parsed;
           } else {
             has_non_dict_elements = true;
           }
         }
 
+        logstream(LOG_PROGRESS) << "Successfully parsed an SArray of size "
```
znation: Grammar: I would say "SArray of n elements" rather than "SArray of size n".

PR #817, opened 2018-07-05
Add NSCopying conformance to TCMPSConvolutionWeights and TCMPSBatchNormData

nickjong:
With the just-released developer seed 18A326g, MPSCNNConvolutionDataSource and MPSCNNBatchNormalizationDataSource now inherit from NSCopying. We must add conformance to our data source implementations to avoid build breakage, since we compile with -Werror.


nickjong:
Good call, I added some assertions

PR #818, opened 2018-07-06
Clean up toolkit progress header names

### LINKED ISSUE #757: Inconsistent naming with supervised learning
In supervised learning, there are a lot of inconsistent names printed.  For example, with logistic regression, it prints:

Step size | Elapsed Time | Training-accuracy | Training-log_loss  | Validation-accuracy

These should all be consistent, e.g. 

Step Size | Elapsed Time | Training Accuracy | Training Log Loss | Validation Accuracy

znation:
* Uses "Training " and "Validation " (capital letter, space at the end)
  as the standard header prefixes.
* Adds a concept of "metric display name", and uses the display name
  whenever there is one defined when adding metric names to the progress
  headers. I.e., "Log Loss" instead of "log_loss".

Fixes #757

### DIFF
--- a/src/logger/assertions.hpp
+++ b/src/logger/assertions.hpp
@@ -375,15 +375,7 @@ extern void __print_back_trace();
 
 
 #define DASSERT_MSG(condition, fmt, ...)                                \
-  do {                                                                  \
-    if (__builtin_expect(!(condition), 0)) {                            \
-      logstream(LOG_ERROR)                                              \
-        << "Check failed: " << #condition << ":\n";                     \
-      logger(LOG_ERROR, fmt, ##__VA_ARGS__);                            \
-      __print_back_trace();                                             \
-      TURI_LOGGER_FAIL_METHOD("assertion failure");                    \
-    }                                                                   \
-  } while(0)
+  ASSERT_MSG(condition, fmt, ##__VA_ARGS__)
 
 #endif
 
--- a/src/unity/toolkits/supervised_learning/supervised_learning.cpp
+++ b/src/unity/toolkits/supervised_learning/supervised_learning.cpp
@@ -350,6 +350,28 @@ std::vector<std::string>
   return tracking_metrics;
 }
 
+std::string supervised_learning_model_base::get_metric_display_name(const std::string& metric) const {
+  const static std::unordered_map<std::string, std::string> display_names = {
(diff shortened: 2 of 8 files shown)


src/logger/assertions.hpp
```diff
@@ -375,15 +375,7 @@ extern void __print_back_trace();
...
-  do {                                                                  \
-    if (__builtin_expect(!(condition), 0)) {                            \
-      logstream(LOG_ERROR)                                              \
-        << "Check failed: " << #condition << ":\n";                     \
-      logger(LOG_ERROR, fmt, ##__VA_ARGS__);                            \
-      __print_back_trace();                                             \
-      TURI_LOGGER_FAIL_METHOD("assertion failure");                    \
-    }                                                                   \
-  } while(0)
+  ASSERT_MSG(condition, fmt, ##__VA_ARGS__)
```
nickjong: Worth commenting why these macros do the same thing? Should DASSERT_MSG depend on NDEBUG?

src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -350,6 +350,28 @@ std::vector<std::string>
   return tracking_metrics;
 }
 
+std::string supervised_learning_model_base::get_metric_display_name(const std::string& metric) const {
```
TobyRoseman: Should this method be static?

src/logger/assertions.hpp
znation: `DASSERT_MSG` doesn't depend on `NDEBUG`, as `ASSERT_MSG` is always present. In `NDEBUG`, `DASSERT_MSG` expands to nothing; otherwise, it shares the same logic as `ASSERT_MSG`.

src/unity/toolkits/supervised_learning/supervised_learning.cpp
znation: Yes, I think so. Originally I had in mind that some model types may want to override it; but the point of this change is to be consistent in how we are naming metrics, so we probably shouldn't allow it to be a per-model-type override.

src/logger/assertions.hpp
hoytak: @nickjong It does; this is in a large ifdef block.

PR #823, opened 2018-07-07
Adopt new MPSNNOptimizer API for updating conv and batch-norm weights

nickjong:
Deeper fix for #742: adopt the newer convenience API for updating our data sources

### DIFF
--- a/src/unity/toolkits/tcmps/mps_weight.mm
+++ b/src/unity/toolkits/tcmps/mps_weight.mm
@@ -24,7 +24,7 @@ API_AVAILABLE(macos(10.14))
 // Batch norm update
 -(void)encodeToCommandBuffer:(id<MTLCommandBuffer>)commandBuffer
      batchNormalizationState:(MPSCNNBatchNormalizationState *)state
-               batchNormData:(TCMPSBatchNormData *)data;
+               batchNormData:(TCMPSBatchNormWeights *)weights;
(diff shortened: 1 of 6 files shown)


src/unity/toolkits/tcmps/mps_weight.mm
```diff
@@ -9,7 +9,116 @@
...
+
+@property (nonatomic, readonly) MPSVector *weightMomentumVector;
+@property (nonatomic, readonly) MPSVector *weightVelocityVector;
+@property (nonatomic, readonly) MPSVector *biasMomentumVector;
+@property (nonatomic, readonly) MPSVector *biasVelocityVector;
+
+@end
+
+// Class extension for TCMPSBatchNormData internal implementation
+@interface TCMPSBatchNormData ()
```
gustavla: Should we perhaps call this `TCMPSBatchNormWeights`? Not sure why we have convolution *weights* but batch norm *data*. I suppose the BN gammas and betas are what we normally consider to be weights, but the moving mean/var don't typically go by this name. However, at the end of the day, during inference, they are weights in the network, so the distinction is too subtle to warrant the -Data name.

nickjong: Good call! I went ahead and renamed to `TCMPSBatchNormWeights`

PR #833, opened 2018-07-09
Use NSNumber for item IDs in TCRecommender

### LINKED ISSUE #832: TCRecommender uses string keys to refer to int item IDs
The TCRecommender implementation always returns a dictionary with string keys, even if the model used int item IDs. So for models trained with int item IDs, the resulting Core ML model will throw (unwrapping optional nil) when accessed from Swift, since the cast to `[Int64: Double]` will fail.

znation:
When the model uses int64 item IDs, we should use NSNumber as the key
type in the output dictionaries.

Fixes #832

### DIFF
--- a/src/objc/TuriCreateObjC.m
+++ b/src/objc/TuriCreateObjC.m
@@ -25,7 +25,6 @@ static bool handleError(tc_error* error_ptr, NSError** error) {
                                          code:MLModelErrorGeneric
                                      userInfo:userInfo];
         }
-        assert(false);
         return true;
     }
     return false;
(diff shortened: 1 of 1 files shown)


src/objc/TuriCreateObjC.m
```diff
@@ -25,6 +25,7 @@ static bool handleError(tc_error* error_ptr, NSError** error) {
                                          code:MLModelErrorGeneric
                                      userInfo:userInfo];
         }
+        assert(false);
```
shantanuchhabra: Do we need the `assert(false)`?

TobyRoseman: Do you want to keep this line? Is the return statement on the line below just to make the compiler happy?

znation: I added it to debug; the return statement is probably the correct behavior here. I'll remove this line.

znation: Probably not (see also Toby's comment). I'll remove it.

PR #836, opened 2018-07-09
Update data-preparation.md

### LINKED ISSUE #720: IG02 Data preparation script for Object Detection throws error
I downloaded the datasets mentioned [here](https://github.com/apple/turicreate/blob/master/userguide/object_detection/data-preparation.md) 

I then ran the Python script there and got the following error:

```
  File "data-prep.py", line 16, in <module>
    info = info.unpack().rename({'X.0': 'name', 'X.1': 'type'})
  File "/Users/schhabra/Downloads/venv/lib/python2.7/site-packages/turicreate/data_structures/sarray.py", line 3422, in unpack
    raise TypeError("Only SArray of dict/list/array type supports unpack")
TypeError: Only SArray of dict/list/array type supports unpack
```

shantanuchhabra:
Fixes #720

### DIFF
--- a/userguide/object_detection/data-preparation.md
+++ b/userguide/object_detection/data-preparation.md
@@ -10,12 +10,23 @@ toolkit, we will rely on many useful SFrame functions.
 The dataset that we will use is [INRIA Annotations for Graz-02
 (IG02)](https://lear.inrialpes.fr/people/marszalek/data/ig02/)[^1][^2]. Our goal is to
 make a detector that can spot bikes and cars. Go ahead and click this link and
-download `ig02-v1.0-bikes.zip` and `ig02-v1.0-cars.zip`. In this tutorial, we
-will assume that you have unzipped these files into `~/Downloads/ig02`. That
-is, you should have the following data structure:
+download `ig02-v1.0-bikes.zip` and `ig02-v1.0-cars.zip`. On unzipping, they will probably 
+get named to `ig02-v1` and `ig02-v1-2` respectively. Open a Terminal session and 
+run the following commands:
 
 ```
-~/Downloads/ig02/
+$ cd ~/Downloads/ig02-v1
+$ mv ~/Downloads/ig02-v1-2/readme.txt ~/Downloads/ig02-v1-2/readme2.txt
+$ mv ~/Downloads/ig02-v1-2/* .
+$ rm -rf ~/Downloads/ig02-v1-2
+```
+Now, you should have the following data structure:
+
+```
+~/Downloads/ig02-v1/
+    bikes_info.txt
+    bikes_test.txt
+    bikes_train.txt
     bikes/
         bike_001.image.png
         bike_001.mask.0.png
@@ -24,11 +35,16 @@ is, you should have the following data structure:
         bike_001.mask.3.png
         bike_002.image.png
         ...
+    cars_info.txt
+    cars_test.txt
+    cars_train.txt
     cars/
         carsgraz_001.image.png
         carsgraz_001.mask.0.png
         carsgraz_002.image.png
         ...
+    readme.txt
+    readme2.txt
 ```
 
 This dataset contains masks marking the exact pixels belonging to each object


userguide/object_detection/data-preparation.md
```diff
@@ -10,12 +10,23 @@ toolkit, we will rely on many useful SFrame functions.
 The dataset that we will use is [INRIA Annotations for Graz-02
 (IG02)](https://lear.inrialpes.fr/people/marszalek/data/ig02/)[^1][^2]. Our goal is to
 make a detector that can spot bikes and cars. Go ahead and click this link and
-download `ig02-v1.0-bikes.zip` and `ig02-v1.0-cars.zip`. In this tutorial, we
-will assume that you have unzipped these files into `~/Downloads/ig02`. That
-is, you should have the following data structure:
+download `ig02-v1.0-bikes.zip` and `ig02-v1.0-cars.zip`. On unzipping, they will probably
```
znation: Is this true on various operating systems/browsers? I know Safari on macOS behaves this way with multiple zipfile downloads that extract in the same place, but I suspect other OSes behave differently (either by appending a different set of characters, or unpacking into the same folder and clobbering, etc.). If we're going to add this text, we should make sure it works across all browsers on all of the OSes we support. Alternately, let's go with fully terminal-based commands using `curl` that will work consistently.

srikris: +1 to Zach's comments. Use curl.

userguide/object_detection/data-preparation.md
```diff
@@ -9,21 +9,23 @@ toolkit, we will rely on many useful SFrame functions.
...
-$ mv ~/Downloads/ig02-v1-2/readme.txt ~/Downloads/ig02-v1-2/readme2.txt
-$ mv ~/Downloads/ig02-v1-2/* .
-$ rm -rf ~/Downloads/ig02-v1-2
+$ mkdir -p ~/Downloads/ig02
+$ cd ~/Downloads/ig02
+$ curl https://lear.inrialpes.fr/people/marszalek/data/ig02/ig02-v1.0-bikes.zip > bikes.zip
+$ curl https://lear.inrialpes.fr/people/marszalek/data/ig02/ig02-v1.0-cars.zip > cars.zip
+$ tar -xvf bikes.zip
+$ mv readme.txt readme2.txt
+$ tar -xvf cars.zip
```
gustavla: Shouldn't this be `unzip`?

shantanuchhabra: I think the `-x` flag in the `tar` command is to specify that we're trying to extract files, so unzip isn't necessary. I also tried this locally and it works.

srikris:
@znation @gustavla Seems like all the issues are addressed here.  @shantanuchhabra Has this been verified on both Linux and Mac

shantanuchhabra:
@srikris Yes, verified on both Linux and Mac.

userguide/object_detection/data-preparation.md
```diff
@@ -9,13 +9,26 @@ toolkit, we will rely on many useful SFrame functions.
...
-make a detector that can spot bikes and cars. Go ahead and click this link and
-download `ig02-v1.0-bikes.zip` and `ig02-v1.0-cars.zip`. In this tutorial, we
-will assume that you have unzipped these files into `~/Downloads/ig02`. That
-is, you should have the following data structure:
+make a detector that can spot bikes and cars. Open a new terminal window and run the following commands:
+
+```
+$ mkdir -p ~/Downloads/ig02
+$ cd ~/Downloads/ig02
+$ curl https://lear.inrialpes.fr/people/marszalek/data/ig02/ig02-v1.0-bikes.zip > bikes.zip
```
znation: I think it's more common to use `curl -O output_name` than piping to a file... but I'm not sure if piping to a file has any drawbacks. This might be fine as is.

gustavla: :+1: This caught my eye too.

PR #846, opened 2018-07-12
Unit tests and adding multiple column join

znation:
* Adds C API unit tests for join (single column case so far)
* Adds a multiple column join implementation in C API
* Implements single column join in terms of multi column join, so our
  single column unit test has some hope of decent coverage.

### DIFF
--- a/src/capi/impl/capi_sframe.cpp
+++ b/src/capi/impl/capi_sframe.cpp
@@ -159,6 +159,32 @@ EXPORT tc_flex_list* tc_sframe_column_names(const tc_sframe* sf,
   ERROR_HANDLE_END(error, NULL);
 }
 
+EXPORT tc_sframe* tc_sframe_join_on_multiple_columns(
+    tc_sframe* left, tc_sframe* right,
+    tc_flex_list* join_columns,
+    const char* how, tc_error** error) {
+  ERROR_HANDLE_START();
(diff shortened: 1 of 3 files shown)


src/capi/impl/capi_sframe.cpp
```diff
@@ -169,21 +195,24 @@ EXPORT tc_sframe* tc_sframe_join_on_single_column(tc_sframe* left,
...
-
-  return new_tc_sframe(
-      left->value.join(right->value, {std::string(column)}, how));
+  CHECK_NOT_NULL(error, column, "column", NULL);
+  CHECK_NOT_NULL(error, how, "how", NULL);
+
+  tc_flex_list *join_columns = tc_flex_list_create(error);
+  if (join_columns == NULL) { assert(false); return NULL; } // error should be populated already
+  tc_flexible_type *flex_column = tc_ft_create_from_cstring(column, error);
+  if (flex_column == NULL) { assert(false); return NULL; } // error should be populated already
```
nickjong: In theory, an error here will leak `join_columns`. But the likelihood is quite low. We should have a cleanup issue to simplify writing code here.

Since we can use C++ in the implementation of this C API, probably we should define some aliases for e.g. std::unique_ptr<tc_flex_list, /* appropriate deleter */>

PR #854, opened 2018-07-16
Fix a Docstring

### LINKED ISSUE #543: Typos in `help(turicreate.recommender._get)`
1. `from turicreate.recommender` is a typo. Moreover, when you try to do `from turicreate.recommender import *`, you get `ImportError: No module named tc.recommender.factorization_recommender`, so need to come up with an alternative for that which imports `factorization_recommender`.
2. `U2 = d['item_id']` instead of `U2 = d['movie_id']`

screen shot 2018-04-17 at 10 52 15 pm

TobyRoseman:
fixes #543. Also fixes some indentation in a C++ unit test.

### DIFF
--- a/src/unity/python/turicreate/toolkits/recommender/util.py
+++ b/src/unity/python/turicreate/toolkits/recommender/util.py
@@ -569,13 +569,12 @@ class _Recommender(_Model):
         Examples
         --------
         >>> data = turicreate.SFrame({'user_id': ["0", "0", "0", "1", "1", "2", "2", "2"],
-        ...                         'item_id': ["a", "b", "c", "a", "b", "b", "c", "d"],
-        ...                         'rating': [1, 3, 2, 5, 4, 1, 4, 3]})
-        >>> from turicreate.recommender
-        >>> m = factorization_recommender.create(data, "user_id", "item_id", "rating")
-        >>> d = m.get("coefficients")
+                                      'item_id': ["a", "b", "c", "a", "b", "b", "c", "d"],
+                                      'rating': [1, 3, 2, 5, 4, 1, 4, 3]})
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/toolkits/recommender/util.py
```diff
@@ -569,13 +569,12 @@ def _get(self, field):
...
         >>> data = turicreate.SFrame({'user_id': ["0", "0", "0", "1", "1", "2", "2", "2"],
-        ...                         'item_id': ["a", "b", "c", "a", "b", "b", "c", "d"],
-        ...                         'rating': [1, 3, 2, 5, 4, 1, 4, 3]})
-        >>> from turicreate.recommender
-        >>> m = factorization_recommender.create(data, "user_id", "item_id", "rating")
-        >>> d = m.get("coefficients")
+                                      'item_id': ["a", "b", "c", "a", "b", "b", "c", "d"],
+                                      'rating': [1, 3, 2, 5, 4, 1, 4, 3]})
+        >>> m = turicreate.factorization_recommender.create(data, "user_id", "item_id", "rating")
+        >>> d = m._get("coefficients")
```
srikris: Please change this doc string to 

```
m.coefficients
```

TobyRoseman: I don't think that makes sense. This is a docstring for `_get(...)`.

Given that there is a better way to get coefficients, perhaps it would be best to get something other than coefficients using `_get`. What do you think? Any idea what would make sense to get?

srikris: Oh. Okay. Never mind then.

PR #862, opened 2018-07-18
Clean up error messages on model save

znation:
* Adds a turi::error namespace and io_error class. This can be used to
  distinguish between ios_base::failure coming from the C++ standard
  library, and I/O errors we throw ourselves with a crafted message.
* Use the new io_error from our own log_and_throw macros.
* In some key places related to model saving, avoid handling and
  throwing a new error when we already have a turi::error.

The result is that in a few cases, the previous user-facing error
message about unknown errors has been replaced with meaningful error
messages on model save:

* When the directory path already exists as a file.
* When the directory fails to be created, and does not exist at all.

### DIFF
--- a/src/fileio/fs_utils.cpp
+++ b/src/fileio/fs_utils.cpp
@@ -272,6 +272,39 @@ EXPORT bool create_directory(const std::string& path) {
   return false;
 }
 
+EXPORT bool create_directory_or_throw(const std::string& path) {
+  bool result = create_directory(path);
+
+  // this function throws if the directory still doesn't exist
+  // at that location
+  file_status status = get_file_status(path);
+  switch (status) {
+    case file_status::MISSING:
+      log_and_throw_io_failure(
+        "Unable to create directory structure at " +
+        path +
+        ". Ensure that you have write permission to this location, or try again with a different path."
+      );
+      break;
+    case file_status::REGULAR_FILE:
+      log_and_throw_io_failure(
+        "Unable to create directory at " +
+        path +
+        ". A non-directory file already exists there. Delete that file, or try again with a different path."
+      );
+    case file_status::DIRECTORY:
+      // happy path, return below
+      break;
+    default:
+      // not sure what error message to give; fall back to current I/O error
+      log_and_throw_current_io_failure();
+  }
+
+  // result will still indicate whether it was created or not
+  // by this function call
+  return result;
+}
+
 EXPORT bool delete_path(const std::string& path,
                         file_status stat) {
   if (stat == file_status::FS_UNAVAILABLE) stat = get_file_status(path);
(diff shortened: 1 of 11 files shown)


src/logger/error.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+/*
+ * Error types to distinguish Turi Create errors from arbitrary C++ exceptions
+ */
+
+#include <ios>
+
+namespace turi {
+    namespace error {
+        class io_error: public std::ios_base::failure {
```
nickjong: Formatting nits:

I prefer the precedent in other areas of our code, in which the contents of namespaces are not indented. This makes it easier to change the nesting level of a class's name without worrying about the indent level of the class itself.

Most of our code also seems to use an indent level of 2. I don't have strong feelings either way, although it would be nice to be consistent.

src/logger/error.hpp
```diff
@@ -0,0 +1,24 @@
...
+        private:
+            std::string m_message;
```
nickjong: I also with we had a consistent way of naming class data members. Honestly, anything is better than having data members be named the same way as local variables, but I think I've seen more instances of `data_member_` than `m_data_member`. This is more food for thought for later, I'm happy this is just not `message`.

src/logger/logger.hpp
```diff
@@ -354,12 +355,19 @@
...
+      throw(turi::error::io_error(message));                         \
     };                                                                \
     throw_error();                                                    \
   } while(0)
 #endif
 
+#define log_and_throw_current_io_failure()                            \
+  do {                                                                \
+    auto error_code = errno;                                          \
+    std::string error_message = std::strerror(error_code);            \
```
hoytak: We should clear errno with `errno = 0;` after "handling" the error (here handling is done by raising an exception).

src/fileio/fs_utils.cpp
```diff
@@ -272,6 +272,39 @@ EXPORT bool create_directory(const std::string& path) {
...
 }
 
+EXPORT bool create_directory_or_throw(const std::string& path) {
+  bool result = create_directory(path);
+
+  // this function throws if the directory still doesn't exist
+  // at that location
+  file_status status = get_file_status(path);
+  switch (status) {
+    case file_status::MISSING:
```
hoytak: Case readonly?

srikris:
@znation Can we merge this?

znation:
@srikris I'll address some feedback and then it should be mergeable after my next commit.

src/fileio/fs_utils.cpp
znation: That isn't one of the cases in `file_status`.

src/logger/logger.hpp
znation: Done.

PR #864, opened 2018-07-19
Automatic regression should attempt multiple models and compare the r…

### LINKED ISSUE #858: tc.regression.create should try multiple models and selected based on validation RMSE
For parity with tc.classifier.create, which trains multiple models and selects based on validation accuracy

nickjong:
…esults

The automatic classification feature already trains multiple models, returning the one with the highest validation accuracy. We should do the same thing for regression, using RMSE to choose the final model.

Note that the current automatic regression logic simply uses boosted trees for any dataset with fewer than 200 columns, and linear otherwise. For small datasets, linear is much more likely to be the best bet.

These changes unify the Python (turicreate.regression.create) and C API (create_automatic_regression_model) implementations. The automatic classification implementations both train multiple models but have some subtle differences, so unification there is deferred for now to minimize risk.

Fixes #858

### DIFF
--- a/src/unity/toolkits/supervised_learning/automatic_model_creation.cpp
+++ b/src/unity/toolkits/supervised_learning/automatic_model_creation.cpp
@@ -1,5 +1,12 @@
-#include <unity/lib/api/unity_sframe_interface.hpp>
 #include <unity/toolkits/supervised_learning/automatic_model_creation.hpp>
+
+#include <algorithm>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <unity/lib/api/unity_sframe_interface.hpp>
+#include <unity/lib/unity_sframe.hpp>
 #include <unity/toolkits/supervised_learning/boosted_trees.hpp>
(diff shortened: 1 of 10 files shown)


src/unity/toolkits/supervised_learning/automatic_model_creation.cpp
```diff
@@ -1,17 +1,155 @@
...
+}
+
+// Returns the validation Root Mean Squared Error if it exists, otherwise
+// returns training RMSE. Throws if neither field exists.
+double get_regression_rmse(
+    const std::shared_ptr<supervised_learning_model_base>& model) {
+
+  variant_type rmse;
+  std::vector<std::string> model_fields = model->list_fields();
+  bool success = false;
```
TobyRoseman: instead of
```
bool success = false;
if (!success) {
    success = read_model_field(model, model_fields, "validation_rmse", &rmse);
}
```
Maybe just:
`bool success = read_model_field(model, model_fields, "validation_rmse", &rmse);`

PR #869, opened 2018-07-19
Fixed seed for automatic model creation.

hoytak:
Ideally, this should be consolidated from python, and have a random seed
that can be set or pulled through.  For now, though, just fix the random
seed.


hoytak:
This code path isn't touched by python.  Ideally, we should consolidate the seed logic AND pass in the random seed through this path.  This change does not affect any non-C-API functionality ATM, but I agree it's not the right approach.   Will close and open an issue about that consolidation.

PR #870, opened 2018-07-19
Add default constructor to table printer

hoytak:
There is no reason the table printer shouldn't have a default
constructor

### DIFF
--- a/src/table_printer/table_printer.hpp
+++ b/src/table_printer/table_printer.hpp
@@ -200,6 +200,8 @@ struct progress_time {
 class table_printer {
 
  public:
+  /** Constructor.  Must be initialized elsewise using copy assignment ops. */
+  table_printer(){}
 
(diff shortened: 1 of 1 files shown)


src/table_printer/table_printer.hpp
```diff
@@ -200,6 +200,8 @@ struct progress_time {
 class table_printer {
 
  public:
+  /** Constructor.  Must be initialized elsewise using copy assignment ops. */
```
znation: What does `Must be initialized elsewise using copy assignment ops` mean? Could you elaborate on this? If the `table_printer` is non-functional when default constructed, perhaps we should assert false on some operations in case someone attempts to use it that way?

hoytak: Those assertions are already present -- it's simply initialized to have no columns, so attempts to use it in any useful way will result in assertion failures.

PR #873, opened 2018-07-20
Avoid passing nil for nonnull parameter of -encodeBatchToCommandBuffer:...

nickjong:
As a result of a change to MetalPerformanceShaders in the last seed, our passing a nil state into `-[MPSCNNBatchNormalization -encodeBatchToCommandBuffer:sourceImages:batchNormalizationState:destinationImages:]` now results in a crash. (This is an error on our end, since the relevant parameter is marked nonnull.)

This changes fixes a crash in test_activity_classifier.py, when testing on a Mac with a discrete AMD GPU on the current seed.

### DIFF
--- a/src/unity/toolkits/tcmps/mps_layers.mm
+++ b/src/unity/toolkits/tcmps/mps_layers.mm
@@ -356,8 +356,6 @@ void BNLayer::Forward(MPSImageBatch *_Nonnull src,
                       id<MTLCommandBuffer> _Nonnull cb,
                       bool is_train) {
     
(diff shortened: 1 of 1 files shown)


nickjong:
Yeah, this method doesn't appear in `MPSCNNBatchNormalization.h`, but it turns out that it's inherited from `MPSKernel` (and declared in `MPSKernel.h`).

At the very least I checked that the method exists in the 18A326 seeds, so this change doesn't introduce a new dependency on a more recent build version.

igiloh:
👍

PR #875, opened 2018-07-20
Fix wheel

abhishekpratapa:
- initialized variable fixes bug


znation:
@abhishekpratapa P.S. Please try to use more descriptive PR titles and descriptions, and commit messages. We may want to later track down what this change was related to, why it went in, etc. and linking to the issue will help. "Fix wheel" is a bit too broad of a description. 😃

PR #887, opened 2018-07-24
Added user metadata to identify TC Version in CoreML Models

### LINKED ISSUE #718: Add version information to MLModel export utilities.

abhishekpratapa:
- added user metadata to identify TC version 
- fixes #718

### DIFF
--- a/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
+++ b/src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
@@ -623,4 +623,11 @@ class ImageSimilarityModel(_CustomModel):
         mlmodel.short_description = _coreml_utils._mlmodel_short_description(model_type)
         mlmodel.input_description[self.feature] = u'Input image'
         mlmodel.output_description[output_name] = u'Distances between the input and reference images'
+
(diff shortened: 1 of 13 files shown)


src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -623,4 +623,11 @@ def export_coreml(self, filename):
         mlmodel.short_description = _coreml_utils._mlmodel_short_description(model_type)
         mlmodel.input_description[self.feature] = u'Input image'
         mlmodel.output_description[output_name] = u'Distances between the input and reference images'
+
+        _coreml_utils._set_model_metadata(mlmodel, self.__class__.__name__, {
+            'model': self.model,
+            'num_examples': str(self.num_examples),
+            'training_time': str(self.training_time)
```
TobyRoseman: I don't understand how training time is helpful information. 

Also it doesn't look like you're including turicreate_version like you are elsewhere.

abhishekpratapa: `turicreate_version` for higher order toolkits gets added when the `_set_model_metadata` function gets called. Training time is probably not helpful, I thought it might be cool to have to potentially see what kinds of devices people were training on.

PR #888, opened 2018-07-24
Style transfer fix data iterator

### LINKED ISSUE #636: Style transfer should work with 1 style image
I tried once with no max iterations, default to 5000, it was too long with CPU, at around 75, I stopped it.

So then I tried max iteration option at 100, it seems to hang at around 17...
model = tc.style_transfer.create(style, content, max_iterations = 100)

Is this iteration something that can be stopped at any time and then saved? Because currently it seems to freeze at random process iteration and I started again from 0...

shreyasvj25:
This PR fixes #636 
* when the number of content images is less than the batch_size, data iterator hangs! Fixed data iterator
* remove unused variables

### DIFF
--- a/src/unity/python/turicreate/toolkits/style_transfer/style_transfer.py
+++ b/src/unity/python/turicreate/toolkits/style_transfer/style_transfer.py
@@ -251,7 +251,6 @@ def create(style_dataset, content_dataset, style_feature=None,
     gram_chunks = [[] for _ in range(num_layers)]
     for s_batch in style_images_loader:
         s_data = _gluon.utils.split_and_load(s_batch.data[0], ctx_list=ctx, batch_axis=0)
-        results = []
         for s in s_data:
             vgg16_s = _vgg16_data_prep(s)
             ret = vgg_model(vgg16_s)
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -284,92 +283,96 @@ def create(style_dataset, content_dataset, style_feature=None,
...
-                    print('Total Loss: {:6.3f} | Content Loss: {:6.3f} | Style Loss: {:6.3f}'.format(cur_loss, print_content_loss, print_style_loss))
-                last_time = cur_time
-            if iterations == max_iterations:
-                print(hr)
-                break
+        try:
+            c_batch = content_images_loader.next()
+        except StopIteration:
+            print('Number of content images is less than the batch_size')
+            raise
```
gustavla: So this situation is necessarily a fatal error? Why can't we handle it more gracefully? What will happen when the `StopIteration` is re-raised?

shreyasvj25: Currently, StopIteration will be raised only when the number of content images is less than the batch_size.  
I don't see any other place else that StopIteration could be raised

src/unity/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -179,8 +179,11 @@ def create(style_dataset, content_dataset, style_feature=None,
         max_iterations = 5000
         if verbose:
             print('Setting max_iterations to be {}'.format(max_iterations))
-    # data loader
 
+    if len(content_dataset) < batch_size:
+        raise _ToolkitError('Number of content images is less than the batch_size')
```
znation: The user doesn't explicitly choose the batch size - so this doesn't seem like a reasonable error to give. Since the batch size is not set by the user anyway, can't we just arbitrarily change it under the hood to avoid this error?

gustavla: Batch size is a parameter for ST/OD/IC now, so the user can and might have set this value.

However, I feel like we just replicate the content images to avoid this altogether.

znation: It appears to be an optional parameter. So we should think about the desired UX when:

* The user passes in an explicit batch size that is larger than `len(content_images)`
* The user does not pass in a batch size, but our default is larger than `len(content_images)`.

In both cases, could we simply set batch_size to len(content_images) (and/or, throw in the csae that the user explicitly provides a nonsensical batch size w.r.t. the number of content images)?

PR #901, opened 2018-07-25
Update test_recommender.py to reflect changes from #833

nickjong:
I believe this change should get us back to all-passing tests again...


znation:
@nickjong Thanks for fixing that!

PR #904, opened 2018-07-26
Fixes #608

### LINKED ISSUE #608: Turicreate - Training process not proceed after 91 iteration , it loading for more than an hour @ GPU
I have trained around 3.5 k images for single category in GPU version, it took around 10 mins to create Sframe data, after that training process log started, it reached 91 iteration and keep on loading not proceed to next step - PFA

turilog

Total dataset size (3500 images)  - 91 MB, Maxi image size - 80 kb
Machine type : AWS deep learning

Any help appreciated !!

shantanuchhabra:
Got rid of the raw image column while shuffling. 

Using the data that @dhgokul sent to us, we noticed that just the raw image column was causing the size of the SFrame to blow up (from around 40 MB to around 760 MB) because of which `shuffle` was taking a long (but finite) amount of time. 

Since that was the bottleneck, we now get rid of the raw image column just during the shuffle. After a lightning fast shuffle, we add the raw image column back and training proceeds like before.

Consequently, training time for that model went down from 79 minutes to 39 minutes for Turi Create 4.3.2 on a GPU in Linux.

### DIFF
--- a/src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
+++ b/src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
@@ -79,8 +79,10 @@ class _SFrameDataSource:
         # If we're about to begin a new epoch, shuffle the SFrame if requested.
         row_index = self.cur_sample % len(self.sframe)
         if row_index == 0 and self.cur_sample > 0 and self.shuffle:
+            self.sframe = self.sframe.remove_column(_TMP_COL_RAW_IMAGE)
             self.sframe[_TMP_COL_RANDOM_ORDER] = _np.random.uniform(size=len(self.sframe))
             self.sframe = self.sframe.sort(_TMP_COL_RANDOM_ORDER)
+            self.sframe[_TMP_COL_RAW_IMAGE] = self.sframe[feature_column].apply(_convert_image_to_raw)
         self.cur_sample += 1
 
         # Copy the image data for this row into a NumPy array.
--- a/src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
+++ b/src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
@@ -52,6 +52,7 @@ class _SFrameDataSource:
     def __init__(self, sframe, feature_column, annotations_column,
                  load_labels=True, shuffle=True, samples=None):
         self.annotations_column = annotations_column
+        self.feature_column = feature_column
         self.load_labels = load_labels
         self.shuffle = shuffle
         self.num_samples = samples
@@ -61,7 +62,7 @@ class _SFrameDataSource:
         self.sframe = sframe.copy()
 
         # Convert images to raw to eliminate overhead of decoding
-        self.sframe[_TMP_COL_RAW_IMAGE] = self.sframe[feature_column].apply(_convert_image_to_raw)
+        self.sframe[_TMP_COL_RAW_IMAGE] = self.sframe[self.feature_column].apply(_convert_image_to_raw)
 
     def __iter__(self):
         return self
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/toolkits/object_detector/_sframe_loader.py
```diff
@@ -79,8 +80,10 @@ def _next(self):
         # If we're about to begin a new epoch, shuffle the SFrame if requested.
         row_index = self.cur_sample % len(self.sframe)
         if row_index == 0 and self.cur_sample > 0 and self.shuffle:
+            self.sframe = self.sframe.remove_column(_TMP_COL_RAW_IMAGE)
             self.sframe[_TMP_COL_RANDOM_ORDER] = _np.random.uniform(size=len(self.sframe))
             self.sframe = self.sframe.sort(_TMP_COL_RANDOM_ORDER)
+            self.sframe[_TMP_COL_RAW_IMAGE] = self.sframe[self.feature_column].apply(_convert_image_to_raw)
```
gustavla: Great work on reproing and coming up with a potential fix! However, I don't think this is the right one. The reason for pre-decoding images in a batch before training, is so that we don't have to do it every time we read an image. We could theoretically also do it using memoization, which might reduce initial latency. Each image gets read once per epoch, so if we re-decode the images every epoch, then we are back to decoding every image every time. The only difference now is that we do all of them in a batch, which I don't see the point of, since that will leave likely leave the GPU idle longer.

I have been discussing with @nickjong that perhaps after the performance improvements he made, we no longer need to pre-decode images. I think we should benchmark this, since then the fix is simply to remove this pre-processed column altogether.

I also want to point out that this only reduces the size of each image. If you make your images so large that their encoded size is the size of the decoded one in the repro, then presumably this issue would re-appear, right?

srikris: @gustavla All this does is to speed up the shuffle (because shuffling with images is currently slow). The net overall GPU idle time is shorter.

You are right, we should benchmark not using RAW and see if that's faster and if that's the case then remove RAW all together.

gustavla: I'm sure it might be faster than the current solution. However, since this decodes every time, it shouldn't be faster than simplifying it even further. I'm running some benchmarks on Linux with CUDA now. Maybe someone can run this on an iMac Pro too?

gustavla: On my tests with an SFrame where shuffling of raw images is actually not that slow, it was still slow enough to eliminate the benefits of pre-decoding. In the end, pre-decoding was only about 1% faster, which is a very modest gain (and not worth the risk of memory issues). On the Mac side, since we have better IO threading there, skipping pre-decoding should be even faster.

I vote to remove pre-decoding altogether!

nickjong: TL;DR I agree with everything Gustav is saying, with respect to the long-term solution here. However, I think this fix seems quite reasonable for our 5.0 release. A deeper change is riskier, and may be subsumed by future refactoring plans (for later releases).

The correctness of this change assumes only that removing the raw column, sorting, and decoding all the images again, as a single (parallelized) batch operation, is faster than sorting the entire SFrame including raw images.

Given the current architecture in this file, a naive (low-risk) attempt to remove decoding turns a parallelized batch decode (right after the infrequent shuffle) into a sequence of single-threaded decode operations (on each call to `_next`). This very possibly is still a win, but it's less obvious and might require more testing than it's worth doing right now.

Also, FWIW, on iMac Pro I'm seeing a slight performance improvement with these changes (exercising the MPS code path). So my vote is to both take this change as quick fix for 5.0 and plan to further improve the SFrame loading as part of a focus on SFrame performance

nickjong: Although let me test this on MacBook Pro, to validate the assumption that "sort encoded then decode" is faster than "sort decoded".

shantanuchhabra:
On an iMac Pro, a model trained without a raw image column altogether took 8m 3s, whereas a model trained by adding a raw image column after each shuffle took 13m 7s, so the speedup was significant.

gustavla:
Great, then this change is a no-brainer, really.

PR #909, opened 2018-07-27
Warning for CUDA GPUs if not enough free memory

### LINKED ISSUE #137: Turi For Cats and Dogs Image Classification
I was trying to use turi for cats and dogs image classification. My gpu is nvidia 970 , ram 4GB . i have installed mxnet_cu80 0.11.0 ..but the error comes out to be 

**RuntimeError: simple_bind error. Arguments:
data: (512, 3, 224, 224)
[01:47:51] src/storage/./pooled_storage_manager.h:102: cudaMalloc failed: out of memory**

### LINKED ISSUE #656: Using CPU to create model - MacBook Pro (Retina, 13-inch, Early 2015)
I'm running the object detection task.
====================================
MacOS Version 10.14 Beta  (18A293u)

MacBook Pro  (Retina, 13-inch, Early 2015) 
Processor 2.7 GHz Intel Core i5 Memory 8 GB 1867 MHz DDR3 
Startup Disk Macintosh HD 
Graphics Intel Iris Graphics 6100 1536 MB

gustavla:
This PR primarily introduces warnings if a user tries to use a CUDA GPU with insufficient free memory. This may happen if for instance Tensorflow has allocated the memory unbeknownst the user. 

- Note, the estimated memory requirements are experimental estimates. These experiments were conducted by first allocating a specific amount of memory and then trying to train the model. I did this for values that the should influence the memory usage (batch size, and number of style in ST) and then fit a simple line to it. Since it's only an estimate, I don't want to cause a fatal error, so this is just a warning that should help the user understand if there is subsequent fatal error.
- Also show name of CUDA GPUs (just like for MPS)
- Consolidate device printing code between OD/ST
- Provide feedback on macOS if CPU is used that an upgrade to 10.14+
  could potentially enable faster model creation. We don't know this for
  sure, because the code that checks for a capable GPU is in libtcmps,
  which is exactly what is not even compiled before 10.14.
- Facilitate use of static graph in OD. This reduces memory usage by
  13%. This allows training on a 2 GiB GPU with batch_size=16.

Note, this PR does not extend device verbosity to image classifier/similarity and only modifies OD and ST that currently prints this information. I think it would be reasonable to also add this to IC/IS, but did not include it in this PR.

Fixes #137
Fixes #656

### DIFF
--- a/src/unity/python/turicreate/toolkits/_internal_utils.py
+++ b/src/unity/python/turicreate/toolkits/_internal_utils.py
@@ -620,3 +620,27 @@ def _mac_ver():
         return tuple([int(v) for v in ver_str.split('.')])
     else:
         return ()
+
+def _print_neural_compute_device(cuda_gpus, use_mps, cuda_mem_req=None, has_mps_impl=True):
+    """
+    Print a message making it clear to the user what compute resource is used in
+    neural network training.
+    """
+    num_cuda_gpus = len(cuda_gpus)
+    if num_cuda_gpus >= 1:
+        gpu_names = ', '.join(gpu['name'] for gpu in cuda_gpus)
+
+    if use_mps:
+        from ._mps_utils import mps_device_name
+        print('Using GPU to create model ({})'.format(mps_device_name()))
+    elif num_cuda_gpus >= 1:
+        from . import _mxnet_utils
+        plural = 's' if num_cuda_gpus >= 2 else ''
+        print('Using GPU{} to create model ({})'.format(plural, gpu_names))
+        if cuda_mem_req is not None:
+            _mxnet_utils._warn_if_less_than_cuda_free_memory(cuda_mem_req, max_devices=num_cuda_gpus)
+    else:
+        import sys
+        print('Using CPU to create model')
+        if sys.platform == 'darwin' and _mac_ver() < (10, 14) and has_mps_impl:
+            print('NOTE: If available, an AMD GPU can be leveraged on macOS 10.14+ for faster model creation')
--- a/src/unity/python/turicreate/toolkits/_mxnet_utils.py
+++ b/src/unity/python/turicreate/toolkits/_mxnet_utils.py
@@ -44,20 +44,30 @@ def get_mxnet_state(model):
     return state
 
 
+def get_gpu_ids_in_use(max_devices=None):
+    from turicreate.util import _CUDA_GPUS
+    gpus = _CUDA_GPUS
+    num_gpus = get_num_gpus_in_use(max_devices=max_devices)
+    if num_gpus >= 0:
+        gpus = gpus[:num_gpus]
+    if max_devices is not None:
+        gpus = gpus[:max_devices]
+    return [gpu_info['index'] for gpu_info in gpus]
+
+
 def get_mxnet_context(max_devices=None):
-    from turicreate.util import _CUDA_GPU_IDS
+    from turicreate.util import _CUDA_GPUS
     import mxnet as _mx
     assert_valid_num_gpus()
     num_gpus = _tc_config.get_num_gpus()
-    if num_gpus == 0 or not _CUDA_GPU_IDS:
+    if num_gpus == 0 or not _CUDA_GPUS:
         return [_mx.cpu()]
     else:
-        if num_gpus == -1:
-            ctx = [_mx.gpu(i) for i in _CUDA_GPU_IDS]
-        elif num_gpus > 0:
-            ctx = [_mx.gpu(i) for i in range(num_gpus)]
+        gpu_indices = [gpu['index'] for gpu in _CUDA_GPUS]
+        if num_gpus > 0:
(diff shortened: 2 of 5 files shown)


src/unity/python/turicreate/toolkits/_mxnet_utils.py
```diff
@@ -99,6 +120,65 @@ def assert_valid_num_gpus():
...
+    available = all(gpu['memory_free'] > mem_mb for gpu in gpus)
+    warning = None
+    if capable and not available:
+        # Get GPU with the least amount of free memroy
+        gpu_mems = [gpu['memory_free'] for gpu in gpus]
+        gpu_idx = np.argmin(gpu_mems)
+
+        warning = '''
+        You may not have enough free GPU memory to create this model, which may
+        result in a fatal crash or lower performance. We recommend that you
```
TobyRoseman: `lower performance` -> `longer training time`?

src/unity/python/turicreate/toolkits/_mxnet_utils.py
```diff
@@ -99,6 +120,65 @@ def assert_valid_num_gpus():
...
+
+    if warning:
+        print('WARNING:')
+        print()
+        # Re-wrap text
+        lines = textwrap.wrap(textwrap.dedent(warning).lstrip(), width=75)
+        for line in lines:
+            print('   ', line)
+        print()
+        time.sleep(3)
```
TobyRoseman: Why the sleep?

src/unity/python/turicreate/util/__init__.py
```diff
@@ -717,19 +717,41 @@ def subprocess_exe(exe, args, setup=None, teardown=None,
 # Automatic GPU detection
 def _get_cuda_gpus():
     """
-    Returns a list of 0-based integer indices of available CUDA GPUs.
+    Returns a list of dictionaries, with the following keys:
+    - index (integer, device index of the GPU)
+    - name (str, GPU name)
+    - memory_free (float, free memory in MiB)
```
TobyRoseman: `MiB` -> `MB`? Also same question about the next line.

src/unity/python/turicreate/toolkits/_mxnet_utils.py
```diff
@@ -99,6 +120,65 @@ def assert_valid_num_gpus():
...
+    warning = None
+    if capable and not available:
+        # Get GPU with the least amount of free memroy
+        gpu_mems = [gpu['memory_free'] for gpu in gpus]
+        gpu_idx = np.argmin(gpu_mems)
+
+        warning = '''
+        You may not have enough free GPU memory to create this model, which may
+        result in a fatal crash or lower performance. We recommend that you
+        have at least {mem_req} MiB available on each GPU, while currently you
```
TobyRoseman: `MiB` -> `MB`?

src/unity/python/turicreate/toolkits/_internal_utils.py
```diff
@@ -620,3 +620,27 @@ def _mac_ver():
...
+    num_cuda_gpus = len(cuda_gpus)
+    if num_cuda_gpus >= 1:
+        gpu_names = ', '.join(gpu['name'] for gpu in cuda_gpus)
+
+    if use_mps:
+        from ._mps_utils import mps_device_name
+        print('Using GPU to create model ({})'.format(mps_device_name()))
+    elif num_cuda_gpus >= 1:
+        from . import _mxnet_utils
+        plural = 's' if num_cuda_gpus >= 2 else ''
```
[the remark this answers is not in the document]
gustavla: I like this change. Fixed.

src/unity/python/turicreate/toolkits/_mxnet_utils.py
gustavla: Agreed! Fixed.

src/unity/python/turicreate/toolkits/_mxnet_utils.py
gustavla: The thought is that fatal memory crashes typically fill up the screen with a backtrace, so waiting a few seconds will increase the chance of the user seeing this warning. I can remove it if you do not think it will have this effect.

src/unity/python/turicreate/util/__init__.py
gustavla: I don't feel strongly about this, so I'm OK switching it to MB. I think MiB is clearer since it's unambiguous, but again, I'm happy to change this.

src/unity/python/turicreate/toolkits/_mxnet_utils.py
TobyRoseman: Sleeping for 3 seconds is ok. I would just add a comment explaining why.

gustavla: Good idea! Pushed.

igiloh:
You should probably apply the same changes to the AC as well (at least all the changes in prints etc.)

PR #912, opened 2018-07-27
Resolve some more python3 compatibility issues with activity classification

nickjong:
- Pass list, not dict_keys, to SFrame.filter_by
- Use SArray.hash to avoid python3 hash randomization of strings
- Change the computation of random seed from time for compatibility with SArray.hash (this computation is lifted from `sarray.py`

### DIFF
--- a/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -202,7 +202,7 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
         _tkutl._raise_error_if_sframe_empty(validation_set, 'validation_set')
         validation_set = _tkutl._toolkits_select_columns(
             validation_set, features + [session_id, target])
-        validation_set = validation_set.filter_by(target_map.keys(), target)
+        validation_set = validation_set.filter_by(list(target_map.keys()), target)
         validation_set, mapping = _encode_target(validation_set, target, target_map)
         chunked_validation_set, _ = _prep_data(validation_set, features, session_id, prediction_window,
                                             predictions_in_chunk, target=target, verbose=False)
--- a/src/unity/python/turicreate/toolkits/activity_classifier/util.py
+++ b/src/unity/python/turicreate/toolkits/activity_classifier/util.py
@@ -78,12 +78,16 @@ def random_split_by_session(dataset, session_id, fraction=0.9, seed=None):
         print ("The dataset has less than the minimum of", _MIN_NUM_SESSIONS_FOR_SPLIT, "sessions required for train-validation split. Continuing without validation set")
         return dataset, None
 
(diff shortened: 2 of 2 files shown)


src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -202,7 +202,7 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
         _tkutl._raise_error_if_sframe_empty(validation_set, 'validation_set')
         validation_set = _tkutl._toolkits_select_columns(
             validation_set, features + [session_id, target])
-        validation_set = validation_set.filter_by(target_map.keys(), target)
+        validation_set = validation_set.filter_by(list(target_map.keys()), target)
```
znation: Hmm. This fix is fine for now, but I wonder if `filter_by` should accept any iterable (not just a list) which would allow the code to work as is?

src/unity/python/turicreate/toolkits/activity_classifier/util.py
```diff
@@ -78,26 +78,31 @@ def random_split_by_session(dataset, session_id, fraction=0.9, seed=None):
...
-        random.seed(hash(session_id) + seed)
+    # The boolean filter is a pseudorandom function of the session_id and the
+    # global seed above, allowing the train-test split to vary across runs using
+    # the same dataset.
+    def random_session_pick(session_id_hash):
+        random.seed(session_id_hash)
         return random.uniform(0, 1) < fraction
-    
-    chosen_filter = dataset[session_id].apply(random_session_pick)
+    chosen_filter = dataset[session_id].hash(seed).apply(random_session_pick)
```
znation: This appears to be outside the function containing the line immediately above. Please precede it with an empty line.

src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
nickjong: It turns out that this boils down to the fact that you cannot pass dict_keys to the SArray initializer:

```
>>> dict = {'alpha': 0, 'beta':1}
>>> sa = tc.SArray(dict.keys())
  File "/Users/jong/Source/turicreate/debug/src/unity/python/turicreate/data_structures/sarray.py", line 422, in __init__
    raise TypeError("Unexpected data source. " \
TypeError: Unexpected data source. Possible data source types are: list, numpy.ndarray, pandas.Series, and string(url)
```

Yeah, not ready to change SArray.__init__ now, but is this worth tracking?

znation: Yeah, I think so. Let's open another issue for it. I think `init` should allow any iterable for this parameter.

nickjong: Sounds good, filed #915

igiloh:
FWIW - LGTM! 😄

PR #923, opened 2018-07-30
Upgraded Vega

### LINKED ISSUE #456: Initial window size for canvas should be at least as big as the plot that appears

abhishekpratapa:
- Fixes #456
- Upgrades Vega

### DIFF
--- a/src/visualization/Turi Create Visualization/src/user_interface/package.json
+++ b/src/visualization/Turi Create Visualization/src/user_interface/package.json
@@ -12,8 +12,8 @@
         "react": "^16.2.0",
         "react-dom": "^16.2.0",
         "react-json-pretty": "^1.7.5",
-        "vega": "^3.2.1",
-        "vega-embed": "^3.2.0",
+        "vega": "4.2.0",
+        "vega-embed": "3.18.2",
         "vega-tooltip": "^0.6.1"
     },
     "scripts": {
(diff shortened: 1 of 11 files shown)


src/visualization/Turi Create Visualization/src/user_interface/package.json
```diff
@@ -12,8 +12,8 @@
         "react": "^16.2.0",
         "react-dom": "^16.2.0",
         "react-json-pretty": "^1.7.5",
-        "vega": "^3.2.1",
-        "vega-embed": "^3.2.0",
+        "vega": "4.2.0",
+        "vega-embed": "3.18.2",
         "vega-tooltip": "^0.6.1"
```
znation: Do we still need vega-tooltip?

abhishekpratapa: Yeah for now with the implementation we have, no issues there, but if we need to change it, it should be a separate PR

znation: So we do still need it, but our existing version is still compatible?

abhishekpratapa: yeah, we still need it and our existing version is compatible.

PR #932, opened 2018-07-31
Support for initializing SArray using numpy.matrix

### LINKED ISSUE #357: NDArray: np.matrix crashes Python
I don't really expect `np.matrix` to work here, nor do I recommend anyone to ever use `np.matrix` (Ever.). However, this is a bit alarming.

Python 3 (macOS, python from python.org):

```python
>>> tc.SArray([np.matrix([[1, 2], [3, 4]])])
[1]    11438 illegal hardware instruction  python
```

Python 2 (macOS, system's python):
```
>>> tc.SArray([np.matrix([[1, 2], [3, 4]])])
[1]    11597 segmentation fault  python
```

nickjong:
Operations on a numpy.matrix object yield another numpy.matrix object, and each instance always has two-dimensional shape. When our SArray type-deduction or data-loading code attempts to traverse this data structure recursively, we get an unbounded recursion. The simplest solution seems to be to guard explicitly against numpy.matrix and convert to numpy.ndarray up front.

Fixes #357

### DIFF
--- a/src/unity/python/turicreate/data_structures/sarray.py
+++ b/src/unity/python/turicreate/data_structures/sarray.py
@@ -353,6 +353,12 @@ class SArray(object):
             self.__proxy__ = data.__proxy__
         else:
             self.__proxy__ = UnitySArrayProxy()
+
+            # Guard against numpy.matrix, which cannot be traversed recursively,
+            # since subscripting a matrix always yields another (2d) matrix.
+            if HAS_NUMPY and isinstance(data, numpy.matrix):
+                data = numpy.array(data)
+
             # we need to perform type inference
             if dtype is None:
                 if HAS_PANDAS and isinstance(data, pandas.Series):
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -353,6 +353,12 @@ def __init__(self, data=[], dtype=None, ignore_cast_failure=False, _proxy=None):
             self.__proxy__ = data.__proxy__
         else:
             self.__proxy__ = UnitySArrayProxy()
+
+            # Guard against numpy.matrix, which cannot be traversed recursively,
+            # since subscripting a matrix always yields another (2d) matrix.
+            if HAS_NUMPY and isinstance(data, numpy.matrix):
+                data = numpy.array(data)
```
gustavla: I recommend using `data = numpy.asarray(data)` here. I think this is the recommended way to go from a matrix to an array. This is *much* faster on large data. In my measurements, `numpy.array` scales linearly with data size, while `numpy.asarray` is constant (it should be constant, since it should just re-cast the same memory as an array).

nickjong:
Er, what .gitlab-ci.yml file? Nothing to see here 😄

PR #934, opened 2018-08-01
Style transfer - allow flexible shaped input for coreml model 

shreyasvj25:
* make coreml model accept any input shape (greater than 64 x 64)
* added unit test for flexible input
* update coreml model image in user docs


src/unity/python/turicreate/test/test_style_transfer.py
```diff
@@ -193,26 +193,37 @@ def test_get_styles_success(self):
...
+            coreml_output = coreml_model.predict(
+                {self.content_feature: pil_img, 'index': index_data},
+                usesCPUOnly=True)
             img = next(iter(coreml_output.values()))
             img = np.asarray(img)
             img = img[..., 0:3]
+            return img
 
-            self.assertEqual(img.shape, (256, 256, 3))
+    def test_export_coreml(self):
```
znation: It doesn't need to be done as part of this PR, but I think we should also add an Objective-C unit test (which we don't really have infrastructure for yet, but could possibly piggyback on our C/C++ tests). That will test the real CoreML code interface. The Python interface is much more forgiving when it comes to model input/output mismatches in type/shape/validation.

PR #941, opened 2018-08-03
Allow flexible types to take builtins types

### LINKED ISSUE #859: `from builtins import str` causes Turi Create crash
Repro steps (Python 2.7 only):

```sh
pip install future
```

```python
from builtins import str
import turicreate as tc
sf = tc.SFrame({'a': ['foo', str('bar')]})
```

Expected: works
Actual: segfault

TobyRoseman:
Fixes #859.


znation:
@TobyRoseman Looks great!

PR #956, opened 2018-08-09
Fix symbol exporting in libTuriCreate

### LINKED ISSUE #843: libTuriCreate.dylib does not contain the right symbols.
After running `build_capi.sh`, it does not look like the right symbols are being exported in libTuriCreate.dylib or in the framework. For example, the following command finds nothing:  `nm target/capi/libTuriCreate.dylib | grep -I sarray`. Looking at `nm target/capi/libTuriCreate.dylib` seems it's missing pretty much all of the symbols.

znation:
* Uses OBJECT library syntax in CMake to create a single reusable
  compilation of capi that will export all symbols in targets that link
  this object.
* Requires a new version of CMake to use that OBJECT syntax correctly,
  with 3.12 being the minimum version to properly support downstream
  dependencies of OBJECT libraries.
* Removes CMake automatic installation logic, as installing cmake from
  source is time consuming, and probably not what most developers or
  CI systems want to do anyway; we should just require the developers or
  system maintainers to install a recent version of CMake as a
  prerequisite. (Alternative would be to update our CMake installation
  script, but the cost of updating and testing that script as we take
  dependencies on new CMake versions doesn't seem worth it.)

Fixes #843


hoytak:
I'm not really comfortable updating cmake to require a version later than 3.9; I don't think most linux distributions ship with a later version than that. 

Yucheng had some ideas about this -- let's sync up with him on that today.

znation:
@hoytak is right; this is not the right fix. Most people building the repo won't have a new enough CMake and it'll error out and be quite cumbersome to work around. There is a potential fix that doesn't require a newer CMake. @hoytak can try that one.

PR #958, opened 2018-08-10
Fixed bug with validation sets having additional classes.

hoytak:
This PR adds tests for validation classes having additional classes not
seen in the training classes.  These tests exposed problems with both
the evaluation and exporting to CoreML, both of which are fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,7 +213,7 @@ message(WARNING "OpenMP Libraries were not found")
 #*                                                                        */
 #**************************************************************************/
 if (CLANG)
-set(CPP11_FLAGS "-std=c++11 -stdlib=libc++ -Wno-deprecated-register -Wno-enum-compare -Wno-conversion-null -Wno-constant-logical-operand -Wno-parentheses-equality -ftemplate-depth=900" CACHE STRING "C++11 enabling flags")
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -213,7 +213,7 @@ message(WARNING "OpenMP Libraries were not found")
 #*                                                                        */
 #**************************************************************************/
 if (CLANG)
-set(CPP11_FLAGS "-std=c++11 -stdlib=libc++ -Wno-deprecated-register -Wno-enum-compare -Wno-conversion-null -Wno-constant-logical-operand -Wno-parentheses-equality -ftemplate-depth=900" CACHE STRING "C++11 enabling flags")
+  set(CPP11_FLAGS "-std=c++11 -stdlib=libc++ -Wno-stdlibcxx-not-found -Wno-deprecated-register -Wno-enum-compare -Wno-conversion-null -Wno-constant-logical-operand -Wno-parentheses-equality -ftemplate-depth=900" CACHE STRING "C++11 enabling flags")
```
znation: Why is this needed?

hoytak: The version of clang shipping with the latest Mohave beta warns on if we use libc++ instead of libcxx, which causes compilation to fail with -Werror.

znation: What do you mean, `libc++` instead of `libcxx`? Are those two different names for the same thing, and one is deprecated? If that's the case, can we just use `libcxx` all the time instead of not warning in this situation?

PR #961, opened 2018-08-11
Link all test dependencies dynamically

### LINKED ISSUE #861: Test binaries should dynamically link their dependencies
Current test binaries come out to 50+ MB, because they are statically linking (some? most? all?) their dependencies. We should be dynamically linking where possible to save on incremental build time, disk space used, etc., especially considering our actual Python package dynamically links the same thing (so it's a better test case to link the same way the Python process will).

znation:
Instead of statically linking each of the dependencies into each test
binary, this change now statically links all test dependencies into
a single shared library, then links each of the tests against that
single shared library.

This saves ~6 GB of disk space in debug mode when all tests are built.

Fixes #861


TobyRoseman:
This looks good to me. I also suspect it might solve my linking issues with setting up CI.

Two questions -
1 - Why are you removing the time series tests?
2 - Also are you sure we should be removing the fault tests? I know that directory is commented out, but that's quite a bit of test code, some of which may still be useful.

znation:
@TobyRoseman To answer your questions:

1. Because they don't currently build or run (note that the parent directory doesn't include that directory in its `CMakeLists.txt`, nor does the dependency `timeseries` exist currently). It appears the code-under-test was removed in 4.0.

2. Same reasoning as above; perhaps the code is useful (I'm not sure - what does `fault` actually do/test?). But since it doesn't currently build or run, I'm not sure it's worth the effort to get it building and running again. And I don't think it's worth keeping in the repo if it doesn't build or run. It'll always be in the source history, so if we want to bring it back at some point, we can get it back.

PR #963, opened 2018-08-12
add style transfer to api doc

### LINKED ISSUE #959: Add Style Transfer to API docs
The [API Docs](
https://apple.github.io/turicreate/docs/api) contain no mention of Style Transfer.

We need to add an `.rst` file to `src/unity/python/doc/source/`.

shreyasvj25:
* fixes #959

### DIFF
--- a/src/unity/python/doc/source/turicreate.toolkits.style_transfer.rst
+++ b/src/unity/python/doc/source/turicreate.toolkits.style_transfer.rst
@@ -6,7 +6,7 @@
 .. currentmodule:: turicreate.style_transfer
 
 
-object detector
+style transfer
 ---------------
 .. autosummary::
   :toctree: generated/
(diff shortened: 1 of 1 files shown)


src/unity/python/doc/source/turicreate.toolkits.style_transfer.rst
```diff
@@ -0,0 +1,16 @@
+:mod:`style_transfer`
+==========================
+
+.. automodule:: turicreate.toolkits.style_transfer
+
+.. currentmodule:: turicreate.style_transfer
+
+
+object detector
```
TobyRoseman: This line needs to be change, probably to `style transfer`.

shreyasvj25: Nice catch, fixed it

PR #967, opened 2018-08-13
Updated JSON reader to throw error on missing file.

hoytak:


src/unity/lib/unity_sarray.cpp
```diff
@@ -234,66 +234,80 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
...
+          break;
         }
-      } else {
-        logstream(LOG_PROGRESS) << "Unable to read " << sanitize_url(p.first) << std::endl;
-      }
-    }
+      case fileio::file_status::DIRECTORY:
+        break; 
+      case fileio::file_status::MISSING:
+        log_and_throw("File '" + p.first + "' not found.");
```
znation: Should these be `log_and_throw_io_failure`?

src/unity/lib/unity_sarray.cpp
```diff
@@ -234,66 +234,80 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
...
+                << "Unable to read " << sanitize_url(p.first) << std::endl;
+          }
+          break;
         }
-      } else {
-        logstream(LOG_PROGRESS) << "Unable to read " << sanitize_url(p.first) << std::endl;
-      }
-    }
+      case fileio::file_status::DIRECTORY:
+        break;
```
znation: Should reading from a directory error out with a user-facing message? It seems this logic may result in an empty SArray/SFrame when attempting to read a directory, which is not quite what the user might expect.

hoytak: Updated in PR

hoytak:
Updated comments.

PR #972, opened 2018-08-14
Fix symbol exporting in libTuriCreate

### LINKED ISSUE #843: libTuriCreate.dylib does not contain the right symbols.

hoytak:
* Uses OBJECT library syntax in CMake to create a single reusable
  compilation of capi objects that will export all symbols in targets that link
  this object.

  Fixes #843


configure
```diff
@@ -22,9 +22,6 @@ function print_help {
   echo
   echo "  --install-python-toolchain        Install python in local virtualenv."
   echo  
-  echo "  --with-system-cmake (default)     Use system cmake, if avaliable."
```
znation: Probably want to bring this back in.

znation: Along with all the other cmake-related stuff and the shell script that builds it.

hoytak: Oops, I thought I had.  Good catch.

PR #983, opened 2018-08-16
style transfer toolkit update to calculate iterations based on style images

### LINKED ISSUE #680: Style transfer should set the number of iterations automatically
- We are automatically set to 5K. We should set based on # of images.

abhishekpratapa:
- Updated max iterations to be based on the number of style images
- fixes #680


abhishekpratapa:
Got it, changed it. haha

Issue #991, opened 2018-08-17
Activity Classifier auto validation split should not require 100 sessions


nickjong:
See #964 for full context.

Currently, with `validation_set='auto'`, we only create a validation set if the data contains 100 sessions. But each session comprises an unbounded number of rows, which we'll eventually downsample according to the `prediction_window` parameter, and also chunk into 20-prediction-window segments. We don't actually support sequential reasoning across more than 20 windows at a time.

It's not clear what exactly the right threshold is, but it should probably be some function of prediction windows or segments, not sessions.

Probably this should wait until after the core AC implementation moves to C++.

PR #998, opened 2018-08-17
Move activity_classifier logic for whether to create an auto validation set to create function

### LINKED ISSUE #964: Activity Classifier random_split_by_session should support datasets with fewer than 100 sessions
Hello, making this new thread since i erroneously discussed this in an unrelated thread (https://github.com/apple/turicreate/issues/813)

First post: 

Total machine learning noob here but came across Turi because I have a task that requires somewhat reliably going from standing-to-kneeling posture (and ideally vice versa). I was following along with the documentation, and made sure my preprocessed CSV mirrors the format used in the table in the "advanced usage" section. 

I'm running the following python script to generate a model: 

```
import turicreate as tc
dataSFrame = tc.SFrame('motion_data_all.csv')

# Train/test split by recording sessions
train, test = tc.activity_classifier.util.random_split_by_session(dataSFrame, session_id='exp_id', fraction=0.8)

# Create an activity classifier
motion_model = tc.activity_classifier.create(train, session_id='exp_id', target='motion', prediction_window=50)

metrics = motion_model.evaluate(test)
print(metrics['accuracy'])

# Save the model for later use in Turi Create
motion_model.save('motion_model.model')

# Export for use in Core ML
motion_model.export_coreml('MotionActivityClassifier.mlmodel')

# try out the model 
model_test_data = dataSFrame[(dataSFrame['motion'] == 'sit') & (dataSFrame['exp_id'] == 1)][500:1000]
motion_model.predict(model_test_data, output_frequency='per_window')
```

My training data is extremely small, as far as ML training data goes. Approx 3.5k rows of standing/sitting/kneeling, each accurately labeled. In the doc you use the HAPT data and then process it into a format with `exp_id` and the activity label. My data set mirrors that end result. 

When I run the following line in spyder: 

```
metrics = motion_model.evaluate(test)
print(metrics['accuracy'])
```

**I get the following error: `ToolkitError: Size of prediction probability vector(3) != number of classes(1).`**

Just to be clear I'm not completely messing up from the start, I've included some screenshots from the file I use with the following: `dataSFrame = tc.SFrame('motion_data_all.csv')`.

screenshot 2018-08-11 18 24 19
screenshot 2018-08-11 18 24 34

When I run the line: `motion_model = tc.activity_classifier.create(train, session_id='exp_id', target='motion', prediction_window=50)` I get the following output: 
screenshot 2018-08-11 18 28 13

Every project is unique so I don't expect anyone to debug this for me but any advice pointing me on what to read to learn how to resolve/understand that following error (`ToolkitError: Size of prediction probability vector(3) != number of classes(1).`) would be great.

nickjong:
We expose tc.activity_classifier.util.random_split_by_session as public API, so it should perform the split unconditionally. The logic to only create an auto validation set if we have 100 sessions belongs in the activity_classifier.create function itself

Fixes #964

### DIFF
--- a/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -166,6 +166,10 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
     _tkutl._raise_error_if_sarray_not_expected_dtype(dataset[session_id], session_id, [str, int])
 
     if isinstance(validation_set, str) and validation_set == 'auto':
+        # Computing the number of unique sessions in this way is relatively
+        # expensive. Ideally we'd incorporate this logic into the C++ code that
+        # chunks the raw data by prediction window.
+        # TODO: https://github.com/apple/turicreate/issues/991
         unique_sessions = _SFrame({'session': dataset[session_id].unique()})
         if len(unique_sessions) < _MIN_NUM_SESSIONS_FOR_SPLIT:
             print ("The dataset has less than the minimum of", _MIN_NUM_SESSIONS_FOR_SPLIT, "sessions required for train-validation split. Continuing without validation set")
--- a/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -28,6 +28,7 @@ from turicreate.toolkits._model import CustomModel as _CustomModel
 from turicreate.toolkits._model import PythonProxy as _PythonProxy
 
 from .util import random_split_by_session as _random_split_by_session
+from .util import _MIN_NUM_SESSIONS_FOR_SPLIT
 
 
 def create(dataset, session_id, target, features=None, prediction_window=100,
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -28,6 +28,7 @@
 from turicreate.toolkits._model import PythonProxy as _PythonProxy
 
 from .util import random_split_by_session as _random_split_by_session
+from .util import _MIN_NUM_SESSIONS_FOR_SPLIT
```
igiloh: Actually - you could just move this param to this file. It's not needed in the util file anymore.

src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -165,7 +166,12 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
     _tkutl._raise_error_if_sarray_not_expected_dtype(dataset[session_id], session_id, [str, int])
 
     if isinstance(validation_set, str) and validation_set == 'auto':
-        dataset, validation_set = _random_split_by_session(dataset, session_id)
+        unique_sessions = _SFrame({'session': dataset[session_id].unique()})
```
igiloh: This calculation of the number of sessions is actually quite heavy, and takes a relatively very long time for larger datasets. That's the main reason we've added a `num_sessions` output to `_prep_data()`.

As you've mentioned in https://github.com/apple/turicreate/issues/964#issuecomment-413922347 it would probably be a better idea to do the train-test split as part of the chunking step.

src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
nickjong: I'm not a python expert, but AFAICT moving this constant from util to _activity_classifier means that either
* test_activity_classifier must copy the constant, so that it is defined in two places, or
* toolkits/activity_classifier/__init__.py must export it as a fourth symbol, after `['create', 'ActivityClassifier', 'util']`.

For now, I just went with leaving it in the util.py module

src/unity/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
nickjong: Good idea, I added a TODO linking to #991

PR #1001, opened 2018-08-17
Build should error out if no lapack

### LINKED ISSUE #175: Lapack not properly detected on build
When building on ubuntu 16.04, the build process does not error out if lapack is not present.  Everything compiles fine, but the use of most of the machine learning algorithms results in `std::logic_error: solve(): use of LAPACK must be enabled`.

TobyRoseman:
fixes #175

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,5 @@
 project(Turi)
 
-
 # We require the most recent version of cmake and automatically
 # install the correct version when using the cmake lists
 cmake_minimum_required(VERSION 2.8.3)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -42,6 +41,12 @@ else()
   message("Skipping python libraries.")
 endif()
 
+if(NOT APPLE)
```
znation: Why if not Apple? We don't need lapack on macOS?

TobyRoseman: Yes, I believe so. It seems we never use BLAS or LAPACK directly. It's just a dependency of armadillo. [It looks like](https://github.com/apple/turicreate/blob/master/src/external/armadillo/CMakeLists.txt#L162) on macOS armadillo uses the Accelerate Framework instead of the BLAS and LAPACK.

@hoytak or @srikris - can you confirm?

znation: I see -- sounds right to me.

Issue #876, opened 2018-07-20
Warn on uninitialized struct/class members


znation:
Currently, we don't warn or error in the build on uninitialized struct/class members. When those structs/classes are put on the heap, those members may be pointing to random bits in memory, and therefore give different behavior in debug and release mode, and weird unreliable behavior in release mode (especially with optimizations turned on). This is the root cause of #777.

We should turn on these build warnings/errors so we are able to prevent this issue from reoccurring in general.

hoytak:
Resolved by https://github.com/apple/turicreate/pull/1003

PR #1003, opened 2018-08-17
Added -Wunitialized to flags.

hoytak:
Turns out this didn't reveal any failures, which makes sense -- a month ago, I ran through things with -fsanitize=undefined and fixed a few of these.


hoytak:
Resolves https://github.com/apple/turicreate/issues/876

Issue #986, opened 2018-08-17
Activity classifier data prep computes majority, not mode


nickjong:
The activity classifier downsamples the raw data, so that *k* consecutive rows are assigned the same target classification. To compute the chunk's label from the *k* raw labels, it performs the [Boyer-Moore majority vote algorithm](https://en.wikipedia.org/wiki/Boyer–Moore_majority_vote_algorithm). However, this algorithm only guarantees that if the data has a majority value (occurs at least *k/2* times), then it outputs that value. Otherwise, it outputs an arbitrary element.

In particular, suppose *k* = 9 and the raw labels are `ABABABABC`. Then the Boyer-Moore algorithm outputs `C`, instead of one of the two modes.

Since we convert the labels into integers in 0..(num classes - 1), we should just generate a histogram. Computing the mode in this fashion is *O(n+k)* time and *O(k)* space, which for small *k* is just as good as Boyer-Moore's *O(n)* time and *O(1)* space.

znation:
@nickjong This sounds like it would affect the accuracy of the resulting model - do you have a sense for how much effect it would have?

nickjong:
It depends on the data. Honestly, I suspect that this is fine for most datasets most of the time. If the window only has two unique labels, then the algorithm is correct. So it's only wrong if the window has three or more labels, and none are in the majority.

For example, the HAPT dataset in the user's guide has a prediction window of one second's worth of samples. Then the algorithm _could_ do the wrong thing if the window contains an interval of less than half a second labeled with A, preceded and followed by two distinct labels B and C. You would think that for most domains, you would choose a prediction window whose granularity matches the data labeling granularity to some degree.

But while I think this may not be a critical bug to fix (AFAICT, we've had it for a while), it should not be hard to implement a fix. The histogram may even be faster than Boyer-Moore, since it involves fewer branch instructions.

igiloh:
@nickjong nice catch! I wasn't aware of this disadvantage of Boyre-Moore algorithm.

@znation - As Nick has mentioned, this shouldn't really affect the correctness too much (if at all). Most prediction windows are expected to contain a single label, besides transition windows that may have two.  
In the HATP example Nick mentioned - it's not likely that a user has done 3 completely different activities (e.g. walking, running and swimming) within a single second. This is probably indicates either wrong data labeling or selecting an inappropriate window size - both of which are errors of a much larger magnitude.

[...]

PR #1005, opened 2018-08-17
Fixed mode calculation for data pre-processing Activity Classifier

abhishekpratapa:


src/unity/toolkits/activity_classification/sequence_iterator.cpp
```diff
@@ -30,21 +30,20 @@ static std::map<std::string,size_t> generate_column_index_map(const std::vector<
 
  * \return    The most frequent value within the given vector.
  */
+
 static double vec_majority_value(const flex_vec& input_vec) {
-    size_t counter = 0;
-    double candidate = 0.0;
+    std::map<double, int> count_map;
```
znation: Let's use `std::vector<int>` instead of `std::map` here. We shouldn't need a map, when the indices are themselves integers (though the fact that they come into this function as doubles is quite weird).

znation: Or `vector<pair<int, int>>` if that makes it easier to preserve the `max_element` logic below.

abhishekpratapa:
Fixes #986

src/unity/toolkits/activity_classification/sequence_iterator.cpp
```diff
@@ -24,27 +24,29 @@ static std::map<std::string,size_t> generate_column_index_map(const std::vector<
...
-        if (counter == 0) {
-            candidate = value;
-            counter = 1;
-        } else if (candidate == value) {
-            counter++;
-        } else {
-            counter--;
+        int value = static_cast<int>(input_vec[i]);
+
+        assert(static_cast<double>(static_cast<double>(input_vec[i])) == input_vec[i]);
```
nickjong: I think you want the inner cast here to return `int`, since casting to `double` twice is redundant. Might also be worth a simple comment, something like "Each value should be in the index of a class label."

src/unity/toolkits/activity_classification/sequence_iterator.cpp
```diff
@@ -24,27 +24,29 @@ static std::map<std::string,size_t> generate_column_index_map(const std::vector<
...
+
+        if(histogram.size() < (value + 1)){
+          histogram.resize(value + 1);
```
nickjong: It's a real shame that we have to potentially resize the vector multiple times on each call to `vec_mode`. This is the main way in which your version of this code is much slower than the old version!

It might be worth creating a single std::vector<int> instance and reusing it across all the calls to `vec_mode` inside a single call to `_activity_classifier_prepare_data_impl`. In the same file, you can see how there's one `curr_window_targets` vector being created in `_activity_classifier_prepare_data_impl` and e.g. passed by reference to `finalize_chunk`. Similarly, you could create a std::vector<int> (maybe called something like `histogram_buffer`) and pass it into `finalize_chunk` or `vec_mode` as appropriate.

For bonus points, note that we could determine the largest value of `input_vec` as it's being populated. (In practice, `input_vec` always seems to be `curr_window_targets`, which is just a cache of the class labels we've seen since the last chunk.) So you could even determine the correct size of the histogram vector up front.

nickjong: After giving this some more thought, although what I suggested would definitely be faster, it's not clear that it matters, given the other work going on in `_activity_classifier_prepare_data_impl`. And there's something to be said for keeping the code simpler. So feel free to take this comment as just an idea, could go either way. (The cast in the assertion should be fixed, though.)

src/unity/toolkits/activity_classification/sequence_iterator.cpp
```diff
@@ -144,7 +145,7 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
 
     // Build a dict pf the column order by column name, to later access within the iterator
     auto column_index_map = generate_column_index_map(data.column_names());
-
+
```
nickjong: Nit: drop this whitespace change. It looks like we prefer no trailing whitespace here....

src/unity/toolkits/activity_classification/sequence_iterator.cpp
```diff
@@ -24,27 +24,30 @@ static std::map<std::string,size_t> generate_column_index_map(const std::vector<
...
+          histogram.resize(value + 1);
         }
+
+        histogram[value]++;
     }
-    return candidate;
+
+    auto majority = std::max_element(histogram.begin(), histogram.end());
+    
+    return (majority - histogram.begin());
```
znation: What is this subtraction ...? I would have expected something like `return majority;` here. Or if majority is an iterator, `return *majority;`.

src/unity/toolkits/activity_classification/sequence_iterator.cpp
```diff
@@ -24,27 +24,30 @@ static std::map<std::string,size_t> generate_column_index_map(const std::vector<
...
-            candidate = value;
-            counter = 1;
-        } else if (candidate == value) {
-            counter++;
-        } else {
-            counter--;
+        int value = static_cast<int>(input_vec[i]);
+
+        // Each value should be in the index of a class label
+        assert(static_cast<double>(static_cast<int>(input_vec[i])) == input_vec[i]);
```
znation: Use `DASSERT_EQ`. I think with those macros in C API, we return an error through the `tc_error**` parameter, instead of crashing in debug mode.

src/unity/toolkits/activity_classification/sequence_iterator.cpp
abhishekpratapa: It's the offset. std::max_element returns the iterator. Subtracting the iterator at the max element with the iterator in the beginning gives the offset. I'm changing it to std::distance to make it more evident what I'm doing.

abhishekpratapa: Since it's a random access iterator, the std::distance function uses the `-` operator

nickjong: I think we can clarify things by changing the variable name from `majority` to something like `largest_count`. Then it's more clear that we can't return `*largest_count` but must instead compute the corresponding index into the vector (getting the argmax instead of the max). Might also be worth a comment

Issue #992, opened 2018-08-17
Python unit test failure in test_json


znation:
Reproduced on Linux in Python 2.7; may reproduce elsewhere.

```
_________________________ JSONTest.test_malformed_json _________________________

self = <turicreate.test.test_json.JSONTest testMethod=test_malformed_json>

    def test_malformed_json(self):
>       self.assertRaises(IOError, SArray.read_json, './malformed.json')

test_json.py:254: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../data_structures/sarray.py:585: in read_json
    proxy.load_from_json_record_files(_make_internal_url(filename))
turicreate/cython/cy_sarray.pyx:72: in turicreate.cython.cy_sarray.UnitySArrayProxy.load_from_json_record_files
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   RuntimeError: Unable to parse /builds/turi/turicreate/deps/env/lib/python2.7/site-packages/turicreate/test/malformed.json. It does not appear to be in JSON record format. A list of dictionaries is expected

turicreate/cython/cy_sarray.pyx:75: RuntimeError
----------------------------- Captured stdout call -----------------------------
Parsing JSON records from /builds/turi/turicreate/deps/env/lib/python2.7/site-packages/turicreate/test/malformed.json
--------------------------- Captured stdout teardown ---------------------------
```

TobyRoseman:
Looks related to #967

znation:
@TobyRoseman Agreed, it does. Assigning to @hoytak.

znation:
@hoytak I'm not sure what you mean re: "old test" and "new version". This failure is from a single commit of the code (`f6840ebb45394140db65db74cc031eafdaeec075`), built and running tests.

znation:
Fixed with #1010.

PR #1010, opened 2018-08-20
Fixed malformed json testing.

hoytak:


src/unity/python/turicreate/test/test_json.py
```diff
@@ -251,8 +252,36 @@ def test_variant_to_json(self):
         ]]
 
     def test_malformed_json(self):
-        self.assertRaises(IOError, SArray.read_json, './malformed.json')
```
TobyRoseman: Do you want to also delete `malformed.json`?

znation: Or keep it, and just switch out `IOError` for `RuntimeError`?

TobyRoseman: Yeah, in a standalone file it is much easier to see why it's malformed.

srikris:
Fixes #992

src/unity/python/turicreate/test/test_json.py
hoytak: The problem is that there is actually an additional hidden error -- if the test is not run from the current directory, then it can't find the file and raises an IOError.  If it is run from the current directory, it finds it and raises a RuntimeError.  Thus the current test wasn't correct -- in many cases, the actual error raised was the file-not-found.  Writing out the file explicitly solves this.

PR #1017, opened 2018-08-21
Enhance evaluation toolkit to accept explicit index map

### LINKED ISSUE #969: Activity classifier should support test dataset with fewer unique labels than the training set
As described in #964 - the user has introduced larger number of labels in the test set than the number of labels the model have been trained on.  
Instead of giving him an explicit error right during the `predict()` portion of `evaluate()` - he got a less intuitive error:
```
ToolkitError: Size of prediction probability vector(3) != number of classes(1)
```

We should probably detect these cases explicitly, right in the beginning of `predict()`, and give an informative error.

### LINKED ISSUE #1007: Evaluation toolkit assumes class index map can be inferred from test data
When evaluating the log loss of a model trained with _k_ classes, an exception is thrown if the test data doesn't happen to span all _k_ classes:
```
>>> targets = tc.SArray(['a', 'b', 'a'])
>>> preds = tc.SArray([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [1.0, 0.0, 0.0]])
>>> tc.evaluation.log_loss(targets, preds)
  File "/Users/jong/Source/turicreate/debug/src/unity/python/turicreate/toolkits/evaluation.py", line 199, in log_loss
    predictions, "multiclass_logloss", {})
  File "/Users/jong/Source/turicreate/debug/src/unity/python/turicreate/extensions.py", line 170, in <lambda>
    return lambda *args, **kwargs: _run_toolkit_function(fn, arguments, args, kwargs)
  File "/Users/jong/Source/turicreate/debug/src/unity/python/turicreate/extensions.py", line 159, in _run_toolkit_function
    raise _ToolkitError(ret[1])
turicreate.toolkits._main.ToolkitError: Size of prediction probability vector(3) != number of classes(2).
```

Even when the targets do span all _k_ classes, the evaluation toolkit assumes that the prediction vector uses a canonical ordering of classes (by sorting the unique values among the targets).

We may want to rethink the API here. For example, we could require a class index map argument, and ensure that every model exposes this.

nickjong:
The log loss, ROC curve, and AUC metrics support multiclass classification by accepting probability vectors as predictions. They infer the semantics of the probability vector by sorting all the labels found in their targets input. If the number of distinct labels is smaller than the probability vectors' length, these metrics throw an error.

With this change, these metrics support an explicit mapping from label to index into the probability vector. The activity classifier's evaluate method adopts this new API, making it robust to evaluation data that doesn't span all the training labels.

Fixes #969, fixes #1007 

Note: if we aren't comfortable committing to this new API for the evaluation toolkit (at the last minute before cutting 5.0), we could instead implement it on an interval version of these functions (and handle #1007 in the future).

(Maybe instead of an index map we should take a list of target labels, of same size as the probability vectors? Note also that we're currently limited in our ability to (conveniently) validate the index map. We don't check that the labels in the map match the labels in the targets SArray, and the backend currently ignores any labels not found in the index map.)

### DIFF
--- a/src/unity/python/turicreate/test/test_activity_classifier.py
+++ b/src/unity/python/turicreate/test/test_activity_classifier.py
@@ -340,6 +340,23 @@ class ActivityClassifierTest(unittest.TestCase):
             expected_len = self._calc_expected_predictions_length(self.data.head(100), top_k=5)
             self.assertEqual(len(preds), expected_len)
 
+    def test_evaluate_with_incomplete_targets(self):
+        """
+        Check that evaluation does not require the test data to span all labels.
+        """
+
+        # Arbitrarily filter out all rows whose label matches the first row's.
+        filtered_label = self.data[self.target][0]
+        filtered_data = self.data[self.data[self.target] != filtered_label]
+
+        # Run evaluation.
+        evaluation = self.model.evaluate(filtered_data)
+
+        # Verify that all metrics were computed and included in the result.
+        for metric in ['accuracy', 'auc', 'precision', 'recall', 'f1_score',
+                       'log_loss', 'confusion_matrix', 'roc_curve']:
+            self.assertIn(metric, evaluation)
+
     def test__list_fields(self):
         """
         Check the list fields function.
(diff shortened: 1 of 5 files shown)


srikris:
@nickjong Amazing work. This is a great fix. I think we do this under the hood automatically for other models.

nickjong:
Yup, it's worth noting that just about all the other toolkits run evaluation through C++ logic (`supervised_learning_model_base::api_evaluate`) that handles all of this by referencing stored metadata. We'll have to see to what degree this can all be unified down the road....

PR #1018, opened 2018-08-21
Fixes #995

### LINKED ISSUE #995: Python unit test failure in test_style_transfer on macOS 10.13
```
_____________________ StyleTransferTest.test_export_coreml _____________________

self = <turicreate.test.test_style_transfer.StyleTransferTest testMethod=test_export_coreml>

    def test_export_coreml(self):
        import coremltools
        filename = tempfile.mkstemp('my_style_transfer.mlmodel')[1]
        model = self.model
>       model.export_coreml(filename)

test_style_transfer.py:214: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../toolkits/style_transfer/style_transfer.py:843: in export_coreml
    mlmodel = coremltools.models.MLModel(spec)
../../coremltools/models/model.py:209: in __init__
    self.__proxy__ = _get_proxy_from_spec(filename)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

filename = '/var/folders/yb/4phfy8f539ng78zzgw90b6sh0000gn/T/tmpU60fLm.mlmodel'

    def _get_proxy_from_spec(filename):
        try:
            from ..libcoremlpython import _MLModelProxy
        except:
            _MLModelProxy = None
    
        if _MLModelProxy:
            # check if the version is supported
            engineVersion = _MLModelProxy.maximum_supported_specification_version()
            spec = _load_spec(filename)
            if spec.specificationVersion > engineVersion:
                # in this case the specification is a newer kind of .mlmodel than this version of the engine can support
                # so we'll not try to have a proxy object
                return None
            # check if there are custom layers
            if _has_custom_layer(spec):
                # custom layers can't be supported directly by compiling and loading the model here
                return None
    
>           return _MLModelProxy(filename)
E           RuntimeError: Error compiling model: "Error reading protobuf spec. validator error: The .mlmodel supplied is of version 3, intended for a newer version of Xcode. This version of Xcode supports model version 2 or earlier.".

../../coremltools/models/model.py:103: RuntimeError
```

This seems to be due to https://github.com/apple/coremltools/issues/198 -- until we can take a dependency on a newer coremltools version, we should work around this in the test so it can pass on 10.13.

shantanuchhabra:
- Avoids call to `coremltools.models.MLModel`
- Only runs flexible shape code if macOS version is >= 10.14

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,8 +235,17 @@ endif()
 # Shared compiler flags used by all builds (debug, profile, release)
 # Allow COMPILER_FLAGS to be used as options from the ./configure without
 # forcing all the other options I want to add to be lost
-set(C_REAL_COMPILER_FLAGS "-Wall -pthread -g ${COMPILER_FLAGS}" CACHE STRING "common compiler options")
-set(CPP_REAL_COMPILER_FLAGS "-Wall -pthread -g ${CPP11_FLAGS} ${COMPILER_FLAGS}" CACHE STRING "common compiler options")
+set(C_REAL_COMPILER_FLAGS "-Wall -g ${COMPILER_FLAGS}" CACHE STRING "common compiler options")
+set(CPP_REAL_COMPILER_FLAGS "-Wall -g ${CPP11_FLAGS} ${COMPILER_FLAGS}" CACHE STRING "common compiler options")
+
+if (CLANG)
+  # clang 9.0.0 and earlier will warn (error with -Werror) on unused -pthread during linking
+  # so adding only to compile options is what we need.
+  add_compile_options(-pthread)
+else()
+  set(C_REAL_COMPILER_FLAGS "${C_REAL_COMPILER_FLAGS} -pthread" CACHE STRING "common compiler options")
+  set(CPP_REAL_COMPILER_FLAGS "${CPP_REAL_COMPILER_FLAGS} -pthread" CACHE STRING "common compiler options")
+endif()
 
 # Disable address space randomization for OSX lion and above
 if (APPLE)
--- a/src/unity/python/turicreate/toolkits/style_transfer/style_transfer.py
+++ b/src/unity/python/turicreate/toolkits/style_transfer/style_transfer.py
@@ -836,13 +836,14 @@ class StyleTransfer(_CustomModel):
         coremltools.utils.rename_feature(spec,
                 'transformer__mulscalar0_output', stylized_image, True, True)
 
-        # Support flexible shape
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -235,8 +235,17 @@ endif()
 # Shared compiler flags used by all builds (debug, profile, release)
 # Allow COMPILER_FLAGS to be used as options from the ./configure without
 # forcing all the other options I want to add to be lost
-set(C_REAL_COMPILER_FLAGS "-Wall -pthread -g ${COMPILER_FLAGS}" CACHE STRING "common compiler options")
-set(CPP_REAL_COMPILER_FLAGS "-Wall -pthread -g ${CPP11_FLAGS} ${COMPILER_FLAGS}" CACHE STRING "common compiler options")
+set(C_REAL_COMPILER_FLAGS "-Wall -g ${COMPILER_FLAGS}" CACHE STRING "common compiler options")
```
srikris: Why does this change require compiler flag changes?

znation: This fixes the Xcode 9.4 build break (with `-Werror` now added) - prior to Xcode 10, clang will warn about an unused parameter on `-pthread` at link time, whereas we need it at compile time. On gcc, we need it at both compile and link time.

src/unity/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -832,30 +832,33 @@ def export_coreml(self, path, image_shape=(256, 256)):
...
                 'transformer__mulscalar0_output', stylized_image, True, True)
 
-        # Support flexible shape
-        flexible_shape_utils = _mxnet_converter._coremltools.models.neural_network.flexible_shape_utils
-        img_size_ranges = flexible_shape_utils.NeuralNetworkImageSizeRange()
-        img_size_ranges.add_height_range((64, -1))
-        img_size_ranges.add_width_range((64, -1))
-        flexible_shape_utils.update_image_size_range(spec, feature_name=self.content_feature, size_range=img_size_ranges)
-        flexible_shape_utils.update_image_size_range(spec, feature_name=stylized_image, size_range=img_size_ranges)
+        if _mac_ver() >= (10, 14):
```
znation: Let's think more about this -- I think instead of hardcoding it based on macOS version (and implicitly then Linux drops flexible shape support!), we should take an optional parameter to the export_coreml method, and behave consistently across operating systems.

PR #1022, opened 2018-08-22
Fix build break on Linux

znation:
Use size_t instead of int in `vec_mode` so that the comparison
between `histogram.size()` and the value doesn't conflict with
respect to sign. The values should always be natural numbers
(array indices) so size_t is appropriate.


znation:
Without this, on Linux we get this build error:

```
turicreate/src/unity/toolkits/activity_classification/sequence_iterator.cpp: In function 'double turi::sdk_model::activity_classification::vec_mode(const flex_vec&)':
/builds/turi/turicreate/src/unity/toolkits/activity_classification/sequence_iterator.cpp:42:29: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
         if(histogram.size() < (value + 1)){
```

PR #1052, opened 2018-08-31
C API for visualization: Round 1

znation:
In progress: C API for visualization. Should build and tests should pass as of this commit, but there is work left to do. If this looks good, please feel free to merge as-is and we can iterate on it in master.


src/capi/impl/capi_visualization.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <capi/TuriCreate.h>
+#include <capi/impl/capi_wrapper_structs.hpp>
+#include <capi/impl/capi_error_handling.hpp>
+#include <capi/impl/capi_initialization_internal.hpp>
+#include <flexible_type/flexible_type.hpp> 
+#include <export.hpp>
+
+extern "C" {
+
+    tc_plot* tc_plot_create_1d(const tc_sarray* sa,
```
hoytak: To future-proof this, I would also pass in a tc_parameters* options to allow you to pack it other options later on as needed.

test/capi/capi_utils.hpp
```diff
@@ -221,6 +221,15 @@ static tc_sframe* make_sframe_double(const std::vector<std::pair<std::string, st
     return sf;
   }
 
+/*
```
hoytak: Is this needed?

src/capi/impl/capi_visualization.cpp
znation: Not sure I am understanding the purpose of this:

* Are C API interfaces within a major release not expected/allowed to change arbitrarily? If not, I think I'd prefer to do some more up-front API design on the whole of visualization C API before proceeding.
* Should all parameters go through a single `tc_parameters*` parameter, or be passed explicitly, or a mix of both? If it's a mix of both, when should a parameter be passed explicitly vs. included in the tc_parameters struct?

test/capi/capi_utils.hpp
znation: Nope, I will remove it.

src/capi/impl/capi_visualization.cpp
hoytak: It's a mix of both -- the main parameters I like to put in explicitly, the ones not really used often I delegate to the optional parameters. 

And yes, once an API is in an internal release, then it has to always be valid. We had some symbol naming tricks to redirect things, so it's not totally set in stone, but it's a pain to deprecate old things.

znation: Got it, thanks. I'll add `tc_parameters *` to all new functions in this PR to future-proof, and consider offloading some less-used parameters there.

znation:
@hoytak Comments addressed; please re-review, thanks!

src/capi/TuriCreate.h
```diff
@@ -874,26 +874,27 @@ tc_variant* tc_function_call(
 
 // Single SArray view (`.show` on an SArray)
 tc_plot* tc_plot_create_1d(const tc_sarray* sa,
-                           const char *title,
-                           const char *x_axis_title,
-                           const char *y_axis_title,
+                           const char* title,
+                           const char* x_axis_title,
+                           const char* y_axis_title,
+                           tc_parameters* params,
```
hoytak: Let's make this `const tc_parameters*`, then good to go.

znation: Thanks, will do.

PR #1074, opened 2018-09-07
mlmodel_image_feature_extractor should delete and replace cached models that cannot be loaded successfully

### LINKED ISSUE #1065: Image deep feature extractor vulnerable to invalid model cache
I finally reproduced the error we've been seeing in some CI tests:
```
RuntimeError: CoreML Error: Unable to extract model type from stream in compiled model: Error opening file stream: /var/folders/y5/9ghgwpqs77s9k8s7s881ng_m0000gn/T/model_cache/squeezenet_v1.1_modified.mlmodelc/coremldata.bin: unspecified iostream_category error
```
Inspection of this file reveals that it doesn't exist: there's an empty `squeezenet_v1.1_modified.mlmodelc` directory in the model cache. (Well, the directory contains some empty subdirectories.)

I don't know whether this invalid compiled model is the result of ordinary cleanup within a temp directory (returned by `NSTemporaryDirectory`) or something like a crash while writing the model in a previous run.

Regardless, we should make this code path robust to model cache corruption. For example, if the cached model cannot be loaded, follow the no-cache path (and regenerate the model).

nickjong:
Otherwise, a corrupted model cache will leave the machine in a state where feature extraction will repeatedly fail, across process launches, until the corrupted files are removed from the temporary directory

Fixes #1065

### DIFF
--- a/src/toolkits/image_deep_feature_extractor/mlmodel_image_feature_extractor.mm
+++ b/src/toolkits/image_deep_feature_extractor/mlmodel_image_feature_extractor.mm
@@ -175,7 +175,7 @@ static MLModel *create_model(const std::string& download_path,
     NSError* error = nil;
     result = [MLModel modelWithContentsOfURL:compiledModelURL error:&error];
 
-    if (error) {
+    if (error || !result) {
 
       // The compiled model appears to be corrupted. Attempt to delete it.
       if (!fileio::delete_path_recursive(compiled_modified_model_path)) {
--- a/src/toolkits/image_deep_feature_extractor/mlmodel_image_feature_extractor.mm
+++ b/src/toolkits/image_deep_feature_extractor/mlmodel_image_feature_extractor.mm
@@ -162,10 +162,33 @@ API_AVAILABLE(macos(10.13))
 static MLModel *create_model(const std::string& download_path,
 			     const std::string& model_name) {
 
+  @autoreleasepool {
+
+  MLModel* result = nil;
+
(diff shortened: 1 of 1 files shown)


src/toolkits/image_deep_feature_extractor/mlmodel_image_feature_extractor.mm
```diff
@@ -162,10 +162,33 @@ void build_vision_feature_print_screen_spec(const std::string& model_path) {
...
   const std::string compiled_modified_model_path = download_path + "/" + model_name + "_modified.mlmodelc";
+  NSURL* compiledModelURL = [NSURL fileURLWithPath:[NSString stringWithUTF8String:compiled_modified_model_path.c_str()]];
+
+  // If the compiled modified model already exists on disk, attempt to load it.
+  if (boost::filesystem::exists(compiled_modified_model_path)) {
+
+    NSError* error = nil;
+    result = [MLModel modelWithContentsOfURL:compiledModelURL error:&error];
+
+    if (error) {
```
TobyRoseman: Do you want to check `result` here too?

nickjong: Good call! Once upon a time, Cocoa APIs had a standard behavior of returning nil for exactly one of the return value and the NSError out parameter, but I'm not sure this is really an invariant anymore. We may as well define success here as _both_ `error` being nil and `result` being non-nil.

PR #1091, opened 2018-09-21
SArray type-casting bug fix 

### LINKED ISSUE #1070: Unintuitive Behaviour in SArray dtype specification
The following creates an SFrame with type `array` in its singular column:
```
x = tc.SFrame({'a': [[1,2], None, [3,4], None]})
```
In an attempt to change the column type, using the following code throws no error, and does not succeed:
```
x['a'] = tc.SArray(x['a'], list)
```
This behaviour is unintuitive.

fareeha:
Fixes #1070

### DIFF
--- a/src/unity/python/turicreate/data_structures/sarray.py
+++ b/src/unity/python/turicreate/data_structures/sarray.py
@@ -349,8 +349,8 @@ class SArray(object):
 
         if (_proxy):
             self.__proxy__ = _proxy
-        elif type(data) == SArray:
-            self.__proxy__ = data.__proxy__
+        elif isinstance(data, SArray):
+            self.__proxy__ = data.astype(dtype).__proxy__
         else:
             self.__proxy__ = UnitySArrayProxy()
             # we need to perform type inference
--- a/src/unity/python/turicreate/data_structures/sarray.py
+++ b/src/unity/python/turicreate/data_structures/sarray.py
@@ -350,7 +350,10 @@ class SArray(object):
         if (_proxy):
             self.__proxy__ = _proxy
         elif isinstance(data, SArray):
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -349,8 +349,8 @@ def __init__(self, data=[], dtype=None, ignore_cast_failure=False, _proxy=None):
 
         if (_proxy):
             self.__proxy__ = _proxy
-        elif type(data) == SArray:
-            self.__proxy__ = data.__proxy__
+        elif isinstance(data, SArray):
+            self.__proxy__ = data.astype(dtype).__proxy__
```
hoytak: We should test if dtype is None or not.  If it is None, we should just do the old assignment, but otherwise we should call `astype`.

znation: If `astype(None)` is a no-op, then why not go with the simpler (no conditional) logic?

hoytak: It's not: 

```
In [2]: import turicreate as tc

In [3]: x = tc.SArray([1,2,3,4])

In [4]: x
Out[4]:
dtype: int
Rows: 4
[1, 2, 3, 4]

In [5]: y = x.astype(None)
Exception AttributeError: "'NoneType' object has no attribute '__name__'" in 'turicreate.cython.cy_flexible_type._secondary_get_translation_code' ignored
```

Thus we should add the above logic to this code, then add a test to this PR to make sure it behaves correctly when dtype is not supplied.

znation: Oh, I see, good call.

Issue #1072, opened 2018-09-07
The data argument to add_column feels overly restrictive


drewfrank:
Currently, `SFrame.add_column` requires that the `data` argument be an `SArray` with the same length as those already existing. It would nice if it would:

  * Automatically convert from non-SArray iterables.
  * Fill the entire SArray with a constant if a non-iterable value is passed.

Direct assignment to a column using the index operator (`sf['a'] = 3`) supports these already, so I found it surprising that `add_column` does not.

PR #1099, opened 2018-09-24
SFrame add_column FR

fareeha:
For feature request #1072 

add_column() in SFrame now handles non-SArray iterables as input and maps constant into SArray when noniterable passed.

### DIFF
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -3144,8 +3144,18 @@ class SFrame(object):
         [3 rows x 3 columns]
         """
         # Check type for pandas dataframe or SArray?
+        
+        from collections import Iterable
+        
         if not isinstance(data, SArray):
-            raise TypeError("Must give column as SArray")
+            if isinstance(data, Iterable):
+                data = SArray(data)
+            else:
+                if len(self.column_names()) == 0:
+                    data = SArray([data])
+                else:
+                    data = SArray([data for _ in xrange(self.shape[0])])
+    
         if not isinstance(column_name, str):
             raise TypeError("Invalid column name: must be str")
 
--- a/src/unity/python/turicreate/test/test_sframe.py
+++ b/src/unity/python/turicreate/test/test_sframe.py
@@ -3444,6 +3444,20 @@ class SFrameTest(unittest.TestCase):
                 {'long': [12], 'unicode': [unicode('foo')]}))
 
         _assert_sframe_equal(sf, sf2)
+    
+    def test_add_column_nonSArray(self):
+        sf = SFrame()
+        sf.add_column([1,2,3,4],'x')
+    
+    def test_add_column_noniterable1(self):
+        sf = SFrame()
+        sf.add_column([1,2,3,4],'x')
+        sf.add_column(5,'y')
+
+    def test_add_column_noniterable2(self):
+            # If SFrame is empty then the passed data should be treated as an SArray of size 1
+            sf = SFrame()
+            sf.add_column(5,'y')
 
 
 if __name__ == "__main__":


src/unity/python/turicreate/test/test_sframe.py
```diff
@@ -3444,6 +3444,20 @@ def test_builtins(self):
                 {'long': [12], 'unicode': [unicode('foo')]}))
 
         _assert_sframe_equal(sf, sf2)
+    
+    def test_add_column_nonSArray(self):
+        sf = SFrame()
+        sf.add_column([1,2,3,4],'x')
```
znation: Please add an assertion below, to make sure the behavior is correct and doesn't just fail silently. Assign the result to a variable, and compare that variable to an expected (constant) value using `_assert_sframe_equal`. And do this for each of the new tests.

znation:
Note that if you put `Fixes #1072` in your commit message or PR description, GitHub will automatically close that issue for you when the PR goes into master: https://blog.github.com/2013-01-22-closing-issues-via-commit-messages/

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3144,8 +3144,18 @@ def add_column(self, data, column_name="", inplace=False):
...
+        
         if not isinstance(data, SArray):
-            raise TypeError("Must give column as SArray")
+            if isinstance(data, Iterable):
+                data = SArray(data)
+            else:
+                if len(self.column_names()) == 0:
+                    data = SArray([data])
+                else:
+                    data = SArray([data for _ in xrange(self.shape[0])])
```
TobyRoseman: `xrange` has been removed as a keyword in Python 3. So this code will not work in Python 3. Writing code which works in both Python 2 and Python 3 can be a pain.

I would recommend just doing: `data = SArray([data] * self.shape[0])`

fareeha:
Thank you for the feedback Zach and Toby! I’ll work on that.

I’m having doubts about whether test_add_column_noniterable2(self) which I added should exist. I noticed that passing a noniterable to an empty SFrame does not store the element as a size one SArray: 

>>> import turicreate as tc
>>> sf = tc.SFrame()
>>> sf['x']=3
>>> sf
Columns:
	x	int

Rows: 0

Data:
	[]

Whereas test_add_column_noniterable2 checks if an noniterable is passed to an empty SFrame it is stored as a size 1 SArray.

znation:
@fareeha Hmm, that actually looks like a bug (`sf['x']=3`). I am not sure what it should do there, but it definitely shouldn't give you an empty int column. I think there are two things we could do here (not sure which I prefer):

1. Raise an exception when attempting to assign a non-iterable to an SFrame column, or when attempting to do the equivalent through `add_column`.
2. Consistently treat non-iterables as a list of length 1 in both of the above cases.

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3144,8 +3144,18 @@ def add_column(self, data, column_name="", inplace=False):
...
         # Check type for pandas dataframe or SArray?
+        
+        from collections import Iterable
+        
         if not isinstance(data, SArray):
-            raise TypeError("Must give column as SArray")
+            if isinstance(data, Iterable):
+                data = SArray(data)
+            else:
+                if len(self.column_names()) == 0:
```
hoytak: Use self.num_columns() instead here.

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3144,8 +3144,18 @@ def add_column(self, data, column_name="", inplace=False):
         [3 rows x 3 columns]
         """
         # Check type for pandas dataframe or SArray?
+        
+        from collections import Iterable
```
hoytak: For these, we should have as many imports at the module level instead of inside a function; here, it can slow it down quite a bit.  Let's put `from collections import Iterable as _Iterable` at the top and then use _Iterable below.

src/unity/python/turicreate/data_structures/sframe.py
hoytak: Use SArray.from_const here.  We want to do as much stuff not in memory as possible.

hoytak: You can also use self.num_rows() here instead of self.shape[0], but shape[0] is fine.

fareeha:
@hoyt Thank you! Those are very good points. The FR is closed but I’m checking out a branch to address the changes you’ve suggested.

PR #1103, opened 2018-09-25
1072b

fareeha:
Changes made:

* Replaced "len(self.column_names())" with "self.num_columns()"
* Placed "from collections import Iterable as _Iterable" at the top
* Replaced "self.shape[0]" with "self.num_rows()"
* Replaced "SArray([data] * self.shape[0])" with "SArray.from_const(data, self.num_rows())"

### DIFF
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -3155,7 +3155,6 @@ class SFrame(object):
                 if len(self.column_names()) == 0:
                     data = SArray([data])
                 else:
-                    #data = SArray([data] * self.shape[0])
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -3145,16 +3146,17 @@ def add_column(self, data, column_name="", inplace=False):
...
         
-        from collections import Iterable
+        
         
         if not isinstance(data, SArray):
-            if isinstance(data, Iterable):
+            if isinstance(data, _Iterable):
                 data = SArray(data)
             else:
                 if len(self.column_names()) == 0:
```
hoytak: Do we want to change this one as well?

Issue #1096, opened 2018-09-24
Tuples get parsed as arrays which are unhashable


shantanuchhabra:
Right now, even if I have an SArray of dictionaries with tuple keys, displaying it doesn't work because tuples automatically get interpreted as arrays. This is calamitous because arrays are not a hashable type. Furthermore, if I try to save/create/explore this SArray/SFrame, I don't get any errors but the data structure is pretty unusable. 

I think the true solution to this would be to support tuple types in Turi Create. 

Minimal Repro:

```
d = {(0,0):"lol"}  
arr = [d]
sa = tc.SArray(arr) # works
sa.save("no_error.sarray") # works
sa.explore() # works
print(sa[0]) # uh oh
print(sa) # uh oh
```

hoytak:
Tuple types are supported; the C++ types don't match up one-to-one to the python types, so the main solution to this seems to be to have the translation from C++ into python depend somewhat on the context.

hoytak:
Fixed in https://github.com/apple/turicreate/pull/1106

PR #1106, opened 2018-09-28
Fix for #1096

hoytak:
Casts lists and arrays to tuples when they are the keys for a dictionary type.

### DIFF
--- a/src/unity/python/turicreate/test/test_flexible_type.py
+++ b/src/unity/python/turicreate/test/test_flexible_type.py
@@ -560,7 +560,7 @@ class FlexibleTypeTest(unittest.TestCase):
             sa_int_words.astype(int)
 
     def test_hashable_dict_keys(self):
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/test/test_flexible_type.py
```diff
@@ -559,6 +559,13 @@ def test_type_conversions(self):
         with self.assertRaises(RuntimeError):
             sa_int_words.astype(int)
 
+    def test_hashable_dict_keys(self):
+        # testing valid sarray of inf's (inf is a float)
```
TobyRoseman: What is `inf`?

hoytak: Inf and -inf are specific floating point values such that -Inf < x < Inf for other valid floating point values of x.

hoytak: However, this comment is wrong.

PR #1113, opened 2018-10-01
Const assignment to empty SFrame 

### LINKED ISSUE #1102: Assignment to SFrame column with a scalar value differs in behavior from `add_column`
Repro steps:

```python
import turicreate as tc
sf = tc.SFrame()
sf = sf.add_column(5, 'x') # note that sf is len(1)

sf2 = tc.SFrame()
sf2['x'] = 5 # note that sf2 is len(0)
```

Expected:
`sf` and `sf2` are equivalent

Actual:
`sf2` contains no data (and the value `5` was dropped).

fareeha:
Reconciles empty SFrame assignment fix #1102 with DBscan test failure on Travis.

### DIFF
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -3146,7 +3146,7 @@ class SFrame(object):
         """
         # Check type for pandas dataframe or SArray?
         
-       
+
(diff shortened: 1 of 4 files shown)


fareeha:
As per offline discussion, the original behavior will be kept. Closing this PR.

PR #1116, opened 2018-10-02
Feature Request for filter_by() in SArray

### LINKED ISSUE #1061: SArray should have filter_by function
This would work just like the SFrame `filter_by` but would not take a `column_name` parameter.

fareeha:
Closes #1061

### DIFF
--- a/src/unity/python/turicreate/data_structures/sarray.py
+++ b/src/unity/python/turicreate/data_structures/sarray.py
@@ -4399,6 +4399,92 @@ class SArray(object):
         agg_op = "__builtin__cum_var__"
         return SArray(_proxy = self.__proxy__.builtin_cumulative_aggregate(agg_op))
 
+    def filter_by(self, values, exclude=False):
+        
+        """
+        Filter an SArray by values inside an iterable object. Result is an SArray that
+        only includes (or excludes) the values in the given ``values`` :class:`~turicreate.SArray`.
+        If ``values`` is not an SArray, we attempt to convert it to one before filtering.
+        
+        Parameters
+        ----------
+        values : SArray | list | numpy.ndarray | pandas.Series | str
+        The values to use to filter the SArray. The resulting SArray will
+        only include rows that have one of these values in the given
+        column.
+        
+        exclude : bool
+        If True, the result SArray will contain all rows EXCEPT those that
+        have one of the ``values``.
+        
+        Returns
+        -------
+        out : SArray
+        The filtered SArray.
+        
+        Examples
+        --------
+        >>> sa = SArray(['dog', 'cat', 'cow', 'horse'])
+        >>> sa.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'])
+        dtype: str
+        Rows: 2
+        ['dog', 'cat']
+        
+        >>> sa.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'], exclude=True)
+        dtype: str
+        Rows: 2
+        ['horse', 'cow']
+        """
+    
+        from .sframe import SFrame as _SFrame
+        
+        column_name = 'sarray'
+        
+        #convert values to SArray
+        if type(values) is not SArray:
+            # If we were given a single element, try to put in list and convert
+            # to SArray
+            if not _is_non_string_iterable(values):
+                values = [values]
+            values = SArray(values)
+    
+        #convert values to SFrame
+        value_sf = _SFrame()
+        value_sf.add_column(values, column_name, inplace=True)
+        given_type = value_sf.column_types()[0] #value column type
+        
+        existing_type = self.dtype
+        sarray_sf = _SFrame()
+        sarray_sf.add_column(self, column_name, inplace=True)
+        
+        if given_type != existing_type:
+            raise TypeError("Type of given values does not match type of the SArray column")
+        
+        # Make sure the values list has unique values, or else join will not
+        # filter.
+        value_sf = value_sf.groupby(column_name, {})
+        print(value_sf)
+        
+        with cython_context():
+            if exclude:
+                id_name = "id"
+                # Make sure this name is unique so we know what to remove in
+                # the result
+                id_name += "1"
+                
+                value_sf = value_sf.add_row_number(id_name)
+                tmp = _SFrame(_proxy=sarray_sf.__proxy__.join(value_sf.__proxy__,
+                                                        'left',
+                                                        {column_name:column_name}))
+                ret_sf = tmp[tmp[id_name] == None]
+                del ret_sf[id_name]
+                return ret_sf[column_name]
+            else:
+                ret_sf = _SFrame(_proxy=sarray_sf.__proxy__.join(value_sf.__proxy__,
+                                                             'inner',
+                                                             {column_name:column_name}))
+                return ret_sf[column_name]
+
     def __copy__(self):
         """
         Returns a shallow copy of the sarray.
--- a/src/unity/python/turicreate/data_structures/sarray.py
+++ b/src/unity/python/turicreate/data_structures/sarray.py
@@ -4402,7 +4402,7 @@ class SArray(object):
     def filter_by(self, values, exclude=False):
         
         """
-        Filter an SArray by values inside an iterable object. Result is an SArray that
+        Filter an SArray by values inside an iterable object. The result is an SArray that
         only includes (or excludes) the values in the given ``values`` :class:`~turicreate.SArray`.
         If ``values`` is not an SArray, we attempt to convert it to one before filtering.
         
@@ -4440,15 +4440,15 @@ class SArray(object):
         
         column_name = 'sarray'
         
-        #convert values to SArray
-        if type(values) is not SArray:
+        # Convert values to SArray
+        if not isinstance(values, SArray): #type(values) is not SArray:
             # If we were given a single element, try to put in list and convert
             # to SArray
             if not _is_non_string_iterable(values):
                 values = [values]
             values = SArray(values)
     
-        #convert values to SFrame
+        # Convert values to SFrame
         value_sf = _SFrame()
         value_sf.add_column(values, column_name, inplace=True)
         given_type = value_sf.column_types()[0] #value column type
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
...
+        Rows: 2
+        ['horse', 'cow']
+        """
+    
+        from .sframe import SFrame as _SFrame
+        
+        column_name = 'sarray'
+        
+        #convert values to SArray
+        if type(values) is not SArray:
```
hoytak: I think this should be isinstance(values, SArray):

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
...
+        Rows: 2
+        ['dog', 'cat']
+        
+        >>> sa.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'], exclude=True)
+        dtype: str
+        Rows: 2
+        ['horse', 'cow']
+        """
+    
+        from .sframe import SFrame as _SFrame
```
hoytak: Just a style thing -- if this is a local variable, then you don't need the underscore on the _SFrame.  That is just needed at the global level to keep from polluting the module namespace -- making something start with an underscore means it's private to a module or class, so it doesn't show up with tab completion or in our documentation.

hoytak: Actually, probably better to move the import to the global scope and leave the _ on it.

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
...
+            raise TypeError("Type of given values does not match type of the SArray column")
+        
+        # Make sure the values list has unique values, or else join will not
+        # filter.
+        value_sf = value_sf.groupby(column_name, {})
+        print(value_sf)
+        
+        with cython_context():
+            if exclude:
+                id_name = "id"
```
TobyRoseman: You don't need to worry about the row number column having a name collision because you just created `value_sf`; you know it doesn't have a column named `id` which is the default column name value for `add_row_number`.

Remove `id_name` variable altogether to make the code simpler.

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
...
+                # Make sure this name is unique so we know what to remove in
+                # the result
+                id_name += "1"
+                
+                value_sf = value_sf.add_row_number(id_name)
+                tmp = _SFrame(_proxy=sarray_sf.__proxy__.join(value_sf.__proxy__,
+                                                        'left',
+                                                        {column_name:column_name}))
+                ret_sf = tmp[tmp[id_name] == None]
+                del ret_sf[id_name]
```
TobyRoseman: don't bother deleting column from the sframe cause you're just returning a single column from the sframe not the sframe itself.

src/unity/python/turicreate/test/test_sarray.py
```diff
@@ -3164,4 +3164,27 @@ def test_type_casting(self):
...
+        self.assertTrue(np.array_equal(x.filter_by([2,2,3,44]),SArray([2,3])))
+        x = SArray([1,2,2,3,4,5,6,7])
+        self.assertTrue(np.array_equal(x.filter_by([2,2,3,44]), SArray([2,2,3])))
+    
+        #strings
+        x = SArray(['dog', 'cat', 'cow', 'horse'])
+        self.assertTrue(np.array_equal(x.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake']), SArray(['dog', 'cat'])))
+        self.assertTrue(np.array_equal(x.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'], exclude=True), SArray(['horse', 'cow'])))
+    
+
```
TobyRoseman: Could you add a test case where you pass a single str value into the filter_by. Something like:
`x.filter_by('dog')`

I'm a little worried about this working properly in Python 3. I'm not sure you want to use `_is_non_string_iterable(...)` to test if it's not an iterable.

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
         agg_op = "__builtin__cum_var__"
         return SArray(_proxy = self.__proxy__.builtin_cumulative_aggregate(agg_op))
 
+    def filter_by(self, values, exclude=False):
+        
+        """
+        Filter an SArray by values inside an iterable object. Result is an SArray that
+        only includes (or excludes) the values in the given ``values`` :class:`~turicreate.SArray`.
```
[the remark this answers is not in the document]
fareeha: The doc string I used here was the same as the one in the original filter_by() in the SFrame. I switched out the SFrame column for SArray wherever applicable since we're using the same logic/code to create this function. That's why I missed removing ":class:~turicreate.SArray", as you mentioned it makes sense to not have it here. Ty for the other feedback too, I'll make those edits as well.

src/unity/python/turicreate/test/test_sarray.py
```diff
@@ -3164,4 +3164,27 @@ def test_type_casting(self):
...
+        self.assertTrue(np.array_equal(x.filter_by([11,7,2,8,4]), SArray([2,4,7])))
+        self.assertTrue(np.array_equal(x.filter_by([11,7,2,8,4,3], exclude=True), SArray([1,5,6])))
+    
+        #empty SArray
+        self.assertTrue(np.array_equal(x.filter_by([77,22,18,42]), SArray([])))
+        self.assertTrue(np.array_equal(x.filter_by([77,22,18,42], exclude=True).sort(), x))
+        
+        #duplicates
+        self.assertTrue(np.array_equal(x.filter_by([2,2,3,44]),SArray([2,3])))
+        x = SArray([1,2,2,3,4,5,6,7])
```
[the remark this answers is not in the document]
fareeha: Do you mean the test case? Because this x contains a duplicate value which the previous x does not.

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
...
+        sarray_sf = _SFrame()
+        sarray_sf.add_column(self, column_name, inplace=True)
+        
+        if given_type != existing_type:
+            raise TypeError("Type of given values does not match type of the SArray column")
+        
+        # Make sure the values list has unique values, or else join will not
+        # filter.
+        value_sf = value_sf.groupby(column_name, {})
+        print(value_sf)
```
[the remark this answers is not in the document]
fareeha: Good catch! Ty

src/unity/python/turicreate/test/test_sarray.py
hoytak: I think that function is tested in python 3 -- it's used in the SFrame filter_by --  so it should be fine.  However, yes, we should have more test cases.

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
         agg_op = "__builtin__cum_var__"
         return SArray(_proxy = self.__proxy__.builtin_cumulative_aggregate(agg_op))
 
+    def filter_by(self, values, exclude=False):
+        
+        """
+        Filter an SArray by values inside an iterable object. Result is an SArray that
```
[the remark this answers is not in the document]
fareeha: This was in the original doc string, which I believe refers to "values" passed in filter_by. Would you like me to change it?

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
...
+        
+        #convert values to SArray
+        if type(values) is not SArray:
+            # If we were given a single element, try to put in list and convert
+            # to SArray
+            if not _is_non_string_iterable(values):
+                values = [values]
+            values = SArray(values)
+    
+        #convert values to SFrame
```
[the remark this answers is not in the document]
fareeha: Thank you for the guide! I'll make sure to follow the conventions in the future.

src/unity/python/turicreate/data_structures/sarray.py
```diff
@@ -4399,6 +4399,92 @@ def cumulative_var(self):
...
+        value_sf = _SFrame()
+        value_sf.add_column(values, column_name, inplace=True)
+        given_type = value_sf.column_types()[0] #value column type
+        
+        existing_type = self.dtype
+        sarray_sf = _SFrame()
+        sarray_sf.add_column(self, column_name, inplace=True)
+        
+        if given_type != existing_type:
+            raise TypeError("Type of given values does not match type of the SArray column")
```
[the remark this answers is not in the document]
fareeha: Done.

src/unity/python/turicreate/data_structures/sarray.py
fareeha: That's certainly better. Thanks! and done.

src/unity/python/turicreate/data_structures/sarray.py
fareeha: Retained the change, as per offline discussion.

src/unity/python/turicreate/data_structures/sarray.py
fareeha: Done.

src/unity/python/turicreate/test/test_sarray.py
fareeha: Added a single str test case, and it passes. If you have any other suggestions for test cases please do let me know. So far I have cases for integer, strings, empty return SArrays and duplicates in "values" and the caller SArray.

src/unity/python/turicreate/data_structures/sarray.py
fareeha: The way I understood left join here is that it returns an SFrame with ids mapped to None for values not in the caller SFrame column.  

tmp: +--------+------+
| sarray |  id  |
+--------+------+
|   2    |  0   |
|   3    |  1   |
|   4    |  4   |
|   7    |  5   |
|   1    | None |
|   5    | None |
|   6    | None |
+--------+------+

Without the row number we don't get any information back to retrieve from:

tmp: +--------+
| sarray |
+--------+
|   1    |
|   2    |
|   3    |
|   4    |
|   5    |
|   6    |
|   7    |
+--------+
[7 rows x 1 columns]

Please correct me if I'm missing something.

TobyRoseman: Right, you still need to call `row_number()`. However you don't need to pass `row_number()` a parameter because you're fine with it using the default value for the new column name that it creates; you know that default value will not be the same as any column name already in the SFrame.

If you want to still pass in a name, that's fine. But please remove the comment about making sure it's unique and remove the `id_name += "1"` line.

fareeha: Thank you for the feedback. I've removed the comment and the id_name += "1" line.

Issue #1031, opened 2018-08-23
Travis CI is not running style transfer unit tests


TobyRoseman:
Running the style transfer unit test in Travis CI causes a core dump. The likely cause is that it's running out of memory. Travis containers only have 4 GB of RAM which is the minimum allowable for style transfer.

znation:
4 GB should be sufficient for style transfer; we should not crash in this case. Relabeling as a bug.

TobyRoseman:
Just to be totally clear the [travis documentation](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) claims that Ubuntu Trusty containers have 4GB of memory. I have not verified that's what we are getting.

I think the deeper issue here is that we just don't know if it has too little memory or not. Currently the Style Transfer Toolkit does not verify the system has the minimum amount of memory. I created #1040 to track that issue.

TobyRoseman:
As of #2063 test_one_shot_object_detector.py has also been removed from Travis CI.

PR #1118, opened 2018-10-03
Skip object detector unit tests in Travis CI for now

nickjong:
We need to get our Travis CI build under two hours more reliably. We can skip the OD tests for now, since I'll likely be running then manually a bunch anyway in the near future (aside from any other CI coverage).

Builds on the black list we already added to cover issues with style transfer (see #1031)


nickjong:
If this doesn't suffice, next up are probably the recommender tests and the image similarity tests, just in terms of shaving the most minutes off with one test suite. Meanwhile, we need to start investing in a more sustainable test design strategy.

TobyRoseman:
Good idea @nickjong. When this is merged please update #1031.

Making the image classification and image similarity test run much faster is on my list of things to do. Hoping to have time to get to that next week.

Issue #755, opened 2018-06-25
Object-oriented interface for MPS models


nickjong:
The original interface between Python and the Objective-C++ backend simply passed raw float pointers to the backend to populate the output. The first attempt at an asynchronous interface treated the backend as a queue of compute requests and a queue of results, the latter of which can be read synchronously by passing a raw float pointer. This API has become quite brittle, as the Python side must carefully match the nth result retrieved against the nth request submitted.

Instead, the backend can provide an API more similar to the MXNet approach to asynchronous dispatch. The compute request can return a handle or wrapper around the result MPSImage, and an attempt to read that image (into a provided float array) will trigger a wait for the compute to finish on the GPU.

PR #1121, opened 2018-10-03
Refactor TCMPS interface for object detection

nickjong:
Before these changes, clients of the TCMPS backend (for graph-API-based models, such as object detection) had to reason about the backend as having a consumer queue of inputs and a producer queue of outputs, leaving it to the client to track with output corresponds to which input.

With these changes, submitting an input to the backend immediately returns an object that will hold the future output.

I originally wanted to get the low-level-API side too, for activity classification, but I wanted to focus on OD work for now since I'm behind schedule. (I'd rather get one model completely converted sooner and the other later, than have both finish simultaneously later.) For that reason, this addresses only half of #755 (although it includes the work common to both cases).

### DIFF
--- a/src/unity/python/turicreate/toolkits/_mps_utils.py
+++ b/src/unity/python/turicreate/toolkits/_mps_utils.py
@@ -243,6 +243,33 @@ def _xavier_init(weight):
     c = _np.sqrt(3. / (0.5 * (n_in * scale + n_out * scale)))
     return _np.random.uniform(-c, c, shape).astype(_np.float32)
 
+def _shape_tuple_from_ctypes(shape_ptr, dim):
+    # size_t* shape_ptr
+    assert isinstance(shape_ptr, _ctypes.POINTER(_ctypes.c_size_t))
+
+    # size_t dim
+    assert isinstance(dim, _ctypes.c_size_t)
+
+    # Wrap size_t* as size_t[dim]
+    shape_buf = (_ctypes.c_size_t * dim.value).from_address(
+        _ctypes.addressof(shape_ptr.contents))
+
+    # Convert size_t[dim] to tuple
+    return tuple(shape_buf)
+
+def _numpy_array_from_ctypes(data_ptr, shape_ptr, dim):
+    # float* data_ptr
+    assert isinstance(data_ptr, _ctypes.POINTER(_ctypes.c_float))
+
+    shape = _shape_tuple_from_ctypes(shape_ptr, dim)
+
+    # Wrap float* to float[size]
+    size = _np.prod(shape)
+    data_buf = (_ctypes.c_float * size).from_address(
+        _ctypes.addressof(data_ptr.contents))
+
+    # Convert float[size] to numpy
+    return _np.fromiter(data_buf, _np.float32, size).reshape(shape)
 
 class MpsFloatArray(object):
     """
@@ -254,13 +281,18 @@ class MpsFloatArray(object):
     """
 
     def __init__(self, x):
-        """Wrap a numpy array"""
-
-        assert isinstance(x, _np.ndarray)
+        """Wrap an existing TCMPSFloatArrayRef or a numpy array"""
 
         # Load TCMPS backend library.
         self._LIB = _load_tcmps_lib()
         assert self._LIB is not None, "Cannot use MpsFloatArray without libtcmps.dylib"
+
+        # If `x` is a void*, assume it's the right type and just wrap it.
+        if isinstance(x, _ctypes.c_void_p):
+            self.handle = x
+            return
+
+        assert isinstance(x, _np.ndarray)
 
         # Convert the input if necessary to contain a contiguous float array.
         self.data = x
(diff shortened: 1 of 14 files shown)


src/unity/python/turicreate/toolkits/_mps_utils.py
```diff
@@ -418,57 +471,85 @@ def init(self, n, c_in, h_in, w_in, c_out, h_out, w_out, config=None, weights=No
...
+            self.handle, input_array.handle, grad_array.handle,
+                _ctypes.byref(result_handle))
+
+        assert status_code == 0, "Error calling TCMPSTrainReturnGradGraph"
+        assert result_handle, "TCMPSTrainReturnGradGraph unexpectedly returned NULL pointer"
+
+        result = MpsFloatArray(result_handle)
+        assert result.shape() == self._ishape
+
+        return result
```
gustavla: A few questions about `train_return_grad`. Are we using this for anything other than testing? Passing out gradients is only necessary for hybrid models (part MPS, part something else). I'm just curious if this is actually used or planned to be used anywhere. Of course, testing is very important too.

Can I also get the loss here? It seems to me that in an ideal world, there would only be `train(..., return_input_gradient=False)`, where you could flip on `return_input_gradient` to have it return a tuple of `(loss, grad)`, instead of just `loss`. I'm calling it "input" gradient here, so as not to imply it returns all gradients.

src/unity/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -543,9 +545,8 @@ def numpy_worker():
             iteration = batch['iteration']
 
         # Wait for any pending batches and finalize our progress updates.
-        while batches_finished < batches_started:
-            batches_finished += 1
-            cur_loss = mps_net.wait_for_batch().sum() / mps_loss_mult
+        while len(batch_queue) > 0:
+            cur_loss = wait_for_batch()
```
gustavla: Shouldn't `update_progress(cur_loss, iteration)` be inside this while loop? The `update_progress` function computes a smoothed value of the loss, and the way this while loop is structured, it would seem that it could miss losses from being included in the running average.

gustavla: Not sure what would happen with `iteration` in that case though. The batch queue shouldn't really ever be more than 1 here though, right? If that's the case, then why the while loop?

src/unity/toolkits/tcmps/mps_graph_cnnmodule.mm
```diff
@@ -101,40 +120,43 @@ - (instancetype)initWithCommandBuffer:(id<MTLCommandBuffer>)commandBuffer {
     [lossState synchronizeOnCommandBuffer:cb];
   }
 
+  // Schedule copying of the output into a float_array promise.
+  // n.b. The block below must not capture `this`.
+  size_t loss_size = static_cast<size_t>(batch.lossState.count);
+  auto loss_promise = std::make_shared<std::promise<shared_float_array>>();
+  NSMutableArray<MPSImageBatch *> *recycled_inputs = recycled_inputs_;
```
gustavla: Is this local copy of the ref needed in case it changes? Maybe you can say something about why this is needed.

gustavla: Or is it simply so that the closure (or *block*, I guess they are called) doesn't have to wrap the entire `this`?

src/unity/toolkits/tcmps/mps_graph_cnnmodule.mm
```diff
@@ -151,35 +173,36 @@ - (instancetype)initWithCommandBuffer:(id<MTLCommandBuffer>)commandBuffer {
     [image synchronizeOnCommandBuffer:cb];
   }
 
+  // Schedule copying of the output into a float_array promise.
+  // n.b. The block below must not capture `this`.
```
gustavla: Ah yes, this comment reaffirms what I thought. Oh, and the same comment is in Train too already, I just missed it. OK, all good here! :+1:

igiloh: > TCMPS is becoming really cool! 

👍 👍   
@nickjong for the rescue! 👏

src/unity/python/turicreate/toolkits/_mps_utils.py
nickjong: Yeah, good question. Honestly, I only included `train_return_grad` because I've been erring on the side of preserving functionality until I can take a serious pass on adding tests here (using these new interfaces). Although ideally I would love to be able to write tests that just use the "normal" interfaces for training/inference. I also want to give more thought to what we need for debugging future networks; honestly my focus right now is on adapting OD.

src/unity/python/turicreate/toolkits/object_detector/object_detector.py
nickjong: `update_progress` implicitly assumes that it's called only once per iteration, with the loss for the last batch in each iteration. In practice, `batch_queue` should have length 2 here almost all of the time, for the last two batches in the last iteration through the data (due to double buffering).

It looks like we'll do the wrong thing if the entire dataset fits into one batch, but I'm not going to worry about that now. Maybe I can clean up some of the mess I made here when we re-implement this logic in C++ for the C API version of OD.

PR #1127, opened 2018-10-05
Implementation of ndarray in ml_data.

hoytak:
This PR allows ndarray to be a valid type that is understood by all the toolkits relying on ml_data.


znation:
@hoytak with the "Final revision" commit, is this still WIP?

PR #1134, opened 2018-10-12
Fix strange JSON parsing bug

znation:
When there are exactly 128-char strings within a nested JSON
structure, they were silently dropped due to an off-by-one
comparison. This change fixes the bug, and adds a unit test to
ensure correct behavior. (Verified that this test failed prior
to this change, and passes now with this change.)


znation:
@ylow Thanks for the fix!

PR #1135, opened 2018-10-12
Create new object_detector extension to serve as basis for new C++ implementation of the OD toolkit

nickjong:
- Build libtcmps as a static library to be rolled up into libunity_shared, just like all the other backend libraries
- Define a new pure C++ interface to wrap the Objective-C++ MPSGraphModule
- Apply C++ naming_convention to touched symbols in libtcmps for consistency (let's leave Objective C names in CamelCase, to help parse Objective-C++ code)
- Create a new object_detector C++ class, registered with the C API

For now, this is just enough to instantiate a MPSGraphModule instance via
```
  model = tc.extensions.object_detector()
  model.init_options({})
```
while preserving the behavior of the Python-based toolkit

Note that some care must be taken to ensure that the new object_detector library builds on Linux, where the tcmps library will not be built.


src/unity/toolkits/tcmps/cnn_module.hpp
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Performs one forward-backward pass.
+   */
+  virtual deferred_float_array train(const float_array& input_batch,
+                                     const float_array& label_batch) = 0;
+
+  /**
+   * Performs a forward pass.
+   */
+  virtual deferred_float_array predict(const float_array& input_batch) = 0;
```
TobyRoseman: Can this be a const method too?

TobyRoseman: I really don't know what would be best here.

Could we and would it make sense to mark the member variables which get mutated as `mutable`?

nickjong: I went ahead and made this method const, since I personally like the operational definition of "does not change the behavior of future method calls".

In this case, there's no need to mark any data structures as mutable, given C++ semantics. (I don't like relying 100% on the compiler to define const, since the following compiles:
```
class foo {
public:
  int read() const { return *x_; } 
  void increment() const {  // O RLY?
    *x_ += 1;
  }
private:
  int* x = new int(0);
};
```
)

PR #1137, opened 2018-10-15
WIP: MPS Style Transfer

abhishekpratapa:
- First WIP is all the neural network layer files.


nickjong:
I'm concerned that this design duplicates much of the existing TCMPS library (note that we already have a `turi::mps` namespace defined in that directory, we should be wary of defining it across multiple directories).

I could see not using TCMPS if we needed to sidestep the approach there, which defines a C++ object as a wrapper/intermediary for each layer of the network. But so far this seems to replicate that approach, just in a different library with almost the same name.

Given the direction you're going here, what's the reason for not using TCMPS?

znation:
@nickjong I discussed this with @abhishekpratapa a while back. I think the goal was to simply get this working, with a sane architecture, and refactor later to share with TCMPS where possible. However, since you're noticing so much similarity now, it might make sense for @abhishekpratapa to rebase on top of https://github.com/apple/turicreate/pull/1135 and refactor to share the same approach? Once TCMPS is in libunity_shared, it should be more straightforward to expose C++ types through the registration mechanism and use the structured C++ API in Python (instead of the low-level ctypes stuff).

PR #1144, opened 2018-10-17
Introduce platform-agnostic unity_neural_net library to wrap tcmps

nickjong:
The primary change here is the introduction of unity_neural_net, which provides a minimal interface and builds on all platforms. Clients can safely depend on this library unconditionally, although for now it's only functional on Apple platforms, where it wraps the MPS-based backend. Eventually we will also wrap MXNet as well (adjusting the minimal interface as necessary).

Most of the changed files reflect a secondary change: renaming the `turi::mps` namespace to something more generic, `turi::neural_net`, and renaming the `temps` directory to `neural_net`.

Open to suggestions for other names besides "neural_net"


igiloh:
LGTM. As you've mentioned - it's mostly namespace changes.

src/unity/toolkits/neural_net/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+# When building for macOS, build the MetalPerformanceShaders-based backend.
+if(APPLE AND HAS_MPS AND NOT TC_BUILD_IOS)
+    find_library(METAL NAMES Metal)
+    message("Metal found at ${METAL}.")
+
+    find_library(METAL_PERFORMANCE_SHADERS NAMES MetalPerformanceShaders)
+    message("MetalPerformanceShaders found at ${METAL_PERFORMANCE_SHADERS}.")
+
+    make_library(tcmps
```
znation: Should this be renamed `neural_net`?

znation: Nevermind, looks like we are intentionally keeping a `neural_net` (platform-agnostic) and `tcmps` (macOS-specific) split within this directory.

PR #1157, opened 2018-10-24
Adopt CoreML/MXNet data layout (NCHW) at library interfaces (for object detection)

nickjong:
We can push the MXNet-to-MPS-to-MXNet conversions into TCMPS, where this particular transpose can be implemented as a conversion between planar and "chunky" float images. In this way, we can standardize on the CoreML layouts where possible, for consistency, leaving our MPS usage of NHWC as an implementation detail.

### DIFF
--- a/src/unity/python/turicreate/toolkits/_mps_utils.py
+++ b/src/unity/python/turicreate/toolkits/_mps_utils.py
@@ -150,6 +150,19 @@ def _prepare_network_parameters(arg_dict):
         arr[i] = _ctypes.c_void_p(items[i][1].ctypes.data)
     return items, name, arr, sz
 
+def _prepare_graph_network_parameters(arg_dict):
+    items = []
+    for name, arr in arg_dict.items():
+        if not isinstance(arr, _np.ndarray):
+            arr = _np.array(arr, dtype=_np.float32)
+        items.append((name, MpsFloatArray(arr)))
+
+    name = (_ctypes.c_char_p * len(items))()
+    arr = (_ctypes.c_void_p * len(items))()
+    for i in range(len(items)):
+        name[i] = _ctypes.c_char_p(items[i][0].encode())
+        arr[i] = items[i][1].handle
+    return items, name, arr
 
 _g_TCMPS_LIB = None
 
@@ -450,8 +463,8 @@ class MpsGraphAPI(object):
         self._mode = int(config.get('mode', MpsGraphMode.TrainReturnGrad))
         self._is_train = self._mode in {MpsGraphMode.TrainReturnGrad, MpsGraphMode.Train}
 
-        config_items, config_name, config_arr, config_sz = _prepare_network_parameters(config)
-        weights_items, weights_name, weights_arr, weights_sz = _prepare_network_parameters(weights)
+        config_items, config_name, config_arr = _prepare_graph_network_parameters(config)
+        weights_items, weights_name, weights_arr = _prepare_graph_network_parameters(weights)
         self._LIB.TCMPSInitGraph(
             self.handle,
             self.network_id,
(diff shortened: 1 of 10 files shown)


igiloh:
LGTM - but I would wait for @gustavla's review. He knows this code best.

igiloh:
On a complete side note - when you do end up doing this for the AC as well - note that the AC doesn't actually need 4 dimension shapes. One of the dimension (Height by my convention) is always `1`.  
If no other toolkits are using the `LowLevelApi` (none should) - you can probably skip these conversions altogether, and stick with a single 3-dim shape. 

As you might recall - the MXnet implementation has some [redundant transposes and reshapes](https://github.com/apple/turicreate/blob/master/src/unity/python/turicreate/toolkits/activity_classifier/_mx_model_architecture.py#L25), which further complicate the weights layout.  
These operations can be removed Using MXnet>0.12 APIs, and then it should be quite easy to find a simple, common layout for both MXnet and MPS.

nickjong:
Yeah, after getting further along in downstream work, I realized that MXNet data augmentation is already working with images in HWC, which is also what MPS wants. So for now, I'm going to leave the batch inputs and outputs as NHWC as this interface boundary, but standardize the model parameters (e.g. for Conv weights) to the CoreML format, since that's persistent.

So this last commit reverts almost half of the original PR....

src/unity/python/turicreate/toolkits/_mps_utils.py
```diff
@@ -150,6 +143,19 @@ def _prepare_network_parameters(arg_dict):
         arr[i] = _ctypes.c_void_p(items[i][1].ctypes.data)
     return items, name, arr, sz
 
+def _prepare_graph_network_parameters(arg_dict):
+    items = []
+    for name, arr in arg_dict.items():
+        if not isinstance(arr, _np.ndarray):
+            arr = _np.array(arr, dtype=_np.float32)
```
gustavla: I think `np.asarray` is more common for this use (this is what numpy uses itself). For instance, if you put in a `np.matrix`, the `asarray` will do a very cheap translation without copying the underlying memory while `np.array` will iterate the matrix and copy manually.

src/unity/toolkits/neural_net/mps_utils.mm
```diff
@@ -11,9 +11,126 @@
...
+    // Use Accelerate framework, interpreting each channel as an image plane.
+
+    const unsigned c = static_cast<unsigned>(image.shape()[0]);
+    const vImagePixelCount h = static_cast<vImagePixelCount>(image.shape()[1]);
+    const vImagePixelCount w = static_cast<vImagePixelCount>(image.shape()[2]);
+
+    std::vector<vImage_Buffer> image_planes(c);
+    std::vector<const vImage_Buffer*> image_plane_ptrs(c);
+    std::vector<void*> output_channels(c);
+    for (size_t i = 0; i < c; ++i) {
```
gustavla: Any reason why `c` is an `unsigned (int)` and not `size_t`, just like `i` here?

src/unity/python/turicreate/toolkits/_mps_utils.py
nickjong: Yeah, looks like `np.asarray` makes sense here. The usage of `np.array` was copied from the existing `_prepare_network_parameters` function, so I can update that usage, too.

src/unity/toolkits/neural_net/mps_utils.mm
nickjong: Ah, this is mostly because I use `c` down below as the `channelCount` argument to [vImageConvert_PlanarToChunkyF](https://developer.apple.com/documentation/accelerate/1533074-vimageconvert_planartochunkyf?language=objc), which takes `unsigned int`. Implicit conversions from `unsigned int` to possibly larger unsigned integer types should be free; supplying a `size_t` value for an `unsigned int` argument could require a cast to squelch warnings.

nickjong:
Yeah, my thought is that the transpose isn't terribly costly; this is entirely about having consistent semantics at the main interfaces between components. Right now, this means just being able to load convolution parameters straight from CoreML (or MXNet params) into a `cnn_module` without any transpose, relying the `cnn_module` to transpose if necessary.

Especially with the most recent version of this PR, we're only moving the transpose into TCMPS for the network parameters init/export. This way, the training images can remain HWC all the way from SFrame  to augmentation to loading into the network (for MPS).

igiloh:
We did sort of an MPS transpose in the [LSTM layer implementation](https://github.com/apple/turicreate/blob/master/src/unity/toolkits/neural_net/mps_layers.mm#L984), by playing around with offsets.  
But I guess that was a sort of a special case, since the MPS `encode()` command we were using was sort-of designed for that, having a set of offsets part of its API.  
It's probably very "cheap" in the sense that we didn't copy any data around, but I'm not sure how efficient it is with caching, etc.

PR #1170, opened 2018-10-30
Clarify Linux build requirements

znation:
Be explicit about user-installed dependencies on Linux, including some that we forgot to mention like lapack and Python headers.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -27,7 +27,7 @@ You will need:
 * On Linux:
   * A C++ compiler toolchain with C++11 support (provided by `build-essential` on Ubuntu)
   * blas and lapack development libraries (typically provided by the `liblapack-dev` package)
-  * `pyfpe.h` (typically provided by the `libpython2.7-dev` package)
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -23,15 +23,22 @@ Build Dependencies
 
 You will need:
 
-* On macOS, [Xcode](https://itunes.apple.com/us/app/xcode/id497799835) with command line tools (tested with Xcode 9 and Xcode 10 beta 3)
+* On macOS, [Xcode](https://itunes.apple.com/us/app/xcode/id497799835) with command line tools (tested with Xcode 9 and later)
 * On Linux:
-  * A C++ compiler toolchain with C++11 support
+  * A C++ compiler toolchain with C++11 support (provided by `build-essential` on Ubuntu)
+  * blas and lapack development libraries (typically provided by the `liblapack-dev` package)
+  * `pyfpe.h` (typically provided by the `libpython2.7-dev` package)
```
TobyRoseman: What if it's python 3 ....

znation: Good point, I will address that here.

PR #1171, opened 2018-10-30
Recommender C-API

fareeha:
Update to the Recommender C++ class to use the new extensions mechanism.

* registered new wrapper functions
* fixed python tests: test_recsys_api.py and test_recommender.py
* added new C-API tests


src/unity/lib/extensions/model_base.cpp
```diff
@@ -246,4 +241,32 @@ void model_base::register_docstring(
...
+        std::dynamic_pointer_cast<model_base>(this->shared_from_this());
+
+    turi::get_unity_global_singleton()->save_model(m, side_data, url);
+  }
+
+  /**
+   * Save a toolkit class to a data stream.
+   *
+   * \param sidedata Any additional side information
+   * \param url The destination url to store the class.
```
shantanuchhabra: Documentation doesn't match function signature. Probably a copy-paste error.

fareeha: Good catch! Thank you

src/unity/python/turicreate/toolkits/recommender/util.py
```diff
@@ -595,9 +596,9 @@ def get_num_items_per_user(self):
         --------
 
         """
-        opts = {'model': self.__proxy__}
-        response = _turicreate.extensions._recsys.get_num_items_per_user(opts)
-        return response['data']
+
+        response = self.__proxy__.get_num_items_per_user()
+        return response
```
shantanuchhabra: Remove trailing space please.

src/unity/python/turicreate/toolkits/recommender/popularity_recommender.py
```diff
@@ -79,30 +79,27 @@ def create(observation_data,
     --------
     PopularityRecommender
     """
-
-    opts = {'model_name': 'popularity'}
-    response = _turicreate.extensions._recsys.init(opts)
-    model_proxy = response['model']
+
```
shantanuchhabra: Delete tab please.

src/unity/python/turicreate/toolkits/recommender/util.py
```diff
@@ -1312,9 +1293,7 @@ def check_type(arg, arg_name, required_type, allowed_types):
                }
 
         with QuietProgress(verbose):
-            response = _turicreate.extensions._recsys.recommend(opt)
-
-        recs = response['data']
+            recs = self.__proxy__.recommend(users, exclude, items, new_observation_data, new_user_data, new_item_data, exclude_known, k, diversity, random_seed)
```
shantanuchhabra: This is just a matter of code style, not super important, and we're not very strict about this throughout our codebase -- I think we should maintain the 80 character per line limit as much as possible, but I'll let @znation , @nickjong , and @hoytak chime in here.

src/unity/toolkits/recsys/recsys_model_base.hpp
```diff
@@ -352,9 +384,87 @@ class EXPORT recsys_model_base : public ml_model_base {
   /// Serialization -- load
   void load_version(turi::iarchive& iarc, size_t version) override;
 
+ // Implement the bare minimum of the pure virtual methods
+  //virtual void init_options(const std::map<std::string, flexible_type>&_options) =0;
```
shantanuchhabra: Please remove dead code.

src/unity/toolkits/recsys/recsys_model_base.cpp
```diff
@@ -1721,6 +1746,200 @@ sframe recsys_model_base::get_num_users_per_item() const {
...
+
+  sframe predictions = this->predict(this->create_ml_data(sf, new_user_data_sf, new_item_data_sf));
+
+  return predictions;
+}
+
+/*
+variant_map_type api_get_value(flexible_type field) {
+  log_func_entry();
+  variant_map_type ret;
```
shantanuchhabra: Dead code.

src/unity/toolkits/recsys/recsys_model_base.cpp
```diff
@@ -1239,6 +1238,26 @@ std::shared_ptr<unity_sframe_base> recsys_model_base::recommend_extension_wrappe
 
 }
 
+std::shared_ptr<unity_sframe_base> recsys_model_base::get_num_users_per_item_extension_wrapper() const {
+
```
shantanuchhabra: Some documentation here would be helpful for people who inherit this code. One-line function headers for all the functions you added in this file (and even elsewhere) would be great. Furthermore, wherever you think something complicated and non-trivial is happening, feel free to add additional documentation. Think of how you can make the ramp-up time for the person who works on this code a year from now, easier!

src/unity/toolkits/recsys/recsys_model_base.hpp
```diff
@@ -250,11 +252,37 @@ class EXPORT recsys_model_base : public ml_model_base {
...
+  gl_sframe new_item_data, bool exclude_training_interactions, size_t top_k, double diversity, size_t random_seed);
+
+  gl_sframe api_get_item_intersection_info(gl_sframe item_pairs);
+
+  gl_sframe api_precision_recall_stats(gl_sframe indexed_validation_data, gl_sframe recommend_output, const std::vector<size_t>& cutoffs);
+
+  variant_map_type api_get_train_stats();
+
+  EXPORT variant_map_type api_get_data_schema();
+  //variant_map_type api_get_value(flexible_type field);
```
shantanuchhabra: Dead code.

src/unity/python/turicreate/toolkits/recommender/util.py
```diff
@@ -499,7 +501,8 @@ def random_split_by_user(dataset,
             'item_test_proportion': item_test_proportion,
             'random_seed': random_seed}
 
-    response = _turicreate.extensions._recsys.train_test_split(opts)
+    response = _turicreate.extensions._recsys.train_test_split(dataset, user_id, item_id, max_num_users, item_test_proportion, random_seed)
```
TobyRoseman: This line is a bit long. I try to limit my line length to under a 100 characters. I think the official guidance is under 80. I'd recommend breaking this up into more than one line.

src/unity/python/turicreate/toolkits/recommender/util.py
TobyRoseman: I didn't see this before my previous comment. [Here is what PEP8 say about it](http://docs.bigchaindb.com/projects/contributing/en/latest/cross-project-policies/python-style-guide.html). I don't think we need to be super strict by keeping it under 100 characters is a good idea.

src/unity/python/turicreate/toolkits/recommender/util.py
fareeha: Thanks for pointing it out!

src/unity/toolkits/recsys/recsys_model_base.cpp
```diff
@@ -1721,6 +1746,200 @@ sframe recsys_model_base::get_num_users_per_item() const {
...
+EXPORT variant_map_type train_test_split(gl_sframe _dataset,
+                                         const std::string& user_column,
+                                         const std::string& item_column,
+                                         flexible_type max_num_users,
+                                         double item_test_proportion,
+                                         size_t random_seed) {
+  variant_map_type ret;
+  sframe dataset = _dataset.materialize_to_sframe();
+  //flexible_type _max_users;
```
TobyRoseman: delete this line

test/capi/capi_models.cxx
```diff
@@ -280,6 +279,220 @@ BOOST_AUTO_TEST_CASE(test_boosted_trees_double) {
...
+        {"target", {0, 1, 1, 1, 1, 0}}};
+
+    tc_sframe* sf = make_sframe_integer(data);
+
+    tc_parameters* args = tc_parameters_create_empty(&error);
+
+    CAPI_CHECK_ERROR(error);
+
+    // Add in the sframe; then destroy it when we're done with it.
+    {
```
TobyRoseman: I'm not really understanding why your creating a namespace here (i.e. using the curly braces) or why you're doing it on line 307 and line 312. I just don't think it adds anything and find it's kind of confusing. 

I like your usage on line 318 and 329 were you allocation something at the beginning and deallocate it at the end.

test/capi/capi_models.cxx
```diff
@@ -280,6 +279,220 @@ BOOST_AUTO_TEST_CASE(test_boosted_trees_double) {
...
+      CAPI_CHECK_ERROR(error);
+
+
+      tc_release(fd);
+    }
+
+    tc_model* model; 
+
+      // We now have enough to create the model.
+      model = tc_model_new(model_name, &error);
```
TobyRoseman: The indentation is wrong here.

fareeha: Ty! Corrected.

test/capi/capi_models.cxx
fareeha: [Line 65](https://github.com/apple/turicreate/blob/master/test/capi/capi_models.cxx) of the original capi_models.cxx, for instance, does this:
   
> // Add in the sframe; then destroy it when we're done with it.
>     {
>       tc_parameters_add_sframe(args, "data", sf, &error);
>       CAPI_CHECK_ERROR(error);
>     }

I assumed this format to be by convention, so I was keeping consistent with it. But I'll remove the namespaces for tc_parameters_add_sframe() in lines 302, 307 and 312 as you've mentioned.

src/unity/toolkits/recsys/recsys_model_base.cpp
fareeha: Done

src/unity/toolkits/recsys/recsys_model_base.hpp
fareeha: Done. Thank you

src/unity/lib/extensions/model_base.hpp
```diff
@@ -83,7 +83,7 @@ class EXPORT model_base: public cppipc::ipc_object_base {
    * Serializes the toolkit class. Must save the class to the file format
    * version matching that of get_version().
    */
-  virtual void save_impl(oarchive& oarc) const;
+  virtual void save_impl(oarchive& oarc) const {};
```
znation: Why do we want to provide an empty implementation here? This doesn't seem like the right behavior - if all subclasses are expected to implement this method, it should be pure virtual (`= 0`), or if there is some reason it can't be, perhaps it should throw an exception at runtime in case it somehow gets called.

src/unity/lib/extensions/model_base.hpp
```diff
@@ -95,14 +95,30 @@ class EXPORT model_base: public cppipc::ipc_object_base {
    * Loads a toolkit class previously saved at a particular version number.
    * Should raise an exception on failure.
    */
-  virtual void load_version(iarchive& iarc, size_t version);
+  virtual void load_version(iarchive& iarc, size_t version) {} ;
```
znation: Same comment as above.

src/unity/toolkits/recsys/recsys_model_base.hpp
fareeha: Done.

src/unity/python/turicreate/toolkits/recommender/util.py
fareeha: Thanks everyone

test/capi/capi_models.cxx
```diff
@@ -280,6 +279,220 @@ BOOST_AUTO_TEST_CASE(test_boosted_trees_double) {
...
+    }
+
+    // Test saving and loading the model.
+    {
+      std::string model_path;
+      std::string error_message;
+      std::string expected_substr;
+
+#ifdef MECHANISM_FOR_TRIGGERING_AN_ERROR_IN_DOCKER
+      // sad path 1 - attempting to save without permission to location
```
znation: You don't need to test all the sad paths here - I think the purpose of that was actually to make sure our low level filesystem handling code had the correct behavior (mostly correct error messages) for various permission issues. Since this test is really testing the same thing across all different models, I would skip it for this one and go straight to the "happy path" (the one where the model actually saves), if you want to test save and load for this model.

test/capi/capi_utils.hpp
```diff
@@ -197,6 +197,36 @@ static tc_sarray* make_sarray_double(const std::vector<double>& v) {
   return sa;
 }
 
+static tc_sarray* make_sarray_integer(const std::vector<int64_t>& v) {
```
znation: You may need `__attribute__((__unused__))` if this header is included in any test files that don't call this function -- otherwise you may get build errors on various platforms.

test/capi/capi_utils.hpp
```diff
@@ -270,6 +300,44 @@ static tc_sframe* make_sframe_double(const std::vector<std::pair<std::string, st
     return sf;
   }
 
+  static tc_sframe* make_sframe_integer(const std::vector<std::pair<std::string, std::vector<int64_t> > >& data ) {
```
znation: Same comment as above.

src/unity/lib/extensions/model_base.hpp
```diff
@@ -95,14 +95,30 @@ class EXPORT model_base: public cppipc::ipc_object_base {
...
+   * \param url The destination url to store the class.
+   */
+  void save_model_to_file(const variant_map_type& side_data,
+                          const std::string& url);
+
+  /**
+   * Save a toolkit class to a data stream.
+   *
+   * \param sidedata Any additional side information
+   * \param url The destination url to store the class.
```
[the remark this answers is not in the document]
fareeha: Thanks for pointing this out!

src/unity/lib/extensions/model_base.hpp
hoytak: Eh, model_base isn't expected to be serializable, but ml_model base is.  We had to make a few of these changes for weird reasons.  I think the best update is to undo the removal of the pure virtual methods in ml_model.hpp.

znation: @hoytak I agree, putting back the pure virtual methods in ml_model.hpp sounds right. Should we do that as part of this PR?

test/capi/capi_utils.hpp
fareeha: Done. Thank you so much for the note!

test/capi/capi_utils.hpp
fareeha: Done.

src/unity/lib/extensions/model_base.hpp
fareeha: Thank you guys. A comment note added to the empty implementations as per offline discussion.

src/unity/lib/extensions/model_base.hpp
fareeha: A comment note added to the empty implementations as per offline discussion.

test/capi/capi_models.cxx
TobyRoseman: It's more than just formatting. Adding those curly braces introduces a namespace; things created inside can not be accessed from the outside. This can be very useful for keeping long functions sane.

src/unity/toolkits/recsys/recsys_model_base.cpp
fareeha: There isn't a documentation for this wrapper code because this is a hack adapted from @znation 's recommend_extension_wrapper() above to get sframe across over to python for specific methods. We don't want to encourage workarounds for each individual method and want to fix this issue in a separate PR. Thanks for bringing this up, I've added a comment above the code explaining what the wrapper is here for.

test/capi/capi_models.cxx
znation: Also note that it controls variable construction/destruction implicitly with scoping - meaning a variable inside this block won't run its constructor until the block runs, and will run its destructor at the end of the block. This can be useful (especially in switch/case statements, where you don't want to initialize variables for cases that aren't currently the one being evaluated). Note that `{ }` always creates a scoping block in C/C++, whether it's for a `for`, `if`, a function, or just by itself (like here), and the semantics are the same. (Python and JavaScript are quite different in this respect.)

test/capi/capi_models.cxx
fareeha: Thanks for letting me know! Sad paths removed :)

test/capi/capi_models.cxx
fareeha: Thank you so much for explaining @TobyRoseman @znation

nickjong: FYI, to avoid future confusion, the C++ construct we're discussing is just a [C++ block](https://en.cppreference.com/w/cpp/language/scope) (compound statement).

In C++, a [namespace](https://en.cppreference.com/w/cpp/language/namespace) is a specific construct always introduced by the `namespace` keyword.

As Toby and Zach discussed, the primary reason to introduce a block (that is not otherwise required by C++ syntax) is to control the lifetime of variables declared inside.

TobyRoseman: That's completely correct: I should have said "scope" not "namespace".

Issue #1030, opened 2018-08-23
Travis CI should also run in Python 3


TobyRoseman:
https://docs.travis-ci.com/user/languages/python/

PR #1175, opened 2018-10-30
Init Python 3 Continuous Integration

TobyRoseman:


TobyRoseman:
@znation, @nickjong - I think this is now ready to go in.

TobyRoseman:
fixes #1030

PR #1177, opened 2018-10-31
Passing in predictions to evaluate

hoytak:


znation:
Not sure I understand what's going on here. @nickjong should probably review.

nickjong:
What's the context for these changes?

FWIW, the test failures that Travis CI is finding are the same ones that prompted the related changes in #894

hoytak:
(Adding in offline discussion here). The purpose of this change is to allow predictions to be passed in to the evaluate function to save time; certain sub-methods (e.g. image classification) take a long time, and this allows already made predictions to be optionally recycled.  This is essentially a slightly hacky alternative to exposing the evaluation toolkit to the C-API.

src/unity/python/turicreate/test/test_boosted_trees.py
```diff
@@ -127,7 +127,7 @@ def test__list_fields(self):
         """
         model = self.model
         fields =  model._list_fields()
-        self.assertEqual(set(fields), set(self.fields_ans))
+        self.assertTrue(set(self.fields_ans).issubset(fields))
```
nickjong: Is the addition of new fields intentional then? (The context provided for this PR only mentioned avoiding re-computation of predictions.)

If it is intentional, shouldn't we document the new expected fields, both in the test and in the Pydoc?

src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -1107,46 +1113,58 @@ gl_sframe supervised_learning_model_base::api_classify(
...
-  sframe test_data = data.materialize_to_sframe();
-  sframe X = setup_test_data_sframe(test_data, model, missing_value_action);
-  sframe y = test_data.select_columns({get_target_name()});
-  ml_data m_data = setup_ml_data_for_evaluation(
-      X, y, model, missing_value_action);
-
+  variant_map_type ret;
 
-  if(metric == "report") {
+  if(metric == "report" || metric == "auto") {
```
nickjong: I found all this logic very difficult to follow.

AFAICT, the original logic basically said:
- If the requested metric is "report" and the model is a classifier, then just return the confusion matrix, report by class, and accuracy.
- If the requested metric is "report" and the model is not a classifier, then treat "report" as "auto
- For non-"report" metrics, just call the underlying `evaluate` method

The new logic basically says:
1. If the metric is "report" or "auto", and the model is a classifier, then compute the confusion matrix, report by class, and accuracy
2. If the metric is NOT "report" OR if the model is NOT a classifier, then compute the results of calling `evaluate`
3. Return the combined the results, if any, from 1 and 2 above

Is this the intent? (That is, add confusion matrix, report by class, and accuracy to all classification toolkits by default?) If so, could we make the logic more like the above? Just create a single accumulator map for all the results. If desired, call out to one function to add the classification reports. Then if necessary, call out to `evaluate` to add the traditional stuff?

src/unity/python/turicreate/test/test_boosted_trees.py
hoytak: Yes, I agree.  I can add that documentation.

src/unity/toolkits/supervised_learning/supervised_learning.cpp
hoytak: Yes, I agree.  This logic is an evolutionary walk of patches to get stuff to work.  I think it should be completely refactored at this point.  Let me do that.

hoytak:
Closing as this is for internal stuff, and this code path will be redone shortly..

src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -1107,19 +1107,12 @@ gl_sframe supervised_learning_model_base::api_classify(
  *  Evaluate the model
  */
 variant_map_type supervised_learning_model_base::api_evaluate(
-    gl_sframe data, std::string missing_value_action_str, std::string metric) {
+    gl_sframe data, std::string missing_value_action_str, std::string metric, gl_sarray predictions) {
```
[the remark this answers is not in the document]
znation: Or a `shared_ptr` that defaults to `nullptr`?

nickjong:
Just to clarify @hoytak, did you intend to push any more commits addressing the feedback last year, after reopening this PR?

src/unity/toolkits/supervised_learning/supervised_learning.cpp
nickjong: In my other comment, I am especially referring to the logic here

PR #1178, opened 2018-10-31
Implement basic training for the new C++ implementation of object detection

nickjong:
With these changes, I can now train a model on the data from our "breakfast" demo:
```
import turicreate as tc

data = tc.SFrame('breakfast-data.sframe')
m = tc.extensions.object_detector()
m.train(data, 'annotation', 'image', {'mlmodel_path':'/Users/jong/Desktop/od-init.mlmodel', 'max_iterations' : 1000})
```

I get output like:
```
| Iteration    | Loss         | Elapsed Time |
| 1            | 7.98467      | 10.86s       |
| 2            | 7.9668       | 14.37s       |
| 3            | 7.96799      | 17.87s       |
...
| 999          | 0.560803     | 1h 1m        |
| 1000         | 0.630218     | 1h 1m        |
```

I can't actually export the model yet or run prediction, but it's clearly learning something, right? Well, there's no data augmentation, so it's probably overfitting. And all the image resizing is happening on a single thread (the same one feeding MPS), so it's slow.

### DIFF
--- a/src/unity/toolkits/neural_net/CMakeLists.txt
+++ b/src/unity/toolkits/neural_net/CMakeLists.txt
@@ -49,8 +49,10 @@ make_library(unity_neural_net
   SOURCES
     cnn_module.cpp
     coreml_import.cpp
+    image_augmentation.cpp
     float_array.cpp
   REQUIRES
+    image_type
     logger
     unity_coreml_model_export
     ${additional_unity_neural_net_requirements}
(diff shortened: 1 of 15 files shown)


nickjong:
I rediscovered virtual functions and cleaned up the dependency injection into object_detector, along the way to fleshing out the unit tests. We could probably stand to add more, but this should be enough to get us off the ground.

nickjong:
Thanks for the review, Hoyt! Going to keep this open a bit longer in case others want to offer feedback.

Also, disregard my comments re performance. When building for release, even single-threaded image processing seems to keep up with the MPS backend. I'm optimistic that we can handle full data augmentation with the appropriate parallelization.

PR #1185, opened 2018-11-02
Fixes Copyright Headers

TobyRoseman:


src/unity/toolkits/coreml_export/MLModel/src/BayesianProbitRegressionValidator.cpp
```diff
@@ -1,11 +1,8 @@
-//
-//  BayesianProbitRegressionValidator.cpp
-//  mlmodelspec
-//
-//  Created by Roman Holenstein on 7/18/17.
-//  Copyright © 2017 Apple. All rights reserved.
-//
-
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: All the changes under `coreml_export/MLModel` will need to go into [coremltools](https://github.com/apple/coremltools/tree/master/mlmodel) as well.

src/objc/TuriCreateObjC.h
```diff
@@ -1,3 +1,8 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: The year should probably say 2018.

TobyRoseman: Let's handle that in a separate PR. That needs to be fixed everywhere.

PR #1189, opened 2018-11-06
Added EXPORT keywords to visualization C-API.

hoytak:
This PR simply added EXPORT keywords to the plot functions in the C API.  No other non-indent changes. 

(Also ended up removing the indent of the extern "C" block.)


znation:
D'oh - sorry I forgot about those. Thanks for adding them @hoytak! LGTM.

PR #1192, opened 2018-11-07
Enable debug optimization mode in travis.

hoytak:
- Tests run much, much faster.  Code should be the same.  -DNDEBUG is not enabled, but -Os is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,4 +58,4 @@ script:
   - rm src/unity/python/turicreate/test/test_style_transfer.py src/unity/python/turicreate/test/test_object_detector.py
 
   # Build the wheel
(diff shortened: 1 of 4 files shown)


hoytak:
Superseded by docker travis builds.

znation:
@hoytak Actually, we may still want this change -- the Docker build change in #1750 won't actually change our Travis config yet to use Docker. Plus, we may want to build debug-opt even in Docker.

znation:
Actually I misunderstood the purpose of this change. I think we have already lost most of the advantage with the switch to `OBJECT` based builds (since test binaries are much smaller now without static linking). Closing.

PR #1193, opened 2018-11-08
Allow visualization server to build without client

znation:
The macro TC_BUILD_VISUALIZATION_CLIENT should only be used for the
client (tcviz app on macOS or Linux); the C API for visualization
should work regardless as it's purely data structure transformations
and doesn't depend on any client app.


abhishekpratapa:
Looks good to me!

hoytak:
@abhishekpratapa -- I think you need to just approve the PR, then Zach can merge it instead of closing it.

znation:
Thanks @hoytak. You could approve it as well. 😄

PR #1199, opened 2018-11-09
Introduce neural_net::model_spec as primary storage for a learned NN

nickjong:
With this change, the C++ object_detector maintains (an encapsulated version of) a CoreML NeuralNetwork proto specifying the trained weights. This proto can be initialized with the pre-trained weights before training, and updated with the results of training when necessary. CoreML export will be able just to copy this proto into the output proto.


nickjong:
The first step in getting CoreML export working in the new C++ object detector: just use CoreML protobuf as the main representation of the model, much as the Python-based implementation stores a MXNet model. (This is a bit of a departure from my original plan, which was to use these `float_array_map` values instead.)

Much of this change is a rename from neural_net/coreml_import to model_spec. For reviewing convenience, I’ve isolated the file rename into its own commit (you’ll get a cleaner diff if you only look at the other commit)

The new model_spec class is conceptually similar to the coremltools.model.neuralNetwork.Builder class, but AFAICT that’s all in Python. For now I’m not claiming (or attempting) anything more general than what we need for this toolkit…

PR #1207, opened 2018-11-10
SArray test_filter_by should not assume filter_by preserves order

nickjong:
Resolves non-deterministic test failures (that happened to pass on Travis)

### DIFF
--- a/src/unity/python/turicreate/test/test_sarray.py
+++ b/src/unity/python/turicreate/test/test_sarray.py
@@ -3168,24 +3168,20 @@ class SArrayTest(unittest.TestCase):
         
         #integer example
         x = SArray([1,2,3,4,5,6,7])
-        self.assertTrue(np.array_equal(x.filter_by([11,7,2,8,4]), SArray([2,4,7])))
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/test/test_sarray.py
```diff
@@ -3168,24 +3168,20 @@ def test_filter_by(self):
         
         #integer example
         x = SArray([1,2,3,4,5,6,7])
-        self.assertTrue(np.array_equal(x.filter_by([11,7,2,8,4]), SArray([2,4,7])))
-        self.assertTrue(np.array_equal(x.filter_by([11,7,2,8,4,3], exclude=True), SArray([1,5,6])))
+        self.assertTrue(sorted(x.filter_by([11,7,2,8,4])) == [2,4,7])
```
znation: Now that this uses the == operator, we'll get better test output on failure using `self.assertEqual` instead.

nickjong: Duh, good catch

Thanks!

PR #1208, opened 2018-11-10
Fix object detector export parameters when set to 0

gustavla:
It should be possible to set `confidence_threshold=0.0` and not have it silently fall back to 0.25.


nickjong:
I manually tested the Python tests that depend on this code, and everything looks good there.

PR #1211, opened 2018-11-13
Allow empty plot labels in TC plots

znation:
Previously, empty string in TC plots was treated as a special
placeholder to indicate default value, and null values weren't
supported in plot APIs below Python.

With this change, null now means to omit the label, empty string
is actually empty string, and there is a __TC_DEFAULT_LABEL
placeholder string to indicate that the default should be used.


hoytak:
<Sorry to take so long to get to the review here. > 

I'm not sure why the default label handling needs to get pushed all the way down to the C++ code.  Instead of the special handling of __TURI_DEFAULT_LABEL, which is a bit opaque to the user in python, what if we would do a scheme like 

DEFAULT_LABEL = "$(axis)s" 

Then in the plot function, do:

```
xlabel = xlabel % {"axis" : "x"}  # Or column name, whatever. 
ylabel = ylabel % {"axis" : "y"}
...
```
 Then it'd be handled in python and we wouldn't have to special case values.

znation:
@hoytak The Python API-level defaults still need to be opaque to the user, since they are different at runtime depending on (for instance) the dtype of the SArray(s). Does that change your opinion of whether I should remove the C++ plumbing for the defaults and move it up to Python? I'm not sure there is any user-facing benefit to that now.

PR #1213, opened 2018-11-14
Remove nodejs/npm build dependency

znation:
With this commit, we use a different strategy for modifying and building the UI
assets for Turi Create. Previously, we would invoke npm to install dependencies
and run the build as part of the Turi Create build. However, this introduced a
few problems:

* Different versions of node/npm on developer machines gave different behavior
  and some combinations did not work.
* The out-of-the-box Ubuntu setups are one of the configurations that don't work,
  resulting in a fresh clone of Turi Create failing to compile on a fresh install
  of Ubuntu.
* In Travis CI, the `npm install` or `npm run build` commands tend to fail for
  strange reasons (possibly networking issues, disk issues, or timeouts).
* All developers of Turi Create now need to have node/npm, even though few
  developers are modifying the JS/HTML/CSS code.

With this change, the npm build is skipped as part of the regular Turi Create build,
and instead, the output of the npm build is committed directly to the repository.

For developers who don't modify HTML/CSS/JS, nothing should change.

For developers who do, you will now need to run `npm install` and `npm run build
explicitly on your dev machine whenever modifying `package.json`, and any JS/
HTML/CSS code, respectively. You will then need to commit the output it generates
into the repo, along with any other changes you are making, for those changes
to get picked up by others.


abhishekpratapa:
When we make the wheel though, do we need to do anything different, to make sure the wheel has visualization packaged into it?

znation:
@abhishekpratapa Good question. I don't think so, but I'll verify that before merging this PR.

PR #1214, opened 2018-11-15
OD: Make sure setting imageCropAndScaleOption is included in example code

### LINKED ISSUE #1016: Different behavior between Model and Core ML model
I successfully trained a Object Detection model (default TC YOLO) with TC b3 and exported in CoreML format.
My model has 8000 iterations and 0.8 final loss.

I then validate it with some images using TC and bounding box drawing util and it recognizes them better that I expected!

I then downloaded [the sample project for recognizing objects in live capture](https://developer.apple.com/documentation/vision/recognizing_objects_in_live_capture) presented by @znation during the WWDC and replaced the model in the project with my new model.

What it's weird is that the object are no longer recognized. Is NOT a problem of VNDetectedObjectObservation because they are correctly returned, but it seems that the classing and the bounding box does not represent the detected object correctly (different class and wrong bounding box).
I used iOS 12 beta 9 and Xcode 10 beta 6 as developing environment, with an iPad Pro 2017 (or 2016, I don't remember).

From my first test seems this could be a rotation issue, but I don't know if that is the real issue and eventually how to fix it.

Does anybody faced a similar issue or can eventually help me with that?

gustavla:
This is included in our deployment code for iOS 11, but missing from iOS 12. This option is important for correct handling of input.

Fixes #1016 (!!)

### DIFF
--- a/userguide/object_detection/export-coreml.md
+++ b/userguide/object_detection/export-coreml.md
@@ -30,7 +30,7 @@ let objectRecognition = VNCoreMLRequest(model: visionModel,
         print(bestLabel.identifier, bestLabel.confidence, objectBounds)
     }
 })
-
+objectRecognition.imageCropAndScaleOption = .scaleFill
 ```
 
 For more details on the integration with Core ML and a sample app to get


gustavla:
Actually, to fix #1016, maybe we should add a note that this line needs to be added to fix the demo code, since most people in #1016 seems to be using Vision's example project, which also has this issue.

PR #1229, opened 2018-11-30
11 28 lbfgs update

hoytak:
Updated the LBFGS solver for iterative solving and checkpointing.

    This is one isolatable step on the road to revamping the supervised
    learning process to support checkpointing and incremental updates.
    with this change, the LBFGS solver is implemented as a class with a
    setup and a next_iteration() method on it.  Each time the
    next_iteration method is called, the solver performs one step of
    optimization, returning true on error.

    In addition, three bugs were fixed in the lbfgs solver that will hopefully
    improve optimization performance.  There was an indexing error in how the previous
    gradients were stored resulting in the solver performing something like
    gradient descent until the lbfgs memory was full.  Second, the order in
    which these previous gradients were accumulated was done in the wrong
    order.  Third, the initial step size calculated by more_thuente was
    never actually stored beyond its initial use.

    A compatibility function that performs comparable progress printing to
    the previous implementation is provided until supervised learning adopts
    the new interface.


src/optimization/lbfgs.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+  /** Sets up (or resets) the solver with the initial conditions.  
+   *
+   *
+   */
+  void setup(const DenseVector& init_point,
+             const std::map<std::string, flexible_type>& opts,
+             const std::shared_ptr<smooth_regularizer_interface>& reg = nullptr);
+
+  /** Call this method repeatedly, with each HERE
```
TobyRoseman: I'm not understanding this comment.

src/optimization/lbfgs.hpp
```diff
@@ -0,0 +1,171 @@
...
+// Old version for backwards compatibility with the previous interface.
+// Includes printing.
+solver_return lbfgs_compat(
+    std::shared_ptr<first_order_opt_interface> model,
+    const DenseVector& init_point,
+    const std::map<std::string, flexible_type>& opts,
+    const std::shared_ptr<smooth_regularizer_interface>& reg = nullptr);
+
+
+// TODO: document this.
```
TobyRoseman: might be a good thing to do now

PR #1235, opened 2018-12-05
Fix logger output in debug builds

znation:
Seems that perhaps we were relying on undefined behavior, and the
behavior has changed in newer compilers. (Unclear cause.)

The result is that in debug mode, the expression
`endltype(f) == endltype(std::endl)` always returns false,
even when the modifier passed in is in fact std::endl.
As a result, in debug mode, we never flush the stream.

This fix is to ignore which modifier is being passed in, and
always flush the stream as though it's std::endl. In practice,
it's likely to be endl most of the time, so the performance hit
from flushing more often should not be too bad.

### DIFF
--- a/src/logger/logger.hpp
+++ b/src/logger/logger.hpp
@@ -498,7 +498,9 @@ class file_logger{
       std::stringstream& streambuffer = streambufentry->streambuffer;
       bool& streamactive = streambufentry->streamactive;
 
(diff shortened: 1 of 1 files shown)


znation:
@ylow Thanks, that's a good point. I added a comment explaining why we flush on all stream modifiers now, and added the output of `clang --version`.

PR #1239, opened 2018-12-05
Prototype of a Sparse NN model for pure categorical features

hoytak:
A simple implementation of an extremely efficient brute-force nearest neighbors algorithm meant for single-query nearest neighbors query with Jaccard distance over categorical variables.  

This also is a simple implementation meant to demonstrate the ease of using the extensions to create a minimal model.  A new directory, prototype, is designed for these simple implementations.  

In addition, a few bugs preventing the proper saving and loading of pure extension models were fixed.

### DIFF
--- a/src/unity/python/turicreate/test/test_prototype_models.py
+++ b/src/unity/python/turicreate/test/test_prototype_models.py
@@ -1,0 +1,31 @@
+import turicreate as tc
+import unittest
+import tempfile
+import shutil
+
+class SparseNNTest(unittest.TestCase):
+
+    def test_sparse_nn(self):
+        X = tc.util.generate_random_sframe(100, "ssszzz")
+        Y = X.copy()
+        X = X.add_row_number()
+
+        m = tc.extensions._sparse_nn()
+
+        m.train(X, "id")
+
+        for i, row in enumerate(Y): 
+            res = m.query(Y[i], 1)
+            self.assertEqual(res, {i : 1.0})
+
+        # Save and load
+        model_file = tempfile.gettempdir() + "/sparse_nn.model"
+        m.save(model_file)
+        m2 = tc.load_model(model_file)
+        
+        for i, row in enumerate(Y): 
+            res = m2.query(Y[i], 1)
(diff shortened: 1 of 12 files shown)


src/unity/python/turicreate/test/test_prototype_models.py
```diff
@@ -0,0 +1,31 @@
...
+class SparseNNTest(unittest.TestCase):
+
+    def test_sparse_nn(self):
+        X = tc.util.generate_random_sframe(100, "ssszzz")
+        Y = X.copy()
+        X = X.add_row_number()
+
+        m = tc.extensions._sparse_nn()
+
+        m.train(X, "id")
```
znation: Shouldn't we try to mirror the standard model APIs here, with a static function `create` rather than an instance function `train`?

hoytak: Unlike in python, the C++ classes have to be instantiated in many situations where they are not trained (e.g. just to get the name) and need a basically constructor.  We could do this, but it should be something separate.  Let's discuss but keep this for now.

src/unity/toolkits/prototype/sparse_nn.cpp
```diff
@@ -0,0 +1,150 @@
...
+    log_and_throw("Model not trained yet.");
+  }
+
+  atomic<size_t> current_index = 0;
+
+  std::vector<atomic<uint32_t> > hit_counts(ids.size(), 0);
+
+  in_parallel([&](size_t thread_idx, size_t num_threads) GL_HOT_FLATTEN {
+
+    // Each thread takes the next one
```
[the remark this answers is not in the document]
hoytak: The current implementation of parallel_for doesn't do very well when the iteration-per-index speed is highly variable; this way of doing things is more resilient to that.  That could be something to resolve.

src/unity/python/turicreate/test/test_prototype_models.py
znation: Oh, right - I forgot about that. Let's keep that in mind for the long term, but for now this is fine.

PR #1240, opened 2018-12-06
Call os_log from table_printer

znation:
* All types that are passed into table printer are handled.
* Logs start, header, values, and end.
* Apple-platform only; on others, this will no-op.

### DIFF
--- a/src/table_printer/table_printer.cpp
+++ b/src/table_printer/table_printer.cpp
@@ -73,8 +73,17 @@ void table_printer::print_header() const {
 
   std::ostringstream ss;
 
+#ifdef __APPLE__
+    os_log_info(
+      OS_LOG_DEFAULT,
+      "source: %lu, event: %lu",
+      1ul, /* source: table printer */
+      0ul); /* event: started */
+#endif
+
   ss << '|';
 
+  size_t i = 0;
   for(const auto& p : format) {
 
     ss << ' ';
@@ -85,6 +94,17 @@ void table_printer::print_header() const {
       ss << ' ';
 
     ss << ' ' << '|';
+
+#ifdef __APPLE__
+    os_log_info(
+      OS_LOG_DEFAULT,
+      "source: %lu, event: %lu, column: %lu, value: %{public}s",
+      1ul, /* source: table printer */
+      1ul, /* event: header */
+      i, /* column index */
+      p.first.c_str());
+#endif
+    i++;
   }
 
   _p(ss);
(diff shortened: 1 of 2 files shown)


src/table_printer/table_printer.hpp
```diff
@@ -18,6 +18,12 @@
 #include <util/code_optimization.hpp>
 #include <table_printer/table_element_printers.hpp>
 
+#ifdef __APPLE__
+#include <os/log.h>
+#undef MIN
+#undef MAX
```
nickjong: It makes me nervous when I see this kind of preprocessor fiddling in header files. Do we really need to import <os/log.h> in the header file, or can we just import it into table_printer.cpp?

src/table_printer/table_printer.hpp
```diff
@@ -228,8 +234,22 @@ class table_printer {
...
+  static void _os_log_value(size_t column_index, unsigned long value);
+  static void _os_log_value(size_t column_index, unsigned int value);
+  static void _os_log_value(size_t column_index, long long value);
+  static void _os_log_value(size_t column_index, long value);
+  static void _os_log_value(size_t column_index, int value);
+  static void _os_log_value(size_t column_index, double value);
+  static void _os_log_value(size_t column_index, float value);
+  void _os_log_value(size_t column_index, const progress_time& value) const;
+  static void _os_log_value(size_t column_index, char* value);
+  static void _os_log_value(size_t column_index, const char* value);
```
nickjong: Nit: the `const char*` override should supersede the `char*` override entirely, right?

src/table_printer/table_printer.cpp
```diff
@@ -244,4 +273,96 @@ size_t table_printer::set_up_time_printing_interval(size_t tick) {
   return _tick_interval; 
 }
 
+#define _os_log_value_impl(format, column_index, value) \
+  os_log_info(                                          \
```
nickjong: Don't we need `#ifdef __APPLE__` here?

I'm assuming this needs to be a macro so that line numbers are preserved?

Also, could we centralize the preprocessor dependency on `__APPLE__` here by having the calls to `os_log_info` earlier in this file use this macro instead?

src/table_printer/table_printer.hpp
znation: Good call; I'll try to move it to the .cpp file.

src/table_printer/table_printer.hpp
znation: Thanks, I'll try removing the non-const version.

src/table_printer/table_printer.cpp
znation: Yes, we do need ifdef. I'll try to centralize the uses of ifdef at the top of the file.

znation: P.S. Yes, it does need to be a macro, and not just for line numbers to be preserved, but because there is a compile-time static assert that the format argument must be a string literal and not a char pointer (which could be dynamic).

znation:
@nickjong Feedback addressed; please re-review, thanks!

PR #1252, opened 2018-12-12
Fix plot get_data for SFrame summary view

### LINKED ISSUE #1248: Visualizations are incomplete
Unless the target for visualization is set explicitly to 'gui', inline visualizations render incompletely. In the course Machine Learning Foundations, the instructor uses the show() command to visualize plots for the First Name, Last Name, Country, and age. However, only First Name renders. The rest of the inline window shows a very large whitespace at the bottom of which has the buttons for exporting and so on. If I click the button Open in Vega Editor, I still see only one plot. 
But if I set the target to 'gui' the plots all render correctly in a separate window.
I am reporting this issue o behalf of a dozen users who asked me for assistance.

znation:
This change picks up two commits from a branch that were conceptually
necessary for other recent changes, but somehow didn't break the
build, or fail any tests, or break anything in practice except for
issue 1248. The necessary commits were:

https://github.com/znation/turicreate/commit/616de4b641389ce7ccc4e621850996af8d82781
https://github.com/znation/turicreate/commit/3e02db92ed685414f1be25f21bf23a0b9af3d32c

Fixes #1248

TBD: what specific commit went in that broke this? It worked in the 5.1
release.

### DIFF
--- a/src/unity/lib/visualization/boxes_and_whiskers.cpp
+++ b/src/unity/lib/visualization/boxes_and_whiskers.cpp
@@ -94,7 +94,7 @@ std::shared_ptr<Plot> turi::visualization::plot_boxes_and_whiskers(
   temp_sf[x_name] = x;
   temp_sf[y_name] = y;
 
-  bw.init(temp_sf);
+  bw.init(temp_sf, 5000000 /* batch_size */);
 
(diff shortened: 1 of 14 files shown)


znation:
@abhishekpratapa Good idea; I'll do that in another PR.

Looks like the commit that caused #1248 was https://github.com/apple/turicreate/commit/a9246c3197019f68037dc5ed3ccd6cd2ffcd27d1.

PR #1259, opened 2018-12-14
Update CoreML spec and tools

nickjong:
Synced to coremltools/mlmodel at master: https://github.com/apple/coremltools/commit/e6fb03be90f25a42ca8836207f7ea55007bfb9f0

(With some minor build fixes to be merged upstream)


nickjong:
@fareeha FYI

nickjong:
For reviewing convenience: the first commit here is the literal result of performing rsync, and then committing only the modified (existing) files and adding the new UtilsTest.cpp.

The second commit is build fixes which I'll put into a PR for coremltools.

The third commit re-applies some copyright header fixes, which coremltools should probably perform globally as well at some point.

nickjong:
Sigh, looks like this will need a few more changes for the linux build

Issue #907, opened 2018-07-26
Reading CSV: newlines in quoted text should be ignored


TobyRoseman:
Reading the following CSV will result in an error:
```
id,title
11,"this is fine"
12,this is also fine
13,"this is 
not fine!"
```
Also the title value for the last row will be just `this is`.

Expected result: no error and the final title value is `this is\nnot fine!`.

srikris:
The impact of this bug is pretty big for anyone doing text processing. Bumping up to P2

PR #1263, opened 2018-12-19
Support newlines in CSV Strings

ylow:
This is a long awaited feature. This PR implements support for newlines
appearing inside a string field in a CSV.

For instance, this allows the parsing of

```
user, review
a, "this is a good string"
b, "This is a
gooder string"
c, "yet another string"
```


TobyRoseman:
fixes #907

PR #1265, opened 2018-12-19
Non Maximum Suppression implemented for Object Detection  

fareeha:
Added NMS support for the new implementation of Object Detection. On the new implementation:

import turicreate as tc
data = tc.SFrame('breakfast-data.sframe')
m = tc.extensions.object_detector()
m.train(data, 'annotation', 'image', {'mlmodel_path':'darknet.mlmodel', 'max_iterations' : 1})

 support for NMS is added through export_to_coreml() and within neural_nets_model_exports.cpp, and can be used as follows:

- Using NMS with custom values:

m.export_to_coreml('od.mlmodel', {'include_non_maximum_suppression': 'True', 'iou_threshold': 0.55, 'confidence_threshold': 0.15})

- Using NMS with default values:

m.export_to_coreml('od2.mlmodel', {'include_non_maximum_suppression': 'True'})

If values are not provided, default values of iou_threshold (0.45) and confidence_threshold (0.25) are used.

- Not using NMS:

m.export_to_coreml('od.mlmodel', {'include_non_maximum_suppression': 'False'})

### DIFF
--- a/src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
+++ b/src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
@@ -9,9 +9,8 @@
 #include <logger/assertions.hpp>
 #include <unity/toolkits/coreml_export/mlmodel_include.hpp>
 
-#include <unity/toolkits/coreml_export/mldata_exporter.hpp> //
-//#include <unity/toolkits/coreml_export/mlmodel_wrapper.hpp> //
-#include <unity/toolkits/coreml_export/coreml_export_utils.hpp> //
+#include <unity/toolkits/coreml_export/mldata_exporter.hpp>
+#include <unity/toolkits/coreml_export/coreml_export_utils.hpp>
 
 using CoreML::Specification::ArrayFeatureType;
 using CoreML::Specification::DoubleFeatureType;
@@ -21,9 +20,6 @@ using CoreML::Specification::ModelDescription;
 using CoreML::Specification::NeuralNetworkLayer;
 using turi::coreml::MLModelWrapper;
 
-
-//This one...
-//Change here return type 
   
 namespace turi {
 
@@ -33,7 +29,7 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
     flex_dict user_defined_metadata, flex_list class_labels, std::map<std::string, flexible_type> options) {
 
   
-  CoreML::Specification::Model model_nn; //coreml_spec_model2
+  CoreML::Specification::Model model_nn;
 
   // Scale pixel values 0..255 to [0,1]
   NeuralNetworkLayer* first_layer =
@@ -96,7 +92,7 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
     output_coordinates_pb_feature->set_datatype(ArrayFeatureType::DOUBLE);
 
     // Set CoreML spec version.
-    model_nn.set_specificationversion(1); 
+    model_nn.set_specificationversion(1);
     auto model_wrapper = std::make_shared<MLModelWrapper>(
       std::make_shared<CoreML::Model>(model_nn));
 
@@ -112,7 +108,7 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
 
   // Write FeatureDescription for the raw confidence output.
   FeatureDescription* output_rconfidence_pb = model_desc->add_output();
-  output_rconfidence_pb->set_name("raw_confidence"); 
+  output_rconfidence_pb->set_name("raw_confidence");
   ArrayFeatureType* output_rconfidence_feature =
       output_rconfidence_pb->mutable_type()->mutable_multiarraytype();
   output_rconfidence_feature->add_shape(num_predictions);
@@ -127,9 +123,7 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
 
   // Write FeatureDescription for the coordinates output.
   FeatureDescription* output_rcoordinates_pb = model_desc->add_output();
-  output_rcoordinates_pb->set_name("raw_coordinates"); //raw_coordinates
-  //output_rcoordinates_pb->set_shortdescription(
-  //    "Boxes × [x, y, width, height] (relative to image size)");
+  output_rcoordinates_pb->set_name("raw_coordinates");
   ArrayFeatureType* output_rcoordinates_pb_feature =
       output_rcoordinates_pb->mutable_type()->mutable_multiarraytype();
   output_rcoordinates_pb_feature->add_shape(num_predictions);
@@ -143,10 +137,9 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
       ->add_sizeranges();
   rcoordinates_shape_pb2->set_lowerbound(4);
   rcoordinates_shape_pb2->set_upperbound(4);
-
   
   // Create Non Maximum Suppression model
-  CoreML::Specification::Model model_nms; //coreml_spec_model3
+  CoreML::Specification::Model model_nms;
   model_nms.set_specificationversion(3);
 
   ModelDescription* nms_desc = model_nms.mutable_description();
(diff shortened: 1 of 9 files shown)


src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -326,16 +340,21 @@ std::shared_ptr<MLModelWrapper> object_detector::export_to_coreml(
...
+      {"training_iterations", read_state<flex_int>("training_iterations")},
+      {"include_non_maximum_suppression", "False"},
+      {"feature", read_state<flex_string>("feature")},
+      {"annotations", read_state<flex_string>("annotations")},
+      {"classes", class_labels_str},
+      {"type", "object_detector"},
+    };
+
+  if (options["include_non_maximum_suppression"] == "True") {
+    user_defined_metadata.push_back(std::make_pair<std::string, std::string>("include_non_maximum_suppression", options["include_non_maximum_suppression"]));
```
nickjong: You should be able to streamline this as e.g.
`user_defined_metadata.emplace_back("include_non_maximum_suppression", options["include_non_maximum_suppression"])`

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -346,7 +365,8 @@ std::shared_ptr<MLModelWrapper> object_detector::export_to_coreml(
                                    GRID_SIZE * SPATIAL_REDUCTION,
                                    class_labels.size(),
                                    GRID_SIZE*GRID_SIZE * anchor_boxes().size(),
-                                   std::move(user_defined_metadata));
+                                   std::move(user_defined_metadata), class_labels,
+                                   options);
```
nickjong: We may as well use std::move for `class_labels` and `options`, the way we have for `user_defined_metadata`. Since these symbols are not being used anywhere after this in the function, it's safe to transfer ownership from them and leave those symbols in undefined states

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -301,19 +301,33 @@ std::unique_ptr<model_spec> object_detector::init_model(
...
+    confidence_str = "raw_confidence";
+    //Set default values if thresholds not provided.
+    if (options.find("iou_threshold") == options.end())
+      options["iou_threshold"] = 0.45;
+    if (options.find("confidence_threshold") == options.end())
+      options["confidence_threshold"] = 0.25;
+  }
+  else {
+    coordinates_str = "coordinates";
+    confidence_str = "confidence";
```
nickjong: Just to make this code a bit more concise, can we initialize these variables assuming no NMS and then override in the if statement above? Then we don't need an else case at all.

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -9,25 +9,31 @@
 #include <logger/assertions.hpp>
 #include <unity/toolkits/coreml_export/mlmodel_include.hpp>
 
+#include <unity/toolkits/coreml_export/mldata_exporter.hpp>
+#include <unity/toolkits/coreml_export/coreml_export_utils.hpp>
```
nickjong: What are these two includes for? I don't see any instances of the two symbols they declare, `turi::setup_pipeline_from_mldata` and `turi::add_metadata`

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -9,25 +9,31 @@
...
+
 using CoreML::Specification::ArrayFeatureType;
+using CoreML::Specification::DoubleFeatureType;
```
nickjong: I also don't see any usage of this symbol below

src/unity/toolkits/neural_net/model_spec.hpp
```diff
@@ -20,6 +20,8 @@
 namespace CoreML {
 namespace Specification {
 class NeuralNetwork;
+class NonMaximumSuppression;
+class Pipeline;
```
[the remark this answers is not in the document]
fareeha: Thank you for pointing out. This is an oversight.

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -37,65 +43,275 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
...
+  output_coordinates_pipeline_feature->set_datatype(ArrayFeatureType::DOUBLE);
+  auto *coordinates_shape_pipeline1 = output_coordinates_pipeline_feature->mutable_shaperange()->add_sizeranges();
+  coordinates_shape_pipeline1->set_upperbound(-1);
+  auto *coordinates_shape_pipeline2 = output_coordinates_pipeline_feature->mutable_shaperange()->add_sizeranges();
+  coordinates_shape_pipeline2->set_lowerbound(4);
+  coordinates_shape_pipeline2->set_upperbound(4);
+  
+  // Add NeuralNetwork model to pipeline
+  auto* model1 = model_pipeline.mutable_pipeline()->add_models();
+  *model1 = model_nn;
```
nickjong: This line copies around ~60 MB of data from `model_nn` into the pipeline, which we should make some effort to avoid. For this performance reason, and also for the sake of clarity, I would suggest a little reorganization of this function.

In particular, the protocol buffer APIs are designed to facilitate a top-down approach to filling out a protobuf value. Note that in the original code, we began by allocating the highest level Model instance, and then filling in everything inside using calls to `mutable_*` or `add_*` as appropriate.

Similarly, I suspect this will be simpler if in the NMS case, we follow this outline:
```
CoreML::Specification::Model model_pipeline;
auto* model_nn = model_pipeline->mutable_pipeline->add_models();
// Populate *model_nn.
auto* model_nms = model_pipeline->mutable_pileine->add_models();
// Populate *model_nms
```

I would also really recommend breaking this down into helper functions, to keep things organized. Offhand, one for populating `model_nn` (which can be shared with the non-NMS case), and perhaps some functions for reducing duplication in setting the model inputs and outputs (since Pipeline models explicitly duplicate these).

src/unity/toolkits/neural_net/model_spec.cpp
```diff
@@ -25,6 +25,7 @@ using CoreML::Specification::ConvolutionLayerParams;
 using CoreML::Specification::Model;
 using CoreML::Specification::NeuralNetwork;
 using CoreML::Specification::NeuralNetworkLayer;
+using CoreML::Specification::NonMaximumSuppression;
```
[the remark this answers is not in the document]
fareeha: Another oversight. Thank you for the catch.

src/unity/toolkits/object_detection/object_detector.cpp
fareeha: Good catch.

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
fareeha: That's a really great point, I've updated the code accordingly. Ty

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -301,19 +301,33 @@ std::unique_ptr<model_spec> object_detector::init_model(
...
+    std::string filename, std::map<std::string, flexible_type> options) {
 
   // Initialize the result with the learned layers from the cnn_module.
   model_spec yolo_nn_spec(nn_spec_->get_coreml_spec());
 
   // Add the layers that convert to intelligible predictions.
-  add_yolo(&yolo_nn_spec, "coordinates", "confidence", "conv8_fwd",
+  std::string coordinates_str;
+  std::string confidence_str;
+  if (options["include_non_maximum_suppression"]=="True") {
```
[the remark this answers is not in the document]
fareeha: Working now, ty!

src/unity/toolkits/object_detection/object_detector.hpp
```diff
@@ -69,7 +69,8 @@ class EXPORT object_detector: public ml_model_base {
   );
   // TODO: Addition training options: batch_size, max_iterations, etc.
 
-  REGISTER_CLASS_MEMBER_FUNCTION(object_detector::export_to_coreml, "filename");
+  REGISTER_CLASS_MEMBER_FUNCTION(object_detector::export_to_coreml, "filename",
+    "options");
```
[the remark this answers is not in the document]
fareeha: Done.

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -9,31 +9,86 @@
...
+constexpr char COORDINATES_STR[] = "Boxes × [x, y, width, height] (relative to image size)";
+constexpr char IOU_THRESHOLD_STR[] = "(optional) IOU Threshold override (default: 0.45)";
+constexpr char CONFIDENCE_THRESHOLD_STR[] = "(optional) Confidence Threshold override (default: 0.25)";
   
 namespace turi {
 
+  namespace {
+    void set_predictions_feature(FeatureDescription* feature_desc, std::string feature_name, 
+        size_t num_predictions, size_t num_classes, bool include_shape, bool include_bounds,
+        size_t bound, std::string short_desc) {
```
nickjong: When `bound` is used, isn't it always equal to `num_classes`?

For example, when num_predictions = 2535 and num_classes = 6, then we're describing a multi-array feature where the multi-array is described as either having the shape [2535, 6] or [x, 6] where x is unbounded.

Maybe `include_bounds` should just be something like `use_flexible_shape`

fareeha: Yes you're right, in the case of 'confidence' it is num_classes and in the case of 'coordinates' it is the number of coordinates i.e. 4. Based on this observation I've simplified the code by removing bounds and using number of classes or number of coordinates through num_classes. I've also changed include_bounds to use_flexible_shape.

PR #1266, opened 2018-12-19
More CSV Parsing Improvements

ylow:
- Allow escape_char to be optional: push a use_escape_char option all the
  way through all code.
 - Add csv parsing options true_values and false_values which accept a list of strings to interpret as the numeric value 1, and the numeric value 0 respectively.
 - pushed through na_values to the recursive parsers so that missing values inside of dict/lists are handled correctly.
 - Filled in the read_json parsing options to correctly map na_values = ["null"], true_values=["true"] and false_values=["false"]
 - Better parser diagnostics on parser failure. For instance:
```
 Unable to interpret ""hello"" as a integer
Parse failed at token ending at:
	2,"hello"^
Successfully parsed 1 tokens:
	0: 2
```

### DIFF
--- a/src/flexible_type/string_parser.hpp
+++ b/src/flexible_type/string_parser.hpp
@@ -17,21 +17,23 @@
 
 namespace parser_impl { 
 
-/*
+/**
  * \internal
  * The string parsing configuration.
  *
  */
 struct parser_config {
-  // If any of these character occurs outside of quoted string, 
-  // the string will be terminated
+  /// If any of these character occurs outside of quoted string, 
+  /// the string will be terminated
   std::string restrictions;
-  // If the delimiter string is seen anywhere outside of a quoted string,
-  // the string will be terminated.
+  /// If the delimiter string is seen anywhere outside of a quoted string,
+  /// the string will be terminated.
   std::string delimiter;
-  // The character to use for an escape character
+  /// Whether escape char should be used
+  bool use_escape_char = true;
+  /// The character to use for an escape character
   char escape_char = '\\';
-  /* whether double quotes inside of a quote are treated as a single quote.
+  /** whether double quotes inside of a quote are treated as a single quote.
    * i.e. """hello""" => \"hello\"
    */
   char double_quote = true;
@@ -136,7 +138,7 @@ struct string_parser
     }
     return true;
   }
-#define PUSH_CHAR(c) ret.add_char(c); escape_sequence = (c == config.escape_char);
+#define PUSH_CHAR(c) ret.add_char(c); escape_sequence = config.use_escape_char && (c == config.escape_char);
 
 // insert a character into the field buffer. resizing it if necessary
 
@@ -229,8 +231,9 @@ struct string_parser
       if (!quote_char) boost::algorithm::trim_right(final_str);
       else if (quote_char) {
         // if was quoted field, we unescape the contents
-        turi::unescape_string(final_str, config.escape_char,
-                                  quote_char, config.double_quote);
+        turi::unescape_string(final_str, config.use_escape_char, 
+                              config.escape_char,
+                              quote_char, config.double_quote);
       }
       attr = std::move(final_str);
       first = cur;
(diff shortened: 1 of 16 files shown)


ylow:
Do not merge. Fixing a CSV issue in #1263 when there are a lot of CPUs.

ylow:
ok. ready for review.

src/sframe/csv_line_tokenizer.cpp
```diff
@@ -199,6 +214,33 @@ size_t csv_line_tokenizer::tokenize_line(char* str, size_t len,
...
+    if (tokenizer_impl_fail_pos >= 0 && (size_t)tokenizer_impl_fail_pos <= len) {
+      strm << "Parse failed at token ending at: \n";
+      std::string line(str, len);
+      // colorize the line.
+      line.insert(tokenizer_impl_fail_pos, "\033[1m\033[31m^\033[00m");
+      if (line.length() > 256) {
+        // truncate around the fail pos
+        ssize_t start = std::max<ssize_t>(0, tokenizer_impl_fail_pos - 60);
+        ssize_t end = std::min<ssize_t>(len, tokenizer_impl_fail_pos + 60);
+        line = line.substr(start, end - start);
```
hoytak: Do we need to turn off the colorization?  (Also, I'm worried about error reporting in XCode; the log there doesn't support colorization.)

src/flexible_type/string_parser.hpp
```diff
@@ -229,10 +235,19 @@ struct string_parser
       if (!quote_char) boost::algorithm::trim_right(final_str);
       else if (quote_char) {
         // if was quoted field, we unescape the contents
-        turi::unescape_string(final_str, config.escape_char,
-                                  quote_char, config.double_quote);
+        turi::unescape_string(final_str, config.use_escape_char, 
+                              config.escape_char,
+                              quote_char, config.double_quote);
+      }
+      if (!config.na_val.empty() && config.na_val.count(final_str)) {
```
hoytak: Might be faster to just use a single hash map of key-to-value instead of 3 successive lookups in separate tables.

src/flexible_type/string_parser.hpp
```diff
@@ -229,10 +235,19 @@ struct string_parser
...
+        attr = turi::flexible_type(turi::flex_type_enum::UNDEFINED);
```
hoytak: Is there a possible reason to separate missing and float NA?  It seems that na_value, if specified, might not be interpreted as missing in a float column.

ylow: It should work correctly in a float column actually. 
```
a,b
1.0,2.0
2.0,3.0
```
```
tc.SFrame.read_csv('z.csv',na_values=['2.0'])
Data:
+------+------+
|  a   |  b   |
+------+------+
| 1.0  | None |
| None | 3.0  |
+------+------+
```
This is actually handled in the CSV tokenization phase.

Where this particular bit is used is to support Nones in recursive parses.
```
a,b
{'a':1},{'b':null}
{'b':null},{'b':2}
tc.SFrame.read_csv('z.csv',na_values=['null'])
```

Indeed in this case, if you do na_values=['1'] that will probably not handle correctly. But handling this correctly in the recursive parse is problematic since that does not fit in the grammar nicely.

src/sframe/csv_line_tokenizer.cpp
ylow: what do you recommend here?

ylow: Is there a flag I can check to optionally turn off colorization?

ylow: There seems to be definition in logger.hpp `#define COLOROUTPUT`. But it seems like it is forced to be set. Are you using this?

src/flexible_type/string_parser.hpp
hoytak: My question is more about whether we should have a separate na_values field, with goes to NAN, and a missing_values field, which goes to None.  Having na_values go to None seems like it could confuse people.

ylow: This was a design decision a long time ago that expands to all of SFrame. That NaN is a valid float value and should not be confused with None.

hoytak: Followup:  Did not realize that this behavior is unchanged from previous version; no reason to change.

PR #1280, opened 2019-01-02
Bug Fix: tokenization

### LINKED ISSUE #1268: NameError in text_analytics.tokenize
text_analytics.tokenize() creates SFrame with undefined 'sa' instead of passed text.

TobyRoseman:
Fixes #1268 
Also adds a unit test for tokenization. There were not any unit tests for tokenization.

### DIFF
--- a/src/unity/python/turicreate/test/test_text_analytics.py
+++ b/src/unity/python/turicreate/test/test_text_analytics.py
@@ -10,7 +10,7 @@ import math
 import unittest
 import warnings
 import turicreate as tc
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/test/test_text_analytics.py
```diff
@@ -41,6 +41,15 @@ def setUpClass(self):
 
         self.sframe_comparer = util.SFrameComparer()
 
+
+    def test_tokinze(self):
```
znation: `test_tokinze` should be `test_tokenize`?

PR #1293, opened 2019-01-08
WIP: Changes to allow building with clang on Linux

znation:
Some consolidated changes (built up over a year or so in a branch) to allow building with clang on Linux.

I think some further changes are needed to make sure we don't break existing builds.


src/unity/python/turicreate/CMakeLists.txt
```diff
@@ -73,4 +73,17 @@ elseif(LINUX)
...
     COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_SOURCE_DIR}/src/unity/python/turicreate/visualization/vg2svg" "${CMAKE_CURRENT_BINARY_DIR}"
   )
+
+  if(CLANG)
+    find_library(LIBCXX c++)
+    find_library(LIBCXXABI c++abi)
+    find_library(LIBUNWIND unwind)
+    add_custom_target(
+      copy_libcxx_deps ALL
+      COMMAND cp ${LIBCXX}.1.0 ${CMAKE_CURRENT_BINARY_DIR}/libc++.so.1
```
nickjong: Is this harmless in the APPLE case?

znation: Probably not. That's why this is still a WIP PR: I want to try building on all our supported configurations currently and make sure it doesn't break anything. I think this probably does break on APPLE -- good catch!

hoytak: What are the holdups here?  How can we get this in?

znation: Ok, this line actually should be harmless in the APPLE case since it's nested in an `elseif(LINUX)`. I'll test locally to make sure.

znation:
TODO: incorporate a Travis build matrix so that we are running and testing with both gcc and clang.

znation:
These changes are currently causing the build to fail for iOS dylibs:
```
[ 12%] Performing install step for 'ex_boost'
Performing configuration checks

    - default address-model    : 64-bit (cached)
    - default architecture     : arm (cached)
    - symlinks supported       : yes (cached)
    - has_icu builds           : no  (cached)
ld: unknown option: --start-group
clang: error: linker command failed with exit code 1 (use -v to see invocation)
...failed updating 1 target...
make[2]: *** [../deps/build/boost/src/ex_boost-stamp/ex_boost-install] Error 1
make[1]: *** [CMakeFiles/ex_boost.dir/all] Error 2
make: *** [all] Error 2
```

hoytak:
@znation With your docker changes, does this need to be open anymore?

znation:
@hoytak Nope, we are now building on clang 8 in Docker -- so these changes are no longer needed for Clang on Linux. There are some additional changes in here to pull libc++ and other dylibs into the built wheel, but those changes are fairly broken at the moment. Closing.

PR #1296, opened 2019-01-09
WIP: Make the Vega JSON templating more flexible

znation:
* Uses named instead of position arguments.
* Takes a dictionary of placeholder -> value to apply the template.
* This will allow us to swap in different templated Vega specs for any
  plot type, and the named format parameters can be applied optionally,
  multiple times, and in any order.
* Drops the boost::format dependency (incidentally - since boost::format
  can only handle positional numbered arguments.


znation:
@hoytak @abhishekpratapa Please review, thanks!

znation:
Abandoning in favor of #1308.

PR #1300, opened 2019-01-10
WIP: infrastructure updates

hoytak:
Relaying commit from Joe Zimmerman:

    Infrastructure updates:

   We relocate a few other utility functions (i.e., hash-related methods
          from util to flexible_type, and filesystem-related methods from util to
          fileio) to avoid the need to merge additional modules into util.

    Support building without extra optimization flags

        - This allows us to trade off some running time for build time during
          development.

    Provide annotated stack traces via LLVM/libunwind

        - When present, we can use the LLVM and libunwind libraries to provide
          full source file and line information for debugging stack traces.

    Add several basic utility functions and type aliases in src/util

    Debug-enabled version of turi::in_parallel

        - In order to facilitate debugging of parallel code, we support two
          additional modes of running lambda functions in parallel: one in which all
          iterations are run sequentially, and one in which a single iteration is
          run on the main thread (with the others in parallel as usual).

    Filesystem utility functions

    String-to-integer conversion function

    Implement Boost assertion handler

        - This causes Boost assertion failures (e.g., dereference of empty optional
          type) to trigger our own error handling machinery instead of the default,
          when the corresponding macro (BOOST_ENABLE_ASSERT_HANDLER) is defined.


src/logger/assertions.hpp
```diff
@@ -382,5 +385,25 @@ extern void __print_back_trace();
 #undef ERROR      // may conflict with ERROR macro on windows
 #endif
 
+#define BOOST_ENABLE_ASSERT_HANDLER
+
+namespace boost {
+  inline void assertion_failed(
```
znation: Is this not defined in the version of Boost we're using? Or are we relying on this header being included before Boost headers so that we take this version of the code?

src/util/basic_types.hpp
```diff
@@ -41,9 +95,103 @@ template<> struct truncate_check_impl<int64_t, size_t, 0> {
...
+  struct X; \
+  using X ## _p = shared_ptr<X>; \
+  bool operator==(shared_ptr<X>, shared_ptr<X>); \
+
 // For integer values, returns ceil(n / m);
 template<typename T> T ceil_divide(T n, T m) {
   return (n + (m-1)) / m;
 }
 
+using i8 = int8_t;
```
znation: Not sure I like these typedefs; now we have two ways of referring to each of these types within our code. Let's stick to one or the other throughout? (I'd be fine with either dropping the typedefs, or a second PR that converts all uses throughout turicreate of `[s]int[n]_t` to the typedefs. But let's have a plan for how to reconcile the two uses.)

src/util/llvm_lib.cpp
```diff
@@ -0,0 +1,28 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: I would consider adding a new directory `src/stack_trace` or something like that, and moving the llvm_lib and stack_trace stuff there. It seems a bit large to add into the existing utils.

src/logger/assertions.hpp
znation: Ah, that makes sense. Let's move the flag to a global compile flag, in case there are ordering issues in headers being included.

hoytak: Turns out this causes other issues.  I misunderstood; it's actually fine to have this here.

ylow:
@hoytak The changes in fs_util duplicate existing code relativize_path --> make_relative_path, and also do not work with our network filesystems like s3:// hdfs:// . It is awkward that the existing create_directory works with the network filesystems, but create_directories here does not.

The str_util methods like starts_with, ends_with etc duplicate functionality in boost strings which we already do use and are unnecessary.

ylow:
Could consider using boost::format as opposed to C format strings + C style arguments.

PR #1301, opened 2019-01-10
WIP: Indexed sframe

hoytak:
SFrame random access backend:

    Initial prototype of SFrame random access backend

        - This implementation provides an alternate binary representation of
          (strongly-typed) SArray and SFrame values, which supports fast random
          access and indirect references, as well as the corresponding query
          execution and optimization machinery for the usual relational operations
          (join, groupby, filter, etc.). This also enables us to implement an
          in-memory hash-based index for relatively small SFrames, which speeds up
          repeated relational queries over the same database.

    Implement time_spec struct and corresponding operations

        - This enables simple speed benchmarking via the now() function.

    Minor fix for capi/test_utils

        - Previously, the gl_sframe equality test in capi/test_utils was not
          checking that the two SFrames have the same number of rows. This could
          cause the equality check to pass incorrectly when one SFrame is actually a
          proper subset of the other.

    Test for SFrame random access

### DIFF
--- a/src/unity/toolkits/util/sframe_test_util.hpp
+++ b/src/unity/toolkits/util/sframe_test_util.hpp
@@ -1,0 +1,65 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+#ifndef TURI_UNITY_SFRAME_TEST_UTIL_H_
+#define TURI_UNITY_SFRAME_TEST_UTIL_H_
+
+#include <boost/test/unit_test.hpp>
+#include <unity/lib/gl_sarray.hpp>
+#include <unity/lib/gl_sframe.hpp>
+#include <util/test_macros.hpp>
+
+#include <string>
+
+__attribute__((__unused__))
+inline bool check_equality_gl_sframe(
+  turi::gl_sframe sf_gl, turi::gl_sframe ref_gl, bool check_row_order=true) {
+
+  size_t num_columns_sf = sf_gl.num_columns();
+  size_t num_columns_ref = ref_gl.num_columns();
+
+  TS_ASSERT_EQUALS(num_columns_sf, num_columns_ref);
+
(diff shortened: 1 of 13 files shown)


src/unity/toolkits/util/sframe_test_util.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <boost/test/unit_test.hpp>
+#include <unity/lib/gl_sarray.hpp>
+#include <unity/lib/gl_sframe.hpp>
+#include <util/test_macros.hpp>
+
+#include <string>
+
+__attribute__((__unused__))
+inline bool check_equality_gl_sframe(
```
znation: Do we not have this already somewhere? I feel like I've seen this exact code before.

src/util/basic_types.hpp
```diff
@@ -41,9 +95,103 @@ template<> struct truncate_check_impl<int64_t, size_t, 0> {
...
+
+template<typename T, typename U> std::ostream& operator<<(
+  std::ostream& os, const std::pair<T, U>& x) {
+
+  os << "(" << x.first << ", " << x.second << ")";
+  return os;
+}
+
+namespace std {
+  template<typename T, typename U> struct hash<pair<T, U>> {
```
znation: Not sure it's valid in C++ to template-specialize hash for built in types. See https://stackoverflow.com/a/20602159. Does the C++ standard not define hash<pair<T,U>> for us, where T and U are hashable themselves?

src/util/stl_common.hpp
```diff
@@ -0,0 +1,58 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#ifndef TURI_STL_COMMON_H_
+#define TURI_STL_COMMON_H_
+
+#include <algorithm>
```
znation: Combining all the STL headers (and especially adding `using` statements for the types within) doesn't really fit the code style of existing code. I think we should prefer explicit STL includes at each point of use (at least for now, perhaps until we have PCH turned on in the build), and explicit references to `std::`.

hoytak:
Needs different approach before going in.

Issue #1056, opened 2018-09-02
VisionFeaturePrint - Screen or Scene? 


hipwelljo:
Everywhere in the Create ML documentation, the feature extractor is termed the **scene** print feature extractor, but everywhere in the Turi Create documentation it is termed **screen** print. Is this interchangeable, is one of them incorrect, or are they actually different extractor types?

[Create ML](https://developer.apple.com/documentation/createml/mlimageclassifier/featureextractortype/sceneprint_revision)
[Turi Create](https://apple.github.io/turicreate/docs/userguide/image_classifier/how-it-works.html)

srikris:
This is a typo. It should be scene. We will fix it in the next release along with all the documentation.

PR #1306, opened 2019-01-10
Fix for https://github.com/apple/turicreate/issues/1056

hoytak:


TobyRoseman:
@hoytak - good caught also making these changes for image similarity. However it doesn't look like the two issues I mentioned previously have been addressed. Have you tested those use cases?

hoytak:
> @hoytak - good caught also making these changes for image similarity. However it doesn't look like the two issues I mentioned previously have been addressed. Have you tested those use cases?

Hey @TobyRoseman, 

The issues above were addressed by intercepting "VisionFeaturePrint_Screen" on model load and on model creation, then changing it to the correct VisionFeaturePrint_Scene.  I don't have old models hanging around, but I'll try to check it.

src/unity/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -124,7 +124,14 @@ def create(dataset, target, feature = None, model = 'resnet-50',
     # Check model parameter
     allowed_models = list(_pre_trained_models.MODELS.keys())
     if _mac_ver() >= (10,14):
-        allowed_models.append('VisionFeaturePrint_Screen')
+        allowed_models.append('VisionFeaturePrint_Scene')
+
+        # Also, to make sure existing code doesn't break, replace incorrect name
+        # with the correct name version
+        if model == "VisionFeaturePrint_Screen":
+            print "WARNING: Correct spelling of model name is VisionFeaturePrint_Scene."
```
TobyRoseman: This will fail in python 3. Need parentheses. 

Something about this argument value being removed in a future release would be good too.

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -105,7 +105,14 @@ def create(dataset, label = None, feature = None, model = 'resnet-50', verbose =
     # Check parameters
     allowed_models = list(_pre_trained_models.MODELS.keys())
     if _mac_ver() >= (10,14):
-        allowed_models.append('VisionFeaturePrint_Screen')
+        allowed_models.append('VisionFeaturePrint_Scene')
+
+        # Also, to make sure existing code doesn't break, replace incorrect name
+        # with the correct name version
+        if model == "VisionFeaturePrint_Screen":
+            print "WARNING: Correct spelling of model name is VisionFeaturePrint_Scene."
```
TobyRoseman: Same point here as I made in the image classifier: this will not work in Python 3 and we should mention we will remove this argument value in a later release.

src/unity/python/turicreate/toolkits/image_classifier/image_classifier.py
hoytak: Great catch!

src/unity/python/turicreate/toolkits/image_similarity/image_similarity.py
hoytak: Great catch!

src/unity/python/turicreate/test/test_image_similarity.py
```diff
@@ -238,9 +238,16 @@ def setUpClass(self):
...
 
-@unittest.skipIf(_mac_ver() < (10,14), 'VisionFeaturePrint_Screen only supported on macOS 10.14+')
-class ImageSimilarityVisionFeaturePrintScreenTest(ImageSimilarityTest):
+@unittest.skipIf(_mac_ver() < (10,14), 'VisionFeaturePrint_Scene only supported on macOS 10.14+')
+class ImageSimilarityVisionFeaturePrintSceneTest(ImageSimilarityTest):
     @classmethod
     def setUpClass(self):
-        super(ImageSimilarityVisionFeaturePrintScreenTest, self).setUpClass(model='VisionFeaturePrint_Screen',
-                                                                            input_image_shape=(3, 299, 299))
+        super(ImageSimilarityVisionFeaturePrintSceneTest, self).setUpClass(model='VisionFeaturePrint_Scene',
```
TobyRoseman: There is a syntax error at the end of this line. That's why the unit tests are failing.

FYI - I've refactored theses unit tests, so they should run pretty fast, if you want to run them locally.

PR #1324, opened 2019-01-15
Fix bad merge in visualization xcodeproj

znation:


znation:
Fixes the syntax error in https://github.com/apple/turicreate/commit/15b14c06d3ffda9ca972c23a710b6a264c45fc90#diff-742897faddeb6373492d4e17370345c8R285

PR #1330, opened 2019-01-16
Fix Travis CI failure on assert(HAS_NUMPY)

znation:
Initial commit: Don't put import numpy behind a try/catch.
Since we're immediately going to fail anyway if numpy is missing, let's just
import it whether it'll fail or not, and let it fail with a proper stack
trace if it's going to. That should at least give us more information in CI.

### DIFF
--- a/src/unity/python/turicreate/cython/cy_flexible_type.pyx
+++ b/src/unity/python/turicreate/cython/cy_flexible_type.pyx
@@ -287,14 +287,10 @@ timedelta_type = datetime.timedelta
 cdef object _image_type
 cdef bint have_imagetype
 
-cdef bint HAS_NUMPY = False
+cdef bint HAS_NUMPY = True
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/cython/cy_flexible_type.pyx
```diff
@@ -287,14 +287,10 @@ timedelta_type = datetime.timedelta
 cdef object _image_type
 cdef bint have_imagetype
 
-cdef bint HAS_NUMPY = False
+cdef bint HAS_NUMPY = True
```
TobyRoseman: Should we maybe just remove this variable entirely?

znation: Probably. That can come in a later change (feel free to do some cleanup here if you'd like). I just merged this to unblock the PR builds now that it's passing.

PR #1335, opened 2019-01-18
Image Classification Evaluation

abhishekpratapa:
- added support for image classification evaluation in TuriCreate


src/unity/toolkits/supervised_learning/supervised_learning.hpp
```diff
@@ -280,6 +280,29 @@ class EXPORT supervised_learning_model_base : public ml_model_base {
...
+   */
+  virtual std::map<std::string, variant_type> evaluate_with_predictions(const ml_data&
+                test_data, const std::string& evaluation_type="");
+
+  /**
+   * Same as evaluate_with_predictions(ml_data), but take SFrame as input.
+   */
+  virtual std::map<std::string, variant_type> evaluate_with_predictions(const sframe& X,
+                const sframe &y, const std::string& evaluation_type="") {
+    ml_data data = construct_ml_data_using_current_metadata(X, y);
```
znation: Implementation for non-template methods should go in the .cpp file.

src/unity/toolkits/supervised_learning/supervised_learning.hpp
```diff
@@ -670,6 +693,15 @@ class EXPORT supervised_learning_model_base : public ml_model_base {
   variant_map_type api_evaluate(
       gl_sframe data, std::string missing_value_action, std::string metric);
 
+  /**
+   *  API interface through the unity server.
+   *
+   *  Evaluate the model and retun with predictions
+   */
+  // TODO: This function should be const
+  variant_map_type api_evaluate_with_predictions(
```
znation: I'm a bit confused as to why we have both `evaluate_with_predictions` and `api_evaluate_with_predictions` (in addition to being confused about why we have both in addition to `evaluate`). Which are intended to be used for which use cases? The API interfaces only seem subtly different.

src/visualization/Turi Create Visualization/src/Pipe.swift
```diff
@@ -98,6 +117,10 @@ class Pipe {
             if let accordion_spec = json["accordion_spec"] as? [String: Any] {
                 self.graph_data.add_accordion(accordion_spec: accordion_spec)
             }
+            
+            if let evaluation_spec = json["evaluation_spec"] as? [String: Any] {
```
znation: This seems duplicated with the code above...?

src/visualization/Turi Create Visualization/src/Pipe.swift
```diff
@@ -69,6 +69,21 @@ class Pipe {
...
+        print("{\"method\":\"get_rows_eval\", \"start\":" + String(start) + ", \"length\": " + String(length) + ", \"row_type\": \"" + row_type + "\", \"mat_type\": \"" + mat_type + "\", \"cells\": " + cells + " }");
+        fflush(__stdoutp)
+    }
+    
+    public func writeIncorrect(label: String){
+        print("{\"method\":\"get_incorrects\", \"label\":\""+label+"\"}");
+        fflush(__stdoutp)
+    }
+    
+    public func getCorrect(){
```
znation: Should be called `writeCorrect` to be consistent with the methods above.

src/visualization/Turi Create Visualization/src/user_interface/package.json
```diff
@@ -1,32 +1,40 @@
...
+  },
+  "devDependencies": {
+    "babel-cli": "^6.26.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.13",
+    "babel-preset-es2017": "^6.24.1",
+    "react-scripts": "2.1.3",
+    "tslib": "^1.9.3"
+  },
+  "browserslist": [
```
znation: What is this section for?

src/visualization/Turi Create Visualization/src/user_interface/src/elements/Explore/Evaluation/Components/TCEvaluationConfusionTable/TCEvaluationConfusionHeader/TCEvaluationConfusionHeaderCells/index.js
```diff
@@ -0,0 +1,69 @@
+import React, { Component } from 'react';
+import './index.scss';
+
+import down from './assets/down.svg';
+import caret_down from './assets/caret-down.svg';
+
+class TCEvaluationConfusionHeaderCell extends Component {
+
+  renderCarret = () => {
```
znation: Spelling: `Caret` (one t)

src/unity/python/turicreate/toolkits/_supervised_learning.py
```diff
@@ -158,6 +158,49 @@ def evaluate(self, dataset, metric="auto",
...
+            - 'auto': Choose a model dependent missing value policy.
+            - 'impute': Proceed with evaluation by filling in the missing
+                        values with the mean of the training data. Missing
+                        values are also imputed if an entire column of data is
+                        missing during evaluation.
+            - 'none': Treat missing value as is. Model must be able to handle missing value.
+            - 'error' : Do not proceed with prediction and terminate with
+                        an error message.
+
+        options : dict
```
TobyRoseman: It doesn't look like `options` or `kwargs` is getting used.

src/unity/python/turicreate/toolkits/image_classifier/_evaluation.py
```diff
@@ -0,0 +1,200 @@
+from __future__ import print_function as _
```
TobyRoseman: This should have the standard copyright header.

src/unity/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -545,12 +546,160 @@ def evaluate(self, dataset, metric='auto', verbose=True, batch_size=64):
...
-        return self.classifier.evaluate(extracted_features, metric = metric)
+        
+        metrics = self.classifier.evaluate_with_predictions(extracted_features, metric = metric)
+        predictions = metrics["predictions"]["probs"]
+        state = self.__proxy__.get_state()
+        labels = state["classes"]
+
+
+        def entropy(probs):
+            return reduce(lambda x, y: x + (y*math.log(1/y, 2) if y > 0 else 0) , probs, 0) / math.log(len(probs),2)
```
TobyRoseman: This will not work in Python 3. `reduce` been moved to `functools`.

src/visualization/Turi Create Visualization/src/user_interface/package.json
abhishekpratapa: In the new version of `create-react-app` by default this gets added to the end of the section. This list shows what browsers that this React app will run on. Without it a warning pops up. However, if you feel it's unnecessary I can remove it.

src/unity/python/turicreate/toolkits/image_classifier/_evaluation.py
```diff
@@ -0,0 +1,200 @@
...
+from __future__ import absolute_import as _
+from ...visualization import _get_client_app_path
+
+import subprocess as __subprocess
+import thread as _thread
+import json as _json
+import base64 as _base64
+import sys as _sys
+import math as _math
+from UserDict import UserDict as _UserDict
```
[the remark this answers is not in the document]
abhishekpratapa: Yup! did so

src/unity/python/turicreate/toolkits/_supervised_learning.py
```diff
@@ -158,6 +158,49 @@ def evaluate(self, dataset, metric="auto",
             dataset, missing_value_action, metric);
         return results
 
+    def evaluate_with_predictions(self, dataset, metric="auto",
```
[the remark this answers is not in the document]
abhishekpratapa: In the supervised learning class the two diffs between `evaluate_with_predictions` and `evaluate` are that evaluate with predictions returns `prob_vectors` while `evaluate` does not. I'll refactor the code to see if it works, just by adding a `prob_vectors` to the `evaluate` method and taking out `evaluate_with_predictions`.

src/unity/toolkits/supervised_learning/supervised_learning.hpp
abhishekpratapa: It's the wrapper around `evaluate_with_predictions` to use the `variant_type` api in Python. This probably should be refactored to use the new function registration mechanism instead of variant type.

abhishekpratapa: Notes to self:
- Removed `api_evaluate_with_predictions` into `evaluate_with_predictions`
- Refactoring `evaluate_with_predictions` into `evaluate`

src/unity/python/turicreate/toolkits/_supervised_learning.py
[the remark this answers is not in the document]
abhishekpratapa: refactored

src/unity/toolkits/supervised_learning/supervised_learning.hpp
abhishekpratapa: refactored

src/unity/python/turicreate/toolkits/image_classifier/_evaluation.py
```diff
@@ -0,0 +1,207 @@
...
+from ...visualization import _get_client_app_path
+
+import subprocess as __subprocess
+import thread as _thread
+import json as _json
+import base64 as _base64
+import sys as _sys
+import math as _math
+
+class Evaluation(dict):
```
znation: I think we had talked about the Evaluation object being returned from all models (not just image classifier). Since we are adding a new API method (explore?) we should make all models return an Evaluation, and have explore throw NotImplementedException in the case of non-image-classifiers (for now).

If that's too major of a refactor, it can be done after this PR goes in. Please either address it in this PR, or open a GitHub issue to reconcile the API return values from evaluate for other models so we don't lose track.

src/unity/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -545,12 +551,163 @@ def evaluate(self, dataset, metric='auto', verbose=True, batch_size=64):
...
-        return self.classifier.evaluate(extracted_features, metric = metric)
+        
+        metrics = self.classifier.evaluate(extracted_features, metric = metric)
+        predictions = metrics["predictions"]["probs"]
+        state = self.__proxy__.get_state()
+        labels = state["classes"]
+
+
+        def entropy(probs):
+            if _sys.version_info >= (3, 0):
```
znation: You can use [six](https://pythonhosted.org/six/index.html?highlight=reduce#module-six.moves) to get a version-independent `reduce`. We already have a dependency on it. I think you would do:

```
from six.moves import reduce
```

And it should work on both Python versions (no need for functools or built-in).

src/unity/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -545,12 +551,163 @@ def evaluate(self, dataset, metric='auto', verbose=True, batch_size=64):
...
+                return _reduce(lambda x, y: x + (y*math.log(1/y, 2) if y > 0 else 0) , probs, 0) / math.log(len(probs),2)
+            else:
+                return reduce(lambda x, y: x + (y*math.log(1/y, 2) if y > 0 else 0) , probs, 0) / math.log(len(probs),2)
+
+        def confidence(probs):
+            return max(probs)
+
+        def relative_confidence(probs):
+            sorted_probs = sorted(probs, reverse=True)
```
znation: Sorting is a relatively expensive operation. Instead, just

```
lp = len(probs)
return sorted_probs[lp-1] - sorted_probs[lp-2]
```

src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -772,12 +772,33 @@ std::map<std::string, variant_type> supervised_learning_model_base::evaluate(
   for(size_t i=0; i < evaluators.size(); i++){
     evaluators[i]->init(n_threads);
   }
+  
+  std::shared_ptr<sarray<flexible_type>> prob_vectors (new sarray<flexible_type>);
```
znation: Use `std::make_shared` to get a new shared pointer (avoiding `new`).

src/unity/python/turicreate/toolkits/image_classifier/_evaluation.py
abhishekpratapa: issue #1352 need to resolve.

src/visualization/Turi Create Visualization/src/user_interface/src/elements/Explore/Evaluation/Components/TCEvaluationConfusionTable/TCEvaluationConfusionRow/TCEvaluationConfusionCell/index.js
```diff
@@ -0,0 +1,73 @@
...
+            return(
+              <TCEvaluationImageCells value={data}/>
+            )
+          })}
+        </div>
+      );
+    }else if(this.props.type == "percent"){
+      return (
+        <div className="TCEvaluationConfusionCell">
+
```
kanitw: Why is the content in the percent empty?

src/visualization/Turi Create Visualization/src/user_interface/src/elements/Explore/Evaluation/Components/TCEvaluationConfusionTable/index.js
```diff
@@ -0,0 +1,96 @@
...
+  filterData = (element) => {
+    if(this.props.filter_confusion != null){
+      return element.actual == this.props.filter_confusion;
+    }else{
+      return true;
+    }
+  }
+
+  sortData = (a, b) => {
+    if(this.props.sort_by_confusions == "actual" && !this.props.sort_direction_confusions){
```
kanitw: It looks like this function can be simplified into:

```
const {sort_by_confusions, sort_direction_confusions} = this.props;
if (sort_by_confusions) {
  if (sort_direction_confusions) {
	return a[sort_by_confusions] < b[sort_by_confusions] ? -1 : 1;
  } else {
    return a[sort_by_confusions] > b[sort_by_confusions] ? -1 : 1;
  }
}
```

I also wonder if the currently code handles ties consistently.  Shouldn't equal return zero?

znation:
@abhishekpratapa It looks like you're getting a test failure:

```
self = <turicreate.test.test_logistic_classifier.LogisticRegressionClassifierModelTest_binary_classification_integer_target testMethod=test__list_fields>
msg = "Items in the first set but not the second:\n'training_predictions'"
```

However, I think the test should simply be updated to accommodate the new field. Seems that training_predictions is added intentionally in this PR, and shouldn't break existing code. What we should consider is:

* What will a serialized model, saved prior to this PR, return when this field is called?
* What will a serialized model, saved after this PR, return when this field is called? (i.e., are we persisting training predictions as part of the model artifact?) If we are persisting to disk, does this massively inflate the model size?
* In either case, if we don't get valid training predictions, is our new evaluation UX broken without it? What should the UX be in that case?

src/unity/python/turicreate/test/test_logistic_classifier.py
```diff
@@ -355,7 +355,7 @@ def test__list_fields(self):
         """
         model = self.model
         fields = model._list_fields()
-        self.assertEqual(set(fields), set(self.fields_ans))
+        self.assertTrue(any(elem in set(fields) for elem in set(self.fields_ans)))
```
znation: I'm not sure this is the right fix. With this change, the behavior will only fail the test if none of the fields listed are in the expected set (if even one is, it seems it will pass, if I'm reading it correctly).

Instead, I think we should use the set methods to test for a subset. If `fields_ans` is a subset of (or same set as) `fields`, the test should pass. See the [docs](https://docs.python.org/2/library/sets.html#set-objects) for subset methods.

znation: We should also make sure we aren't breaking compatibility by adding a new field (which is ostensibly what this test is supposed to be testing):

* Does loading an old model (saved before this field existed) still work and/or give expected results?
* Does saving a new model (after this field exists) still work when you load it in an older version of Turi Create?

PR #1339, opened 2019-01-18
Reduce (C-API) object_detector batch size to 16 for 2GB GPUs

nickjong:
The first commit here does some refactoring, defining a single abstraction for the platform dependencies and hardware resources used by a neural_net toolkit. The second commit plumbs through the memory info from Metal and uses that for choosing a batch size (when not specified by the user).

@tjia1818 FYI


nickjong:
And yes, this is really just a port of #806 to the new version of the toolkit, sigh

Issue #1279, opened 2019-01-02
Segmentation Fault: 11, Stack trace returned 0 entries


varunlakshmanan:
I'm following the official documentation (https://apple.github.io/turicreate/docs/userguide/image_classifier/), and used the first code sample that was displayed. When I run that Python script on my dataset of ~200 images with 6 different categories, I'm getting this error:

```
Segmentation fault: 11

Stack trace returned 0 entries:

Segmentation fault: 11


Segmentation fault: 11


Segmentation fault: 11

Stack trace returned 0 entries:
Stack trace returned 0 entries:
Stack trace returned 0 entries:
```

What is causing this and how can I fix it?

varunlakshmanan:
My mxnet version is 1.1.0.post0.

TobyRoseman:
@varunlakshmanan - Please give us some more details:

1 - What operating system are you using?
2 - What version of Python are you using?
3 - What dataset are you using?

varunlakshmanan:
@TobyRoseman I'm using MacOS 10.13.6 (High Sierra). I'm using Python 2.7.15, and the dataset is a custom dataset of images of different types of fingernails (all of the images are below 256x256 in dimensions). I looked online for help solving this problem, and I altered the code to this:

```
import turicreate as tc
import os

# 1. Load images
data = tc.image_analysis.load_images('/Users/fpladminPGD/Desktop/Nail_images', with_path=True)

# 2. From the path-name, create a label column
data["label"] = data["path"].apply(lambda path: os.path.basename(os.path.split(path)[0]))

# 3. Make a train-test split
train_data, test_data = data.random_split(0.8)

# print "Hello" <- works
# 4. Create the model
model = tc.image_classifier.create(data, target="label") # <- When model is created using train_data (recommended), segfault. When data is used, lambda error.

# 5. Save predictions to an SArray
predictions = model.predict(test_data)

# 6. Evaluate the model and save the results into a dictionary
metrics = model.evaluate(test_data)
print(metrics['accuracy'])

# 7. Save the model for later use in Turi Create
model.save('mymodel.model')

# 8. Export for use in Core ML
model.export_coreml('MyCustomImageClassifier.mlmodel')
```

Running that code gives me this output: 

```
Analyzing and extracting image features.
+------------------+--------------+------------------+
| Images Processed | Elapsed Time | Percent Complete |
+------------------+--------------+------------------+
libpng warning: IDAT: Too much image data

Segmentation fault: 11

[...]

PR #1344, opened 2019-01-22
Handle log and error for unsupported colorspaces

### LINKED ISSUE #1305: Unsupported Colorspace Format Error
This is somewhat related to #1264, but I think it's different enough to warrant its own issue.

I'm using a custom data set, so I can't share it here, but there were 3 images in my data set in the wrong colorspace format, and TC properly warned me about them and told me which images were the offending ones, as expected. After deleting them, a second issue occurs:
```
  File "make_predictions.py", line 14, in <module>
    sf['image'] = tc.image_analysis.resize(sf['image'], 150, 200)
  File "/usr/local/lib/python2.7/site-packages/turicreate/toolkits/image_analysis/image_analysis.py", line 159, in resize
    return image.apply(lambda x: _extensions.resize_image(x, width, height, channels, decode, resample_method))
  File "/usr/local/lib/python2.7/site-packages/turicreate/data_structures/sarray.py", line 1894, in apply
    dryrun = [fn(i) for i in self.head(100) if i is not None]
  File "/usr/local/lib/python2.7/site-packages/turicreate/toolkits/image_analysis/image_analysis.py", line 159, in <lambda>
    return image.apply(lambda x: _extensions.resize_image(x, width, height, channels, decode, resample_method))
  File "/usr/local/lib/python2.7/site-packages/turicreate/extensions.py", line 171, in <lambda>
    return lambda *args, **kwargs: _run_toolkit_function(fn, arguments, args, kwargs)
  File "/usr/local/lib/python2.7/site-packages/turicreate/extensions.py", line 160, in _run_toolkit_function
    raise _ToolkitError(ret[1])
turicreate.toolkits._main.ToolkitError: Unsupported colorspace format. Currently, only RGB and Grayscale are supported.
```

This is fine, and I understand that only RGB and Grayscale formats are supported, but there's no additional information here on which images are the offenders. I was able to solve this by using the `channels` attribute on `tc.Image` with a binary filter on my SFrame:

```python
sf = sf[sf['image'].apply(lambda img: img.channels <= 3)]
```

I still don't think this is the best solution because formats like YUV and HSV will still "pass" this test. Is there any way to get additional debugging information that tells us which images are causing trouble?

fareeha:
This PR fixes the issues emerging from changes made in PR #1212 'Handle resizing of channels of input images'.

Change allows unsupported colorspaces to be detected while loading images. URL for the offending image is now logged with the error message. 

Fixes #1305 and #1279

### DIFF
--- a/src/image/jpeg_io.cpp
+++ b/src/image/jpeg_io.cpp
@@ -36,7 +36,7 @@ jpeg_error_exit (j_common_ptr cinfo)
 
 void parse_jpeg(const char* data, size_t length,
                 size_t& width, size_t& height,
-                size_t& channels, const std::string& url) {
+                size_t& channels) {
   struct jpeg_decompress_struct cinfo;
   struct jpeg_error_mgr jerr;
   memset(&cinfo, 0, sizeof(cinfo));
--- a/src/image/jpeg_io.cpp
+++ b/src/image/jpeg_io.cpp
@@ -36,7 +36,7 @@ jpeg_error_exit (j_common_ptr cinfo)
 
 void parse_jpeg(const char* data, size_t length,
                 size_t& width, size_t& height,
-                size_t& channels) {
+                size_t& channels, const std::string& url) {
   struct jpeg_decompress_struct cinfo;
   struct jpeg_error_mgr jerr;
   memset(&cinfo, 0, sizeof(cinfo));
@@ -49,6 +49,10 @@ void parse_jpeg(const char* data, size_t length,
 
     jpeg_mem_src(&cinfo, (unsigned char*)data, length); // Specify data source for decompression
     jpeg_read_header(&cinfo, TRUE); // Read file header, set default decompression parameters
+
+    if (cinfo.out_color_space != JCS_GRAYSCALE && cinfo.out_color_space != JCS_RGB) {
+      log_and_throw(std::string("Unsupported colorspace format. Currently, only RGB and Grayscale are supported."));
+    }
 
     width = cinfo.image_width;
     height = cinfo.image_height;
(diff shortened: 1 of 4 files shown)


src/image/jpeg_io.cpp
```diff
@@ -36,7 +36,7 @@ jpeg_error_exit (j_common_ptr cinfo)
 
 void parse_jpeg(const char* data, size_t length,
                 size_t& width, size_t& height,
-                size_t& channels) {
+                size_t& channels, const std::string& url) {
```
nickjong: Where is the `url` argument used?

src/image/jpeg_io.cpp
```diff
@@ -50,6 +50,10 @@ void parse_jpeg(const char* data, size_t length,
     jpeg_mem_src(&cinfo, (unsigned char*)data, length); // Specify data source for decompression
     jpeg_read_header(&cinfo, TRUE); // Read file header, set default decompression parameters
 
+    if (cinfo.out_color_space != JCS_GRAYSCALE && cinfo.out_color_space != JCS_RGB) {
+      log_and_throw(std::string("Unsupported colorspace format. Currently, only RGB and Grayscale are supported."));
```
nickjong: How does this interact with the `ignore_failure` parameter of `load_images` (in `src/unity/lib/image_util.cpp`?

fareeha: If ignore_failure parameter is set to true then it logs the error above along with the url of the offending file.

image_util.cpp Line [188:](https://github.com/apple/turicreate/blob/af42916ca1b2816843c56b749e4166e6a6d709b6/src/unity/lib/image_util.cpp#L188)  logprogress_stream << error << "\t" << " file: " << sanitize_url(*iter) << std::endl;

which logs the error as:

"Unsupported colorspace format. Currently, only RGB and Grayscale are supported.	 file: ./images/sample.jpg"

nickjong: Ah, got it, I see now. Thanks for the explanation!

src/image/jpeg_io.cpp
fareeha: This is redundant - my bad. Thank you for the catch! I'll fix this.

image_util has access to url of images which it uses to log the error. So we don't need to include url as a parameter in parse_jpeg.

PR #1346, opened 2019-01-23
Plumbed through IOU and Confidence to evaluate and predict

### LINKED ISSUE #1333: Allow controlling non-maximum suppression threshold in predict
Right now, the documentation doesn't make it clear whether or not [predict](https://apple.github.io/turicreate/docs/api/generated/turicreate.object_detector.ObjectDetector.predict.html#turicreate.object_detector.ObjectDetector.predict) includes non-maximum suppression (NMS). It does and the threshold is 0.45 (which can checked by accessing `ObjectDetector.non_maximum_suppression_threshold`).

We don't expose any public API to override this value, as we do in `export_coreml`. We do allow `confidence_threshold` to be changed though in both, so I see no reason why we don't allow the NMS threshold too (which would make things more consistent).

Question: Should we extend this to `evaluate` too? If not, we give users the power to change these values but not the tools to decide the right value for their use case. If we allow this as an optional parameter anytime it would be used, should we just go ahead and create a function where the user can change the threshold instead?

abhishekpratapa:
Fixes #1333

### DIFF
--- a/src/unity/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/unity/python/turicreate/toolkits/object_detector/object_detector.py
@@ -726,6 +726,7 @@ class ObjectDetector(_CustomModel):
 
     def _predict_with_options(self, dataset, with_ground_truth,
                               postprocess=True, confidence_threshold=0.001,
+                              iou_threshold=None,
                               verbose=True):
         """
         Predict with options for what kind of SFrame should be returned.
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1046,6 +1071,9 @@ class cannot be returned and are ignored under this format.
         >>> print('mAP: {:.1%}'.format(results['mean_average_precision']))
         mAP: 43.2%
         """
+        if iou_threshold is None: iou_threshold = self.non_maximum_suppression_threshold
+        if confidence_threshold is None: confidence_threshold = 0.25
```
nickjong: Before this change, evaluation used a confidence threshold of 0.001, but now it defaults to 0.25. Any reason behind the change in default behavior?

PR #1349, opened 2019-01-23
updated the confidence value for the od evaluate function

abhishekpratapa:
Updated the confidence value for the od evaluate function

### DIFF
--- a/src/unity/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/unity/python/turicreate/toolkits/object_detector/object_detector.py
@@ -1072,6 +1072,7 @@ class ObjectDetector(_CustomModel):
         mAP: 43.2%
         """
         if iou_threshold is None: iou_threshold = self.non_maximum_suppression_threshold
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1072,7 +1072,6 @@ class cannot be returned and are ignored under this format.
         mAP: 43.2%
         """
         if iou_threshold is None: iou_threshold = self.non_maximum_suppression_threshold
-        if confidence_threshold is None: confidence_threshold = 0.25
```
nickjong: With this change, you're still passing a value for `confidence_threshold` below, it just happens to be None. This will still override the default value for this parameter in `_predict_with_options`, right?

abhishekpratapa: My bad, you're right. That's how code works, mental error.

abhishekpratapa: Added confidence threshold

PR #1354, opened 2019-01-24
Include master build status in README.md

znation:
Using the Markdown exported from Travis CI, we can see (and get a link to) the latest build status on Travis.


TobyRoseman:
Checking out the URL, it seems to be saying that master is currently broken:
https://travis-ci.com/apple/turicreate.svg?branch=master

I don't think that's right. Am I missing something?

znation:
@TobyRoseman That is correct, master is actually broken at the moment. Not for any particular reason; just because we happened to hit the 2 hour limit (as we intermittently do) on the latest build. Since there are 4 more queued up, probably the next build will be Passing instead.

TobyRoseman:
@znation - thanks for the explanation. That makes sense.

I think this is a great thing to add to our `README.md` but perhaps we should wait till our Travis CI is more stable. What do other people think?

nickjong:
The timeout failure seems to be an outlier. Even if not, I don't have a problem with publicizing the current health of our build in this way.

PR #1355, opened 2019-01-24
 The experimental query/logical components of the indexed SFrame

hoytak:
The experimental query/logical components of the indexed SFrame

    This PR incorporates comments from code review, as well as Yucheng's suggestion
    from today that we should separate the stable "value"/physical functionality
    from the more experimental "query"/logical functionality.

    This portion of the code contains the query / logical components of the
    indexed SFrame.  It is intended as a stepping stone towards having a the
    query functionality integrated into the existing operator framework.


hoytak:
Needs update; will develop on separate branch.

PR #1357, opened 2019-01-24
Test for Non Maximum Suppression feature in OD

fareeha:
Added test for Non Maximum Suppression feature implementation for the Object Detector in PR #1265.

Test asserts correct field population of the Non Maximum Suppression model.

### DIFF
--- a/test/unity/toolkits/object_detection/test_object_detector.cxx
+++ b/test/unity/toolkits/object_detection/test_object_detector.cxx
@@ -7,6 +7,8 @@
 #define BOOST_TEST_MODULE test_object_detector
 
 #include <unity/toolkits/object_detection/object_detector.hpp>
+#include <unity/toolkits/coreml_export/neural_net_models_exporter.hpp>
+#include <unity/toolkits/coreml_export/MLModel/src/Model.hpp>
 
 #include <array>
 #include <deque>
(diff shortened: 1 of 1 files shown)


test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -445,6 +447,63 @@ BOOST_AUTO_TEST_CASE(test_object_detector_train) {
 
   // Deconstructing `model` here will assert that every expected call to a
   // mocked-out method has been called.
+
+  // Test Non Maximum Suppression model
```
nickjong: Let's add a new test to the test suite, instead of tacking this onto the end of `test_object_detector_train`. The code you're testing has nothing to do with training or the `train` method.

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -445,6 +447,63 @@ BOOST_AUTO_TEST_CASE(test_object_detector_train) {
...
+  auto test_iou_threshold = 0.55;
+  auto test_confidence_threshold = 0.15;
```
nickjong: Nit: generally, `auto` should only be used when it improves the clarity and readability of the code, not for convenience. I think the code here would be strictly more readable with you used `double`.

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -445,6 +447,63 @@ BOOST_AUTO_TEST_CASE(test_object_detector_train) {
...
+  user_defined_metadata.emplace_back("training_iterations", test_max_iterations);
+  user_defined_metadata.emplace_back("include_non_maximum_suppression", "True");
+  user_defined_metadata.emplace_back("feature", test_image_name);
+  user_defined_metadata.emplace_back("annotations", test_annotations_name);
+  user_defined_metadata.emplace_back("classes", "label1, label2");
+  user_defined_metadata.emplace_back("type", "object_detector");
+  user_defined_metadata.emplace_back("confidence_threshold", options["confidence_threshold"]);
+  user_defined_metadata.emplace_back("iou_threshold", options["iou_threshold"]);
+    
+  flex_list t_class_labels = flex_list(test_class_labels.begin(), test_class_labels.end());
```
nickjong: Could we test that the output includes these labels?

test/unity/toolkits/object_detection/test_object_detector.cxx
fareeha: Thanks for the catch. Done!

test/unity/toolkits/object_detection/test_object_detector.cxx
fareeha: That's a good idea, thanks.

test/unity/toolkits/object_detection/test_object_detector.cxx
fareeha: Added. Thanks.

PR #1358, opened 2019-01-24
WIP: Some Travis CI config cleanup

znation:
* Remove apt installations around Python. Now that we are using Python
  configuration directly within the Travis config, it shouldn't be neccessary.

* Add Python 3.5

* Skip doc build


znation:
@hoytak @nickjong @TobyRoseman Looks like this is passing Travis. Please review, thanks!

Issue #1272, opened 2018-12-25
-std= should not be used for setting C++ version


KOLANICH:
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD

hoytak:
Unfortunately, there is a bug in cmake prior to 3.8 that prevents us from using this option until we upgrade the minimum cmake version required.  We'll evaluate doing that.

hoytak:
Resolved with https://github.com/apple/turicreate/pull/1359.

KOLANICH:
@hoytak, you can safely put `20` there, CMake will fall back to the latest version available.

PR #1359, opened 2019-01-25
Switched CMakeLists.txt to use CXX_STANDARD instead of manual flags.

hoytak:
Addresses https://github.com/apple/turicreate/issues/1272


hoytak:
Okay, turns out theres a bug here fixed in CMake 3.8 that is preventing compilation here.  It doesn't look like this is feasible yet.

hoytak:
This can be updated now that our CMake minimum version is 3.9.3

PR #1362, opened 2019-01-28
Move Vega spec including data down to C++

znation:
This change adds C++ API to include (streaming) data in the Vega spec
returned by the Plot.

This was previously implemented only in Python. Implementing in C++ is
both more efficient, and will allow API consumers beneath Python to have
this functionality.


znation:
@TobyRoseman I did; see the changes in `src/unity/python/turicreate/visualization/_plot.py`.

PR #1366, opened 2019-01-29
Add clang warning suppression on self assignment

znation:
This new warning in Xcode 10.2 beta causes an error in our build (since
we have `-Werror` on), but the thing we're trying to test only makes
sense in the presence of compilers that don't have this error/warning.
So for now, let's just suppress the warning in this test.


znation:
@nickjong Good point, I'll copy my commit message into a comment next to the suppression.

PR #1370, opened 2019-01-29
WIP: Use precompiled headers to speed up the build

znation:


znation:
Let's see if the Travis build passes. I've only tested this change locally on macOS with Xcode 9 (so far).

znation:
@hoytak @nickjong FYI, having tried using pre-compiled headers in our project with both C++ stdlib and boost included, it's actually significantly slower than before (in both gcc and clang). Trying it now with just C++ stdlib (boost is about an order of magnitude larger and we probably include it much less often).

znation:
@hoytak @nickjong With only the C++ stdlib in the precompiled header, it did better than before, but only about as good as with no precompiled header. Abandoning this PR.

Issue #1356, opened 2019-01-24
Unit test failures in image classifier


znation:
Repro steps: Run current master Python unit test test_image_classifier.py, on macOS 10.14.
Expected: No errors.
Actual:

```
=================================== FAILURES ===================================
__________________ VisionFeaturePrintScreenTest.test_predict ___________________

self = <turicreate.test.test_image_classifier.VisionFeaturePrintScreenTest testMethod=test_predict>

    def test_predict(self):
        model = self.model
        for output_type in ['class', 'probability_vector']:
            preds = model.predict(data.head(), output_type=output_type)
            _raise_error_if_not_sarray(preds)
            self.assertEqual(len(preds), len(data.head()))
            if output_type == 'class':
>               self.assertTrue(all(preds[:5] == 'white'))
E               AssertionError: False is not true

test_image_classifier.py:132: AssertionError
_______________ VisionFeaturePrintScreenTest.test_save_and_load ________________

self = <turicreate.test.test_image_classifier.VisionFeaturePrintScreenTest testMethod=test_save_and_load>

    def test_save_and_load(self):
        with test_util.TempDirectory() as filename:
    
            self.model.save(filename)
            self.model = tc.load_model(filename)
    
>           self.test_predict()

test_image_classifier.py:230: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test_image_classifier.py:132: in test_predict
    self.assertTrue(all(preds[:5] == 'white'))
E   AssertionError: False is not true
```

TobyRoseman:
This is caused by #1243 

@hoytak - please take a look. Logistic regression just seems to have the labels swapped (i.e. it's always predicting `A` when it should be predicting `B` and always predicting `B` when it should be predicting `A`). This is only happening when the model is VisionFeaturePrintScene. Let me know if I can help you fix this.

PR #1372, opened 2019-01-31
Improvements to LBFGS Solver.

hoytak:
Fix for https://github.com/apple/turicreate/issues/1356.

This fix addresses two issues:

- Occasionally the value jumped after a step.  While the algorithm
  recovered, this could lead to a suboptimal value being reported as the
  final solution.  This PR adds in a bracketing back-tracking routine
  to ensure that every subsequent value is an improvement on the
  previous.

- more_thuente is called too infrequently, causing the algorithm to be
  stuck at an inoptimal step size for two long.  This expands the
  conditions that trigger a step size search to include differences in
  the line search value or steps that could have been much more
  significant.

- The input function was not scaled, leading to non-trivial numerical 
  issues in some cases.


hoytak:
> @hoytak What is a `more_thuente` routine? And what issue is this fixing?

This is the routine that sets the step size.  It requires a few passes over the data, so we do it as little as possible, but if we don't do it enough it will converge too slowly.  This fix puts in additional conditions under which to call that routine, which helps the convergence.

znation:
@hoytak Reviewed and builds are passing. Is this ready to go in?

src/optimization/lbfgs.cpp
```diff
@@ -137,20 +135,17 @@ void lbfgs_solver::setup(
...
     auto it = opts.find(n);
     return (it != opts.end()) ? it->second : default_solver_options.at(n);
   };
 
-
   lbfgs_memory_level = get_param("lbfgs_memory_level");
   convergence_threshold = get_param("convergence_threshold");
-  m_status.step_size = 1; // get_param("step_size");
-
+  m_status.step_size = 1.0;  // / std::sqrt(num_variables);
```
znation: What's with the comment? The value doesn't seem to match the comment...

srikris:
LGTM; Have we tested this on many datasets. Just to make sure we haven't regressed.

src/optimization/lbfgs.cpp
hoytak: Bad comment.  It's set automatically now.

PR #1374, opened 2019-02-01
Added CTRL-C interupt checking to logging command 

hoytak:
This way nothing gets printed without a check.


hoytak:
This is a good fix, but it doesn't solve the issue.   It appears that CTRL-C is broken/unreliable everywhere.

PR #1376, opened 2019-02-01
WIP: Use precompiled headers (round 2)

znation:
This time, I want to see if it's any faster with all explicit headers removed, and the PCH provided explicitly.


znation:
After some more thorough investigation, this really doesn't speed up our build noticeably anywhere. Closing without merging.

PR #1381, opened 2019-02-01
WIP: Don't make copies of anything during the build

znation:
This should (in theory) speed up the build, and take up less disk space, by avoiding copying anything. Prior to this change, we had a macro `make_copy_target`, which could copy targets, files, and/or directories into the current binary directory. With this change, we now have `make_symlink_target` instead, which makes symlinks (no copies) of files and directories, and does not support targets.

Note: targets that were being copied have been moved to be produced directly in their intended location, so they are no longer copied.


znation:
Abandoning this PR, as it causes test failures and doesn't actually seem to improve build times now that #1413 is in.

PR #1385, opened 2019-02-04
Reverts the LBFGS solver to a known state:

hoytak:
This PR reverts the LBFGS Solver to before commit #1229.  The solver at
that time had some minor issues; however, the subsequent fixes have
other unknown issues which need more testing.  The decision was made to
hold off the updates for this release to ensure we release with a known
quantity.


src/optimization/CMakeLists.txt
```diff
@@ -3,7 +3,6 @@ project(optimization)
 make_library(optimization
   SOURCES
     utils.cpp
-    lbfgs.cpp
```
znation: Technically for this commit to be a revert, it should be deleting lbfgs.cpp as well, right? Since as far as I can tell in 5.2, the repo did not have this file. I think it could be confusing that master will now have both lbfgs.cpp and lbfgs-inl.hpp (as they seem largely duplicated in terms of content and purpose).

znation:
I diffed this against the 5.2.1 release tree, and the only remaining diff is the existence of `lbfgs.cpp` and `lbfgs.hpp`. As they are unused as of this PR, it wouldn't hurt correctness to leave them in, but may cause confusion.

src/optimization/CMakeLists.txt
nickjong: Is there any reason not just to literally use `git revert` on the appropriate commits we want to roll back?

znation: Good point, I'm guessing that would help with the build break too:

```
E   ImportError: /home/travis/build/apple/turicreate/debug/src/unity/python/turicreate/cython/../libunity_shared.so: undefined symbol: _ZN4turi12optimization12lbfgs_compatESt10shared_ptrINS0_25first_order_opt_interfaceEERKN4arma3ColIdEERKSt3mapISsNS_13flexible_typeESt4lessISsESaISt4pairIKSsSA_EEERKS1_INS0_28smooth_regularizer_interfaceEE
```

hoytak:
Updated to do the revert method -- should be cleaner now.

PR #1391, opened 2019-02-05
The minimal change needed to compile and link with internal libs.

hoytak:


src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -1176,18 +1181,21 @@ variant_map_type supervised_learning_model_base::api_evaluate(
       std::string target = "class"; 
       std::string pred_column = "predicted_class";
 
-
       gl_sframe out;
 
       out["class"] = data[variant_get_ref<flexible_type>(state.at("target")).get<flex_string>()];
-      out["predicted_class"] = api_predict(data, missing_value_action_str, "class");
+      if(predictions.empty()) {
+        out[pred_column] = api_predict(data, missing_value_action_str, "class");
```
znation: This should use `target` rather than `"class"`, right?

src/unity/toolkits/supervised_learning/supervised_learning.hpp
```diff
@@ -750,7 +750,10 @@ class EXPORT supervised_learning_model_base : public ml_model_base {
 
   register_defaults("evaluate",
                     {{"metric", std::string("_report")},
-                     {"missing_value_action", std::string("auto")}});
+                     {"missing_value_action", std::string("auto")},
+                     {"predictions", gl_sarray()}, 
+                     {"with_predictions", true}
```
znation: This will change the default when called from Python to `with_predictions=True`, right? (Or does Python explicitly pass that in?)

src/unity/toolkits/supervised_learning/supervised_learning.cpp
hoytak: "class" is the type of prediction requested, not the target column.

src/unity/toolkits/supervised_learning/supervised_learning.hpp
hoytak: Ah, good catch.  I think python almost always passes it in explicitly, but we might as well set this to false.

src/vega_renderer/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 project( vega_renderer )
 
-if(APPLE AND NOT TC_BUILD_IOS)
+if(APPLE)
```
znation: This is not the right change, I think... vega_renderer does not currently build on iOS.

hoytak: Fixed.

PR #1393, opened 2019-02-05
Don't build AppKit on IOS

abhishekpratapa:
Fixes a build break such that AppKit isn't being built as part of the recommender dylib.

### DIFF
--- a/src/vega_renderer/CMakeLists.txt
+++ b/src/vega_renderer/CMakeLists.txt
@@ -1,6 +1,6 @@
 project( vega_renderer )
 
-if(APPLE)
+if(APPLE AND NOT ${TC_BUILD_IOS})
 
    find_library( FOUNDATION_LIBRARY Foundation)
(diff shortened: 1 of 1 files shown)


abhishekpratapa:
fixes #1394

configure
```diff
@@ -242,6 +242,11 @@ else
   CMAKE_CONFIG_FLAGS="$CMAKE_CONFIG_FLAGS -DTC_BUILD_CAPI=0"
 fi
 
+# If we don't need to build for IOS don't make the flag nothing
+if [[ $with_ios == 0 ]]; then
+  CMAKE_CONFIG_FLAGS="$CMAKE_CONFIG_FLAGS -DTC_BUILD_IOS=0"
```
nickjong: Why do we need this?

On a potentially related note, the existing similar checks in our CMakeLists files tend to be written as  `if(APPLE AND NOT TC_BUILD_IOS)`, no `${}`.

abhishekpratapa: Ahh that would work as well, and then we wouldn't explicitly need that

PR #1395, opened 2019-02-05
Added missing __bridge symbol.

hoytak:


nickjong:
Why is this needed?

If it's not needed for 5.3, can we hold off until we finish releasing?

znation:
It seems to be needed on newest (edge) clang, but I'm actually confused why existing clang builds didn't warn us about this. The build passes without this fix on both my machine and our builds, so I think let's take this fix after we release 5.3.

hoytak:
@znation -- can you approve this?

znation:
@hoytak Approved!

PR #1398, opened 2019-02-06
Commented out vega_renderer tests temporarily

abhishekpratapa:


znation:
@abhishekpratapa By any chance, do you remember why these were disabled "temporarily"? Should we re-enable them now?

PR #1402, opened 2019-02-07
Replace Armadillo with Eigen

hoytak:
Eigen provides benefits over Armadillo in terms of speed, stability, and
    portability.  This PR replaces Armadillo with Eigen 3.  All
    functionality should be essentially uneffected.


src/external/Eigen/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+project(TuriExternalDependencies)
+
+make_empty_library(eigen)
```
znation: Why is it empty? Is Eigen a header-only library? If so, you may not even need a target for it at all.

src/unity/extensions/CMakeLists.txt
```diff
@@ -47,4 +48,11 @@ make_extension(dictionary_transformation_extensions SOURCES
 # )
 # target_link_libraries(timeseries PUBLIC grouped_sframe)
 
+#----------------------------
```
znation: Where did this come from? Doesn't look related to this change.

src/ml_data/column_statistics.cpp
```diff
@@ -237,11 +237,8 @@ void column_statistics::_finalize_global(
             // Adjust for many of the elements being zero
             sa.var_sum += std::pow(sa.mean, 2) * count * (1 - scale);
           }
-          if (total_row_count > 1) {
-            s.stdev = std::sqrt(sa.var_sum / (total_row_count - 1));
-          } else {
-            s.stdev = 0;
-          }
+
+          s.stdev = std::sqrt(sa.var_sum / (total_row_count - 1));
```
znation: Is this change intentional? It doesn't look related to Eigen.

hoytak: Good catch.

src/unity/extensions/CMakeLists.txt
hoytak: Good catch.

src/external/Eigen/CMakeLists.txt
hoytak: True.  Earlier versions had a small configuration step; that is gone now.  IMO it's better to declare the dependency explicitly if possible.  That makes it easier to, e.g. move the package to the deps/ folder.

znation: 🤷‍♂️ 

Personally I don't see why we would want to move this to the deps folder, or why we want anything to explicitly depend on an empty library, but it doesn't really hurt either.

Note that in #1403 I'm actually removing most of the existing empty libraries, as that change splits compile and link-time dependencies apart, and the empty libraries are neither a compile-time nor a link-time dependency of... well, anything, really.

hoytak: I see what you're saying.  I'm okay doing it that way.

hoytak:
> Do you want to delete `src/external/armadillo`?

It's not already deleted?

src/serialization/serialize_eigen.hpp
```diff
@@ -0,0 +1,235 @@
+/**
+ * Copyright (C) 2016 Turi
```
znation: Replace this with the Apple copyright header.

Issue #1405, opened 2019-02-07
Exception when importing TuriCreate on a machine with no GPUs


steverichey:
We have a Python script that trains an image classifier using the latest version of TuriCreate. On a MacBook Pro, iMac Pro, or Linux instance with a GPU running in Amazon EC2, this works fine. However, as we have found that training the classifier does not use the GPU, we attempted to train on a compute-optimized machine with no GPU (named `c5d.18xlarge` in the AWS interface, running Ubuntu 18.04) and received the following error:

```
Traceback (most recent call last):
  File "build_model.py", line 1, in <module>
    import turicreate as tc
  File "/home/ubuntu/projects/.../venv/local/lib/python2.7/site-packages/turicreate/__init__.py", line 19, in <module>
    from turicreate.data_structures.sgraph import Vertex, Edge
  File "/home/ubuntu/projects/.../venv/local/lib/python2.7/site-packages/turicreate/data_structures/__init__.py", line 18, in <module>
    from . import sframe
  File "/home/ubuntu/projects/.../venv/local/lib/python2.7/site-packages/turicreate/data_structures/sframe.py", line 16, in <module>
    from ..connect import main as glconnect
  File "/home/ubuntu/projects/.../venv/local/lib/python2.7/site-packages/turicreate/connect/main.py", line 13, in <module>
    from ..cython.cy_unity import UnityGlobalProxy
  File "turicreate/cython/cy_sarray.pxd", line 92, in init turicreate.cython.cy_unity
  File "turicreate/cython/cy_model.pxd", line 28, in init turicreate.cython.cy_sarray
  File "turicreate/cython/cy_model.pyx", line 1, in init turicreate.cython.cy_model
  File "turicreate/cython/cy_graph.pxd", line 62, in init turicreate.cython.cy_variant
  File "turicreate/cython/cy_graph.pyx", line 20, in init turicreate.cython.cy_graph
  File "/home/ubuntu/projects/.../venv/local/lib/python2.7/site-packages/turicreate/util/__init__.py", line 750, in <module>
    _CUDA_GPUS = _get_cuda_gpus()
  File "/home/ubuntu/projects/.../venv/local/lib/python2.7/site-packages/turicreate/util/__init__.py", line 731, in _get_cuda_gpus
    universal_newlines=True)
  File "/usr/lib/python2.7/subprocess.py", line 574, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
[...]

PR #1406, opened 2019-02-07
Handle exception thrown when calling `nvidia-smi` when no GPU is installed

steverichey:
This fixed the issue noted in the related ticket; seems the exception type is just not as expected.


TobyRoseman:
fix for #1405

PR #1410, opened 2019-02-11
Re-factor Activity Classifier MPS backend to adopt same backend interface as Object Detection

nickjong:
This is the first step towards creating a C++ implementation of the AC toolkit, to parallel the new OD implementation. I had to generalize the `neural_net::cnn_module` interface I had created for OD, since AC requires an additional input (weights) for each training batch, and the training model returns both the forward output and the loss. To this end, I made the interface more like the MPS and CoreML interfaces, in that they take named inputs and return named outputs. Along the way, I renamed `cnn_module` to `model_backend`, since "CNN" was always a bit of a misnomer.

Note, this constitutes a dramatic narrowing of the interface for the "low-level" MPS backend. AFAICT much of this interface surface supported debugging and testing during the original development, but should not generally be needed. Much of this functionality would be straightforward to add back via configuration parameters of the revamped class.

It may be easier to review this work one commit at a time; the commit boundaries were chosen with an eye towards reviewable/logical chunks.

@igiloh,@tjia1818 FYI

### DIFF
--- a/src/unity/python/turicreate/toolkits/_mps_utils.py
+++ b/src/unity/python/turicreate/toolkits/_mps_utils.py
@@ -494,7 +494,7 @@ class MpsGraphAPI(object):
         result_handle = _ctypes.c_void_p()
         status_code = self._LIB.TCMPSTrainGraph(
             self.handle, input_array.handle, label_array.handle,
-                _ctypes.byref(result_handle))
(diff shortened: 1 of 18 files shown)


TobyRoseman:
Looks good to me.

PR #1413, opened 2019-02-13
CMake simplification & build speedup

znation:
The general idea behind these changes is to prevent linking targets that
aren't actually shipped from this repo. Prior to this PR, the repo
contains quite a few static libraries that aren't shipped anywhere (and
may or may not be included in various other targets), which result in an
extra cost (to CPU and disk) both in assembling and preserving the `.a` files.

This change reorients building and linking along a different axis.

Previously: each module within the build (typically a single folder
under `src/`) would get compiled and linked into its own static library.
Those static libraries would then get (usually) get linked into
libunity_shared.so or another Python lib, another library for execution
(libTuriCreate/libRecommender/etc.), or a test executable.

With this PR: each module within the build only gets compiled (into a
`OBJECT` library, which is a way of tracking collections of `.o` files
as a single library in CMake). Each final link target (i.e.
libunity_shared.so) then compiles with all the .o files directly
(essentially bypassing compilation and just linking instead). This also
allows us to get rid of the usage of things like `make_shared_from_static`.

Bonus change: all tests now link against
`libunity_shared_for_testing.so` instead of statically linking all their
dependencies, so they are much smaller now.

Disk usage (debug, macOS):

* Before: 17.0 GB
* After: 4.3 GB

Build time (debug, 15" 2015 MBP, primed ccache):

* Before: 927.47s user 487.59s system 318% cpu 7:23.84 total
* After: 575.39s user 406.67s system 256% cpu 6:22.35 total

Build time (Travis CI, approximate):

* Before: About 2 hours total (including running tests).
* After: About 1:51 total (including running tests).


CMakeLists.txt
```diff
@@ -48,6 +48,16 @@ set(TC_PLATFORM_WIN32 false)
 if (APPLE)
   set(TC_PLATFORM_APPLE true)
   add_definitions(-DTC_PLATFORM_APPLE)
+  find_library(ACCELERATE NAMES Accelerate)
+  find_library(APPKIT AppKit)
+  find_library(CORE_GRAPHICS CoreGraphics)
+  find_library(JAVASCRIPTCORE JavaScriptCore)
```
nickjong: Okay, I'm going to be that guy. Shouldn't this be `JAVA_SCRIPT_CORE`?

src/external/libjson/CMakeLists.txt
```diff
@@ -16,11 +15,6 @@ SOURCES
...
-COMPILE_FLAGS_EXTRA
-  -Wno-reorder
-COMPILE_FLAGS_EXTRA_GCC
-  -Wno-cpp
-COMPILE_FLAGS_EXTRA_CLANG
-  -Wno-header-guard
-  -Wno-\#warnings
 )
+
+target_compile_options(libjson PRIVATE -w)
```
nickjong: Also going to be that guy: let's append the newline here (and in other instances below), so that any future addendums to this file don't necessarily mutate this line.

src/unity/toolkits/coreml_export/MLModel/build/format/CategoricalMapping_enums.h
```diff
@@ -8,6 +8,7 @@ enum MLCategoricalMappingMappingType: int {
     MLCategoricalMappingMappingType_NOT_SET = 0,
 };
 
+__attribute__((__unused__))
```
nickjong: This is ultimately generated code, right? Will we have to make these modifications each time we update our snapshot of the proto?

src/unity/toolkits/neural_net/CMakeLists.txt
```diff
@@ -7,68 +7,35 @@ project(unity_toolkits)
...
-  COMPILE_FLAGS_EXTRA_GCC
-    -Wno-unknown-pragmas  # NOTE: used for auto-generated protobuf source files
-    -Wno-unused-function  # NOTE: used for auto-generated protobuf source files
+add_library(unity_neural_net OBJECT
+  compute_context.cpp
+  float_array.cpp
+  image_augmentation.cpp
+  model_spec.cpp
 )
+add_dependencies(unity_neural_net boost)
```
nickjong: Why do we need to add this to each object library?

src/unity/toolkits/coreml_export/MLModel/build/format/CategoricalMapping_enums.h
znation: Yes. I plan to put up a coremltools PR to fix it in the generator.

src/unity/toolkits/neural_net/CMakeLists.txt
znation: To enforce ordering. Otherwise the boost headers don’t exist yet (in the place we’re looking for them). It’s possible we could refer to them elsewhere but I haven’t tried.

nickjong: Makes sense. Room for future work

CMakeLists.txt
znation: Actually I think it should be JAVASCRIPT_CORE (since JavaScript is all one word).

znation:
@nickjong Comments addressed; please re-review, thanks!

ylow:
A suggestion is to keep the original make_library macro but making it just a thin wrapper around "add_library( ... OBJECT ...)". We can then maintain the original REQUIRE flags in all the CMakeLists.txt, but we just ignore the REQUIRE arguments until we upgrade to a newer CMake which can handle the transitive dependencies correctly.

znation:
Thanks @ylow, I'll do that.

znation:
@ylow @nickjong Comments have been addressed. Please re-review, thanks!

src/flexible_type/CMakeLists.txt
```diff
@@ -6,14 +6,14 @@ project(flexible_type)
...
     string_escape.cpp
     flexible_type_conversion_utilities.cpp
     flexible_type_spirit_parser.cpp
   REQUIRES
-    logger
-    image_type
-    EXTERNAL_VISIBILITY
-  )
+    logger		
+    image_type
```
nickjong: Nit: I'm assuming this trailing whitespace is accidental?

src/logger/CMakeLists.txt
```diff
@@ -6,18 +6,18 @@ project(Turi)
...
     log_level_setter.cpp
   REQUIRES
-    boost
-    timer
-    parallel
-    stack_trace
-    cppipc
-    EXTERNAL_VISIBILITY
-  )
+    boost
```
nickjong: Here too

src/nanosockets/CMakeLists.txt
```diff
@@ -1,24 +1,21 @@
...
+    async_reply_socket.cpp
+    async_request_socket.cpp
+    print_zmq_error.cpp
+    zmq_msg_vector.cpp
+    socket_config.cpp
+    subscribe_socket.cpp
+    publish_socket.cpp
+    get_next_port_number.cpp
+  REQUIRES
+    network
```
nickjong: And here

src/flexible_type/CMakeLists.txt
znation: Thanks, good catch. Will fix.

src/unity/toolkits/coreml_export/MLModel/build/format/CategoricalMapping_enums.h
znation: Fixed on the coremltools side: https://github.com/apple/coremltools/pull/327

Issue #1400, opened 2019-02-06
Size of Recommender Model


dogahe:
I am looking at ways that I can reduce the size of a recommender model I am working on. You can assume I want to train a model for recommending topics to readers. Using one dataset which has 6452 observations of (TopicID, UserID), there are 342 unique topics and 3281 unique users. The size of the model that was trained by using `tc.recommender.create` is 442 KB which is fine. 

Now I have a larger dataset with 3,549,113 observations, 2174 unique topics and 251,514 unique users, the size of the model is 32.2 MB which is really large for my application. I was wondering if you have any recommendations on what the different things that I can try are to reduce the size of the model. 

One thing I had in mind was to let say only try to make recommendations for the top topics (623 out of 2174), therefore removing all the observations for topics which are not among top 623. That resulted in 3,395,846 observations with 245,963 users and 623 topics and the size of the model becoming 28.8 MB, just a small decrease in size. 

Next thing, I am going to try is to remove all the observations by users who have only read one topic. Meaning if a user has only one row in the observations I am going to remove the data (that one row) for that user. Does this make sense to you? Would this make the model less effective? Any thoughts on this approach that I yet have to try to see how much it will reduce the size of the model?

TobyRoseman:
@dogahe - To make sure we provide you with useful information, please provide some additional details: 

1 - Which recommender model type are you using? Is it `ItemSimilarityRecommender`?

2 - Are you concerned about the size of the output file when you call `model.save(...)` or `model.export_coreml(...)`?

dogahe:
Hi @TobyRoseman - Yes, I believe based on the fact that data has only the two mentioned columns it picks ItemSimilarityRecommender. I am concerned about the size of the `.mlmodel` which is the output of `model.export_coreml`

dogahe:
Even doing what I asked you if it makes sense to do only reduces the size of the model to 24.7 MB. I still would like to hear your thoughts on whether " remove all the observations by users who have only read one topic." affects the performance of recommendation? Even though for my case does not reduce the size of the model by too much. Also I would like to hear if there is anything you can suggest for me to do to reduce the size of the model?

hoytak:
[...]

PR #1415, opened 2019-02-14
Strip unneeded indexing tables from CoreML exported recsys models.

hoytak:
- Added ability to drop user data from the recommender coreml export.
- Export to coreml now does this.

Fixes https://github.com/apple/turicreate/issues/1400


TobyRoseman:
Would it be possible to get a unit test which reproduces the bug?

hoytak:
This doesn't actually work -- set up a unit test (thanks @TobyRoseman).  Realized I need to strip the tables from the statistics side of things too.

hoytak:
This new version works and in my tests resolves https://github.com/apple/turicreate/issues/1400.

PR #1416, opened 2019-02-15
Fixes S3 write path for SFrames

ylow:
S3 has no directories, hence create_directory always succeeds when used on S3 filesystems.

create_directory_or_throw should therefore not check file status when S3 is used.


srikris:
Fixes #57

Issue #1419, opened 2019-02-15
Remove references to blas and lapack


znation:
Now that we have replaced Armadillo with Eigen, we can remove the dependencies on libblas and libpack. We still mention these dependencies in:

```
BUILD.md
INSTALL_ISSUES.md
LINUX_INSTALL.md
```

And we still check for LAPACK in `CMakeLists.txt`:
```
CMakeLists.txt:  include(FindLAPACK)
CMakeLists.txt:  if(NOT ${LAPACK_FOUND})
CMakeLists.txt:    message(FATAL_ERROR "lapack is not installed")
```

And we still install it in Travis CI:
```
.travis.yml:      - liblapack-dev
```

znation:
We can probably also remove `-DARMA_NO_DEBUG`, right?

PR #1424, opened 2019-02-16
19 02 remove lapack blas deps

hoytak:
Resolves https://github.com/apple/turicreate/issues/1419

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ addons:
       - gcc-5
       - g++-5
       - libstdc++6
-      - libblas-dev
-      - liblapack-dev
       - python-setuptools
       - libgif-dev
       - vim-common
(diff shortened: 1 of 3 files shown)


src/unity/lib/gl_sframe.cpp
```diff
@@ -266,9 +265,8 @@ std::vector<flexible_type> gl_sframe::operator[](int64_t i) {
   if (i < 0 || (size_t)i >= size()) {
     throw std::string("Index out of range");
   }
-  ensure_has_sframe_reader();
   std::vector<std::vector<flexible_type> > rows(1);
-  size_t rows_read  = m_sframe_reader->read_rows(i, i + 1, rows);
```
znation: These changes look unrelated... should they be in a separate branch?

hoytak: Ooops -- should not have pulled those in.  IDK

PR #1426, opened 2019-02-18
Make ccache work on Travis CI

znation:


znation:
@TobyRoseman @hoytak Please review. With 4GB of cache (mostly/fully primed), we're now looking at a floor of about 1h 15m for the Travis build.

PR #1432, opened 2019-02-19
Updates packaged CMake to version 3.13.4.

hoytak:
This enables the use of object libraries, which would substantially help the build process.

    Several GPL-ish files in the LexerParser directory had to be replaced.
    The log for how this was done is checked in.

Also, the minimum cmake version is now 3.12.


znation:
Please also update the Travis CI config to download and use the CMake binary from 
https://github.com/Kitware/CMake/releases/ (hopefully it should work on Travis). Otherwise we'll be compiling CMake from source on every Travis CI build, which pushed us over the 2 hour limit on most builds last time I accidentally caused it.

PR #1434, opened 2019-02-19
Fixed a failing test on the data od iterator so it works with shuffling.

hoytak:


nickjong:
Let's either skip testing a second pass through the data, or add a parameter to disable shuffling.

hoytak:
Alternate solution

PR #1436, opened 2019-02-20
New Toolkit: Drawing Classifier 

shantanuchhabra:
This is the base Pull Request for the Drawing Classifier Toolkit. Other Pull Requests that are upcoming, won't be part of this Pull Request, but will build on top of it are:
- [docs] Add docstrings for all functions.
- [cleanup] Change `drawing_recognition` to `drawing_classifier`.
- [linux] Add rasterization code from strokes to bitmap for non-macOS code paths: #1449


src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,483 @@
...
+        from turicreate.toolkits import _coreml_utils
+        model_type = "drawing classifier"
+        coreml_model.short_description = _coreml_utils._mlmodel_short_description(model_type)
+        coreml_model.input_description['bitmap'] = u'bitmap'
+        coreml_model.output_description['probabilities'] = 'Prediction probabilities'
+        coreml_model.output_description['classLabel'] = 'Class Label of Top Prediction'
+        coreml_model.save(filename)
+        return coreml_model
+
+    def _predict_with_options(self, input_dataset, verbose=True):
```
TobyRoseman: `verbose` does nothing.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,262 @@
...
+template<typename T>
+std::vector<T> slice(std::vector<T> const &v, int m, int n) {
+    auto first = v.cbegin() + m;
+    auto last = v.cbegin() + n + 1;
+    std::vector<T> vec(first, last);
+    return vec;
+}
+
+flex_list ramer_douglas_peucker(flex_list stroke, float epsilon) {
+    float dmax = 0;
```
TobyRoseman: A comment with a link to wikipedia would be cool: https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm

src/unity/toolkits/drawing_recognition/drawing_recognition.cpp
```diff
@@ -0,0 +1,6 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
TobyRoseman: Why do we have these two empty files?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,483 @@
...
+           **kwargs):
+    """
+    Create a :class:`DrawingRecognition` model.
+    """
+    start_time = _time.time()
+
+    is_stroke_input = (input_dataset[feature].dtype != _tc.Image)
+
+    if is_stroke_input:
+        # This only works on macOS right now
```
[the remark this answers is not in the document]
shantanuchhabra: There is code to make it work for non-macOS code paths, in another PR: #1449

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,262 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
shantanuchhabra: Please leave comments/feedback for the `src/unity/toolkits/drawing_recognition` code path in #1449 since that is more up to date.

shantanuchhabra:
Please leave comments/feedback for the `src/unity/toolkits/drawing_recognition` code path (i.e. all C++ code) in #1449 since that is more up to date.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,483 @@
...
+    def hybrid_forward(self, F, x):
+        x = self.pool1(self.conv1(x))
+        x = self.pool2(self.conv2(x))
+        x = self.pool3(self.conv3(x))
+        x = self.flatten(x)
+        x = self.fc1(x)
+        x = self.fc2(x)
+        return F.softmax(x)
+
+def create(input_dataset, annotations=None, num_epochs=100, feature="bitmap", model=None,
```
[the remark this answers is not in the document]
shantanuchhabra: num_iterations = num_epochs * len(dataset) / batch_size

src/unity/python/turicreate/toolkits/_mxnet_to_coreml/_layers.py
```diff
@@ -81,6 +81,9 @@ def convert_reshape(net, node, module, builder):
     name = node['name']
     param = _get_attr(node)
     target_shape = literal_eval(param['shape'])
+    if param['shape'] == '(0, -1)':
```
nickjong: Could you explain what this logic does?

src/unity/python/turicreate/toolkits/drawing_recognition/__init__.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Apple Inc. All rights reserved.
```
nickjong: New code may as well use '2019' here

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
```diff
@@ -0,0 +1,282 @@
...
+            self.sframe = self.sframe.sort(_TMP_COL_RANDOM_ORDER)
+        self.cur_sample += 1
+
+        # Copy the image data for this row into a NumPy array.
+        row = self.sframe[row_index]
+
+        drawing_feature = row[self.feature_column]
+        is_stroke_input = (type(drawing_feature) != _tc.Image)
+        if is_stroke_input:
+            pixels_01 = drawing_feature
```
nickjong: Didn't you just test that `drawing_feature` is NOT an image here? How are you interpreting it as pixels? Should there be a TODO here to implement rasterization?

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
```diff
@@ -0,0 +1,282 @@
...
+        else:    
+            image = _tc.image_analysis.resize(drawing_feature, 28, 28, 1)
+            pixels_01 = image.pixel_data.reshape(1, 28, 28) / 255.
+        
+        if self.load_labels:
+            label = row[self.target_column]
+            if label is None:
+                label = []
+            elif type(label) is dict:
+                label = [label]
```
nickjong: Why is `label` a list? Shouldn't our labels be strings or integers?

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
```diff
@@ -0,0 +1,282 @@
...
+
+        return pixels_01, label, row_index
+
+    def reset(self):
+        self.cur_sample = 0
+
+
+# A wrapper around _SFrameDataSource that uses a dedicated worker thread for
+# performing SFrame operations.
+class _SFrameAsyncDataSource:
```
nickjong: Does the async data prep actually help? Might be worth seeing if the performance is fine without it. This adds significant complexity that seemed needed for object detection, but I'm not convinced we need it here, especially if we can't factor it out and share it among toolkit data iterators

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
```diff
@@ -0,0 +1,282 @@
...
+        # waiting to put something into it (and therefore never receives the
+        # new buffer).
+        if self.buffer is not None:
+            while self.buffer.get() is not None:
+                pass
+
+        self.buffer = buffer
+
+
+class SFrameRecognitionIter(_mx.io.DataIter):
```
nickjong: Why is this called a "recognition" iterator?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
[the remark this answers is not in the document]
nickjong: Note that some of the match from the OD toolkit was due to each mini-batch being so slow, that we yielded progress updates at the granularity of mini-batch. Note that in contrast, AC yields progress at the granularity of epoch.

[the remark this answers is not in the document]
nickjong: Do any of our other toolkits use "epoch" as a parameter?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,480 @@
...
+           classes=None, batch_size=256, max_iterations=0, verbose=True, 
+           **kwargs):
+    """
+    Create a :class:`DrawingRecognition` model.
+    """
+    start_time = _time.time()
+
+    is_stroke_input = (input_dataset[feature].dtype != _tc.Image)
```
nickjong: Should we throw an error if the input is neither images nor stroke data? What if you an SArray of integers or something?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,480 @@
...
+        classes = dataset['label'].unique()
+    classes = sorted(classes)
+    class_to_index = {name: index for index, name in enumerate(classes)}
+
+    def update_progress(cur_loss, iteration):
+        iteration_base1 = iteration + 1
+        if progress['smoothed_loss'] is None:
+            progress['smoothed_loss'] = cur_loss
+        else:
+            progress['smoothed_loss'] = (0.9 * progress['smoothed_loss']
```
nickjong: If I'm not mistaken, smoothed loss in OD was again a concession to the length of an epoch. Most of our other toolkits yield statistics at the granularity of training step. I.e., what was the average loss during epoch n? There's also no reason we shouldn't output metrics like accuracy. (This could be a followup PR though)

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,480 @@
...
+                time=elapsed_time , width=column_width-1))
+            progress['last_time'] = cur_time
+
+    loader = _SFrameRecognitionIter(dataset, batch_size,
+                 feature_column='bitmap',
+                 target_column='label',
+                 class_to_index=class_to_index,
+                 load_labels=True,
+                 shuffle=True,
+                 io_thread_buffer_size=0,
```
nickjong: If we're not actually using an IO thread, let's just not include the relevant code in the iterator implementation. If you decide we need it, you can copy it then. Hopefully will have a better, shared solution before then

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,480 @@
...
+        cur_loss = loss.mean().asscalar()
+        
+        update_progress(cur_loss, batch.iteration)
+        iteration = batch.iteration
+
+    training_time = _time.time() - start_time
+    if verbose:
+        print(hr)   # progress table footer
+    state = {
+        'model': model,
```
nickjong: Let's make the MXNet model an implementation detail, `_model`.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,480 @@
...
+        from turicreate.toolkits import _coreml_utils
+        model_type = "drawing classifier"
+        coreml_model.short_description = _coreml_utils._mlmodel_short_description(model_type)
+        coreml_model.input_description['bitmap'] = u'bitmap'
+        coreml_model.output_description['probabilities'] = 'Prediction probabilities'
+        coreml_model.output_description['classLabel'] = 'Class Label of Top Prediction'
+        coreml_model.save(filename)
+        return coreml_model
+
+    def _predict_with_options(self, input_dataset):
```
nickjong: What are the options? Why do we need a separate function here?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,480 @@
...
+            
+            Parameters
+            ----------
+            dataset : SFrame | SArray | turicreate.Image
+            The images on which to perform object detection.
+            If dataset is an SFrame, it must have a column with the same name
+            as the feature column during training. Additional columns are
+            ignored.
+            
+            confidence_threshold : float
```
nickjong: How is this parameter passed?

Can the user get the class likelihoods?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,480 @@
...
+            
+            Examples
+            --------
+            .. sourcecode:: python
+            
+            # Make predictions
+            >>> pred = model.predict(data)
+            
+            # Stack predictions, for a better overview
+            >>> turicreate.object_detector.util.stack_annotations(pred)
```
nickjong: Please remove dead copy-pasta

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
shantanuchhabra: This is a little confusing. 

In this PR, _drawing_recognition_prepare_data in C++ returns a flex_list of pixel values, which can directly be passed in to the neural network which is why the sframe_loader doesn't do anything.

As of #1449,  _drawing_recognition_prepare_data in C++ returns a flex_image, which sframe_loader would need to extract pixel_data from.

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
shantanuchhabra: To be in line with "drawing_recognition". As part of the "recognition" to "classifier" rename, I expect to rename this as well. Is there a better name you have in mind for this iterator?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Absolutely. We should throw a ToolkitError.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: A better name for this would be `predict_with_probabilities`. The code here is shared by both `evaluate` and `predict`. While `predict` only cares about the label column in the SFrame that's returned, `evaluate` cares about the `p robabilities` column as well (to compute metrics like AUC). I'll let @fareeha chime in if there's something I'm missing.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: I think this is dead copy-paste that came in during `evaluate` and `predict` development. Will delete until we get better docs.

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
shantanuchhabra: Yes, these type checks are irrelevant for us. Deleting.

src/unity/python/turicreate/toolkits/_mxnet_to_coreml/_layers.py
shantanuchhabra: For flatten layers, the mxnet converter calls `convert_reshape` (which is fine because flatten is a kind of reshape) but it doesn't actually call `convert_flatten` and instead throws a `NotImplemented` Error since the shape parameters for a flatten layer when `convert_reshape` is called are `(0,-1)`. @gustavla looked at this code during our code review and recommended that I change this to not do a string comparison but to do a value comparison for the tuple, i.e. target_shape. Will make that change.

src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
shantanuchhabra: Removed Async Data Prep.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Noted. Will make a followup PR to address this. Tracking this issue in #1464

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,483 @@
...
+        # calculate training metrics
+        cur_loss = loss.mean().asscalar()
+        
+        update_progress(cur_loss, batch.iteration)
+        iteration = batch.iteration
+
+    training_time = _time.time() - start_time
+    if verbose:
+        print(hr)   # progress table footer
+    state = {
```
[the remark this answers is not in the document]
shantanuchhabra: Added.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -0,0 +1,483 @@
+import turicreate as _tc
+import urllib as _urllib
+import os as _os
+import glob as _glob
+import numpy as _np
+import mxnet as _mx
```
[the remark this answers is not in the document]
shantanuchhabra: Hmm, I do need all the mxnet imports because I have a Model class that inherits HybridBlock. Should I defer all the other mxnet imports to where I need them?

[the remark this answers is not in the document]
TobyRoseman: I think you're probably going to need to move that model class to a separate file, and not import that file until it's needed.

src/unity/python/turicreate/toolkits/_mxnet_to_coreml/_mxnet_converter.py
```diff
@@ -124,7 +124,8 @@ def _get_layer_converter_fn(layer):
 
 
 def convert(model, input_shape, class_labels=None, mode=None,
-            preprocessor_args=None, builder=None, verbose=True):
+            preprocessor_args=None, builder=None, verbose=True,
```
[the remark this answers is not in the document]
shantanuchhabra: Done.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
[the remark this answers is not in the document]
shantanuchhabra: Thanks for the suggestion, fixed now.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -229,23 +219,35 @@ def export_coreml(self, filename, verbose=False):
                                 class_labels=self.classes,
                                 input_shape=[('bitmap', image_shape)],
                                 builder=None, verbose=verbose,
-                                preprocessor_args={'image_input_names':['bitmap']},
-                                is_drawing_recognition=True)
-
+                                preprocessor_args={'image_input_names':['bitmap']})
+
+        DESIRED_OUTPUT_NAME = "probabilities"
```
nickjong: Should use `self.target + 'Probabilities'`?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -229,23 +219,35 @@ def export_coreml(self, filename, verbose=False):
...
+        class_label_output_index = 1
```
nickjong: Dead assignment?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -229,23 +219,35 @@ def export_coreml(self, filename, verbose=False):
...
-        coreml_model.short_description = _coreml_utils._mlmodel_short_description(model_type)
-        coreml_model.input_description['bitmap'] = u'bitmap'
-        coreml_model.output_description['probabilities'] = 'Prediction probabilities'
-        coreml_model.output_description['classLabel'] = 'Class Label of Top Prediction'
-        coreml_model.save(filename)
-        return coreml_model
-
-    def _predict_with_options(self, input_dataset):
+        spec.description.metadata.shortDescription = _coreml_utils._mlmodel_short_description(model_type)
+        spec.description.input[0].shortDescription = u'bitmap'
```
nickjong: Should use self.feature?

src/unity/python/turicreate/toolkits/_mxnet_to_coreml/_layers.py
```diff
@@ -942,10 +936,10 @@ def convert_instancenorm(net, node, model, builder):
 
     Parameters
     ----------
-    network: net
+    net: network
         A mxnet network object.
 
-    layer: node
+    node: layer
```
TobyRoseman: Thanks for fixing all of these. I'm a big believer in always trying to leave existing code in a better state than you found it.

src/unity/python/turicreate/toolkits/drawing_recognition/_model_architecture.py
```diff
@@ -0,0 +1,31 @@
+import mxnet as _mx
```
TobyRoseman: Copyright header.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Done

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Done.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Replaced 'bitmap' with `self.feature` everywhere.

src/unity/python/turicreate/toolkits/_mxnet_to_coreml/_layers.py
shantanuchhabra: Thank you for pointing it out!

src/unity/python/turicreate/toolkits/drawing_recognition/_model_architecture.py
shantanuchhabra: Added.

src/unity/toolkits/drawing_recognition/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+project(unity_toolkits)
+
+make_library(unity_drawing_recognition
```
nickjong: I believe this needs to be an `OBJECT` library before merging to master

PR #1445, opened 2019-02-21
Updated default compiler flags to enable newer architectures.

hoytak:
Turns out the default one was REALLY old.


CMakeLists.txt
```diff
@@ -563,7 +552,6 @@ endif()
 add_compile_options(-Dgoogle=__tc_google)
 add_compile_options(-DCoreML=__tc_CoreML)
 add_compile_options(-DHAVE_PTHREAD=1)
-add_compile_options(-DEIGEN_MPL2_ONLY)
```
znation: Do we not need this anymore? I thought we did...

hoytak: It was already defined above -- this deletes the duplicate.

PR #1446, opened 2019-02-22
Enable building on GCC 4.8

znation:
Our minimum GCC version to build is currently 5.4. With this PR, we can build on GCC 4.8, which comes out of the box in Ubuntu 14.04.


src/platform/parallel/execute_task_in_native_thread.hpp
```diff
@@ -6,6 +6,9 @@
 #ifndef TURI_PARALLEL_EXECUTE_TASK_IN_NATIVE_THREAD_HPP
 #define TURI_PARALLEL_EXECUTE_TASK_IN_NATIVE_THREAD_HPP
 #include <functional>
+#include <boost/mpl/vector.hpp>
```
ylow: we can drop these 3 includes.

znation: Oh, right. Thanks.

znation:
@ylow Comment addressed, please re-review, thanks!

PR #1448, opened 2019-02-22
Make mxnet import lazy; test for import of expensive packages in lambda workers

hoytak:
It's not currently.


hoytak:
This will fail, but we should fix the code so these tests pass.

znation:
Whoa! I thought it was lazy already. Thanks for fixing this @hoytak!

src/unity/python/turicreate/test/test_mxnet_import.py
```diff
@@ -0,0 +1,15 @@
+import unittest
```
TobyRoseman: Copyright header.

TobyRoseman:
As part of our top level [__init__.py](https://github.com/apple/turicreate/blob/ae10b860d462fdbbdc8bf351d5a58bae9f789854/src/unity/python/turicreate/__init__.py#L124) file we import mxnet. I think we probably want to keep this behavior. 

We do the import inside of a function. However it seems once a module is imported it will stay in `sys.modules` even if the import goes out of scope:
```python
In [1]: import sys

In [2]: 'coremltools' in sys.modules
Out[2]: False

In [3]: def foo():
   ...:     import coremltools
   ...:     

In [4]: foo()

In [5]: 'coremltools' in sys.modules
Out[5]: True
```

znation:
@TobyRoseman Why should we keep the import top-level? I think it would be useful to defer import of mxnet until first use. That way we don't pay the (relatively expensive) cost of importing it until it's actually used, and lots of use cases don't use it at all.

TobyRoseman:
@znation - I think the error message (that is potentially displayed by the top level import) is very helpful to our users. In my experience MXNet as poor API stability. A version other than the ones we explicitly support is unlikely to work. I'm thinking it will probably also fail in a way which makes determining the root cause difficult.

znation:
@TobyRoseman Could we defer that error message until first use as well? Or would users be unlikely to see it then?

TobyRoseman:
@znation - That's an option. My main concern would be that we'd be duplicating this logic in several places (at least five places now and more in the future). We'll likely miss it somewhere.

I do agree that importing mxnet is slow:
```python
In [1]: %time import mxnet
CPU times: user 767 ms, sys: 216 ms, total: 984 ms
Wall time: 742 ms
```

znation:
@TobyRoseman That makes sense; if we do defer import, we'll need to be careful to make sure all uses go through our deferral-with-possible-error-message logic. I think it's probably worth doing in this PR (and keeping Hoyt's tests) to prevent the delay on import turicreate. 742 ms is quite long.

hoytak:
The other issue is that this means mxnet is imported on all the lambda workers, which uses gigs of ram.  We observed that this causes significant swapping in some cases.

TobyRoseman:
I'm convinced. Let's stop always importing mxnet.

hoytak:
@znation @TobyRoseman -- got this done now; could you reexamine this PR?

PR #1449, opened 2019-02-22
Stroke to Bitmap conversion for non-macOS code paths

### LINKED ISSUE #1438: Add code for converting stroke-based drawings to a bitmap for non-Mac code paths.
We have code for converting stroke-based drawings to a bitmap for code paths on macOS. We want to expand that functionality to Linux.

shantanuchhabra:
Fixes #1438


src/unity/python/turicreate/toolkits/drawing_recognition/_sframe_loader.py
```diff
@@ -0,0 +1,283 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Apple Inc. All rights reserved.
```
znation: © 2019 Apple Inc.

src/unity/toolkits/drawing_recognition/class_registrations.cpp
```diff
@@ -0,0 +1,19 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: © 2019 Apple Inc.

src/unity/toolkits/drawing_recognition/class_registrations.hpp
```diff
@@ -0,0 +1,19 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: © 2019 Apple Inc.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,376 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: © 2019 Apple Inc.

src/unity/toolkits/drawing_recognition/drawing_recognition.cpp
```diff
@@ -0,0 +1,6 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: © 2019 Apple Inc.

src/unity/toolkits/drawing_recognition/drawing_recognition.hpp
```diff
@@ -0,0 +1,13 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: © 2019 Apple Inc.

src/unity/toolkits/drawing_recognition/data_preparation.hpp
```diff
@@ -0,0 +1,43 @@
...
+ */
+#ifndef TURI_DRAWING_DATA_PREPARATION_H_
+#define TURI_DRAWING_DATA_PREPARATION_H_
+
+#include <export.hpp>
+#include <unity/lib/gl_sframe.hpp>
+#include <unity/lib/variant.hpp>
+#include <unity/lib/image_util.hpp>
+#ifdef __APPLE__
+#include <CoreGraphics/CoreGraphics.h>
```
znation: Nothing in this header uses CoreGraphics; please remove the include.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,376 @@
...
+    const uint8_t* data_ptr = CFDataGetBytePtr(pixel_data);
+    uint8_t real_data[FINAL_BITMAP_WIDTH * FINAL_BITMAP_HEIGHT];
+    for (size_t row = 0; row < FINAL_BITMAP_WIDTH; row++) {
+        for (size_t col = 0; col < FINAL_BITMAP_HEIGHT; col++) {
+            // check platform here
+            real_data[row*FINAL_BITMAP_WIDTH+col] = (uint8_t)(
+                data_ptr[row * MAC_OS_STRIDE + col]);
+        }
+    }
+    return flex_image((const char *)real_data,
```
znation: This function (and others that use CoreGraphics) looks like it might leak memory. All created CoreGraphics/CoreFoundation objects (like CGImageRef, CFDataRef, etc.) need to be released (CGImageRelease, CFRelease, etc.) to free their memory.

Take a look at the vega_renderer source code for examples of this. Let me know if you want help checking these codepaths in Instruments to see if they are leaking.

src/unity/toolkits/drawing_recognition/data_preparation.hpp
shantanuchhabra: Moved it to data_preparation.cpp

src/unity/toolkits/drawing_recognition/data_preparation.cpp
shantanuchhabra: That's a great point. Thank you for reminding me about that.

src/unity/toolkits/drawing_recognition/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+project(unity_toolkits)
+
+make_library(unity_drawing_recognition
```
nickjong: Note that you'll need to update to the new CMake structure that Zach merged, using OBJECT libraries instead.

src/unity/toolkits/drawing_recognition/class_registrations.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include "class_registrations.hpp"
+#include "data_preparation.hpp"
+
+#include <unity/lib/toolkit_function_macros.hpp>
+
+namespace turi {
+namespace drawing_recognition {
+
+BEGIN_FUNCTION_REGISTRATION
+REGISTER_FUNCTION(_drawing_recognition_prepare_data , "data" , "feature" , "target")
```
nickjong: Nit: don't leave spaces before the commas

src/unity/toolkits/drawing_recognition/data_preparation.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+namespace turi {
+namespace drawing_recognition {
+
+/**
+ *  This function is responsible for converting stroke-based drawing data
+ *  to 28x28 bitmaps to make it ready for training with the Neural Network.
+ *  
+ * 
+ * \param[in] data 					SFrame from the user, which contains
```
nickjong: I don't think we need that much whitespace between the parameter name and its definition

Should we document here the expected input format of the SFrame?

src/unity/toolkits/drawing_recognition/data_preparation.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ *  This function is responsible for converting stroke-based drawing data
+ *  to 28x28 bitmaps to make it ready for training with the Neural Network.
+ *  
+ * 
+ * \param[in] data 					SFrame from the user, which contains 
+ *									stroke-based drawings
+ * \param[in] feature				Name of the feature column
+ * \param[in] target				Name of the target (label) column
```
nickjong: What does this function do with the target column?

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+#include <math.h>
+#include <iomanip>
+#include <logger/assertions.hpp>
+#include <image/image_type.hpp>
+#ifdef __APPLE__
+#include <CoreGraphics/CoreGraphics.h>
+#endif // __APPLE__
+#include "data_preparation.hpp"
+#include "random/random.hpp"
+#include "util/sys_util.hpp"
```
nickjong: It would be nice to clean up the list of includes. For example, it can be helpful to group these into system headers versus Turi headers. It seems weird to use angle brackets for some Turi headers and quotes for others, especially for those that don't live in the same directory as this file

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    }
+    L->a = L->slope;
+    L->b = -1;
+    L->c = start_y - L->slope * start_x;
+}
+
+template<typename T>
+std::vector<T> slice(std::vector<T> const &v, int m, int n) {
+    auto first = v.cbegin() + m;
+    auto last = v.cbegin() + n + 1;
```
nickjong: I recommend embracing the C++ convention of specifying ranges with the second value as being "one past the last value in the range"

Note that at the two call sites for this function, you're subtracting 1 before passing in as `n`, then adding 1 back here.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+        if (d > dmax) {
+            index = i;
+            dmax = d;
+        }
+    }
+    if (dmax > epsilon) {
+        flex_list first_slice = slice(stroke, 0, index-1);
+        flex_list last_slice = slice(stroke, index, num_points-1);
+        compressed_stroke = ramer_douglas_peucker(first_slice, epsilon);
+        flex_list rec_results_2 = ramer_douglas_peucker(last_slice, epsilon);
```
nickjong: Interesting algorithm. This implementation performs a number of unnecessary memory allocations, but honestly it may or may not be important to optimize those away. We should at least handle the low-hanging fruit: the function can take a `const flex_list&` instead of `flex_list` to avoid copying `first_slice` and `last_slice` on each recursive call.

FYI, the optimized way to do this would be to have this function take as an argument an iterator range into a `flex_list`, instead of a manually construct slice. Even you also use an output iterator to write the results, instead of returning a `flex_list`, then the only allocations at all would be for constructing the final output value. This would also avoid construct a large number of `flex_dict` values for all the points.

This would be a good exercise in learning how to work with C++ iterators, or could be needed if this is a performance bottleneck.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    } else {
+        compressed_stroke.push_back(first_point);
+        compressed_stroke.push_back(last_point);
+    }
+    return compressed_stroke;
+}
+
+flex_list simplify_drawing(flex_list raw_drawing) {
+    size_t num_strokes = raw_drawing.size();
+    size_t min_x = INT_MAX;
```
nickjong: In C++ code, prefer [std::numeric_limits<size_t>](https://en.cppreference.com/w/cpp/types/numeric_limits).

In particular, note that `INT_MAX` is not the maximum value that `size_t` can store

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    Point P;
+    // Compute bounds of the drawing
+    for (size_t i = 0; i < num_strokes; i++) {
+        const flex_list &stroke = raw_drawing[i].get<flex_list>();
+        for (size_t j = 0; j < stroke.size(); j++) {
+            const flex_dict &point = stroke[j].get<flex_dict>();
+            make_point(&P, point);
+            float x = P.x;
+            float y = P.y;
+            if (x < min_x) min_x = x;
```
nickjong: This would be a good place to use `std::min` and `std::max`

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+            if (x > max_x) max_x = x;
+            if (y < min_y) min_y = y;
+            if (y > max_y) max_y = y;
+        }
+    }
+    // Align the drawing to top-left corner and scale to [0,255]
+    for (size_t i = 0; i < num_strokes; i++) {
+        flex_list &stroke = raw_drawing[i].mutable_get<flex_list>();
```
nickjong: You're using `mutable_get` but not actually mutating the result. In this case, since `raw_drawing` was passed by value, you could actually just mutate the result. This would save on a number of allocations.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+                new_y = ((P.y - min_y) * 255.0) / (max_y - min_y);
+            }
+            flex_dict new_point;
+            new_point.push_back(std::make_pair("x", new_x));
+            new_point.push_back(std::make_pair("y", new_y));
+            new_stroke.push_back(new_point);
+        }
+        // Apply RDP Line Algorithm
+        flex_list compressed_stroke = ramer_douglas_peucker(new_stroke, 2.0);
+        simplified_drawing.push_back(compressed_stroke);
```
nickjong: May as well do one line to avoid an unnecessary copy:
```
simplified_drawing.push_back(ramen_douglas_peucker(new_stroke, 2.0f);
```
Or you can use `std::move` if you prefer two lines for clarity

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    }
+    return simplified_drawing;
+}
+
+void swap_points(Point *P1, Point *P2) {
```
nickjong: `Point` is a simple value type, so you can just use [std::swap](https://en.cppreference.com/w/cpp/algorithm/swap)

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    int pad = (int)(stroke_width/2);
+    size_t dimension = bitmap.shape()[1];
+    if (x1 == x2) {
+        // just paint the y axis
+        int min_y = MIN(y1, y2);
+        int max_y = MAX(y1, y2);
+        for (int y = min_y; y < max_y; y++) {
+            for (int dx = -pad; dx < pad; dx++) {
+                for (int dy = -pad; dy < pad; dy++) {
+                    // figure out how to do this with the flex_list
```
nickjong: What does this mean?

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    if (in_bounds(x1+dx, y+dy, dimension)) {
+                        bitmap[(y+dy)*dimension + (x1+dx)] = 1.0;
+                    }
+                }
+            }
+        }
+    } else {
+        for (int x = x1; x < x2; x++) {
+            int y = (int)(slope * (x - x1) + y1);
+            for (int dx = -pad; dx < pad; dx++) {
```
nickjong: I feel like you could factor out a helper function (or maybe just a lambda) that draws a particular (x,y) point (plus padding)

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    // figure out how to do this with the flex_list
+                    if (in_bounds(x1+dx, y+dy, dimension)) {
+                        bitmap[(y+dy)*dimension + (x1+dx)] = 1.0;
+                    }
+                }
+            }
+        }
+    } else {
+        for (int x = x1; x < x2; x++) {
+            int y = (int)(slope * (x - x1) + y1);
```
nickjong: If the slope is steep, doesn't this lead to gaps? For example, if `x1 + 1 == x2` then you're only drawing at two values of `y`, regardless of how far apart they are.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+                blurred_bitmap[index] = (int)(255*bitmap[index]);
+                continue;
+            }
+            double sum_for_blur = 0.0;
+            for (int dr = -pad; dr <= pad; dr++) {
+                for (int dc = -pad; dc <= pad; dc++) {
+                    sum_for_blur += bitmap[(row+dr) * dimension + (col+dc)];
+                }
+            }
+            blurred_bitmap[index] = (int)(255*sum_for_blur/(ksize*ksize));
```
nickjong: You should guard again assigning values greater than 255 here.

Note that if `ksize` is 2, then `pad` is 1, and `ksize*ksize` is 4 but you're adding 9 values here.

nickjong: Huh, why are you casting to int here? `flex_nd_vec` stores doubles

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+        }
+    }
+    uint8_t image_data[dimension * dimension];
+    for (int idx=0; idx < dimension * dimension; idx++) {
+        image_data[idx] = ((uint8_t)(blurred_bitmap[idx]));
```
nickjong: Maybe this is where you should guard against values greater than 255.0

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    }
+    return flex_image((const char *)image_data,
+        dimension, dimension, 1, dimension*dimension, 1, 2);
```
nickjong: Comment the non-self-explanatory arguments?

Note that the pen-ultimate argument is version, but `IMAGE_TYPE_CURRENT_VERSION` is 0.

nickjong: (You should probably just use the constant)

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+}
+
+flex_image rasterize(flex_list simplified_drawing) {
+    flex_image final_bitmap; // 1 x 28 x 28
+    size_t num_strokes = simplified_drawing.size();
+    int INTERMEDIATE_BITMAP_WIDTH = 256;
```
nickjong: Constants (e.g. anything you're naming in all caps) should be declared e.g.
```
static constexpr int INTERMEDIATE_BITMAP_WIDTH = 256;
```

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    int INTERMEDIATE_BITMAP_HEIGHT = 256;
+    int FINAL_BITMAP_WIDTH = 28;
+    int FINAL_BITMAP_HEIGHT = 28;
+    float STROKE_WIDTH = 20.0;
+    int MAC_OS_STRIDE = 64;
+#ifdef __APPLE__
```
nickjong: I would really recommend extracting platform-specific code out into a separate file. Then this implementation can be tested explicitly. (For example, a Mac could explicitly test both this implementation and the portable one.)

So this could just be
```
#ifdef __APPLE__
return call_out_to_apple_specific_version(args)
#endif

File: src/unity/toolkits/drawing_recognition/class_registrations.cpp (hunk 10 above)
Comment (shantanuchhabra) [reply src/unity/toolkits/drawing_recognition/class_registrations.cpp]: Done

File: src/unity/toolkits/drawing_recognition/CMakeLists.txt (hunk 9 above)
Comment (shantanuchhabra) [reply src/unity/toolkits/drawing_recognition/CMakeLists.txt]: Done

File: src/unity/toolkits/drawing_recognition/data_preparation.cpp [hunk 29]
```diff
@@ -0,0 +1,382 @@
...
+#include <CoreGraphics/CoreGraphics.h>
+#endif // __APPLE__
+#include "data_preparation.hpp"
+#include "random/random.hpp"
+#include "util/sys_util.hpp"
+
+namespace turi {
+namespace drawing_recognition {
+
+#define MIN(a,b) (((a)<(b))?(a):(b))
```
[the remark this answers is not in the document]
shantanuchhabra: Done

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+struct Line {
+    // ax + by + c = 0
+    float slope;
+    float a;
+    float b;
+    float c;
+};
+
+void make_point(Point *P, const flex_dict &point_dict) {
+    flex_string x ("x");
```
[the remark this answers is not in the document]
shantanuchhabra: Done.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
shantanuchhabra: I did do a refactor of the function. Are there any more optimizations I could do?

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -0,0 +1,382 @@
...
+    for (auto key_value_pair: point_dict) {
+        const flex_string &key = key_value_pair.first.get<flex_string>();
+        const flex_float &value = key_value_pair.second.get<flex_float>();
+        if (key.compare(x) == 0) {
+            P->x = value;
+        } else {
```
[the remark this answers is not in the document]
shantanuchhabra: Added it here.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
shantanuchhabra: Should be fixed now, per the strategy we discussed earlier today.

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -3,106 +3,145 @@
...
+            return floorf(distance);
+        }
+    };
+
+} // namespace
+
+/* Ramer Douglas Peucker Algorithm:
+ * https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm
+ */
+flex_list ramer_douglas_peucker(flex_list &stroke,
```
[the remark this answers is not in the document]
shantanuchhabra: Yep, we don't!

src/unity/toolkits/drawing_recognition/data_preparation.cpp
```diff
@@ -3,106 +3,145 @@
...
     flex_list compressed_stroke;
-    int num_points = stroke.size();
-    Line L;
-    const flex_dict &first_point = stroke[0].get<flex_dict>();
-    const flex_dict &last_point = stroke[num_points-1].get<flex_dict>();
-    initialize_line(&L, first_point, last_point);
-    for (int i = 0; i < num_points; i++) {
-        float d = line_to_point_distance(&L, stroke[i].get<flex_dict>());
+    const flex_dict &first_point = (*begin).get<flex_dict>();
+    const flex_dict &last_point = (*(end-1)).get<flex_dict>();
```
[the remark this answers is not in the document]
shantanuchhabra: Added a guard here, and also where this function is called from simplify_drawing.

PR #1450, opened 2019-02-23
Audio Preprocessing

TobyRoseman:
* External code: VGGish preprocessing
* Core ML custom model doing the same preprocessing
* Unit test to make sure they produce the same results

### DIFF
--- a/src/objc/AudioPreprocessing.h
+++ b/src/objc/AudioPreprocessing.h
@@ -6,19 +6,22 @@
 
 #import <CoreML/CoreML.h>
 
+NS_ASSUME_NONNULL_BEGIN
 
-__attribute__((visibility("default")))
+API_AVAILABLE(macos(10.13))
 @interface TCSoundClassifierPreprocessing : NSObject<MLCustomModel>
 
 @property(readonly, nullable) NSString* inputFeatureName;
 @property(readonly, nullable) NSString* outputFeatureName;
 
-- (nullable instancetype)initWithModelDescription:(MLModelDescription * _Null_unspecified)modelDescription
-                              parameterDictionary:(NSDictionary<NSString *, id> * _Null_unspecified)parameters
-                                            error:(NSError * _Nullable * _Nullable)error API_AVAILABLE(macos(10.13));
+- (nullable instancetype)initWithModelDescription:(MLModelDescription*)modelDescription
+                              parameterDictionary:(NSDictionary<NSString *, id>*)parameters
+                                            error:(NSError**)error;
 
-- (nullable id<MLFeatureProvider>)predictionFromFeatures:(id<MLFeatureProvider> _Nonnull)input
-                                                 options:(MLPredictionOptions * _Null_unspecified)options
-                                                 error:(NSError * _Nullable * _Nullable)error API_AVAILABLE(macos(10.13));
+- (nullable id<MLFeatureProvider>)predictionFromFeatures:(id<MLFeatureProvider>)input
+                                                 options:(MLPredictionOptions*)options
+                                                   error:(NSError**)error;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/src/objc/AudioPreprocessing.h
+++ b/src/objc/AudioPreprocessing.h
@@ -1,4 +1,4 @@
-/* Copyright © 2017 Apple Inc. All rights reserved.
+/* Copyright © 2019 Apple Inc. All rights reserved.
  *
  * Use of this source code is governed by a BSD-3-clause license that can
  * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
--- a/src/objc/AudioPreprocessing.h
+++ b/src/objc/AudioPreprocessing.h
@@ -1,0 +1,24 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#import <CoreML/CoreML.h>
+
+
+__attribute__((visibility("default")))
+@interface TCSoundClassifierPreprocessing : NSObject<MLCustomModel>
+
+@property(readonly, nullable) NSString* inputFeatureName;
+@property(readonly, nullable) NSString* outputFeatureName;
+
+- (nullable instancetype)initWithModelDescription:(MLModelDescription * _Null_unspecified)modelDescription
+                              parameterDictionary:(NSDictionary<NSString *, id> * _Null_unspecified)parameters
+                                            error:(NSError * _Nullable * _Nullable)error API_AVAILABLE(macos(10.13));
+
+- (nullable id<MLFeatureProvider>)predictionFromFeatures:(id<MLFeatureProvider> _Nonnull)input
+                                                 options:(MLPredictionOptions * _Null_unspecified)options
+                                                 error:(NSError * _Nullable * _Nullable)error API_AVAILABLE(macos(10.13));
+
+@end
--- a/src/objc/AudioPreprocessing.m
+++ b/src/objc/AudioPreprocessing.m
@@ -45,7 +45,7 @@ static inline float hzToMel(float freq) {
 
 @property (readonly) float* PowerSpectrum;
 
--(id)initWithFFTLength:(NSUInteger)fftLength frameLength:(NSUInteger)frameLength;
+-(nullable instancetype)initWithFFTLength:(NSUInteger)fftLength frameLength:(NSUInteger)frameLength;
 -(void)applyHammingWindow:(float*)signal numSamples:(NSUInteger)numSamples;
 -(void)calculatePowerSpectrum:(float*)signal numSamples:(NSUInteger)numSamples;
 
@@ -62,10 +62,11 @@ static inline float hzToMel(float freq) {
 }
 
 
--(id)initWithFFTLength:(NSUInteger)fftLength frameLength:(NSUInteger)frameLength {
+-(nullable instancetype)initWithFFTLength:(NSUInteger)fftLength frameLength:(NSUInteger)frameLength {
     
     self = [super init];
-    
+    if (!self) return nil;
+
     _fftLength = fftLength;
     _frameLength = frameLength;
     
@@ -125,11 +126,11 @@ static inline float hzToMel(float freq) {
 
 @interface TCSoundClassifierMelFrequencyFilterBank : NSObject
 
--(id)initWithMinFrequency:(float)minFreq
-             maxFrequency:(float)maxFrequency
-               sampleRate:(double)sampleRate
-              numMelBands:(NSUInteger)numMelBands
-                  numBins:(NSUInteger)numBins;
+-(nullable instancetype)initWithMinFrequency:(float)minFreq
+                                maxFrequency:(float)maxFrequency
+                                  sampleRate:(double)sampleRate
+                                 numMelBands:(NSUInteger)numMelBands
+                                     numBins:(NSUInteger)numBins;
 
 -(void)apply:(const float*)powerSpectrum
       output:(float *)out;
@@ -144,13 +145,14 @@ static inline float hzToMel(float freq) {
     float** _filterBanks;
 }
 
--(id)initWithMinFrequency:(float)minFreq
-             maxFrequency:(float)maxFreq
-               sampleRate:(double)sampleRate
-              numMelBands:(NSUInteger)numMelBands
-                  numBins:(NSUInteger)numBins {
+-(nullable instancetype)initWithMinFrequency:(float)minFreq
+                                maxFrequency:(float)maxFreq
+                                  sampleRate:(double)sampleRate
+                                 numMelBands:(NSUInteger)numMelBands
+                                     numBins:(NSUInteger)numBins {
 
     self = [super init];
+    if (!self) return nil;
 
     _minFreq = minFreq;
     _maxFreq = maxFreq;
--- a/src/objc/AudioPreprocessing.m
+++ b/src/objc/AudioPreprocessing.m
@@ -1,4 +1,4 @@
-/* Copyright © 2017 Apple Inc. All rights reserved.
+/* Copyright © 2019 Apple Inc. All rights reserved.
  *
  * Use of this source code is governed by a BSD-3-clause license that can
  * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
--- a/src/objc/AudioPreprocessing.m
+++ b/src/objc/AudioPreprocessing.m
@@ -216,8 +216,9 @@ static inline float hzToMel(float freq) {
             filterBank[k] = (k == 0) ? 0.0 : fmax(0.0, fmin(lowerBank[k], upperBank[k]));
         }
 
-        _filterBanks[i] = calloc(spectrumSize, sizeof(float));
         _filterBanks[i] = filterBank;
+        free(lowerBank);
+        free(upperBank);
     }
 
     return self;
--- a/src/objc/AudioPreprocessing.m
+++ b/src/objc/AudioPreprocessing.m
@@ -1,0 +1,365 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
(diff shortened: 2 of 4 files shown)


src/objc/AudioPreprocessing.h
```diff
@@ -0,0 +1,24 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: 2019

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
+/* Copyright © 2017 Apple Inc. All rights reserved.
```
znation: 2019

src/objc/CMakeLists.txt
```diff
@@ -15,8 +15,25 @@ if(APPLE AND HAS_MLCUSTOMMODEL)
...
+    REQUIRES
+      ${ACCELERATE}
+      ${_TC_COMMON_REQUIREMENTS}
+    SHARED
+    SHARED_ALL_DEFINED
+    EXPORT_OSX_MAP_FILE ${CMAKE_CURRENT_SOURCE_DIR}/symbol_exports.ver
+    DEAD_STRIP
+  )
+
+  # Adding sources here allows us to unit test these custom models in python.
```
znation: Thanks for adding this comment!

src/unity/python/turicreate/test/test_audio_functionality.py
```diff
@@ -0,0 +1,63 @@
...
+import unittest
+
+import coremltools
+from coremltools.proto import FeatureTypes_pb2
+import numpy as np
+
+from turicreate.toolkits.sound_classifier import vggish_input
+
+
+@unittest.skipIf(platform != 'darwin', 'Can not test something which usees Core ML.')
```
znation: `s/usees/uses/`

src/unity/python/turicreate/test/test_audio_functionality.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Apple Inc. All rights reserved.
```
znation: 2019

src/objc/CMakeLists.txt
```diff
@@ -15,8 +15,25 @@ if(APPLE AND HAS_MLCUSTOMMODEL)
     DEAD_STRIP
   )
 
+  find_library(ACCELERATE NAMES Accelerate)
```
znation: Since we sometimes share frameworks between different libraries within turicreate, I started moving these to the top level CMakeLists.txt. I think there may be one there already for `Accelerate`.

src/objc/AudioPreprocessing.h
TobyRoseman: Done

src/objc/AudioPreprocessing.m
TobyRoseman: Done

src/unity/python/turicreate/test/test_audio_functionality.py
TobyRoseman: done

src/unity/python/turicreate/test/test_audio_functionality.py
TobyRoseman: done

src/objc/CMakeLists.txt
TobyRoseman: Thanks for letting me know. It's even named exactly the same thing. Done.

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+
+    float scale = 0.5; // Factor of 2 needed due to vDSP_fft_zrip implementation detail.
+    vDSP_zvabs(&_spectrum, 1, _PowerSpectrum, 1, _fftLength/2);
+    vDSP_vsmul(_PowerSpectrum, 1, &scale, _PowerSpectrum, 1, _fftLength/2);
+}
+
+-(void)_generateHammingWindow:(NSUInteger)frameLength {
+    // Don't use vDSP_hamm_window because it uses a slightly different implementation
+    for(size_t i = 0; i < frameLength; i++) {
+        _hammingWindow[i] = 0.5 - 0.5*cos((2*M_PI*i)/(frameLength));
```
nickjong: Another place where you may as well use `float` values (and `cosf`) to minimize work given that you're only storing `float` precision anyway

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+    }
+}
+
+@end // Power Spectrum implementation
+
+
+@interface MelFrequencyFilterBank : NSObject
+
+@property (readonly) float* MelSpectrum;
```
nickjong: Property names generally begin lowercase

nickjong: Hmm, is this used anywhere?

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+                  numBins:(NSUInteger)numBins {
+
+    self = [super init];
+
+    _minFreq = minFreq;
+    _maxFreq = maxFreq;
+    _sampleRate = sampleRate;
+    _numMelBands = numMelBands;
+    _numBins = numBins;
+    _filterBanks = calloc(_numMelBands, sizeof(float *));
```
nickjong: Where is this deallocated?

nickjong: Note that you might consider performing allocations using `NSData` (https://developer.apple.com/documentation/foundation/nsdata?language=objc) so that it can handle memory management for you. It looks like _filterBanks could be `NSMutableArray<NSData*> *`

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+    _fftLength = fftLength;
+    _frameLength = frameLength;
+    
+    _hammingWindow = (float*)calloc(frameLength, sizeof(float));
+    [self _generateHammingWindow:frameLength];
+    
+    _fftSetup = vDSP_create_fftsetup(log2(fftLength*2), FFT_RADIX2);
+    _spectrum.realp = (float*)calloc(fftLength, sizeof(float));
+    _spectrum.imagp = (float*)calloc(fftLength, sizeof(float));
+    _PowerSpectrum = (float*)calloc(fftLength/2 + 1, sizeof(float));
```
nickjong: Strange, I don't understand where `_PowerSpectrum` is declared

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+}
+
+@end   // MelFrequencyFilterBank implementation
+
+
+
+@implementation TCSoundClassifierPreprocessing
+
+PowerSpectrum* _powerSpectrum;
+MelFrequencyFilterBank* _melFilterBank;
```
nickjong: Huh, I note that there aren't curly braces around these instance variables. Are those no longer needed these days? Or are these actually global variables?

Note that it's often cleaner to define internal ("private") variables as class extension properties, e.g.
```
@interface TCSoundClassifierPreprocessing ()
@property (nonatomic) PowerSpectrum *powerSpectrum;
@property (nonatomic) MelFrequenceFilterPank *melFilterBank;
@end
```
This will auto-synthesize underlying instance variables such as `_powerSpectrum` and allow you to use property syntax (e.g. `self.powerSpectrum`) outside of the initializer

src/objc/AudioPreprocessing.h
```diff
@@ -0,0 +1,24 @@
+/* Copyright © 2019 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#import <CoreML/CoreML.h>
+
+
+__attribute__((visibility("default")))
```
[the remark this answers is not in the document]
TobyRoseman: Done.

src/objc/AudioPreprocessing.h
```diff
@@ -0,0 +1,24 @@
...
+@interface TCSoundClassifierPreprocessing : NSObject<MLCustomModel>
+
+@property(readonly, nullable) NSString* inputFeatureName;
+@property(readonly, nullable) NSString* outputFeatureName;
+
+- (nullable instancetype)initWithModelDescription:(MLModelDescription * _Null_unspecified)modelDescription
+                              parameterDictionary:(NSDictionary<NSString *, id> * _Null_unspecified)parameters
+                                            error:(NSError * _Nullable * _Nullable)error API_AVAILABLE(macos(10.13));
```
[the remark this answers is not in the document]
TobyRoseman: Thanks for the link, learned quite a bit.

Done.

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+
+#include <math.h>
+
+#import <Accelerate/Accelerate.h>
+
+#import "AudioPreprocessing.h"
+
+
+// TODO: Make some of these parameters to the model
+const int inputLength = 15600;
```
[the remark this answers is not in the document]
TobyRoseman: Great idea. Done.

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+static void setErrorMsg(const char* errorMsg,  NSError** error) {
+    if(error) {
+      *error = [NSError errorWithDomain:MLModelErrorDomain
+                                   code:MLModelErrorGeneric
+                               userInfo:@{NSLocalizedDescriptionKey: @(*errorMsg)}];
+    }
+}
+
+static inline float hzToMel(float freq) {
+    return 1127.0 * log(1 + freq/700.0f);
```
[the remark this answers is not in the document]
TobyRoseman: Good point. I've changed: `log` -> `logf` and `1127.0` -> `1127.0f`.

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+}
+
+
+@interface PowerSpectrum : NSObject
```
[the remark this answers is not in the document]
TobyRoseman: Ok, for consistency I prepended `TCSoundClassifier` (rather than just two letters) to this class and `MelFrequencyFilterBank`.

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+
+@property (readonly) float* PowerSpectrum;
+
+-(id)initWithFFTLength:(NSUInteger)fftLength withFrameLength:(NSUInteger)frameLength;
```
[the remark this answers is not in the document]
TobyRoseman: Done.

src/objc/AudioPreprocessing.m
TobyRoseman: Couldn't find a float version of `M_PI`, but I have changed the two `0.5` to `0.5f` and the `cos` to `cosf`.

src/objc/AudioPreprocessing.m
TobyRoseman: It is not being used. Good catch. I refactored this out to save some memory. I've removed it.

src/objc/AudioPreprocessing.m
TobyRoseman: Good catch. I've added a `dealloc`. Please take a look at it.

I actually think I probably just shouldn't be dynamically allocating any memory in this file (i.e. no calloc/malloc). I **think** I could just do everything with variable sized arrays. What do you think?

src/objc/AudioPreprocessing.m
TobyRoseman: I don't know what's going on here either. I've added curly braces.

src/objc/AudioPreprocessing.m
TobyRoseman: Yeah, as you mentioned in a later comment, I think it was actually a global. That's fixed now.

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+    float *_hammingWindow, *_workingFrame;
+
+    DSPSplitComplex _spectrum;
+    FFTSetup _fftSetup;
+    
+    NSUInteger _frameLength, _fftLength;
+}
+
+
+-(id)initWithFFTLength:(NSUInteger)fftLength withFrameLength:(NSUInteger)frameLength {
```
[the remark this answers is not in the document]
TobyRoseman: Ok, changed it to `nullable instancetype`. Also made the same change for `TCSoundClassifierMelFrequencyFilterBank`

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,365 @@
...
+    
+    self = [super init];
```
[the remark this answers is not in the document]
TobyRoseman: Ok went with `if (!self) return nil;`. I'll pretty much always take an early return over another layer of nesting. Changed it in the other initializer too.

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,370 @@
...
+        vDSP_vsmul(spectrogramBins, 1, &negative_one, upperBank, 1, spectrumSize);
+        vDSP_vsadd(upperBank, 1, &add, upperBank, 1, spectrumSize);
+        vDSP_vsdiv(upperBank, 1, &div, upperBank, 1, spectrumSize);
+
+        // find intersection
+        for(int k = 0; k < spectrumSize; k++) {
+            filterBank[k] = (k == 0) ? 0.0 : fmax(0.0, fmin(lowerBank[k], upperBank[k]));
+        }
+
+        _filterBanks[i] = calloc(spectrumSize, sizeof(float));
```
nickjong: Something I missed before: this line is a dead assignment, right? You overwrite it in the next line (and leak the allocation).

src/objc/AudioPreprocessing.m
```diff
@@ -0,0 +1,370 @@
...
+    // For each band edge, create a new filter (basically a float array that we vDSP_dot later)
+    for(size_t i = 0; i < _numMelBands; i++) {
+        // Grab the low, center, and bottom edges
+        float lower_edge = melBandEdges[i];
+        float center_edge = melBandEdges[i+1];
+        float upper_edge = melBandEdges[i+2];
+
+        // Compute lower and upper slopes
+        float* lowerBank = (float*)calloc(spectrumSize, sizeof(float));
+        float* upperBank = (float*)calloc(spectrumSize, sizeof(float));
```
nickjong: I don't think lowerBank and upperBank are being freed anywhere, right?

src/objc/AudioPreprocessing.m
nickjong: By variable sized array do you mean e.g.
```
int size = ...;
float my_array[size];
```
?

Note that this is not actually part of the C++ standard (AFAIK), but it just a popular extension that popular compilers have implemented. Since this code will only be compiled on Apple platforms, it may be okay if it compiles now given our current toolchain, but I think you may as well stick with heap allocations, given that you need one anyway to store _filterBanks from initializer to calls to `apply`. 

FWIW, I think it's fairly standard to have e.g.
```
NSData *lowerBankData = [[NSData alloc] initWithLength:spectrumSize*sizeof(float)];
float* lowerBank = (float*)lowerBankData.mutableBytes;
```
Then Objective C will take care of memory management for you, acknowledging that the `lowerBank` pointer will only remain valid as long as `lowerBankData` is around.

TobyRoseman: Yes, I believe we're talking about the same thing. Basically the value of `size` isn't know until runtime. From what I can tell, it's been a part of C since C99 but only a part of C++ since C++11.

Good point about this approach not working because of needing to store the values between the initializer and the call to `apply`. Clearly variable sized arrays are not going to work here.

src/objc/AudioPreprocessing.m
TobyRoseman: Yes, you're right. Fixed. Thanks.

src/objc/AudioPreprocessing.m
TobyRoseman: Yeah, I am immediately leaking that memory. That's pretty bad. I don't like managing my own memory :stuck_out_tongue_closed_eyes:

nickjong: Ha, yeah, there's a reason that every other surviving programming language has automatic memory management!

Issue #1350, opened 2019-01-23
Suggested improvement to syntax of .unpack()


nicoinn:
I tend to face quite often the need to convert a list of dictionaries to an SFrame...

It's actually quite easy - just do
` tc.SFrame([dict1,dict2]).unpack("X1",column_name_prefix="")`

The first part will create a single column SFrame with a column name "X1" and the second will unpack "X1" into separate columns. The column_name_prefix="" means that I do not want any "X1" prefix anywhere.  

Now - imagine an unexperienced person looking at this line out of context --- guaranteed "WTF effect"!!!   The "X1" seems to come from nowhere!

The following would much clearer to read, and faster to write: 

` tc.SFrame([dict1,dict2]).unpack()`

The assumption there being that if an SFrame has only one column, unpack that one without prefix. Otherwise, throw an error!

I don't think that it is currently possible to use .unpack() without a column name argument --- so, such an alteration of the default behaviour should not break any code (?).

hoytak:
That's a great request -- I like it!

znation:
Fixed with #1456.

PR #1456, opened 2019-02-25
#1350

shreyajain17:
- made column name in .unpack optional 
- unit test for the same

### DIFF
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -4866,8 +4866,11 @@ class SFrame(object):
 
         Parameters
         ----------
-        column_name : str
-            Name of the unpacked column
+        column_name : str, optional
+            If provided, it unpacks the column with the given name. If not provided
+            and only one column is present then the column is unpacked. In case of 
+            multiple columns, name must be provided to know which column to be unpacked.
+        
 
         column_name_prefix : str, optional
             If provided, unpacked column names would start with the given
@@ -4973,6 +4976,19 @@ class SFrame(object):
         | 3  |    3.0    |    4.0    |    5.0    |
         +----+-----------+-----------+-----------+
         [3 rows x 4 columns]
+        
+        >>> sf = turicreate.SFrame([{'a':1,'b':2,'c':3},{'a':4,'b':5,'c':6}])
+        >>> sf.unpack()
+        +---+---+---+
+        | a | b | c |
+        +---+---+---+
+        | 1 | 2 | 3 |
+        | 4 | 5 | 6 |
+        +---+---+---+
+        [2 rows x 3 columns]
+
+
+
         """
         if column_name==None:
             if len(self.column_names())==1:
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -4994,7 +4994,7 @@ class SFrame(object):
         if column_name is None:
             if self.num_columns()==0:
                 raise RuntimeError("No column exists in the current SFrame")
-            unpack_columns,column_number=0,0
+            
             for t in range(self.num_columns()):
                 column_type = self.column_types()[t]
                 if column_type==type(dict()) or column_type==type(list()) or column_type==type(array.array('c','he')):
@@ -5003,17 +5003,18 @@ class SFrame(object):
                         column_name = self.column_names()[t]
                     else:
                         raise RuntimeError("Column name needed to unpack")
-            if column_name is not None and column_name_prefix is None:
-                    column_name_prefix=""
+            
 
             if column_name is None:
                 raise RuntimeError("No columns can be unpacked")
-
-        if column_name not in self.column_names():
+            elif column_name_prefix is None:
+                column_name_prefix=""
+        elif column_name not in self.column_names():
             raise KeyError("Column '" + column_name + "' does not exist in current SFrame")
 
         if column_name_prefix is None:
             column_name_prefix = column_name
+
 
         new_sf = self[column_name].unpack(column_name_prefix, column_types, na_value, limit)
 
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -4867,9 +4867,10 @@ class SFrame(object):
         Parameters
         ----------
         column_name : str, optional
-            If provided, it unpacks the column with the given name. If not provided
-            and only one column is present then the column is unpacked. In case of 
-            multiple columns, name must be provided to know which column to be unpacked.
+            Name of the unpacked column, if provided. If not provided
+            and only one column is present then the column is unpacked. 
+            In case of multiple columns, name must be provided to know 
+            which column to be unpacked.
         
 
         column_name_prefix : str, optional
@@ -4992,17 +4993,24 @@ class SFrame(object):
         """
         if column_name is None:
             if self.num_columns()==0:
-                raise RuntimeError("no column exists in the current SFrame")
-            if self.num_columns()==1:
-                column_name=self.column_names()[0]
-                if column_name_prefix==None:
+                raise RuntimeError("No column exists in the current SFrame")
+            unpack_columns,column_number=0,0
+            for t in range(self.num_columns()):
+                column_type = self.column_types()[t]
+                if column_type==type(dict()) or column_type==type(list()) or column_type==type(array.array('c','he')):
+                    
+                    if column_name is None:
+                        column_name = self.column_names()[t]
+                    else:
+                        raise RuntimeError("Column name needed to unpack")
+            if column_name is not None and column_name_prefix is None:
                     column_name_prefix=""
-            else:
-                raise RuntimeError("column name needed to unpack")
-        
+
+            if column_name is None:
+                raise RuntimeError("No columns can be unpacked")
 
         if column_name not in self.column_names():
-            raise KeyError("column '" + column_name + "' does not exist in current SFrame")
+            raise KeyError("Column '" + column_name + "' does not exist in current SFrame")
 
         if column_name_prefix is None:
             column_name_prefix = column_name
--- a/src/unity/python/turicreate/data_structures/sframe.py
+++ b/src/unity/python/turicreate/data_structures/sframe.py
@@ -4990,13 +4990,11 @@ class SFrame(object):
 
 
         """
-        if column_name==None:
-            if len(self.column_names())==1:
+        if column_name is None:
+            if self.num_columns()==1:
+                column_name=self.column_names()[0]
                 if column_name_prefix==None:
-                    column_name=self.column_names()[0]
                     column_name_prefix=""
-                else:
-                    column_name=self.column_names()[0]
             else:
                 raise RuntimeError("column name needed to unpack")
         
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4973,7 +4976,31 @@ def unpack(self, column_name, column_name_prefix=None, column_types=None,
...
+        +---+---+---+
+        | 1 | 2 | 3 |
+        | 4 | 5 | 6 |
+        +---+---+---+
+        [2 rows x 3 columns]
+
+
+
         """
+        if column_name==None:
```
TobyRoseman: Since there is only one `None` value, generally `is` is preferred when checking against `None`. 
So this is probably better: `if column_name is None:`

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4973,7 +4976,31 @@ def unpack(self, column_name, column_name_prefix=None, column_types=None,
...
+            if len(self.column_names())==1:
+                if column_name_prefix==None:
+                    column_name=self.column_names()[0]
```
TobyRoseman: This line is in both the if and else block. You could move it outside and get rid of the else block.

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4973,7 +4976,31 @@ def unpack(self, column_name, column_name_prefix=None, column_types=None,
...
+        | 1 | 2 | 3 |
+        | 4 | 5 | 6 |
+        +---+---+---+
+        [2 rows x 3 columns]
+
+
+
         """
+        if column_name==None:
+            if len(self.column_names())==1:
```
TobyRoseman: I would just do: `if self.num_columns() == 1:`

I wonder if maybe we need a check, and a separate error message, for when the number of columns is zero.

shreyajain17: Cool. I will make the changes. I think it returns a KeyError when there are zero columns. This was already in code though.

shreyajain17: I get it now the case of zero columns.

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4866,8 +4866,11 @@ def unpack(self, column_name, column_name_prefix=None, column_types=None,
 
         Parameters
         ----------
-        column_name : str
-            Name of the unpacked column
+        column_name : str, optional
+            If provided, it unpacks the column with the given name. If not provided
```
nickjong: Nit: this reads a little strangely to me, since the antecedent of "it" isn't immediately clear. Consider just augmenting the original comment, e.g. "Name of the unpacked column, if provided. If not provided..." (or something like that)

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4973,7 +4976,31 @@ def unpack(self, column_name, column_name_prefix=None, column_types=None,
...
+        | 4 | 5 | 6 |
+        +---+---+---+
+        [2 rows x 3 columns]
+
+
+
         """
+        if column_name is None:
+            if self.num_columns()==0:
+                raise RuntimeError("no column exists in the current SFrame")
```
nickjong: Nit: our error messages usually start with a capitalized letter

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4973,7 +4976,31 @@ def unpack(self, column_name, column_name_prefix=None, column_types=None,
...
+            if self.num_columns()==1:
```
nickjong: Should we also support the case where the SFrame has only one column that can be unpacked (type is list/array/dict)?

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4992,17 +4993,24 @@ def unpack(self, column_name=None, column_name_prefix=None, column_types=None,
         """
         if column_name is None:
             if self.num_columns()==0:
-                raise RuntimeError("no column exists in the current SFrame")
-            if self.num_columns()==1:
-                column_name=self.column_names()[0]
-                if column_name_prefix==None:
+                raise RuntimeError("No column exists in the current SFrame")
+            unpack_columns,column_number=0,0
```
nickjong: We can remove this line now

src/unity/python/turicreate/data_structures/sframe.py
```diff
@@ -4992,17 +4993,24 @@ def unpack(self, column_name=None, column_name_prefix=None, column_types=None,
...
+                        column_name = self.column_names()[t]
+                    else:
+                        raise RuntimeError("Column name needed to unpack")
+            if column_name is not None and column_name_prefix is None:
                     column_name_prefix=""
-            else:
-                raise RuntimeError("column name needed to unpack")
-        
+
+            if column_name is None:
```
nickjong: I think you can simplify the logic slightly here:
```
if column_name is None:
    raise RuntimeError(...)
elif column_name_prefix is None:
    column_name_prefix=""

(Looks like you need to update the indent level of `column_name_prefix=""`, too)

File: src/unity/python/turicreate/data_structures/sframe.py [hunk 9]
```diff
@@ -4992,17 +4993,24 @@ def unpack(self, column_name=None, column_name_prefix=None, column_types=None,
...
+                raise RuntimeError("No columns can be unpacked")
 
         if column_name not in self.column_names():
```
nickjong: Nit: maybe this should go into an `elif` from the `if` statement in line 4994, since if we deduced the column name automatically, then it's definitely in `self.column_names()`. What do you think?

shreyajain17:
Hi

I think it failed in Travis. I don’t know why though.

znation:
The failure is in Python 3.5:

```
turicreate/test/test_sframe.py:2321: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    def unpack(self, column_name=None, column_name_prefix=None, column_types=None,
               na_value=None, limit=None):
        """
            Expand one column of this SFrame to multiple columns with each value in
            a separate column. Returns a new SFrame with the unpacked column
            replaced with a list of new columns.  The column must be of
            list/array/dict type.
    
            For more details regarding name generation, missing value handling and
            other, refer to the SArray version of
            :py:func:`~turicreate.SArray.unpack()`.
    
            Parameters
            ----------
            column_name : str, optional
                Name of the unpacked column, if provided. If not provided
                and only one column is present then the column is unpacked.
                In case of multiple columns, name must be provided to know
                which column to be unpacked.
    
    
            column_name_prefix : str, optional
                If provided, unpacked column names would start with the given
                prefix. If not provided, default value is the name of the unpacked
                column.
    
            column_types : [type], optional
                Column types for the unpacked columns.
                If not provided, column types are automatically inferred from first
                100 rows. For array type, default column types are float.  If
                provided, column_types also restricts how many columns to unpack.
    
            na_value : flexible_type, optional
                If provided, convert all values that are equal to "na_value" to
                missing value (None).
    
            limit : list[str] | list[int], optional
                Control unpacking only a subset of list/array/dict value. For
                dictionary SArray, `limit` is a list of dictionary keys to restrict.
                For list/array SArray, `limit` is a list of integers that are
                indexes into the list/array value.
    
            Returns
            -------
            out : SFrame
                A new SFrame that contains rest of columns from original SFrame with
                the given column replaced with a collection of unpacked columns.
    
            See Also
            --------
            pack_columns, SArray.unpack
    
            Examples
            ---------
            >>> sf = turicreate.SFrame({'id': [1,2,3],
            ...                      'wc': [{'a': 1}, {'b': 2}, {'a': 1, 'b': 2}]})
            +----+------------------+
            | id |        wc        |
            +----+------------------+
            | 1  |     {'a': 1}     |
            | 2  |     {'b': 2}     |
            | 3  | {'a': 1, 'b': 2} |
            +----+------------------+
            [3 rows x 2 columns]
    
            >>> sf.unpack('wc')
            +----+------+------+
            | id | wc.a | wc.b |
            +----+------+------+
            | 1  |  1   | None |
            | 2  | None |  2   |
            | 3  |  1   |  2   |
            +----+------+------+
            [3 rows x 3 columns]
    
            To not have prefix in the generated column name:
    
            >>> sf.unpack('wc', column_name_prefix="")
            +----+------+------+
            | id |  a   |  b   |
            +----+------+------+
            | 1  |  1   | None |
            | 2  | None |  2   |
            | 3  |  1   |  2   |
            +----+------+------+
            [3 rows x 3 columns]
    
            To limit subset of keys to unpack:
    
            >>> sf.unpack('wc', limit=['b'])
            +----+------+
            | id | wc.b |
            +----+------+
            | 1  | None |
            | 2  |  2   |
            | 3  |  2   |
            +----+------+
            [3 rows x 3 columns]
    
            To unpack an array column:
    
            >>> import array
            >>> sf = turicreate.SFrame({'id': [1,2,3],
            ...                       'friends': [array.array('d', [1.0, 2.0, 3.0]),
            ...                                   array.array('d', [2.0, 3.0, 4.0]),
            ...                                   array.array('d', [3.0, 4.0, 5.0])]})
            >>> sf
            +-----------------+----+
            |     friends     | id |
            +-----------------+----+
            | [1.0, 2.0, 3.0] | 1  |
            | [2.0, 3.0, 4.0] | 2  |
            | [3.0, 4.0, 5.0] | 3  |
            +-----------------+----+
            [3 rows x 2 columns]
    
            >>> sf.unpack('friends')
            +----+-----------+-----------+-----------+
            | id | friends.0 | friends.1 | friends.2 |
            +----+-----------+-----------+-----------+
            | 1  |    1.0    |    2.0    |    3.0    |
            | 2  |    2.0    |    3.0    |    4.0    |
            | 3  |    3.0    |    4.0    |    5.0    |
            +----+-----------+-----------+-----------+
            [3 rows x 4 columns]
    
            >>> sf = turicreate.SFrame([{'a':1,'b':2,'c':3},{'a':4,'b':5,'c':6}])
            >>> sf.unpack()
            +---+---+---+
            | a | b | c |
            +---+---+---+
            | 1 | 2 | 3 |
            | 4 | 5 | 6 |
            +---+---+---+
            [2 rows x 3 columns]
    
    
    
            """
        if column_name is None:
            if self.num_columns()==0:
                raise RuntimeError("No column exists in the current SFrame")
    
            for t in range(self.num_columns()):
                column_type = self.column_types()[t]
>               if column_type==type(dict()) or column_type==type(list()) or column_type==type(array.array('c','he')):
E               TypeError: cannot use a str to initialize an array with typecode 'c'
turicreate/data_structures/sframe.py:5000: TypeError
```

Looks like it may be related to the changes in this PR. Though I don't know why it only happened on Python 3.5, and not 2.7 or 3.6.

TobyRoseman:
> Hi I think it failed in Travis. I don’t know why though.
> […](#)

Yeah, this isn't straightforward to figure. Click on `Details` in the Travis CI Section below. Under `Jobs and Stages` you'll see there is only a red x next to 3.5; so it's only failing on Python 3.5. Click thru to that build and scroll down to almost the bottom. 

It's not liking the `array.array('c','he')` on line 7860 of your sframe.py.

I'm not really sure what's suppose to be happening there. Rather than `column_type==type(array.array('c','he'))` could you just do `column_type==type(array.array)`? That works in Python 3.5.

shreyajain17:
The type(array.array) didn’t work for me in 2.7 though. I will to figure out the right syntax in 3.5 and try again.

Issue #1455, opened 2019-02-25
mxnet < 1.4.0 brings in an unsupported requests version


znation:
Current mxnet versions (prior to 1.4) bring in a version of requests that is unsupported and has reported vulnerabilities. We should upgrade our max dependency to 1.4.0 to fix this.

znation:
We should also take out the requests dependency that the coremltools source drop brought in. This one is not even used anywhere.

PR #1458, opened 2019-02-25
Drop requests dependency from coremltools

znation:
Port of apple/coremltools#331


znation:
This is part of the fix for #1455.

PR #1462, opened 2019-02-26
Image Classifier Annotate: Python Plumbing

abhishekpratapa:
- Added Class Registrations for critical methods in AnnotationBase (ImageClassifier)
- Added a Python wrapper around the ImageClassifier Annotation Extension in TuriCreate


src/unity/lib/annotation/CMakeLists.txt
```diff
@@ -3,9 +3,14 @@ project(unity)
 set(TC_ANNOTATIONS_PROTO_FILES
   build/format/cpp/annotate.pb.cc
   build/format/cpp/data.pb.cc
+  build/format/cpp/message.pb.cc
   build/format/cpp/meta.pb.cc
 )
 
+set(TC_ANNOTATIONS_CLASS_REGISTRATION_FILE
+  class_registrations.cpp
```
znation: This can probably just go in the list of SOURCE_FILES unless it's used separately elsewhere.

src/unity/lib/annotation/README.md
```diff
@@ -35,7 +35,7 @@ https://github.com/protocolbuffers/protobuf/releases/tag/v3.3.0
 When you've installed the package, regenerate these protobuf files use the command below in the current directory.
 
 ```bash
-protoc --proto_path="./format" --cpp_out="./build/format/cpp" "./format/annotate.proto" "./format/data.proto" "./format/meta.proto"
+protoc --proto_path="./format" --cpp_out="./build/format/cpp" "./format/annotate.proto" "./format/data.proto" "./format/message.proto" "./format/meta.proto"
```
znation: Doesn't need to be addressed in this commit, but it should be possible to put a build rule in CMakeLists.txt that doesn't run by default (so regular `make` won't invoke it) but we could manually run something like `make protosrc` to regenerate.

src/unity/lib/annotation/annotation_base.cpp
```diff
@@ -24,9 +31,10 @@ AnnotationBase::AnnotationBase(const std::shared_ptr<unity_sframe> &data,
 
 void AnnotationBase::annotate(const std::string &path_to_client) {
   visualization::process_wrapper aw(path_to_client);
-  /* TODO: Easier to Implement when working on the Front End. Hold off
-   * implementation until front end PR. Will be more clear what to implement
-   * there */
+
+  aw << this->__serialize_proto<annotate_spec::MetaData>(this->metaData())
```
znation: Shouldn't this be inside the `while` loop below?

src/unity/lib/annotation/annotation_base.cpp
```diff
@@ -24,9 +31,10 @@ AnnotationBase::AnnotationBase(const std::shared_ptr<unity_sframe> &data,
...
+            .c_str();
+
   while (aw.good()) {
     break;
```
znation: Don't we want to keep the connection open now? Break will exit the loop and allow the annotate call to return.

src/unity/lib/annotation/annotation_base.hpp
```diff
@@ -74,6 +105,16 @@ class AnnotationBase {
...
+
+private:
+  /* A little trick to overload the `__serialize_proto` function at compile time
+   * so I don't have to define that for every Annotation::Specification type.
+   *
+   * Using the SFINAE method: https://en.cppreference.com/w/cpp/language/sfinae
+   */
+  template <typename T, typename = typename std::enable_if<std::is_base_of<
+                            ::google::protobuf::MessageLite, T>::value>::type>
+  std::string __serialize_proto(T message);
```
znation: Nice! I see the definition is elsewhere though. Won't that potentially result in undefined symbols at link time? For templates, I've always seen the definition in the header file where it's declared.

src/unity/python/turicreate/toolkits/image_classifier/_annotate.py
```diff
@@ -0,0 +1,216 @@
...
+import turicreate as __tc
+
+def _warning_annotations():
+    print(
+        """
+        *** WARNING: If you don't assign the output of the annotate function
+                     to a variable all annotations are going to be lost.
+                     To recover the last annotated SFrame call:
+
+        `tc.image_classifier._recover_annotation()`
```
znation: Since this is API we intend for users to call, it shouldn't start with an underscore.

src/visualization/Turi Create Visualization/src/VegaContainer.swift
```diff
@@ -278,6 +278,23 @@ class VegaContainer: NSObject, WKScriptMessageHandler {
         }
     }
     
+    public func send_proto(protobuf: String) {
```
znation: Is this used? It looks like elsewhere in the code, proto is only ever encoded as part of a JSON message.

src/visualization/Turi Create Visualization/src/user_interface/src/index.js
```diff
@@ -88,6 +88,12 @@ window.setSpec = function setSpec(value) {
     }
 }
 
+window.setProtoMessage = function setProtoMessage(value){
```
znation: Is this used in this PR?

abhishekpratapa: I forgot to push updates! Apologize

src/unity/lib/annotation/annotation_base.hpp
abhishekpratapa: Ya, so I ran into this issue when compiling this template. However I believe there was an explanation here:

https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file
https://isocpp.org/wiki/faq/templates#templates-defn-vs-decl

src/unity/python/turicreate/toolkits/image_classifier/_annotate.py
abhishekpratapa: Fair point. I always thought of this as an edge-case, something of a recovery state. So kept it hidden. When the user tab completes `tc.image_classifier` recover_annotation doesn't show up. But I see your point. It could be something the user wants to discover.

src/visualization/Turi Create Visualization/src/VegaContainer.swift
abhishekpratapa: Updated code! Sorry forgot to push!

src/unity/python/turicreate/toolkits/image_classifier/_annotate.py
```diff
@@ -0,0 +1,216 @@
...
+
+        """
+    )
+
+def annotate(data, image_column = 'image', annotation_column = 'annotations'):
```
TobyRoseman: In the image classifier the default value for the image column is `None` and we infer the image column (if there is only one column):
https://github.com/apple/turicreate/blob/5aacc9ceabae8bae5536dcafc6a652c8cfbc7b05/src/unity/python/turicreate/toolkits/image_classifier/image_classifier.py#L253
That would be nice to do here too.

src/unity/python/turicreate/toolkits/image_classifier/_annotate.py
```diff
@@ -0,0 +1,216 @@
...
+
+        out : SFrame
+            A new SFrame that contains the newly annotated data.
+
+        Examples
+        --------
+
+        >> import turicreate as tc
+        >> images = tc.image_analysis.load_images("path/to/images")
+        >> images
```
TobyRoseman: Maybe change this line to `print(image)` to make it clear what's happening on the lines below it.

src/unity/python/turicreate/toolkits/image_classifier/_annotate.py
```diff
@@ -0,0 +1,216 @@
...
+            +------------------------+--------------------------+
+            | /Users/username/Doc... | Height: 1712 Width: 1952 |
+            | /Users/username/Doc... | Height: 1386 Width: 1000 |
+            | /Users/username/Doc... |  Height: 536 Width: 858  |
+            | /Users/username/Doc... | Height: 1512 Width: 2680 |
+            +------------------------+--------------------------+
+            [4 rows x 2 columns]
+
+        >> images = tc.image_classifier.annotate(images)
+        >> images
```
TobyRoseman: same comment here too.

src/unity/python/turicreate/toolkits/image_classifier/_annotate.py
znation: Not only discover - we're telling the user to call it. I think of underscore as more of a marker denoting a public or private/internal method, rather than just as a way of making things hidden from tab completion. (There are other ways to hide things from tab completion if you really want to do just that.)

src/visualization/Turi Create Visualization/src/user_interface/src/index.js
znation: No problem!

src/visualization/Turi Create Visualization/src/VegaContainer.swift
znation: No problem!

src/visualization/Turi Create Visualization/src/user_interface/src/index.js
```diff
@@ -73,9 +80,6 @@ window.setSpec = function setSpec(value) {
         case "summary":
             spec_type = SpecType.summary;
             break;
-        case "annotate":
```
znation: Don't we still need this case here?

src/unity/python/turicreate/toolkits/image_classifier/_annotate.py
nickjong: Yes, I would argue that the primary purpose of underscore is tell people that it's an implementation detail and that they _shouldn't_ call it themselves directly

src/unity/lib/annotation/README.md
abhishekpratapa: Noted! Thanks!

src/unity/lib/annotation/annotation_base.cpp
abhishekpratapa: This is the metadata sent before the Two way call is initiated.

src/unity/lib/annotation/CMakeLists.txt
abhishekpratapa: Fixed!

PR #1467, opened 2019-02-27
Link build status badge to branch overview

znation:
Currently, our build status badge links to whatever happens to be the "current" build, which is usually a random pull request build. This commit changes that to link to the branch overview page (which lists the default branch, master, separately and first). I couldn't find a better direct link for just the master branch.


znation:
Thanks @TobyRoseman; I agree with your feedback. Please see my latest commit. I also added a couple more badges that look useful to me.

PR #1469, opened 2019-02-28
Metric

shreyajain17:
- Fixing the inconsistency between python and c++ object detection codes (metric options)

### DIFF
--- a/src/unity/toolkits/object_detection/object_detector.cpp
+++ b/src/unity/toolkits/object_detection/object_detector.cpp
@@ -332,7 +332,7 @@ variant_map_type object_detector::evaluate(
     metrics = {AP50,MAP50};
   }
   else if (metric == "all") {
-    metrics = {AP,AP50,MAP,MAP50};
+    metrics = all_metrics;
   }
   else {
     log_and_throw("Metric " + metric + " not supported");
--- a/src/unity/toolkits/object_detection/object_detector.cpp
+++ b/src/unity/toolkits/object_detection/object_detector.cpp
@@ -331,7 +331,7 @@ variant_map_type object_detector::evaluate(
   else if (metric == "auto") {
     metrics = {AP50,MAP50};
   }
-  else if (metric == "all") {
+  else if (metric == "all" || metric == "report") {
     metrics = all_metrics;
   }
   else {
--- a/src/unity/toolkits/object_detection/object_detector.cpp
+++ b/src/unity/toolkits/object_detection/object_detector.cpp
@@ -319,23 +319,23 @@ variant_map_type object_detector::evaluate(
     gl_sframe data, std::string metric,
     std::map<std::string, flexible_type> opts) {
 
-  // TODO: Support user-configurable metric and options.
   std::vector<std::string> metrics;
   static constexpr char AP[] = "average_precision";
   static constexpr char MAP[] = "mean_average_precision";
   static constexpr char AP50[] = "average_precision_50";
   static constexpr char MAP50[] = "mean_average_precision_50";
   std::vector<std::string> all_metrics = {AP,MAP,AP50,MAP50};
-  if (std::find(all_metrics.begin(), all_metrics.end(), metric) != all_metrics.end() ){
+  if (std::find(all_metrics.begin(), all_metrics.end(), metric) != all_metrics.end()) {
     metrics = {metric};
-    }
-  else if (metric=="auto"){
-    metrics={AP50,MAP50};}
-  else if(metric=="all"){
-    metrics={AP,AP50,MAP,MAP50};
-    }
-  else{
-    log_and_throw("Metric "+ metric + " not supported");
+  }
+  else if (metric == "auto") {
+    metrics = {AP50,MAP50};
+  }
+  else if (metric == "all") {
+    metrics = {AP,AP50,MAP,MAP50};
+  }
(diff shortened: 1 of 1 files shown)


nickjong:
It looks like this PR includes some commits from your last PR. Could you remove those commits from your branch?

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -320,7 +320,25 @@ variant_map_type object_detector::evaluate(
     std::map<std::string, flexible_type> opts) {
 
   // TODO: Support user-configurable metric and options.
```
nickjong: We can remove this TODO now

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -469,10 +487,20 @@ variant_map_type object_detector::evaluate(
     return std::accumulate(v.begin(), v.end(), 0.f) / v.size();
```
nickjong: We can remove the TODO above here, too ("TODO: Respect the `metric` argument...")

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -469,10 +487,20 @@ variant_map_type object_detector::evaluate(
...
   };
   variant_map_type result_map;
-  result_map["average_precision_50"] = create_dict(average_precision_50);
-  result_map["average_precision"] = create_dict(average_precision);
-  result_map["mean_average_precision_50"] = compute_mean(average_precision_50);
-  result_map["mean_average_precision"] = compute_mean(average_precision);
+  
+  if (std::find(metrics.begin(), metrics.end(), AP) != metrics.end() ){
+    result_map["average_precision"] = create_dict(average_precision);
```
nickjong: We may as well use the new constants you defined here, instead of these string literals. So `AP` instead of `"average_precision"`

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -320,7 +320,25 @@ variant_map_type object_detector::evaluate(
...
 
   // TODO: Support user-configurable metric and options.
-
+  std::vector<std::string> metrics;
+  static constexpr char AP[] = "average_precision";
+  static constexpr char MAP[] = "mean_average_precision";
+  static constexpr char AP50[] = "average_precision_50";
+  static constexpr char MAP50[] = "mean_average_precision_50";
+  std::vector<std::string> all_metrics = {AP,MAP,AP50,MAP50};
+  if (std::find(all_metrics.begin(), all_metrics.end(), metric) != all_metrics.end() ){
```
nickjong: Nitpick: While we don't have a strict style guide, I think that we should strive to make the formatting of our code somewhat consistent. The series of if statements here vary on a number of points:
* Whether there's a space between the `if` and the opening parenthesis
* Whether there's a space before the closing parenthesis of the `if`
* The indent level of the closing curly brace

FWIW, the other code in this file:
* Exactly one space immediately after `if` or `else`
* One space between a `)` and `{`
* Closing `}` on its own line (especially when the opening `{` is at the end of a line)
* Closing `}` at the same indent level as the `if` or `else` that opened that scope
* Generally, no space immediately after `(` or immediately before `)`. But definitely the opening and closing spacing should match one another

nickjong: (Someday maybe this can be something we do automatically as a post-processing step, using `clang-format`)

src/unity/toolkits/object_detection/object_detector.cpp
```diff
@@ -319,8 +319,26 @@ variant_map_type object_detector::evaluate(
...
     std::map<std::string, flexible_type> opts) {
 
-  // TODO: Support user-configurable metric and options.
-
+  std::vector<std::string> metrics;
+  static constexpr char AP[] = "average_precision";
+  static constexpr char MAP[] = "mean_average_precision";
+  static constexpr char AP50[] = "average_precision_50";
+  static constexpr char MAP50[] = "mean_average_precision_50";
+  std::vector<std::string> all_metrics = {AP,MAP,AP50,MAP50};
```
nickjong: Let's go ahead and leave `all_metrics` default-initialized in this line:
```
std::vector<std::string> all_metrics;
```
Since you already handle this case in line 335.

znation:
@shreyajain17 FYI, you can do a couple of things to make your PRs more associated with the issue(s) you are fixing:

* The title should reflect either the issue(s) being addressed, or the specific changes in the PR; "Metric" is a bit too vague and could mean lots of things. We will want to be able to look back at the list of PRs and get a sense of what went in, just from the PR titles.

* You can write "Fixes #xyz" anywhere in either the commit message and/or the PR description (but with an issue number instead of "xyz") to have GitHub automatically close the issue for you once the PR goes in. Otherwise, you'll want to manually associate the issue with the PR (by linking from one to the other in either direction, in a comment) and manually close it.

shreyajain17:
Got it.

nickjong:
I think it's also a good idea to edit what Git stores in the log for the commit. Just include what will be useful for someone reading the log in the future.

For example, it's not useful to have a bullet point in the final commit message for "Addressed review feedback"

(The default commit message is mostly just a list of the individual commits that were in the final pull request)

PR #1472, opened 2019-02-28
Added error message for probability output type for multiclass predict. Fixes #1414

### LINKED ISSUE #1414: Logistic Classifier: disable probability output type for multiclass predict
```python
import turicreate as tc

data = tc.SFrame([-100,-100,-100, 0,0,0, 100,100,100])
data['label'] = data['X1']
m = tc.logistic_classifier.create(data, 'label')

m.predict({'X1': 0}, output_type='probability')
```

The call to `predict` produces a long stack trace with the error message:
`TypeError: Cannot convert [0.0773239 0.869718 0.0529583] to float`

We shouldn't be trying to do a `probability` predict if the number of classes is more than two, per the docstring:
```
    - 'probability': Prediction probability associated with the True
      class (not applicable for multi-class classification)
```
We should just print a message about it being disabled for multi-class.

shreyajain17:
- Added an error message for probability output type for multiclass predict in logistic classifier

- It points the user to use probability vector instead

- fixes #1414


src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
```diff
@@ -621,10 +622,11 @@ class as a vector. The probability of the first class (sorted
         >>> class_predictions = model.predict(data, output_type='class')
 
         """
-
-        return super(_Classifier, self).predict(dataset,
-                                                output_type=output_type,
-                                                missing_value_action=missing_value_action)
+        if self.num_classes >2 and output_type=='probability':
+            raise ToolkitError("Not for multiclass")
+        return super(_Classifier, self).predict(dataset,output_type=output_type,missing_value_action=missing_value_action)
```
nickjong: Nit: I don't think you need to touch this line. Leaving it alone avoids cluttering the git blame/history for this line

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
```diff
@@ -621,10 +622,11 @@ class as a vector. The probability of the first class (sorted
         >>> class_predictions = model.predict(data, output_type='class')
 
         """
-
-        return super(_Classifier, self).predict(dataset,
-                                                output_type=output_type,
-                                                missing_value_action=missing_value_action)
+        if self.num_classes >2 and output_type=='probability':
```
nickjong: Nit: for readability, let's leave spaces around the `>` and `==` operators

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
```diff
@@ -10,6 +10,7 @@
 from __future__ import division as _
 from __future__ import absolute_import as _
 import turicreate as _turicreate
+from turicreate.toolkits._main import ToolkitError
```
TobyRoseman: You'll want to hide this import (like the other imports in this file):
`from turicreate.toolkits._main import ToolkitError as _ToolkitError`

Then in the body of your code, refer to it as `_ToolkitError` rather than `ToolkitError`.

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
```diff
@@ -621,10 +622,11 @@ class as a vector. The probability of the first class (sorted
         >>> class_predictions = model.predict(data, output_type='class')
 
         """
-
-        return super(_Classifier, self).predict(dataset,
-                                                output_type=output_type,
-                                                missing_value_action=missing_value_action)
+        if self.num_classes >2 and output_type=='probability':
+            raise ToolkitError("Not for multiclass")
```
nickjong: Could we add a more informative error message for users who may not understand what's going on?

nickjong:
Could we add a unit test for this case?

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
TobyRoseman: I thought your previous error message was better, more-descriptive.

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
shreyajain17: Because when I made the change in C++, predict_topk for multi class classification was also breaking which is not we want, right?

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
shreyajain17: Yeah completely forgot to change it.

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
shreyajain17: Okay

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
shreyajain17: Okay

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
shreyajain17: Yeah I forgot to change it.

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
shreyajain17: > Could we add a unit test for this case?

This is already present for number of classes more than 2

```
try:
      ans =  model.predict(self.sf, output_type='probability')
except ToolkitError:
       pass
```
Should I just remove try except block because it should always in the situation raise an error

shreyajain17:
> Could we add a unit test for this case?
This is already there in test file for more than two classes.
```
try:
                ans =  model.predict(self.sf, output_type='probability')
            except ToolkitError:
                pass
```
Should I just remove try except block because it should always raise an error in this situation?

nickjong:
Ah, I see, we already had a passing test for this case, we're just improving the error message that gets shown to the user. Nevermind, I think the test is fine then as is

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
nickjong: Hmm, upon closer inspection, I believe this code is intended to handle this case:
https://github.com/apple/turicreate/blob/master/src/unity/toolkits/supervised_learning/supervised_learning.cpp#L448

We should understand why the code path doesn't go through this line, where the check should be applied and an appropriate error message already exists

TobyRoseman: You're right that's not what we want. . . . Looking at your first change, I'm very surprised it was having that behavior. Was `log_and_throw` not actually throwing an exception?

src/unity/python/turicreate/toolkits/classifier/logistic_classifier.py
```diff
@@ -621,10 +621,12 @@ class as a vector. The probability of the first class (sorted
         >>> class_predictions = model.predict(data, output_type='class')
 
         """
-
+
```
nickjong: Let's just revert back to the original `logistic_classifier.py`; I don't think we need this PR to touch this file at all

src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -519,6 +519,14 @@ gl_sarray supervised_learning_model_base::fast_predict(
   } else if (output_type == "probability_vector"){
     ret_type = flex_type_enum::VECTOR;
   } else {
+      if (output_type == "probability" && variant_get_value<size_t>(state.at("num_classes")) > 2) {
```
nickjong: Nit: Let's match the surrounding code and only indent by 2 here, not 4.

(We sometimes use an indent level of 4 for code that is essentially wrapped due to line length)

src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -519,6 +519,14 @@ gl_sarray supervised_learning_model_base::fast_predict(
...
+        std::stringstream ss;
+        ss << "Output type '" << output_type
```
nickjong: We don't need to use `output_type` here, since you've already tested that it's equal to `"probability"`. As a result, you can avoid the use of `std::stringstream` entirely

PR #1473, opened 2019-03-01
Correcting anchor tags for relative links to work correctly. Fixes #1143

### LINKED ISSUE #1143: UG relative links don't work well on apple.github.io/
The user guide has many links to specific subtitles in other UG pages. For example is the link at the bottom of https://github.com/apple/turicreate/tree/master/userguide/image_classifier#advanced-usage correctly links to the "Accessing attributes of the model" section in https://github.com/apple/turicreate/blob/master/userguide/supervised-learning/linear-regression.md#linregr-model-access/. 

However, the rendered counterpart of this page, https://apple.github.io/turicreate/docs/userguide/image_classifier/, tries to link to https://apple.github.io/turicreate/docs/userguide/supervised-learning/linear-regression.html#linregr-model-access, while the correct link is in fact https://apple.github.io/turicreate/docs/userguide/supervised-learning/linear-regression.html#accessing-attributes-of-the-model - ending up at the top of that page, instead of the relevant section.

This is just one example I've encountered - but from poking around it seems to be happening across the board

shreyajain17:
- fixes relative links to sections on different pages of documentation

### DIFF
--- a/userguide/image_classifier/README.md
+++ b/userguide/image_classifier/README.md
@@ -77,6 +77,6 @@ Refer to the following chapters for:
 
 In addition, the following chapters contain more information on how to use classifiers:
 
-* [Accessing attributes of the model](../supervised-learning/linear-regression.md#linregr-model-access)
-* [Evaluating Results](../supervised-learning/logistic-regression.md#logregr-evaluation)
-* [Multiclass Classification](../supervised-learning/logistic-regression.md#logregr-multiclass)
+* [Accessing attributes of the model](../supervised-learning/linear-regression.md#accessing-attributes-of-the-model)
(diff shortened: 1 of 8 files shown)


nickjong:
I think this makes these headings appear as links that can be clicked on, but they don't actually lead anywhere. We probably want to find a way to preserve the current appearance

Also, is the problem confined just to linear and logistic regression, or is it throughout the user guide?

shreyajain17:
Appears to be only in these two. I will check for others.

znation:
You can use CSS to prevent them from looking like links. Give them a class like "non_link_anchor", then define it in CSS as:
```css
a.non_link_anchor:hover {
    text-decoration: none;
    /* possibly change other things like color too */
}
```

PR #1474, opened 2019-03-01
turicreate.drawing_recognition.util.draw_strokes

### LINKED ISSUE #1439: Expose a method to convert stroke-based drawings to bitmaps.
This function would be an analog to `tc.util.draw_bounding_boxes` to convert stroke-based drawings to bitmaps.

shantanuchhabra:
Fixes #1439

Note that this PR has a sequential dependency on #1449 and must be merged only after #1449 is merged in.

### DIFF
--- a/src/unity/python/turicreate/toolkits/drawing_recognition/util/_visualization.py
+++ b/src/unity/python/turicreate/toolkits/drawing_recognition/util/_visualization.py
@@ -1,0 +1,17 @@
+import turicreate as _tc
+import numpy as _np
+from turicreate import extensions as _extensions
+
+def draw_strokes(stroke_based_drawings):
+    if type(stroke_based_drawings) != _tc.SArray:
+        raise _ToolkitError("Input to draw_strokes must be of type " 
+            + "turicreate.SArray")
+    is_stroke_input = (stroke_based_drawings.dtype != _tc.Image)
+    if is_stroke_input:
+        sf = tc.SFrame({"drawings": stroke_based_drawings})
+        sf_with_drawings = _extensions._drawing_recognition_prepare_data(
+            sf, "drawings")
+        return sf_with_drawings["drawings"]
+    else:
+        # User passed in drawings that were already drawings
+        return stroke_based_drawings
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/toolkits/drawing_recognition/util/_visualization.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2019 Apple Inc. All rights reserved.
+#
+# Use of this source code is governed by a BSD-3-clause license that can
+# be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+import turicreate as _tc
+from turicreate import extensions as _extensions
+
+def draw_strokes(stroke_based_drawings):
```
nickjong: Doc string?

src/unity/python/turicreate/toolkits/drawing_recognition/util/_visualization.py
```diff
@@ -0,0 +1,21 @@
...
+        raise _ToolkitError("Input to draw_strokes must be of type " 
+            + "turicreate.SArray")
+    is_stroke_input = (stroke_based_drawings.dtype != _tc.Image)
+    if is_stroke_input:
+        sf = _tc.SFrame({"drawings": stroke_based_drawings})
+        sf_with_drawings = _extensions._drawing_recognition_prepare_data(
+            sf, "drawings")
+        return sf_with_drawings["drawings"]
+    else:
+        # User passed in drawings that were already drawings
```
nickjong: Why do we want/need to support this?

shantanuchhabra: In the event that the user passes in an SFrame with drawings to the `draw_strokes` function, I thought it was better to silently return the same SFrame instead of throwing an error. Although now that I think about it, passing in an SFrame with drawings is a user error and we should probably throw an error for it. Thoughts?

nickjong: Yeah, you already have it as an error to pass an SFrame. If anything, you might consider accepting either an SArray of lists of dictionaries, or just one list of dictionaries.

src/unity/python/turicreate/toolkits/drawing_recognition/util/_visualization.py
shantanuchhabra: Added.

PR #1481, opened 2019-03-02
Rendering images directly in jupyter notebook. Fixes #947

### LINKED ISSUE #947: Render images directly in Jupyter notebooks
When showing an SFrame in Jupyter notebook (without calling `explore()`, natively in the notebook), it would be great if the output would render images instead of showing their path.

shreyajain17:


src/unity/python/turicreate/data_structures/image.py
```diff
@@ -184,6 +184,19 @@ def __repr__(self):
...
+    def _repr_png_(self):
+        img = self._to_pil_image()
+        from io import BytesIO
+        b = BytesIO()
+        img.save(b, format='png')
+        data = b.getvalue()
+        res = {"Height" :str(self._height), "Width":str(self._width), "Channels: " :str(self._channels)}
+        return (data,res)
+
+        # ip_img = display.Image(data=data, format='png', embed=True)
```
TobyRoseman: I wouldn't check in commented out code without a good reason. If there is a good reason to have it, please add a comment about why it's important.

shreyajain17: Got it.

znation:
Looks great!

PR #1482, opened 2019-03-02
Changes in make_wheel.sh to create wheels that are pip installable. Fixes #927

### LINKED ISSUE #927: `make_wheel.sh` creates wheels that are not pip installable
`make_wheel.sh` creates wheels that have `linux` as the platform name, which `pip` does not recognize. The platform name should be `manylinux1`

shreyajain17:
- names that wheel file with manylinux1 instead of linux

### DIFF
--- a/scripts/make_wheel.sh
+++ b/scripts/make_wheel.sh
@@ -264,7 +264,7 @@ package_wheel() {
   else
       NEW_WHEEL_PATH=${WHEEL_PATH/-py2-/-cp27-}
   fi
-  NEW_WHEEL_PATH=${WHEEL_PATH/linux/manylinux1}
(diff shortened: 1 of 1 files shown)


scripts/make_wheel.sh
```diff
@@ -264,6 +264,7 @@ package_wheel() {
   else
       NEW_WHEEL_PATH=${WHEEL_PATH/-py2-/-cp27-}
   fi
+  NEW_WHEEL_PATH=${WHEEL_PATH/linux/manylinux1}
```
znation: I think this should be:
```
NEW_WHEEL_PATH=${NEW_WHEEL_PATH/linux/manylinux1}
```

Since we already assigned `NEW_WHEEL_PATH` above and made some modifications to it.

znation:
One minor comment, otherwise looks good!

PR #1483, opened 2019-03-02
Added Annotation - Front End Implementation for Image Classifier

abhishekpratapa:
Working on the annotation front end for Image Classification.


src/visualization/Turi Create Visualization/src/user_interface/src/elements/Annotate/index.js
```diff
@@ -184,6 +168,26 @@ class Annotate extends Component {
...
+      throw Error(err);
+
+    const message = ParcelMessage.create(payload);
+    const buffer = ParcelMessage.encode(message).finish();
+    const encoded = btoa(String.fromCharCode.apply(null, buffer));
+
+    if (window.navigator.platform == 'MacIntel') {
+      window.webkit.messageHandlers["scriptHandler"].postMessage({status: 'writeProtoBuf', message: encoded});
+    } else {
+      window.postMessageToNativeClient(encoded);
```
znation: Does not need to be done in this PR, but in the future, let's abstract away the Linux vs. macOS way we communicate back to the host.

PR #1484, opened 2019-03-02
Added code to use a pre-trained model.

shantanuchhabra:
This is temporary, and will eventually use the published pre-trained model.


src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -103,10 +106,21 @@ def update_progress(cur_loss, iteration):
...
+            temp_filename,
+            ctx=ctx,
+            allow_missing=True)
+        _os.unlink(temp_filename)
     softmax_cross_entropy = _mx.gluon.loss.SoftmaxCrossEntropyLoss()
-    
-    ctx = _mxnet_utils.get_mxnet_context(max_devices=batch_size)
 
     model_params = model.collect_params()
     model_params.initialize(_mx.init.Xavier(), ctx=ctx)
```
nickjong: Doesn't this overwrite the model params defined above?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -103,10 +106,21 @@ def update_progress(cur_loss, iteration):
...
+    if use_pretrained_model:
+        pretrained_model_url = _os.environ["PRETRAINED_MODEL_URL"]
+        temp_filename = _mkstemp()[1]
+        _six.moves.urllib.request.urlretrieve(pretrained_model_url,
+            temp_filename)
+        model.load_params(
+            temp_filename,
+            ctx=ctx,
+            allow_missing=True)
+        _os.unlink(temp_filename)
```
nickjong: Won't this cause us to retrieve the model every time? Why not use the pre-trained model cache used by the object detection, style transfer, and sound classification toolkits?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -103,10 +106,21 @@ def update_progress(cur_loss, iteration):
                  epochs=num_epochs,
                  iterations=None)
 
+    ctx = _mxnet_utils.get_mxnet_context(max_devices=batch_size)
     model = _Model(num_classes = len(classes))
+    model_params = model.collect_params()
+    model_params.initialize(_mx.init.Xavier(), ctx=ctx)
+    if use_pretrained_model:
+        pretrained_model_url = _os.environ["PRETRAINED_MODEL_URL"]
```
nickjong: If we're going to control whether this is enabled via `use_pretrained_model`, why not just make the model URL an argument to `create`? If it's `None`, then you use random initialization

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -30,8 +33,8 @@ def _raise_error_if_not_drawing_classifier_input_sframe(dataset, feature):
             + " where each stroke is a list of points and" 
             + " each point is stored as a dictionary")
 
-def create(input_dataset, annotations=None, num_epochs=100, feature="bitmap", 
-           target="label", model=None, classes=None, batch_size=256, 
+def create(input_dataset, num_epochs=100, feature="bitmap",
```
nickjong: Does `num_epochs` not depend in some way on whether we're using a pretrained model?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -30,8 +33,8 @@ def _raise_error_if_not_drawing_classifier_input_sframe(dataset, feature):
             + " where each stroke is a list of points and" 
             + " each point is stored as a dictionary")
 
-def create(input_dataset, annotations=None, num_epochs=100, feature="bitmap", 
-           target="label", model=None, classes=None, batch_size=256, 
+def create(input_dataset, num_epochs=100, feature="bitmap", 
+           target="label", use_pretrained_model=False, classes=None, batch_size=256, 
            max_iterations=0, verbose=True, **kwargs):
```
nickjong: What does `max_iterations` do?

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Interestingly, mxnet ignores reinitialization until the `force_reinit` flag is passed, so the functionality is unaffected by these lines. However, this code is dead for all practical purposes and shouldn't be checked in so I pushed an update to fix it.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Fixed.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Not at the moment. I was thinking that could come in as part of polish or v2.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Added functionality so user can specify max_iterations.

src/unity/python/turicreate/toolkits/_pre_trained_models.py
```diff
@@ -238,3 +238,19 @@ def get_model_path(self, format):
             )[0]
 
         return model_path
+
+class DrawingClassifierPreTrainedModel(object):
+    def __init__(self, model_url = None):
+        filename = "drawing_model_v1.params"
```
nickjong: "drawing model" is extremely vague. Let's give this some more thought before we actually post anything. Anything we post for 5.4 will probably be closer to "v0" than "v1", too.

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
```diff
@@ -42,6 +43,11 @@ def create(input_dataset, annotations=None, num_epochs=100, feature="bitmap",
     
     start_time = _time.time()
 
+    if max_iterations == 0:
+        max_iterations = num_epochs * len(input_dataset) / batch_size
+    else:
+        num_epochs = max_iterations * batch_size / len(input_dataset)
```
nickjong: Why should we expose both `num_epochs` and `max_iterations` as parameters?

src/unity/python/turicreate/toolkits/_pre_trained_models.py
shantanuchhabra: Tracked in #1489

src/unity/python/turicreate/toolkits/drawing_recognition/drawing_recognition.py
shantanuchhabra: Tracked in #1488

PR #1486, opened 2019-03-02
WIP: Fix broken build with `--no-python`

znation:
The big CMake refactor that went in recently seems to have broken
the build for non-Python builds (`./configure --no-python`). The problem
is that unity_shared now only builds in the `python` directory (which
makes sense, since it's only used in Python), but the extensions all
depended on it and all were expected to build even without Python.

The fix is to move or remove all of the extensions, since the whole
extension mechanism is a bit deprecated now. Now that we are open
source, API consumers should directly link against our public APIs
and/or build into a clone/fork of this repo. Eventually an extension
mechanism may come back but will likely be based on the C API.

* gpu_count: removed (was not built)
* grouped_sframe: removed (was not built)
* internal_demo: removed (was not used; superseded by the examples in
  `src/unity/toolkits/prototype`.
* json: removed (exposed to Python, but not imported or used).
* timeseries: removed (was not built)
* additional_sframe_utilities: moved to `src/unity/lib`
* coreml_extension: moved to `src/unity/toolkits/coreml_export`
* content_interpretation_extensions: moved to `feature_engineering`
* transform_to_flat_dict_extensions: moved to `feature_engineering`
* xgboost_extension: moved to `supervised_learning`


znation:
@hoytak Assigning to you to finish this WIP PR as per our conversation. Please feel free to assign it back to me if it's no longer critical path and you won't have time to do it.

znation:
Marking WIP since this doesn't actually fix the issue yet.

znation:
Could not get this working -- the extension mechanism doesn't work with any of the attempts in this PR. Closing in favor of https://github.com/apple/turicreate/pull/1833.

PR #1497, opened 2019-03-05
Fix for bit mask to make corner case more portable.

hoytak:
Turns out uint64_t(1) << 64 isn't actually well defined.

### DIFF
--- a/src/util/bitops.hpp
+++ b/src/util/bitops.hpp
@@ -129,7 +129,7 @@ static inline T bit_mask(uint64_t n_bits, _ENABLE_IF_UINT(T)) {
   shift_t _n = shift_t(n_bits);
   shift_t _n_mask = ~shift_t(bitsizeof(T) - 1);
 
-  return UNLIKELY(_n & _n_mask) ? ~T(0) : (T(1) << _n) - 1;
+  return UNLIKELY(_n & _n_mask) ? ~T(-1) : (T(1) << _n) - 1;
 }
 
 /**
--- a/src/util/bitops.hpp
+++ b/src/util/bitops.hpp
@@ -10,6 +10,7 @@
 #include <cmath>
 #include <type_traits>
 #include <climits>
+#include <util/code_optimization.hpp>
 // #include <bitset>
 // #include <iostream>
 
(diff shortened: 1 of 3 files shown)


ylow:
There is possibly a branch free version
```
return (T(1) << n_bits) - 1 - (n_bits == bitsizeof(T));
```

hoytak:
> There is possibly a branch free version
> 
> ```
> return (T(1) << n_bits) - 1 - T(n_bits == bitsizeof(T));
> ```

But the first expression sometimes is zero and sometimes a no-op, so this wouldn't work either. 

We could use the branch free version in https://stackoverflow.com/questions/1392059/algorithm-to-generate-bit-mask, though.

ylow:
What do you mean by no-op? 
This seems to work fine.

```
#include <cstdint>
#include <iostream>
#include <ios>

template <typename T>
T bitmask(int n_bits) {
  return (T(1) << n_bits) - 1 - T(n_bits == 8 * sizeof(T));
}
int main() {
  std::cout << std::hex << bitmask<int64_t>(0) << "\n"
                        << bitmask<int64_t>(1) << "\n"
                        << bitmask<int64_t>(63) << "\n"
                        << bitmask<int64_t>(64) << "\n"
                        << bitmask<uint64_t>(0) << "\n"
                        << bitmask<uint64_t>(1) << "\n"
                        << bitmask<uint64_t>(63) << "\n"
                        << bitmask<uint64_t>(64) << "\n"
                        << bitmask<int32_t>(0) << "\n"
                        << bitmask<int32_t>(1) << "\n"
                        << bitmask<int32_t>(31) << "\n"
                        << bitmask<int32_t>(32) << "\n"
                        << bitmask<uint32_t>(0) << "\n"
                        << bitmask<uint32_t>(1) << "\n"
                        << bitmask<uint32_t>(31) << "\n"
                        << bitmask<uint32_t>(32) << "\n";
}
```

Essentially, if n_bits == bitsizeof(T), (T(1) << n_bits) - 1 == 0, so we minus 1 again to get -1.

Though technically this does depend on undefined behavior. Compiling with -fsanitize=undefined does give errors like
```
runtime error: shift exponent 64 is too large for 64-bit type 'long long'
```

Your branching version is fine. Will be great to have some tests for this.

hoytak:
On my machine, lldb says your expression yields the equivalent of -2.

Better version, branch free, no undefined behavior.  This one I'm proud of. 

```
template <typename T>
static inline T bit_mask(uint64_t n_bits, _ENABLE_IF_UINT(T)) {
  return (~T(0)) >> ((-n_bits) & (bitsizeof(T) - 1) );
}
```

PR #1502, opened 2019-03-06
Fixes #1407

### LINKED ISSUE #1407: Per-class accuracy doesn't make any sense
Consider this code from https://apple.github.io/turicreate/docs/userguide/evaluation/classification.html

```
import turicreate as tc

y    = tc.SArray(["cat", "dog", "foosa", "cat"])
yhat = tc.SArray(["cat", "dog", "cat", "dog"])

print(tc.evaluation.accuracy(y, yhat, average = "micro"))
print(tc.evaluation.accuracy(y, yhat, average = "macro"))
print(tc.evaluation.accuracy(y, yhat, average = None))
```

which produces

```
0.5
0.666666666667
{'dog': 0.75, 'foosa': 0.75, 'cat': 0.5}
```

The per class accuracy doesn't make any sense here...

shantanuchhabra:
As per the conversation with @hoytak , we want to return the precision and not the accuracy when `average=None` is passed into `tc.evaluation.accuracy`.

As of this change, when the input is:
```
import turicreate as tc 
y    = tc.SArray(["cat", "dog", "foosa", "cat"])
yhat = tc.SArray(["cat", "dog", "cat", "dog"])
print(tc.evaluation.accuracy(y, yhat, average = "micro"))
print(tc.evaluation.accuracy(y, yhat, average = "macro"))
print(tc.evaluation.accuracy(y, yhat, average = None))
```
The output is:
```
0.5
0.6666666666666666
{'foosa': None, 'dog': 0.5, 'cat': 0.5}
```

### DIFF
--- a/src/unity/python/turicreate/test/test_evaluation.py
+++ b/src/unity/python/turicreate/test/test_evaluation.py
@@ -966,6 +966,8 @@ class MetricsTest(unittest.TestCase):
         self.assertAlmostEqual(skl_score, str_score)
 
         # Act [Average = None]
+        prec_score = precision_score(list(targets), list(predictions),
+            average = None)
         score = turicreate.toolkits.evaluation.accuracy(targets,
                                             predictions,
                                             average = None)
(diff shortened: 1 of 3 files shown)


src/unity/toolkits/evaluation/evaluation_interface-inl.hpp
```diff
@@ -1430,7 +1432,7 @@ class flexible_accuracy : public precision_recall_base {
       // All scores.
       case average_type_enum::NONE:
       {
-        return to_variant(accuracy_scores);
```
srikris: Can we add a small unit test to make sure this stays fixed :)

shantanuchhabra: Absolutely. And I'll also update the outputs in the User Guide, where @hoytak first spotted this issue.

src/unity/python/turicreate/test/test_evaluation.py
```diff
@@ -979,12 +981,12 @@ def test_accuracy_multi_class_score(self):
 
         # Note: Explicitly not putting it into a for loop for ease of
         # debugging when the tests fail.
-        self.assertAlmostEqual(cls_skl_score[0], score[0])
-        self.assertAlmostEqual(cls_skl_score[0], str_score['0'])
-        self.assertAlmostEqual(cls_skl_score[1], score[1])
-        self.assertAlmostEqual(cls_skl_score[1], str_score['1'])
-        self.assertAlmostEqual(cls_skl_score[2], score[2])
-        self.assertAlmostEqual(cls_skl_score[2], str_score['2'])
+        self.assertAlmostEqual(prec_score[0], score[0])
```
srikris: Why are you comparing this to precision score from sklearn. I'm a bit confused.

shantanuchhabra: I think the unit tests use sklearn as the source of truth. I just modified the existing unit tests to compare to precision. What would you recommend?

PR #1504, opened 2019-03-06
Sound Classifier Documentation

TobyRoseman:
User guide and documentation


userguide/sound_classifier/README.md
```diff
@@ -0,0 +1,55 @@
...
+#### Loading Data
+
+For example purposees, we will use the ESC-10 dataset. This dataset
+contain ten classes. Each class has 40 examples with five seconds of
+audio per example.
+
+**Please note:** the ESC-10 dataset is part of a larger [ESC-50 dataset](https://github.com/karoldvl/ESC-50)
+dataset. In order to use the ESC-10, you will need to download the
+[ESC-50 dataset](https://github.com/karoldvl/ESC-50). These two datasets have different licenses.
+Depending on your use case you may be able to the larger ESC-50
```
shantanuchhabra: ... you may be able to *use* the larger ...

userguide/sound_classifier/export-coreml.md
```diff
@@ -0,0 +1,114 @@
...
+      fatalError("Error calling predict")
+   }
+   results.append(modelOutput.categoryProbability)
+   windowNumber += 1
+}
+```
+
+Once all window predictions are saved, to generate the overall
+prediction, calculate the mean accross all prediciton windows:
+```
```
shantanuchhabra: Add `swift` at the end of the three backticks here for proper syntax highlighting?

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
+# How Does this Work?
+
+Training and making predctions, for a sound classifier model, is a three
```
shantanuchhabra: Typo: predictions, not predctions.

Nit: I don't think there should be a comma between "predictions" and "for"? Not sure...

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+stage process. The first stage does signal preprocessing. In
+the second step we use a pretrained neural network to extract deep
+features. Finally in the last step, we use a custom neural network.
```
shantanuchhabra: Nit: Maybe add bullet points here to improve readability? I know the rest of this document goes into each of the stages separately, but adding bullet points here can help the reader ease into that.

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Signal Preprocessing Pipeline Stage
+Several audio transformation happen during this stage. For someone without
```
shantanuchhabra: transformation -> transformations

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+sound domain knowledge, this is probably the most complicated part.
+Nothing about this stage is updated based on the input data (i.e. nothing
+is learned in this stage).
+
+At a high level the preprocessing pipeline does the following:
```
shantanuchhabra: Insert comma between "high level" and "the"?

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+* The raw pulse code modulation data from the wav file is converted to
+floats on a [-1.0, +1.0] scale.
+* The data is resampled to only 16,000 sample per second.
```
shantanuchhabra: sample -> samples

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+* The data is broken up in to several overlapping windows.
```
shantanuchhabra: broken up in to -> broken up into

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+* A [Hamming Window](https://en.wikipedia.org/wiki/Window_function#Hann_and_Hamming_windows) is applied to each windows.
+* The [Power Spectrum](https://en.wikipedia.org/wiki/Spectral_density#Power_spectral_density) is calculated, using a [Fast Fourier Transformation](https://en.wikipedia.org/wiki/Fourier_transform).
+* Frequencies above and below certain thresholds are dropped.
+* Mel Frequency Filter Banks are applied.
```
shantanuchhabra: A Wikipedia link here would be nice: https://en.wikipedia.org/wiki/Mel-frequency_cepstrum

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+* Finally the log is taken of all values.
```
shantanuchhabra: Does this mean logarithm?

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+
+The preprocessing pipeline takes 975ms worth of audio as input (exact
+input lenght depends on sample rate) and produce an array of shape
```
shantanuchhabra: lenght -> length, produce -> produces

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+last three layers of the original VGGish model. We use the widest
+layer, from the original network, as our input data for the final
+stage. This modified VGGish model ouputs a double vector of length
+12,288. On non-Linux systems, the model has also been eight bit
+quantized, to reduce its size.
+
+## Custom Neural Network Stage
+This is the only stage which is updated based on the input data.
+During training of a sound classifier this model is trained using
+the features from VGGish and the input labels. During prediction a
```
shantanuchhabra: "During prediction a only a forward pass is done" -> "During prediction, only a forward pass is done"

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+only a forward pass is done using this network. This custom neural
+network is a simple three layer neural netowrk. The first two
```
shantanuchhabra: netowrk -> network

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+layers are dense layers, with 100 units each. These layers use [RELU activation](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)).
+The final layers is a softmax. The number of units in this layer is
```
shantanuchhabra: "final layers" -> "final layer"

userguide/sound_classifier/how-it-works.md
```diff
@@ -0,0 +1,52 @@
...
+VGGish is a pretrained [Convolutional Neural Network](https://en.wikipedia.org/wiki/Convolutional_neural_network) from Google,
+see [their paper](https://ai.google/research/pubs/pub45611) and [their GitHub page](https://github.com/tensorflow/models/tree/master/research/audioset) for more details. As the name suggests, the architecture of
+this network is inspired by the famous VGG networks used for image
+classification. The network consists of a series of convolution and
+activation layers, optionally followed by a max pooling layer.
+This network contains 17 layers in total.
+
+This network is kept static during model training. We have removed the
+last three layers of the original VGGish model. We use the widest
+layer, from the original network, as our input data for the final
```
shantanuchhabra: Two general questions: 
1. Is the widest layer from the original network the same as the fourth last layer in the original network? 
2. Are the three layers we have in our custom neural network the same as the last three layers in VGGish?

userguide/sound_classifier/README.md
```diff
@@ -0,0 +1,55 @@
...
+dataset, see their [license section](https://github.com/karoldvl/ESC-50#license) for details.
+
+#### Introductory Example
+```python
+import turicreate as tc
+from os.path import basename
+
+# Load the audio data and meta data.
+data = tc.load_audio('./ESC-50/audio/')
```
srikris: Don't we have a with_path = True on this like with load images? It reduces the need for the base name for the join.

userguide/sound_classifier/export-coreml.md
```diff
@@ -0,0 +1,114 @@
...
+   guard let modelOutput = try? model.prediction(input: modelInput) else {
+      fatalError("Error calling predict")
+   }
+   results.append(modelOutput.categoryProbability)
+   windowNumber += 1
+}
+```
+
+Once all window predictions are saved, to generate the overall
+prediction, calculate the mean accross all prediciton windows:
```
shantanuchhabra: accross -> across

userguide/sound_classifier/README.md
TobyRoseman: We do have a `with_path=True` parameter on `tc.load_images(...)`. I've also added it `tc.load_audio(...)` where it is also on by default; I'm using it here.

We still need the `basename` apply before the join because in the meta data CSV only the filenames are stored. `load_images` and `load_audio` both put the full path in the SFrame, when the `with_path` parameter is true.

userguide/sound_classifier/README.md
TobyRoseman: Done.

userguide/sound_classifier/export-coreml.md
TobyRoseman: Good catch. Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done and done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: That's a good idea. I decided to use a numbered list rather than a bulleted.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Yeah, I debated about adding that link here. The thing is that link covers more than what's being mentioned in this bullet point. I agree this bullet point needs a link to more info.

I've added wikipedia links to both "Mel Frequency" and "Filter Banks". I think that more accurate and probably a lot easier to understand too.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Yes. I've clarified that also mentioned the base.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done and done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: Done.

userguide/sound_classifier/how-it-works.md
TobyRoseman: 1 - Yes, the widest layer is the fourth to the last layer in the original VGGish network.
2 - No, those have significantly more (4096) nodes. In prototyping testing I did fewer (100) nodes actually preformed better. It is also of course faster to train and predict too.

userguide/sound_classifier/export-coreml.md
TobyRoseman: Done and done.

TobyRoseman:
Thanks @shantanuchhabra for the thorough review.

PR #1507, opened 2019-03-06
Exclude parameter of recommend checks user_id type and uses it

### LINKED ISSUE #1164: Recommender Exclude Parameter Not Working
The `exclude` parameter for `recommend` does not seem to work: 
```python
import turicreate
sf = turicreate.SFrame({'user_id': ["0", "0", "0", "1", "1", "2", "2", "2"],
                       'item_id': ["a", "b", "c", "a", "b", "b", "c", "d"],
                       'rating': [1, 3, 2, 5, 4, 1, 4, 3]})
m = turicreate.recommender.create(sf, target='rating')

recs1 = m.recommend(users=[1])
print(recs1) # correctly recommending 'c' to 1

exclude = turicreate.SFrame({'item_id': ['c'], 'user_id': [1]})
rec2 = m.recommend(users=[1], exclude=exclude)
print(rec2)  # still recommending 'c' to 1
```

shreyajain17:
fixes #1164

### DIFF
--- a/src/unity/python/turicreate/toolkits/recommender/util.py
+++ b/src/unity/python/turicreate/toolkits/recommender/util.py
@@ -1260,9 +1260,17 @@ class _Recommender(_Model):
                 assert new_observation_data.num_rows() != 0
                 original_user_type = user_column.dtype
                 users[user_id] = user_column.astype(str)
+                user_type=str
 
             elif user_column.dtype != user_type:
                 users[user_id] = user_column.astype(user_type)
+
+        # Cast user specified in exclude to the appropriate type if necessary.
+        # for i in range(exclude.num_columns()):
+        #     if exclude.column_names()[i]==user_id:
+        if user_id in exclude.column_names():
+            if exclude[user_id].dtype!=user_type:
+                exclude[user_id] = exclude[user_id].astype(user_type)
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/toolkits/recommender/util.py
```diff
@@ -1260,10 +1260,16 @@ def check_type(arg, arg_name, required_type, allowed_types):
...
                 original_user_type = user_column.dtype
                 users[user_id] = user_column.astype(str)
+                user_type=str
 
             elif user_column.dtype != user_type:
                 users[user_id] = user_column.astype(user_type)
 
+        # Cast user specified in exclude to the appropriate type if necessary.
+        if user_id in exclude.column_names():
+            if exclude[user_id].dtype!=user_type:
```
shantanuchhabra: Usually, it's good practice to write a conjuncted conditional instead of writing a nested conditional. Replace `if A: if B: do stuff` with `if A and B: do stuff`. Note that Python has short-circuiting so if A is false, B will never get evaluated, like you intend here.

shreyajain17: Got it.

Issue #1047, opened 2018-08-27
Error when trying to 'import turicreate' in Google Cloud Datalab


SoroushKadineh:
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-71-1b519f14fefb> in <module>()
----> 1 import turicreate

/usr/local/envs/py3env/lib/python3.5/site-packages/turicreate/__init__.py in <module>()
     17 from turicreate.version_info import __version__
     18 
---> 19 from turicreate.data_structures.sgraph import Vertex, Edge
     20 from turicreate.data_structures.sgraph import SGraph
     21 from turicreate.data_structures.sarray import SArray

/usr/local/envs/py3env/lib/python3.5/site-packages/turicreate/data_structures/__init__.py in <module>()
     16 
     17 from . import image
---> 18 from . import sframe
     19 from . import sarray
     20 from . import sgraph

/usr/local/envs/py3env/lib/python3.5/site-packages/turicreate/data_structures/sframe.py in <module>()
     14 from __future__ import division as _
     15 from __future__ import absolute_import as _
---> 16 from ..connect import main as glconnect
     17 from ..cython.cy_flexible_type import infer_type_of_list
     18 from ..cython.context import debug_trace as cython_context

/usr/local/envs/py3env/lib/python3.5/site-packages/turicreate/connect/main.py in <module>()
     11 from __future__ import absolute_import as _
     12 
---> 13 from ..cython.cy_unity import UnityGlobalProxy
     14 from ..cython.cy_server import EmbeddedServer
     15 

ImportError: libblas.so.3: cannot open shared object file: No such file or directory

----------------------------------------------------------------------------------------------------
Above is the output of when I try 'import turicreate'. I have trained a few models in Jupyter Notebook using turicreate on my local machine. But, Google Cloud Datalab is basically a cloud version of Jupyter Notebook, and I can't figure out how install 'libblas' as was suggested in #191

TobyRoseman:
@SoroushKadineh - you need to install BLAS, as well as other Linux dependencies. See our [Linux Install Instructions](https://github.com/apple/turicreate/blob/master/LINUX_INSTALL.md) for details.

SoroushKadineh:
@TobyRoseman I get the followings when I do `!pip install libstdc++`

> Invalid requirement: 'libstdc++'
Traceback (most recent call last):
[...]

Issue #191, opened 2018-01-17
Import Error


elseagle:
ImportError                               Traceback (most recent call last)
<ipython-input-5-1b519f14fefb> in <module>()
----> 1 import turicreate

/home/elseagle/venv/local/lib/python2.7/site-packages/turicreate/__init__.py in <module>()
     28 from turicreate.version_info import __version__
     29 
---> 30 from turicreate.data_structures.sgraph import Vertex, Edge
     31 from turicreate.data_structures.sgraph import SGraph
     32 from turicreate.data_structures.sarray import SArray

/home/elseagle/venv/local/lib/python2.7/site-packages/turicreate/data_structures/__init__.py in <module>()
     15 __all__ = ['sframe', 'sarray', 'sgraph', 'sketch', 'image']
     16 
---> 17 from . import image
     18 from . import sframe
     19 from . import sarray

ImportError: cannot import name image

elseagle:
`ImportError                               Traceback (most recent call last)
<ipython-input-5-1b519f14fefb> in <module>()
----> 1 import turicreate

/home/elseagle/venv/local/lib/python2.7/site-packages/turicreate/__init__.py in <module>()
     28 from turicreate.version_info import __version__
     29 
---> 30 from turicreate.data_structures.sgraph import Vertex, Edge
     31 from turicreate.data_structures.sgraph import SGraph
     32 from turicreate.data_structures.sarray import SArray

/home/elseagle/venv/local/lib/python2.7/site-packages/turicreate/data_structures/__init__.py in <module>()
     15 __all__ = ['sframe', 'sarray', 'sgraph', 'sketch', 'image']
     16 
---> 17 from . import image
     18 from . import sframe
     19 from . import sarray

ImportError: cannot import name image`

gustavla:
Hi @elseagle, thanks for trying out Turi Create! Let's see if we can figure out what's going on. If you do:
```
ls /home/elseagle/venv/local/lib/python2.7/site-packages/turicreate/data_structures/
```
Does it list the file `image.py`? It should, among other files. If this is the case, you can try to uninstall and re-install `pip uninstall -y turicreate && pip install turicreate` to see if that fixes it.

elseagle:
it lists `image.py`

elseagle:
`ls /home/elseagle/venv/local/lib/python2.7/site-packages/turicreate/data_structures/
[...]

PR #1570, opened 2019-03-11
Update LINUX_INSTALL.md

parinpshah94:
Adding `libatlas3-base` as a dependency for turicreate. Several user-repoted turicreate imports failed with this error (`ImportError: libblas.so.3: cannot open shared object file: No such file or directory`). Installing `libatlas3-base` fixes that on Ubuntu 18.04. 

For reference, see issues here: https://github.com/apple/turicreate/issues/1047 and https://github.com/apple/turicreate/issues/191

### DIFF
--- a/LINUX_INSTALL.md
+++ b/LINUX_INSTALL.md
@@ -11,7 +11,12 @@ The minimum requirements are:
 
 ### Ubuntu >= 17.10
 
-Follow the instructions for Ubuntu >= 16.04, but also install `libgconf-2-4`. This was previously included in desktop Ubuntu distributions but now needs to be installed separately starting with 17.10.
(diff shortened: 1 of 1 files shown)


TobyRoseman:
@parinpshah94 - Thank you for the pull request. However [we recently edited that file to remove the libblas dependency](https://github.com/apple/turicreate/commit/07a7f04fcd1855ba6c866d5814d525170c2b1906#diff-cd1420011682a9ab888c76702656133c) as it is no longer a dependency. Starting in our next release people will no longer need to have blas or lapack installed in order to use Turi Create. We should be releasing that later this month.

parinpshah94:
Okay, that's great. Thanks for the update @TobyRoseman!

Issue #1527, opened 2019-03-07
Annotation UI doesn't work on an SFrame with more than one image column


shantanuchhabra:
I have an SFrame with two columns of type `tc.Image`. Annotation UI errors out even when I specify the `image_column` to use.

Repro:
```
annotated = tc.image_classifier.annotate(annotated, image_column="image")                                             
---------------------------------------------------------------------------
ToolkitError                              Traceback (most recent call last)
<ipython-input-36-ce890a73f4fa> in <module>
----> 1 annotated = tc.image_classifier.annotate(annotated, image_column="image")

~/Desktop/test_fest/venv/lib/python3.6/site-packages/turicreate/toolkits/image_classifier/_annotate.py in annotate(data, image_column, annotation_column)
    122     """
    123 
--> 124     image_column = _tkutl._find_only_image_column(data)
    125 
    126     # Check Value of Column Variables

~/Desktop/test_fest/venv/lib/python3.6/site-packages/turicreate/toolkits/_internal_utils.py in _find_only_image_column(sframe)
    111     from turicreate import Image
    112     return _find_only_column_of_type(sframe, target_type=Image,
--> 113                                      type_name='image', col_name='feature')
    114 
    115 

~/Desktop/test_fest/venv/lib/python3.6/site-packages/turicreate/toolkits/_internal_utils.py in _find_only_column_of_type(sframe, target_type, type_name, col_name)
     97         if ctype in target_type:
     98             if image_column_name is not None:
---> 99                 raise ToolkitError('No "{col_name}" column specified and more than one {type_name} column in "dataset". Can not infer correct {col_name} column.'.format(col_name=col_name, type_name=type_name))
    100             image_column_name = name
    101     if image_column_name is None:

ToolkitError: No "feature" column specified and more than one image column in "dataset". Can not infer correct feature column.
```

PR #1572, opened 2019-03-11
Image classification/annotation multiple image columns

abhishekpratapa:
- Fixes the case for multiple image columns present in the intended annotation dataset


abhishekpratapa:
Fixes #1527

PR #1575, opened 2019-03-11
Drawing Classifier: Expose only max_iterations to the user.

### LINKED ISSUE #1488: Drawing Classifier: max_iterations vs num_epochs
At the moment, the drawing classifier toolkit exposes both the max_iterations and the num_epochs to the user. We want to think about that some more.

shantanuchhabra:


src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -51,9 +51,13 @@ def create(input_dataset, feature="drawing", target="label",
...
 
     if max_iterations == 0:
-        max_iterations = num_epochs * len(input_dataset) / batch_size
+        # @TODO: Be able to automatically choose number of epochs based on 
+        # data size
+        num_epochs = 100
+        num_iterations = num_epochs * len(input_dataset) / batch_size
     else:
         num_epochs = max_iterations * batch_size / len(input_dataset)
+        num_iterations = max_iterations
```
nickjong: If I'm not mistaken, our other toolkits (besides OD) use "iteration" to refer to a pass through the training data. For example if I specify, `max_iterations=100`, should I expect 100 passes through the training data? Or 100 mini-batches?

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -160,7 +164,7 @@ def update_progress(cur_loss, iteration):
         'batch_size': batch_size,
         'training_loss': cur_loss,
         'training_time': training_time,
-        'max_iterations': max_iterations,
+        'num_iterations': num_iterations,
```
nickjong: I believe our other toolkits store `max_iterations` in the model state

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: The intention here is 100 mini-batches, but if our other toolkits use it to refer to number of epochs, I'll change the code to account for it.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: @nickjong Looks like all toolkits except for the Object Detector and Style Transfer take `max_iterations` to mean "Maximum number of Epochs". However, OD and Style Transfer take it to mean "Maximum number of training iterations ~ number of mini-batches". While the latter is more correct from an academic perspective, I have modified the drawing classifier to take `max_iterations` to mean "Maximum number of Epochs" for two reasons, after discussing with @ylow and @TobyRoseman :
1. An epoch is a more intuitive parameter to pass in for our users.
2. A default value for number of epochs makes more sense for this toolkit, just like it does for the Image Classifier.
3. To align with other toolkits, especially the Image Classifier.

It is a unfortunate that the same parameter name means different things for different toolkits. There are three possible solutions, all API-breaking changes outside the scope of this PR that we should seriously think about before executing:
1. Rename the parameter from `max_iterations` to `max_epochs` for all toolkits except ST/OD. 
2. Maybe change OD and ST logic to also take in number of epochs but that may be suboptimal for the respective toolkits. 
3. For OD and ST, rename the parameter from `max_iterations` to `max_mini_batches` or something distinct from `max_iterations` to not cause confusion.

Note that for any parameter renames, there is a backwards-compatibility supporting path through `kwargs`, as pointed out by @ylow .

nickjong:
Please give pull requests human-readable titles, so it's easier to tell what it does without having to look up an issue number or parsing through the contents

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -51,9 +51,9 @@ def create(input_dataset, feature="drawing", target="label",
     start_time = _time.time()
 
     if max_iterations == 0:
-        max_iterations = num_epochs * len(input_dataset) / batch_size
-    else:
-        num_epochs = max_iterations * batch_size / len(input_dataset)
+        # @TODO: Be able to automatically choose number of epochs based on 
+        # data size
+        max_iterations = 100
```
nickjong: Why not just set the default in line 42?

shantanuchhabra: Fixed.

PR #1579, opened 2019-03-12
Updated coremltools finding routines to be a bit more flexible.

hoytak:


src/external/coremltools_wrap/CMakeLists.txt
```diff
@@ -4,13 +4,11 @@ project(TuriExternalDependencies)
...
-
-file(GLOB _mlmodel_files
+file(GLOB_RECURSE _mlmodel_files
   RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
   CONFIGURE_DEPENDS
-  ${TC_COREMLTOOLS_DIR}/mlmodel/build/format/*.cc
-  ${TC_COREMLTOOLS_DIR}/mlmodel/src/*.cpp
-  ${TC_COREMLTOOLS_DIR}/mlmodel/src/transforms/*.cpp)
+  ${TC_COREMLTOOLS_DIR}/mlmodel/build/*.cc
+  ${TC_COREMLTOOLS_DIR}/mlmodel/src/*.cpp)
```
znation: Do we no longer need `${TC_COREMLTOOLS_DIR}/mlmodel/src/transforms/*.cpp`? Or does GLOB_RECURSE already include that with this line?

hoytak: GLOB recurse includes that line.

PR #1581, opened 2019-03-13
Drawing Classifier: Added support for validation set + fixed training metric reporting

### LINKED ISSUE #1464: Fix Drawing Classifier Training Metric Reporting
- [x] Get rid of smoothed loss in Drawing Classifier.
- [x] Yield statistics at the granularity of a training step.
- [x] Output metrics like accuracy.

### LINKED ISSUE #1493: Add automatic validation set to the Drawing Classifier
To prevent overfitting, we should implement a validation split as part of training in the Drawing Classifier.

shantanuchhabra:
Fixes #1493 
Fixes #1464

### DIFF
--- a/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -20,6 +20,11 @@ BITMAP_HEIGHT = 28
 
 def _raise_error_if_not_drawing_classifier_input_sframe(
     dataset, feature, target):
+    """
+    Performs some sanity checks on the SFrame provided as input to 
+    `turicreate.drawing_classifier.create` and raises a ToolkitError
+    if something in the dataset is missing or wrong.
+    """
     from turicreate.toolkits._internal_utils import _raise_error_if_not_sframe
     _raise_error_if_not_sframe(dataset)
     if feature not in dataset.column_names():
@@ -37,11 +42,72 @@ def _raise_error_if_not_drawing_classifier_input_sframe(
     if len(dataset) == 0:
         raise _ToolkitError("Input Dataset is empty!")
 
-def create(input_dataset, feature="drawing", target="label", 
-            pretrained_model_url=None, classes=None, batch_size=256, 
-            num_epochs=100, max_iterations=0, verbose=True, **kwargs):
+def create(input_dataset, target, feature=None, 
+            pretrained_model_url=None, batch_size=256, 
+            num_epochs=100, max_iterations=0, verbose=True):
     """
     Create a :class:`DrawingClassifier` model.
+
+    Parameters
+    ----------
+    dataset : SFrame
+        Input data. The columns named by the ``feature`` and ``target``
+        parameters will be extracted for training the drawing classifier.
+
+    target : string
+        Name of the column containing the target variable. The values in this
+        column must be of string or integer type.
+
+    feature : string optional
+        Name of the column containing the input drawings. 'None' (the default)
+        indicates the column in `dataset` named "drawing" should be used as the
+        feature.
+        The feature column can contain both bitmap-based drawings as well as
+        stroke-based drawings. Bitmap-based drawing input can be a grayscale
+        tc.Image of any size.
+        Stroke-based drawing input must be in the following format:
+        Every drawing must be represented by a list of strokes, where each
+        stroke must be a list of points in the order in which they were drawn
+        on the canvas.
+        Each point must be a dictionary with two keys, "x" and "y", and their
+        respective values must be numerical, i.e. either integer or float.
+
+    pretrained_model_url : string optional
+        A URL to the pretrained model that must be used for a warm start before
+        training.
+
+    batch_size: int optional
+        The number of images per training step. If not set, a default
+        value of 256 will be used. If you are getting memory errors,
+        try decreasing this value. If you have a powerful computer, increasing
+        this value may improve performance.
+
+    max_iterations : int optional
+        The maximum number of allowed passes through the data. More passes over
+        the data can result in a more accurately trained model. 
+
+    verbose : bool optional
+        If True, print progress updates and model details.
+
+    Returns
+    -------
+    out : DrawingClassifier
+        A trained :class:`DrawingClassifier` model.
+
+    See Also
+    --------
+    DrawingClassifier
+
+    Examples
+    --------
+    .. sourcecode:: python
+
+        # Train a drawing classifier model
+        >>> model = turicreate.drawing_classifier.create(data)
+
+        # Make predictions on the training set and as column to the SFrame
+        >>> data['predictions'] = model.predict(data)
+
     """
     import mxnet as _mx
     from mxnet import autograd as _autograd
@@ -54,6 +120,10 @@ def create(input_dataset, feature="drawing", target="label",
         max_iterations = num_epochs * len(input_dataset) / batch_size
     else:
         num_epochs = max_iterations * batch_size / len(input_dataset)
+
+    # automatically infer feature column
+    if feature is None:
+        feature = _tkutl._find_only_drawing_column(input_dataset)
 
     _raise_error_if_not_drawing_classifier_input_sframe(
         input_dataset, feature, target)
@@ -70,8 +140,7 @@ def create(input_dataset, feature="drawing", target="label",
     progress = {'smoothed_loss': None, 'last_time': 0}
     iteration = 0
 
-    if classes is None:
-        classes = dataset[target].unique()
+    classes = dataset[target].unique()
     classes = sorted(classes)
     class_to_index = {name: index for index, name in enumerate(classes)}
 
@@ -168,7 +237,12 @@ def create(input_dataset, feature="drawing", target="label",
     return DrawingClassifier(state)
 
 class DrawingClassifier(_CustomModel):
+    """
+    A trained model that is ready to use for classification, and to be 
+    exported to Core ML.
 
+    This model should not be constructed directly.
+    """
     def __init__(self, state):
         self.__proxy__ = _PythonProxy(state)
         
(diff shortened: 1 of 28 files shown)


src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -132,7 +146,7 @@ def create(input_dataset, target, feature=None,
     dataset = _extensions._drawing_classifier_prepare_data(
         input_dataset, feature) if is_stroke_input else input_dataset
 
-    column_names = ['Iteration', 'Loss', 'Elapsed Time']
+    column_names = ['Training Step', 'Loss', 'Validation Accuracy', 'Elapsed Time']
```
TobyRoseman: I think `Iteration` makes more sense than `Training Step`. Displaying `Training Accuracy` would be good too. I'm not sure there is much value in showing `Loss`.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -144,7 +158,27 @@ def create(input_dataset, target, feature=None,
...
+            if verbose:
+                print_validation_track_notification()
+            dataset, validation_dataset = dataset.random_split(
+                TRAIN_VALIDATION_SPLIT, seed=seed)
+        else:
+            validation_dataset = _tc.SFrame()
+    elif validation_set is None:
+        validation_dataset = _tc.SFrame()
+    else:
+        raise TypeError("Unrecognized value for 'validation_set'.")
```
TobyRoseman: `value` -> `type`.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -144,7 +158,27 @@ def create(input_dataset, target, feature=None,
     classes = sorted(classes)
     class_to_index = {name: index for index, name in enumerate(classes)}
 
-    def update_progress(cur_loss, iteration):
+    if isinstance(validation_set, _tc.SFrame):
+        _raise_error_if_not_drawing_classifier_input_sframe(
+            validation_set, feature, target)
+        is_validation_stroke_input = (validation_set[feature].dtype != _tc.Image)
```
TobyRoseman: It doesn't look like `is_validation_stroke_input` is getting used.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -42,9 +43,13 @@ def _raise_error_if_not_drawing_classifier_input_sframe(
     if len(dataset) == 0:
         raise _ToolkitError("Input Dataset is empty!")
 
-def create(input_dataset, target, feature=None, 
+def print_validation_track_notification():
```
TobyRoseman: I'm not really understanding why this is its own function. It should probably be a hidden function.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -196,26 +239,42 @@ def update_progress(cur_loss, iteration):
...
+        for validation_batch in validation_loader:
+            validation_batch_data = _mx.gluon.utils.split_and_load(
+                validation_batch.data[0], ctx_list=ctx, batch_axis=0)
+            validation_batch_label = _mx.gluon.utils.split_and_load(
+                validation_batch.label[0], ctx_list=ctx, batch_axis=0)
+            outputs = []
+            for x, y in zip(validation_batch_data, validation_batch_label):
+                z = model(x)
+                outputs.append(z)
+            # Updates internal evaluation
```
TobyRoseman: Not really understanding this comment

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: I avoided saying "Iteration" because of the `max_iterations` parameter where an iteration is used to refer to an epoch. What do you think? I'll add a Training Accuracy column.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: Thanks for the catch, the next line should use it.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: Updated.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
TobyRoseman: Oh, I'm confused. I thought you were printing out one row per iteration/epoch. . . . .  I would recommend doing that rather than print out every ten seconds or so.

TobyRoseman:
Take a look a the [ProgressTablePrinter util](https://github.com/apple/turicreate/blob/master/src/unity/python/turicreate/util/progress_table_printer.py) I just checked. I'm using it [in the sound classifier](https://github.com/apple/turicreate/blob/23e1e8be55f18580939df768c6561c2e22e48ad8/src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py#L108) and it makes things easier and cleaner. Perhaps you want to use it here too.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -42,9 +43,9 @@ def _raise_error_if_not_drawing_classifier_input_sframe(
     if len(dataset) == 0:
         raise _ToolkitError("Input Dataset is empty!")
 
-def create(input_dataset, target, feature=None, 
+def create(input_dataset, target, feature=None, validation_set='auto',
             pretrained_model_url=None, batch_size=256, 
-            num_epochs=100, max_iterations=0, verbose=True):
+            num_epochs=100, max_iterations=0, seed=None, verbose=True):
```
TobyRoseman: I'm not sure there is a lot of value to having a `seed` parameter if it's just being used for the train/validation set split. If the user wants reproducibility they can construct the validation set themselves before calling create. If you do want a `seed` parameter, it should be mentioned in the docstring.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -132,7 +142,7 @@ def create(input_dataset, target, feature=None,
     dataset = _extensions._drawing_classifier_prepare_data(
         input_dataset, feature) if is_stroke_input else input_dataset
 
-    column_names = ['Iteration', 'Loss', 'Elapsed Time']
+    column_names = ['Training Step', 'Loss', 'Train Accuracy', 'Validation Accuracy', 'Elapsed Time']
```
TobyRoseman: So what happens if the user calls `create` with `validation_set=None`? Is that column just empty? It would probably be better for that column to just not appear.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
nickjong: Let's just be consistent with our other toolkits for now

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -196,39 +209,63 @@ def update_progress(cur_loss, iteration):
...
     state = {
         '_model': model,
         '_class_to_index': class_to_index,
         'num_classes': len(classes),
         'classes': classes,
         'input_image_shape': (1, BITMAP_WIDTH, BITMAP_HEIGHT),
         'batch_size': batch_size,
-        'training_loss': cur_loss,
-        'training_time': training_time,
+        'training_loss': train_loss,
```
nickjong: Shouldn't we also include the training accuracy, and validation accuracy if available?

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -196,39 +209,63 @@ def update_progress(cur_loss, iteration):
...
+                # Computes softmax cross entropy loss.
+                loss = softmax_cross_entropy(z, y)
+                # Backpropagate the error for one iteration.
+                loss.backward()
+                train_outputs.append(z)
+            # Compute training accuracy
+            train_accuracy.update(train_batch_label, train_outputs)
+
+        # Compute validation accuracy
+        for validation_batch in validation_loader:
```
nickjong: Doesn't this iterate through the validation loader for _every_ mini-batch (training step)?

nickjong: Doesn't the validation_loader need to be reset? How does it iterate through the data more than once?

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
TobyRoseman: If consistency with other toolkits is our priority, then I don't think we should have a `seed` parameter here. Other toolkits take a `seed` parameter but I believe all of them will produce the same model given the same seed and the same dataset. That isn't true here.

nickjong: If fixing the seed doesn't make model creation deterministic, then I don't think we should expose it, I agree

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: You're right. The validation loader in fact runs the data through only one epoch throughout training. Fixing this now.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: Fixed.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: Fixed.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: Fixed.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -211,6 +210,27 @@ def create(input_dataset, target, feature=None, validation_set='auto',
 
     train_accuracy = _mx.metric.Accuracy()
     validation_accuracy = _mx.metric.Accuracy()
+    
+    def compute_validation_accuracy(validation_accuracy):
+        validation_loader = _SFrameClassifierIter(validation_dataset, batch_size,
```
nickjong: You should also be able just to create the iterator once and call `reset` on it, I think

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -196,39 +208,79 @@ def update_progress(cur_loss, iteration):
...
+                z = model(x)
+                # Computes softmax cross entropy loss.
+                loss = softmax_cross_entropy(z, y)
+                # Backpropagate the error for one iteration.
+                loss.backward()
+                train_outputs.append(z)
+            # Compute training accuracy
+            train_accuracy.update(train_batch_label, train_outputs)
+
+        compute_validation_accuracy(validation_accuracy)
```
nickjong: I believe you're still doing this after every mini-batch

shantanuchhabra: Fixed.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: Fixed.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -194,39 +206,72 @@ def update_progress(cur_loss, iteration):
...
-            loss = softmax_cross_entropy(output, label)
-        loss.backward()
-        # update parameters
-        trainer.step(1)
+            for x, y in zip(train_batch_data, train_batch_label):
+                z = model(x)
+                # Computes softmax cross entropy loss.
+                loss = softmax_cross_entropy(z, y)
+                # Backpropagate the error for one iteration.
+                loss.backward()
```
TobyRoseman: I'm implementing similar functionality for the sound classifier and in my testing I noticed an interesting bug. I believe your code also has the same issue. Perhaps it's also in other toolkits as well....

Try passing in the same SFrame to `create(...)` as both the `input_dataset` and the `validation_set`. The training accuracy and the validation accuracy columns that get printed out should match, since they're the same SFrame, but I don't think they will.

I don't think it's correct to save the predicted labels while you're still doing backward passes. You probably want to do all your backward passes for in one loop and then iterate over your training set again and save the predictions.

shantanuchhabra: Fixed. Thanks for pointing this out!

PR #1583, opened 2019-03-13
Int histogram: align bin boundaries to ints

### LINKED ISSUE #167: Histogram of ints should have integer bin boundaries
See screenshot. When computing a histogram of integers, the bin boundaries should also be integers. Currently they are floating point (and may not align to the int values in the dataset).

screen shot 2018-01-04 at 11 35 47 pm

znation:
Previously they were aligned to non-int values awkwardly, despite the underlying
values being ints.

In this commit:

* Templatize the histogram transformer, with a few special cases of
  logic (and assertions to ensure correct behavior) for the `flex_int`
  case, to ensure that bin boundaries come out int-aligned after
  resizing and rescaling.
* Some extra plumbing into the histogram Vega spec to allow for bin
  boundaries to be declared.
* Sends the first data batch with the Vega spec (previously we waited
  until the vega spec had been received). This is to work around visual
  glitches on first render when signals reference data values. The front
  end UI changes are all to accommodate this change.
* Moves some string escaping logic into escape.hpp/cpp.

Fixes #167


znation:
@abhishekpratapa Please review, thanks!
@kanitw Please review the Vega spec changes.
@nickjong, @hoytak: FYI, this ended up being a way bigger (and probably riskier) change than I had planned on, so I'm thinking we should wait on merging this until 5.5.

znation:
@hoytak @abhishekpratapa When you have a chance, I'd appreciate a review from both of you, thanks!

znation:
@hoytak @abhishekpratapa This has passed internal tests and is otherwise ready to go in. Please review, thanks!

PR #1584, opened 2019-03-13
SArray Builder: doc string fix + minor clean up

TobyRoseman:
If `dtype` is `type(None)`, `self._builder.init(num_segments, history_size, dtype)` fails with error message `RuntimeError: Must specify a valid type.`. So I've removed that case.

### DIFF
--- a/src/unity/python/turicreate/data_structures/sarray_builder.py
+++ b/src/unity/python/turicreate/data_structures/sarray_builder.py
@@ -21,16 +21,15 @@ class SArrayBuilder(object):
 
     Parameters
     ----------
+    dtype : type
+        The type of the elements in the SArray.
+
     num_segments : int, optional
         Number of segments that can be written in parallel.
 
     history_size : int, optional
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/data_structures/sarray_builder.py
```diff
@@ -40,22 +39,19 @@ class SArrayBuilder(object):
...
     >>> sb.close()
     dtype: int
     Rows: 3
     [1, 2, 3]
-
     """
     def __init__(self, dtype, num_segments=1, history_size=10):
         self._builder = UnitySArrayBuilderProxy()
-        if dtype is None:
-            dtype = type(None)
```
hoytak: Does `dtype = None` work ?

TobyRoseman: It does not. Error message: `TypeError: dtype argument cannot be None.`

PR #1586, opened 2019-03-13
WIP: Add missing JS files used to generate documentation images

### LINKED ISSUE #1092: Figure missing from regression user guide
**Observed Behavior:**
User guide page: https://apple.github.io/turicreate/docs/userguide/supervised-learning/regression.html

“The following figure depicts a sample supervised learning model where the training data (depicted in blue) is used to learn a function (depicted in pink) which can then be used in the future to make predictions.”

No figure on page

**Expected Behavior:**
Figure to be present on page.
Figure properly appears on old user guide version: https://turi.com/learn/userguide/supervised-learning/regression.html

syoutsey:
## Summary

This adds the necessary Javascript files used to generate the documentation images like [here](https://turi.com/learn/userguide/supervised-learning/linear-regression.html). Fixes #1092.

## Open questions

- It's unclear if this is the conventional "gitbook-y" way to include these relevant files
- It's unclear how these files were not included during the documentation transition from Turi.com to Github.io
- This only adds back in the Javascript files but does not republish the static site; I believe this is done with `gitbook build .`

### DIFF
--- a/userguide/README.md
+++ b/userguide/README.md
@@ -89,3 +89,7 @@ With Core ML, you can integrate machine learning models into your macOS,
 iOS, watchOS, and tvOS app. Many models created  in Turi Create can be
 exported for use in Core ML.
 
+<script type="text/javascript" src="../turi/js/jquery/jquery-2.1.3.min.js"></script>
+<script type="text/javascript" src="../turi/js/d3/d3.min.js"></script>
+<script type="text/javascript" src="../supervised-learning/images/random-utils.js"></script>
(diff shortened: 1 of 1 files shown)


syoutsey:
This won't work as the script includes tags get rendered by Github on our landing page. Marking as WIP and investigating other solutions.

abhishekpratapa:
This may actually be the solution. Like you pointed out there is precedence. https://github.com/apple/turicreate/blob/master/userguide/supervised-learning/linear-regression.md

Issue #1298, opened 2019-01-09
Linear Regression Gives Incorrect Error Message with Image Column 


TobyRoseman:
Using the [Kaggle Cats and Dogs dataset](https://www.microsoft.com/en-us/download/details.aspx?id=54765), the following code:
```python
import turicreate as tc
import os
data = tc.image_analysis.load_images('PetImages', with_path=True)
data['label'] = data['path'].apply(lambda path: 'dog' if '/Dog' in path else 'cat')
model = tc.linear_regression.create(data, target='label')
```
Give the following error message: `ToolkitError: Column type of target 'label' must be int or float.`

However linear regression does support string target types. The following works fine:
```python
data = tc.SFrame({'x': ['0', '1', '2', '99', '100', '101'], 'y':['low', 'low', 'low', 'high', 'high', 'high']})
model = tc.linear_regression.create(data, target='y')
```

nickjong:
Well, following your example, I get:
```
>>> model.predict(data)
dtype: float
Rows: 6
[1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
>>> model.evaluate(data)
{'max_error': 0.0, 'rmse': 0.0}
```

It doesn't seem like we're doing anything sensible with the string targets. Maybe the model training here should throw an error too?

shantanuchhabra:
I think it's because a validation set is automatically created when the `dataset_size >= 100` and the type of the target column is only checked for then. I think the fix is to do a sanity check on the target column (along with maybe other checks that are independent of creating a validation set) in the code path when no validation set is created.

fareeha:
> I think it's because a validation set is automatically created when the `dataset_size >= 100` and the type of the target column is only checked for then. I think the fix is to do a sanity check on the target column (along with maybe other checks that are independent of creating a validation set) in the code path when no validation set is created.

@shantanuchhabra I already fixed this issue. We went over this fix yesterday - thanks for re-iterating it here..

hoytak:
Huh.  Linear regression should not take string targets.

PR #1598, opened 2019-03-14
Check Linear Regression target column data type

fareeha:
Linear Regression accepts 'int' and 'float' target values. Previously `check_target_column_type()` which checks the type of target values was called on the validation set. Validation set is only created for a dataset size 100 or larger. For cases when data with fewer than 100 examples was given, the check was not applied and target values of type string were being accepted and handled incorrectly. 

The check for target values type is now applied on the target column of the data passed in. Fixes https://github.com/apple/turicreate/issues/1298

### DIFF
--- a/src/unity/toolkits/supervised_learning/supervised_learning.cpp
+++ b/src/unity/toolkits/supervised_learning/supervised_learning.cpp
@@ -1037,6 +1037,8 @@ void supervised_learning_model_base::api_train(
   sframe X = f_data.materialize_to_sframe(); 
     
   sframe y = data.select_columns({target}).materialize_to_sframe();
+    
+  check_target_column_type(this->name(), y);
 
   ml_missing_value_action missing_value_action =
     this->support_missing_value() ? ml_missing_value_action::USE_NAN
@@ -1049,9 +1051,6 @@ void supervised_learning_model_base::api_train(
     valid_X = validation_data.select_columns(f_data.column_names()).materialize_to_sframe();
     
     valid_y = validation_data.select_columns({target}).materialize_to_sframe();
-
-    
-    check_target_column_type(this->name(), valid_y);
 
     auto valid_filter_names = f_data.column_names();
     valid_filter_names.push_back(target);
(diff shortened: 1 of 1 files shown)


src/unity/toolkits/supervised_learning/supervised_learning.cpp
```diff
@@ -1050,9 +1052,6 @@ void supervised_learning_model_base::api_train(
     
     valid_y = validation_data.select_columns({target}).materialize_to_sframe();
 
-    
-    check_target_column_type(this->name(), valid_y);
```
nickjong: Should we no longer be checking the validation set too?

fareeha: Validation set is created from the dataset that is provided by the user. So I feel it might be repetitive to also call the check function again on a section of that dataset.

nickjong: Although validation data can be automatically sampled from the training data, I believe our API also allows users to provide the validation set as a completely unrelated SFrame.

fareeha: Yes, you're right. Thank you for pointing that! It was worth verifying that case as well. So, in the case when a user provides their own validation set: the model `create()` calls `_validate_data(dataset, validation_set)`. The validation_set is set to 'auto' if we want the validation set to the generated automatically or it can be given an SFrame containing the validation set. In the latter case, `_validate_data(dataset, validation_set)` applies append on the two SFrames using `validation_set.head().append(dataset.head())`  to check schema. This allows verifying that the columns and their types of both `dataset` and `validation_data` match.
Since we are now checking the type of the `dataset` target column, and by extension we make sure that the `validation_data` also has the matching valid column type.

hoytak: I think the check should probably be in both places.  That's the easiest fix here.

fareeha: Good idea, to be safe the check on validation data column type should be restored.

Issue #1476, opened 2019-03-01
Sounds Classifier Needs Method to Prevent Overfitting


TobyRoseman:
Currently the sound classifier has no method to prevent overfitting. We should let the user pass in a validation set. Or if we have enough data, automatically make one. Alternatively, perhaps we could take a regularization parameter.

Thanks @ylow for finding this issue.

PR #1599, opened 2019-03-14
Sound classifier validation set

TobyRoseman:


src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -68,16 +76,29 @@ def create(dataset, target, feature, max_iterations=10, verbose=True, batch_size
         raise _ToolkitError("'%s' column is not audio data." % feature)
     if target not in dataset.column_names():
         raise _ToolkitError("Target column '%s' does not exist" % target)
+    if not (isinstance(validation_set, _tc.SFrame) or validation_set == 'auto' or validation_set is None):
+        raise TypeError("Unrecognized value for 'validation_set'")
```
shantanuchhabra: value -> type

shantanuchhabra: value -> type

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -126,19 +163,39 @@ def create(dataset, target, feature, max_iterations=10, verbose=True, batch_size
...
+            outputs = [custom_NN(x) for x in data]
+            train_metric.update(label, outputs)
+        train_data.reset()
+
+        # Calculate validataion metric
+        for batch in validation_data:
+            data = _mx.gluon.utils.split_and_load(batch.data[0], ctx_list=ctx, batch_axis=0, even_split=False)
+            label = _mx.gluon.utils.split_and_load(batch.label[0], ctx_list=ctx, batch_axis=0, even_split=False)
+            outputs = [custom_NN(x) for x in data]
+            validation_metric.update(label, outputs)
```
shantanuchhabra: This for loop and the one above could probably be two function calls.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -105,18 +137,23 @@ def create(dataset, target, feature, max_iterations=10, verbose=True, batch_size
...
         # Setup progress table
-        table_printer = _tc.util._ProgressTablePrinter(['epoch', 'accuracy', 'time'],
-                                                        ['Epoch', 'Accuracy (%)', 'Elapsed Time (seconds)'])
-
-    metric = _mx.metric.Accuracy()
+        row_ids = ['epoch', 'train_accuracy', 'time']
+        row_display_names = ['Epoch', 'Training Accuracy (%)', 'Elapsed Time (seconds)']
+        if validation_data:
+            row_ids.insert(2, 'validation_accuracy')
+            row_display_names.insert(2, 'Validation Accuracy (%)')
```
shantanuchhabra: Nice, I'll do this too. This is cleaner.

TobyRoseman:
Fixes #1476

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: I actually think "value" makes more sense here. If the user passes in `validation_set='foobar'`, it's not the type I have a problem with, it's the value.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: It certainly could be, but I don't think it's worth it. I guess it would look something like this:

```python
def compute_accuracy(data, metric):
    for batch in data:
        features = _mx.gluon.utils.split_and_load(batch.data[0], ctx_list=ctx, batch_axis=0, even_split=False)
        y = _mx.gluon.utils.split_and_load(batch.label[0], ctx_list=ctx, batch_axis=0, even_split=False)
        outputs = [custom_NN(x) for x in features]
        metric.update(label, outputs)
```

```python
compute_accuracy(train_data, train_metric)
train_data.reset()
compute_accuracy(validation_data, validation_metric)
```
It's only a savings of two lines. Not really sure that's worth an inner function that isn't completely straight forward. What do you think?

shantanuchhabra: Absolutely, I wouldn’t hold up merging this PR for that change

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
shantanuchhabra: That’s true, but isn’t it also the case that if I pass in an unrecognized type for the validation set, I’ll get the same error? Regardless, I don’t think this should hold up merging this in. It’s a really small nitpick anyway.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
nickjong: I agree that it's more readable as is.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
nickjong: "Value" seems fine here to me, FWIW

TobyRoseman:
Passed on internal tests.

PR #1602, opened 2019-03-14
numba and llvmlite now have Python 3.5 wheels on PyPI

TobyRoseman:
The latest versions of [numba](https://pypi.org/project/numba/#files) and [llvmlite](https://pypi.org/project/llvmlite/#files) now both have wheels for Python 3.5. So I think it's safe to remove these pins.


TobyRoseman:
Passed on internal tests.

PR #1608, opened 2019-03-15
Prevent multithreaded execution within lambda workers. 

hoytak:
Added environment variables in the lambda workers to try to prevent multithreaded execution by common frameworks on lambda workers.


hoytak:
Passed on internal tests.

PR #1624, opened 2019-03-18
Image Classification Plumbing to Use Image Similarity

abhishekpratapa:


src/unity/lib/annotation/annotation_base.hpp
```diff
@@ -66,6 +66,11 @@ class EXPORT AnnotationBase : public ml_model_base {
 
   std::shared_ptr<annotation_global> get_annotation_registry();
 
+  virtual void calculateSimilarItems() = 0;
+
+  virtual annotate_spec::Data getSimilarItems(annotate_spec::Datum &item,
```
hoytak: Is there a reason that this reference is non-const?  `type&` Implies that the value will get modified by the calling function; `const type&` is how it should be done if that value will not get modified.

src/unity/lib/annotation/image_classification.cpp
```diff
@@ -238,7 +250,7 @@ void ImageClassification::_addAnnotationToSFrame(size_t index, int label) {
 
 void ImageClassification::cast_annotations() {
   size_t annotation_column_index = m_data->column_index(m_annotation_column);
-  if(m_data->dtype().at(annotation_column_index) == flex_type_enum::INTEGER) {
+  if (m_data->dtype().at(annotation_column_index) == flex_type_enum::INTEGER) {
```
hoytak: Instead of `m_data->dtype().at(...)`, use `m_data->column_type(m_annotation_column)`

znation:
@abhishekpratapa is this the same thing as #1638? Which one should we be reviewing?

abhishekpratapa:
@znation It's subtly different. One is using the image similarity function we wrote in the other PR to implement the sorting in the Image Classification Toolkit. The other #1638 is the actual implementation of the image_similarity toolkit in C++.

nickjong:
But it looks like they have overlapping changes. For the sake of sanity, which one should be reviewed first? Then the other should be proposed as the incremental changes on top of the first (ideally, after merging the first).

In general, we should really avoid having two open PRs that share some of the same changes. Sometimes the two PRs should really be one bigger PR. More often than not, one should come first. Or there should be a third PR with just the shared changes.

Offhand recommendation: work bottom-up.

abhishekpratapa:
Nick Good Point. I think this one should be reviewed first then.

znation:
@abhishekpratapa Is this one ready for review? There are a lot of TODO comments in what looks like the core functionality of this PR.

PR #1626, opened 2019-03-18
Drawing Classifier: Added predict_topk and output_type to predict

### LINKED ISSUE #1618: Drawing classifier missing prediction options
The following options are missing in drawing classifier prediction:

- predict_topk (to get more than one prediction)
- predict (with output type)

All other toolkits have the following options.
```
output_type : {'probability', 'class', 'probability_vector'}, optional
    Form of the predictions which are one of:

    - 'class': Class prediction. For multi-class classification, this
      returns the class with maximum probability.
    - 'probability': Prediction probability associated with the True
      class (not applicable for multi-class classification)
    - 'probability_vector': Prediction probability associated with each
      class as a vector. Label ordering is dictated by the ``classes``
      member variable.
```

shantanuchhabra:
Fixes #1618

### DIFF
--- a/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -489,7 +489,7 @@ class DrawingClassifier(_CustomModel):
         _save_spec(spec, filename)
 
 
-    def _predict_with_probabilities(self, input_dataset, verbose = True):
+    def _predict_with_probabilities(self, input_dataset, verbose=True):
         """
         Predict with probabilities. The core prediction part that both 
         `evaluate` and `predict` share.
@@ -567,7 +567,7 @@ class DrawingClassifier(_CustomModel):
         return (_tc.SFrame({self.target: _tc.SArray(all_predicted),
             'probability': _tc.SArray(all_probabilities)}))
 
-    def evaluate(self, dataset, metric = 'auto', verbose = True):
+    def evaluate(self, dataset, metric='auto', verbose=True):
         """
         Evaluate the model by making predictions of target values and comparing
         these to actual values.
@@ -656,7 +656,7 @@ class DrawingClassifier(_CustomModel):
         
         return ret
 
-    def predict(self, data, verbose = True):
+    def predict(self, data, output_type='class', verbose=True):
         """
         Predict on an SFrame or SArray of drawings, or on a single drawing.
 
@@ -670,6 +670,16 @@ class DrawingClassifier(_CustomModel):
             If the data is a single drawing, it can be either of type tc.Image,
             in which case it is a bitmap-based drawing input,
             or of type list, in which case it is a stroke-based drawing input.
+
+        output_type : {'probability', 'class', 'probability_vector'} optional
+            Form of the predictions which are one of:
+            - 'class': Class prediction. For multi-class classification, this
+              returns the class with maximum probability.
+            - 'probability': Prediction probability associated with the True
+              class (not applicable for multi-class classification)
+            - 'probability_vector': Prediction probability associated with each
+              class as a vector. Label ordering is dictated by the ``classes``
+              member variable.
 
         verbose : bool optional
             If True, prints prediction progress.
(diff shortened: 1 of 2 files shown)


src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -656,21 +657,137 @@ def evaluate(self, dataset, metric = 'auto', verbose = True):
...
+        | 2  |   2   |  0.00151345680933 |
+        | 2  |   7   | 0.000637513934635 |
+        | 3  |   1   |   0.998070061207  |
+        | .. |  ...  |        ...        |
+        +----+-------+-------------------+
+        [35688 rows x 3 columns]
+        """
+        _tkutl._check_categorical_option_type("output_type", output_type, 
+            ["probability", "rank"])
+        int_type_error_string = (lambda var_name: ("'{var_name}' must be "
```
TobyRoseman: `lambda` are used to create anonymous functions. So it's a little strange here that you're giving them names. 

I get that you're trying to reduce duplication, but I think having these helper functions build the error messages is hurting your code readability.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -656,21 +657,137 @@ def evaluate(self, dataset, metric = 'auto', verbose = True):
...
+        if output_type == 'probability':
+            results = prob_vector.apply(lambda p: [
+                            {'class': classes[i], 'probability': p[i]}
+                            for i in reversed(_np.argsort(p)[-k:])]
+                        )
+        else:
+            assert(output_type == 'rank')
+            def convert_probability_vector_to_ranks(p):
+                rank, ranks_list = 0, []
+                for index in reversed(_np.argsort(p)[-k:]):
```
TobyRoseman: Could you just do `for rank, index in enumerate(reversed(_np.argsort(p)[-k:])):` and then delete line 750?

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -656,21 +657,137 @@ def evaluate(self, dataset, metric = 'auto', verbose = True):
...
+
+    def predict(self, data, output_type='class', batch_size=None, verbose=True):
         """
         Predict on an SFrame or SArray of drawings, or on a single drawing.
 
         Parameters
         ----------
         data : SFrame | SArray | tc.Image | list
-            The image(s) on which to perform drawing classification.
+            The drawing(s) on which to perform drawing classification.
```
TobyRoseman: It would be best to be consistent about image(s)/drawing(s). It looks like the other docstring in this commit uses "image".

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -701,21 +818,35 @@ def predict(self, data, verbose = True):
             Rows: 10
             [3, 4, 3, 3, 4, 5, 8, 8, 8, 4]
         """
+        output_type = output_type.lower()
```
TobyRoseman: This seems like kind of an odd choice. I could be wrong but I don't think this is something that other toolkits do. I think this parameter should be case sensitive.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -701,21 +818,35 @@ def predict(self, data, verbose = True):
...
+                batch_size,
                 verbose
             )
-        return predicted[self.target]
+        if output_type == "class":
+            return predicted[self.target]
+        elif output_type == "probability":
+            _class_to_index = self._class_to_index
+            target = self.target
+            return predicted.apply(
```
TobyRoseman: An `apply` on just the SArray probably has better performance and I think more readable.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shantanuchhabra: But I need to be able to access both columns in the SFrame, i.e. `target` and `probability`.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
TobyRoseman: ah, I missed the second `row` in the lambda expression.

shantanuchhabra:
Internal tests pass.

PR #1627, opened 2019-03-18
Two sound classifier bug fixes

TobyRoseman:


TobyRoseman:
Internal test passed.

PR #1631, opened 2019-03-19
WIP: Possible fix for numba bug.

hoytak:
This bug is extremely interesting.  It appears to be a numba bug revealed in the cfunc path of numba, which is exposed when numpy is used (as documented in https://github.com/numba/numba/issues/2904. ) This bug is exposed when the a thread other than the main thread on which the python interpreter is started is actually the thread executing the function.  When this happens, PyGILState_Ensure() goes down a different path and attempts to acquire a lock on the interpreter, but the main thread never releases that lock.  

The workaround is to swap the thread states of the interpreter between the evaluation thread and the main thread before evaluation calls to the lambda function.

### DIFF
--- a/src/unity/python/turicreate/test/test_lambda.py
+++ b/src/unity/python/turicreate/test/test_lambda.py
@@ -76,3 +76,20 @@ class LambdaTests(unittest.TestCase):
             return x
         self.assertRaises(RuntimeError, lambda: glconnect.get_unity().parallel_eval_lambda(lambda x: bad_fun(x), ls))
         glconnect.get_unity().parallel_eval_lambda(lambda x: good_fun(x), ls)
+
+    def test_numba(self):
+
+        from numba import jit
+        import numpy as np
+        import turicreate as tc
+
+        @jit
+        def _sum(x):
+            return x[0, 0]
+
+        print("BORK")
+        tc.SArray([np.random.normal(size=(2,2))]*1000).apply(_sum)
+        print("BORK2")
+
+         
+
--- a/src/lambda/pylambda.cpp
+++ b/src/lambda/pylambda.cpp
@@ -17,6 +17,15 @@ namespace turi { namespace lambda {
 void* eval_thread_state;
 pylambda_evaluation_functions evaluation_functions;
 
+static void ensure_thread_state_consistent() { 
+  static bool thread_state_swapped = false; 
+
+  if(!thread_state_swapped) { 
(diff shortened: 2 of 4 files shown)


src/unity/python/turicreate/test/test_lambda.py
```diff
@@ -76,3 +76,20 @@ def bad_fun(x):
...
         glconnect.get_unity().parallel_eval_lambda(lambda x: good_fun(x), ls)
+
+    def test_numba(self):
+
+        from numba import jit
+        import numpy as np
+        import turicreate as tc
+
+        @jit
+        def _sum(x):
```
TobyRoseman: Probably want to call this something other than `_sum` cause it doesn't sum, even `foo` I think would be fine for a unit test.

src/unity/python/turicreate/test/test_lambda.py
```diff
@@ -76,3 +76,20 @@ def bad_fun(x):
...
+            return x[0, 0]
+
+        print("BORK")
```
TobyRoseman: You should get rid of these two print statements.

TobyRoseman:
@hoytak - thanks for working on this! The code looks good to me, just a couple minor things. Although someone more familiar with this part of the code base should really take a look.

I tried testing this fix with the Sound Classifier's preprocessing. However I'm even able to get to the point where I call `tc.sound_classifier.create(...)`. Running the code below out of master is pretty fast (less than a minute). However it takes at least a very long time (maybe never finishes) with your changes rebased on top of master.

The following code uses the [ESC dataset](https://github.com/karoldvl/ESC-50). `data` only has 2,000 rows.
```python
import turicreate as tc
from os.path import basename

# Load the audio data and meta data.
data = tc.load_audio('./ESC-50/audio/')
data['filename'] = data['path'].apply(lambda p: basename(p))
data.__materialize__()
```

hoytak:
Yes, it appears this breaks other code.  It's not ready yet.

ylow:
I am not entirely convinced that the diagnosis is right. I have tried a change where the main thread keeps reentering the Python interpreter thus should avoid any issues with the main thread holding on to the GIL. And that does not seem to work either.

ylow:
I worry that with an incomplete understanding of the problem this fix may cause problems in different places.

hoytak:
I'm abandoning this PR.  This is a won't fix -- there's no easy way I see to get around the numba bug.

Issue #1556, opened 2019-03-08
Annotation should allow automatically sorting by image similarity


znation:
Similar to #1555, but we should also allow "similarity" to be one of the sort orderings.

PR #1638, opened 2019-03-19
Adding Image Similarity to the Annotation Toolkit

abhishekpratapa:
- added featurizer in image_similarity.


abhishekpratapa:
Not labeled WIP anymore. Ready for review!

src/toolkits/image_deep_feature_extractor/image_deep_feature_extractor_toolkit.hpp
```diff
@@ -26,6 +26,7 @@ class EXPORT image_deep_feature_extractor_toolkit : public ml_model_base {
   void init_options(const std::map<std::string, flexible_type>& options) override;
 
   gl_sarray extract_features(gl_sframe data, const std::string& column_name, bool verbose, size_t batch_size) const;
+  gl_sarray sarray_extract_features(gl_sarray data, bool verbose, size_t batch_size) const;
```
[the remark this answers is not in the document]
abhishekpratapa: So I attempted to do that and I got some errors compiling due to something in the Registration Mechanism. Is there any reason in your mind this could happen?

abhishekpratapa:
fixes #1556

src/toolkits/image_deep_feature_extractor/image_deep_feature_extractor_toolkit.hpp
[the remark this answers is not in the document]
znation: Yeah, there is some weirdness with the registration mechanism and overloads. I think it's theoretically possible to get it to work but I'm not sure how. For now, it's fine as is.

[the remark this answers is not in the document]
nickjong: Note that the registration macros don't incorporate any notion of type and are therefore brittle in the face of overloading.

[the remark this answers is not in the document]
nickjong: Hmm, any reason not just to converge to SArray version?

src/unity/lib/annotation/utils.cpp
```diff
@@ -1,10 +1,91 @@
...
+#include <toolkits/image_deep_feature_extractor/image_deep_feature_extractor_toolkit.hpp>
 
-bool turi::annotate::isInteger(std::string s) {
-	if(s.empty() || ((!isdigit(s[0])) && (s[0] != '-') && (s[0] != '+'))) return false;
+namespace turi {
+namespace annotate {
 
-	char * p;
-	strtol(s.c_str(), &p, 10);
+bool isInteger(std::string s) {
```
nickjong: Nit: our C++ function names typically use underscores, not camel case

test/annotation/utils_tests.cxx
```diff
@@ -0,0 +1,82 @@
+#define BOOST_TEST_MODULE
```
nickjong: Should define a test suite name? See https://www.boost.org/doc/libs/1_68_0/libs/test/doc/html/boost_test/tests_organization/test_suite/master_test_suite.html#ref_BOOST_TEST_MODULE

test/annotation/utils_tests.cxx
```diff
@@ -0,0 +1,82 @@
...
+    std::vector<turi::flexible_type> feature_vector =
+        turi::annotate::similar_items(feature_sarray, 1, top_k);
+
+    TS_ASSERT(feature_vector.size() == top_k);
+  }
+};
+
+BOOST_FIXTURE_TEST_SUITE(_utils_test, utils_test)
+BOOST_AUTO_TEST_CASE(test_featurize_images) {
+  utils_test::test_featurize_images();
```
nickjong: There is no need to define the test code in a separate `utils_test` type. You can just move the test code into here.

src/toolkits/image_deep_feature_extractor/image_deep_feature_extractor_toolkit.hpp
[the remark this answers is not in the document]
hoytak: I think an SArray version makes sense.  I think the argument against it is to be consistent with other feature extractors (only boosted trees ATM) that take multiple input columns.   However, this is pretty weak as that doesn't take the column name as part of it.

src/unity/lib/annotation/image_classification.cpp
```diff
@@ -238,7 +238,7 @@ void ImageClassification::_addAnnotationToSFrame(size_t index, int label) {
 
 void ImageClassification::cast_annotations() {
   size_t annotation_column_index = m_data->column_index(m_annotation_column);
-  if(m_data->dtype().at(annotation_column_index) == flex_type_enum::INTEGER) {
+  if (m_data->dtype().at(annotation_column_index) == flex_type_enum::INTEGER) {
```
hoytak: I know this is reformatted old code, but this should have been m_data->column_type(annotation_column_index) in the first place.  Or just m_data->column_type(m_annotation_column).

src/unity/lib/annotation/utils.cpp
```diff
@@ -1,10 +1,91 @@
...
+  if (s.empty() || ((!isdigit(s[0])) && (s[0] != '-') && (s[0] != '+')))
+    return false;
 
-	return (*p == 0);
-}
\ No newline at end of file
+  char *p;
+  strtol(s.c_str(), &p, 10);
+
+  return (*p == 0);
```
hoytak: This code right here is wrong (see the definition of strtol).  It should be `return (*p == s.data() + s.size())`.

src/unity/lib/annotation/utils.cpp
```diff
@@ -1,10 +1,91 @@
...
+        return flexible_type(
+            vectors_distance(target_vector, a.get<flex_vec>()));
+      },
+      flex_type_enum::FLOAT);
+
+  std::vector<flexible_type> indicies;
+
+  for (size_t x = 0; x < distances->size(); x++) {
+    indicies.push_back(x);
+  }
```
hoytak: Avoid using repeated push_back when possible.  It wastes a lot of memory allocations:

```
std::vector<flexible_type> indicies(distances->size());

  for (size_t x = 0; x < distances->size(); x++) {
    indicies[x]=x;
  }
```
Or better yet:
```
std::vector<flexible_type> indicies(distances->size());
std::iota(indices.begin(), indices.end(), 0);

File: src/unity/lib/annotation/image_classification.cpp (hunk 5 above)
Comment (abhishekpratapa) [reply src/unity/lib/annotation/image_classification.cpp]: I don't think `unity_sframe` has this method on it. Unless you're suggesting I use a different data representation.

File: src/unity/lib/annotation/utils.cpp [hunk 8]
```diff
@@ -1,10 +1,91 @@
...
+}
+
+/**
+ * Note: very inefficient way of calculating the distances.
+ */
+std::vector<flexible_type>
+similar_items(std::shared_ptr<unity_sarray> distances, size_t index, size_t k) {
+
+  DASSERT_EQ(distances->dtype(), flex_type_enum::VECTOR);
+  flex_vec target_vector = distances->to_vector().at(index).get<flex_vec>();
```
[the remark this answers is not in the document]
hoytak: Still not fixed.

src/unity/lib/annotation/utils.cpp
```diff
@@ -1,10 +1,91 @@
...
+
+  gl_sarray gl_features =
+      feature_extractor.sarray_extract_features(gl_images, false, 6);
+
+  return std::shared_ptr<unity_sarray>(gl_features);
+}
+
+double vectors_distance(const std::vector<double> &a,
+                        const std::vector<double> &b) {
+  std::vector<double> auxiliary;
```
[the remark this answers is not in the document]
hoytak: We should do this: 

```
DASSERT_EQ(a.size(), b.size()); 
double acc = 0; 
for(size_t i = 0; i < a.size(); ++i) { acc += std::pow(a[i] - b[I], 2); } 
return std::sqrt(acc); 
```
No memory allocations, simpler code, much faster.

src/unity/lib/annotation/utils.cpp
hoytak: Okay, to file an issue on this and ignore it here, but it should get fixed.

src/unity/lib/annotation/utils.cpp
```diff
@@ -1,10 +1,79 @@
...
-}
\ No newline at end of file
+  char *p;
+  strtol(s.c_str(), &p, 10);
+
+  return (*p == 0);
+}
+
+gl_sarray
+featurize_images(const std::shared_ptr<turi::gl_sarray> &images) {
```
hoytak: Just use gl_sarray directly -- it can be passed by value or const gl_sarray&

test/annotation/utils_tests.cxx
```diff
@@ -0,0 +1,93 @@
...
+    std::shared_ptr<turi::gl_sarray> image_gl_sarray =
+        std::make_shared<turi::gl_sarray>(image_sarray);
+
+    turi::gl_sarray feature_sarray =
+        turi::annotate::featurize_images(image_gl_sarray);
+
+    TS_ASSERT(image_gl_sarray->dtype() == turi::flex_type_enum::IMAGE);
+    TS_ASSERT(feature_sarray.dtype() == turi::flex_type_enum::VECTOR);
+
+    std::vector<turi::flexible_type> feature_vector =
```
hoytak: Use the gl_sarray range iterator here to loop through it.  Tests are a little looser, but this should still be good practice.

berrylaw:
anyone can help?

hoytak:
@berrylaw -- feel free to help out, review, etc!  We're working on getting this one in, so please add any comments you'd like.  For new code, you need to open your own PR :-).

syoutsey:
Removing milestone as these changes currently have no client, will go into a future release.

src/unity/lib/annotation/utils.cpp
abhishekpratapa: I’m having a difficult time understanding the reasoning, here’s my understanding:

As it parses the string `strtol` will stop at the first character that’s not considered an integer. The  `p` provided then is populated with the pointer to the first non-integer character, hence in the case if a `NULL` terminated string is 0. if that character isn't 0 or `NULL` than we can safely assume that the string isn't a parsable integer.

-- Likewise--

`(s.data() + s.size())` - is the pointer to the last character of the string. so rather than comparing the dereferenced pointer `*p` we can compare it to the pointer itself `p == (s.data() + s.size())` and we get the same result.

One is a pointer level comparison, the other is a content level comparison.

What am I missing here?

src/unity/lib/annotation/utils.cpp
```diff
@@ -1,10 +1,91 @@
...
-	return (*p == 0);
-}
\ No newline at end of file
+  char *p;
+  strtol(s.c_str(), &p, 10);
+
+  return (*p == 0);
+}
+
+std::shared_ptr<unity_sarray>
```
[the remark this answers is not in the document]
abhishekpratapa: changed to gl_sarray!

src/unity/lib/annotation/utils.cpp
abhishekpratapa: fixed!

test/annotation/utils_tests.cxx
abhishekpratapa: Done!

src/unity/lib/annotation/utils.cpp
abhishekpratapa: done!

src/unity/lib/annotation/utils.cpp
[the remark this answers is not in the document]
abhishekpratapa: Thank you! Works perfectly!

berrylaw:
Your plan is really pretty prefect
You win chees

「Abhishek <notifications@github.com>」在 2019年4月27日 週六，上午1:07 寫道：

berrylaw:
My EQ maximums around 6 times.
But your testing had 10 lesson, so high level for me,  my EQ failed

「berrylaw <notifications@github.com>」在 2019年4月27日 週六，上午1:08 寫道：

berrylaw:
I think other one will be smart and better than me

「Abhishek <notifications@github.com>」在 2019年3月25日 週一，下午6:57 寫道：

src/unity/lib/annotation/utils.cpp
hoytak: Ah, you're right.  Yes.

abhishekpratapa:
Why do the test keep failing on this one (seems like they aren't loading). Maybe some changes to some CMake lists I did? Just interesting that's all.

znation:
@abhishekpratapa Looks like an undefined symbol in unity_shared. From the log:
```
_________ ERROR collecting turicreate/test/test_activity_classifier.py _________
ImportError while importing test module '/home/travis/build/apple/turicreate/debug/src/unity/python/turicreate/test/test_activity_classifier.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
turicreate/__init__.py:19: in <module>
    from turicreate.data_structures.sgraph import Vertex, Edge
turicreate/data_structures/__init__.py:18: in <module>
    from . import sframe
turicreate/data_structures/sframe.py:16: in <module>
    from .._connect import main as glconnect
turicreate/_connect/main.py:13: in <module>
    from .._cython.cy_unity import UnityGlobalProxy
E   ImportError: /home/travis/build/apple/turicreate/debug/src/unity/python/turicreate/_cython/../libunity_shared.so: undefined symbol: _ZTVN4turi28image_deep_feature_extractor36image_deep_feature_extractor_toolkitE
_________ ERROR collecting turicreate/test/test_audio_functionality.py _________
```
Did you test these changes locally? May be a platform-specific linker error?

PR #1654, opened 2019-03-22
Changing deployment target to iOS 10.0

### LINKED ISSUE #1347: Building the app with Deployment Target iOS 11 but Recommendation Feature working for iOS 12
We are working on an app that uses Recommender Systems model. I understand that the recommendation feature of the app won't work in iOS 11 and on the Simulator (because of bug https://github.com/apple/turicreate/issues/822). But we need to keep the deployment target of our app at iOS 11 but still would like our iOS 12 users to be able to take advantage of this feature. Therefore, I have added the following checks in code, before initializing the Recommender model:
```
#if !targetEnvironment(simulator)
        if #available(iOS 12.0, *) {
            let model = MyRecommender()
            // Use the model
           // ...
        }
#endif
```
So everything works fine in Simulator, both iOS 11 and 12 (i.e. it ignores the model). And on iOS 12 device which uses the model and makes recommendations. But the problem is that it crashes on iOS 11 devices even though, I have made `libRecommender.dylib` as Optional instead of Required under "Link Binary With Libraries". The error that I get is:

```
dyld: Symbol not found: _OBJC_CLASS_$_MLArrayBatchProvider
  Referenced from: /private/var/containers/Bundle/Application/.../Frameworks/libRecommender.dylib (which was built for iOS 12.0)
  Expected in: /System/Library/Frameworks/CoreML.framework/CoreML
 in /private/var/containers/Bundle/Application/.../Frameworks/libRecommender.dylib
Message from debugger: Terminated due to signal 6
```

Can you please let me know how I can use this model in iOS 12 devices, yet keep the deployment target as iOS 11 (knowing that, the recommendation feature of the app won't work on iOS 11)?

shreyajain17:
fixes #1347

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,7 +152,7 @@ endif()
 if(APPLE)
   if(TC_BUILD_IOS)
     SET(TC_BASE_SDK iphoneos)
-    SET(COMPILER_FLAGS "${COMPILER_FLAGS} -miphoneos-version-min=12.0 -Wno-c++11-narrowing")
+    SET(COMPILER_FLAGS "${COMPILER_FLAGS} -miphoneos-version-min=11.0 -Wno-c++11-narrowing")
   else()
     SET(TC_BASE_SDK macosx)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,7 +152,7 @@ endif()
 if(APPLE)
   if(TC_BUILD_IOS)
     SET(TC_BASE_SDK iphoneos)
-    SET(COMPILER_FLAGS "${COMPILER_FLAGS} -miphoneos-version-min=11.0 -Wno-c++11-narrowing")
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -152,7 +152,7 @@ endif()
 if(APPLE)
   if(TC_BUILD_IOS)
     SET(TC_BASE_SDK iphoneos)
-    SET(COMPILER_FLAGS "${COMPILER_FLAGS} -miphoneos-version-min=12.0 -Wno-c++11-narrowing")
+    SET(COMPILER_FLAGS "${COMPILER_FLAGS} -miphoneos-version-min=11.0 -Wno-c++11-narrowing")
```
nickjong: Thinking about this some more, does everything still work if we set this to 10.0?

Then it better matches the macOS deployment target (will be the iOS that was released the same year as that macOS).

It may also help ensure that all CoreML symbols are weak-linked, making it more likely that our dylibs can be included in apps targeted older OS versions (even ones that don't include CoreML, in case all the CoreML features are optional).

PR #1657, opened 2019-03-22
Sound Classifier: Allow Reusing Deep Features

### LINKED ISSUE #1601: Sound Classifier needs way to save/use deep features
The Sound Classifier currently spends ~95% of its time doing feature preprocessing and extracting deep features. Once a couple of issues are fixed less time will be spent on both but these will still constitute the majority of total time.

Doing common machine tasks, such as k-fold validation or hyper-parameter search, will result in a large amount of duplicated work; the preprocessing and feature extraction for an example will be identical across folds or hyper-parameter runs.

We need a way to avoid this large amount of duplicated work.

Data augmentation (#1571) will complicate this issue because all data augmentation must be done prior to preprocessing and feature extraction.

TobyRoseman:
Fixes #1601

- Add tc.sound_classifier.get_deep_features.
- Change sound classifier (create & predict) to use get_deep_features.
- Allow sound classifier (create & predict) to use deep features directly.
-----------------------------------------
- Use SArrayBuilder to construct deep features
- Logging: print header info only if printing progress info.
- Any function which could result in deep feature extraction should have `verbose` flag.
- Allow predict to do parameter checking. Rather than doing it, not as well, in a bunch of other
      places that just call predict anyways.
- classify/evaluate/predict_topk should allow SFrame or dict input, in addition to SArray.
-----------------------------------------
- Unit tests: test deep feature reuse.
- Unit tests: copy data rather than regenerate each time.
- Unit tests: make problematic test less sensitive (six decimal places is fine).
- Unit tests: test validation set is actually getting used or not.
-----------------------------------------
- Took a first stab at the user guide section for this feature, more will be added later (#1649).

### DIFF
--- a/src/unity/python/turicreate/test/test_audio_functionality.py
+++ b/src/unity/python/turicreate/test/test_audio_functionality.py
@@ -11,6 +11,7 @@ from __future__ import absolute_import as _
 
 from .util import TempDirectory
 
+from copy import copy
 import math
 from os import mkdir
 import unittest
@@ -145,11 +146,14 @@ def _generate_binary_test_data():
     return data
 
 
+binary_test_data = _generate_binary_test_data()
+
+
 class ClassifierTestTwoClassesStringLabels(unittest.TestCase):
 
     @classmethod
     def setUpClass(self):
-        self.data = _generate_binary_test_data()
+        self.data = copy(binary_test_data)
         self.is_binary_classification = True
         self.model = tc.sound_classifier.create(self.data, 'labels', feature='audio')
 
@@ -213,7 +217,7 @@ class ClassifierTestTwoClassesStringLabels(unittest.TestCase):
         old_model_probs = self.model.predict(self.data['audio'], output_type='probability_vector')
         new_model_probs = new_model.predict(self.data['audio'], output_type='probability_vector')
         for a, b in zip(old_model_probs, new_model_probs):
-            self.assertItemsEqual(a, b)
+            np.testing.assert_array_almost_equal(a, b, decimal=6)
 
     @unittest.skipIf(_mac_ver() < (10,14), 'Custom models only supported on macOS 10.14+')
     def test_export_coreml_with_prediction(self):
@@ -281,14 +285,16 @@ class ClassifierTestTwoClassesStringLabels(unittest.TestCase):
         self.assertTrue(len(unique_ranks) == 1)
         self.assertTrue(unique_ranks[0] == 0)
 
+    def test_validation_set(self):
+        self.assertTrue(self.model.validation_accuracy is None)
+
 
 class ClassifierTestTwoClassesIntLabels(ClassifierTestTwoClassesStringLabels):
     @classmethod
     def setUpClass(self):
-        self.data = _generate_binary_test_data()
+        self.data = copy(binary_test_data)
         self.data['labels'] = self.data['labels'].apply(lambda x: 0 if x == 'white noise' else 1)
         self.is_binary_classification = True
-
         self.model = tc.sound_classifier.create(self.data, 'labels', feature='audio', validation_set=None)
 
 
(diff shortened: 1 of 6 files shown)


userguide/sound_classifier/advanced-usage.md
```diff
@@ -0,0 +1,29 @@
+# Advanced Sound Classifier Usage
+
+
+#### Reusing Deep Features
+Often it is necessary to train more than one instance of a model on the same data. For example this is required by both [K-Folds Cross Validataion](https://en.wikipedia.org/wiki/Cross-validation_(statistics)) and [Hyperparameter Tuning](https://en.wikipedia.org/wiki/Hyperparameter_optimization).
+
+Much of the work accross differnt training run can be reused. Sound Classification is done using a [three stage process](how-it-works.md). The first two of these stages (signal preprocessing and VGGish feature extraction) will be identical and can be reused.
```
shantanuchhabra: differnt training run -> different training runs

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -38,8 +92,11 @@ def create(dataset, target, feature, max_iterations=10, verbose=True,
 
     feature : string, optional
         Name of the column containing the feature column. This column must
-        contain audio data (represented as dicts with key 'data' and
-        'sample_rate').
+        contain audio data or deep audio features.
+        Audio data is represented as dicts with key 'data' and 'sample_rate',
+        see `turicreate.load_audio(...)`.
+        Deep audio features are represent as a list of numpy arrays, each of
```
shantanuchhabra: represent -> represented

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -61,18 +118,17 @@ def create(dataset, target, feature, max_iterations=10, verbose=True,
         If you are getting memory errors, try decreasing this value. If you
         have a powerful computer, increasing this value may improve performance.
     '''
-    from ._audio_feature_extractor import _get_feature_extractor
-    import mxnet as _mx
-    import time as _time
+    import time as time
```
shantanuchhabra: Just a general question: Is it better practice to use the `as` keyword even if we don't want to use a shorthand for the package?

src/unity/python/turicreate/test/test_audio_functionality.py
```diff
@@ -11,6 +11,7 @@
 
 from .util import TempDirectory
 
+from copy import copy
```
shantanuchhabra: I've wondered this earlier looking at our test code as well -- why don't we hide imports in the test code?

TobyRoseman: We hide imports in non test code so users know which methods/classes we provide. Users aren't typically going to import unit test files. So we don't worry about it there.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: Done.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: This was a search and replace error. I've fixed it.

userguide/sound_classifier/advanced-usage.md
TobyRoseman: Thanks. Done.

TobyRoseman:
internal tests passed

nickjong:
Should we be worried about any kind of versioning here? If we change the feature extraction at some point, will our training methods necessarily have to generalize to any version of extracted feature vector?

TobyRoseman:
> Should we be worried about any kind of versioning here? If we change the feature extraction at some point, will our training methods necessarily have to generalize to any version of extracted feature vector?

I've tried to keep the toolkit as separate as possible from the feature extractor. I think things are already well generalized. The name of the feature extractor is saved as a part of the model. This name is used to get a reference to the feature extractor. The output dimensions as well as the feature preprocessing is a part of the feature extractor, not a part of the toolkit. I don't think there is anything stopping us from adding another feature extractor tomorrow. 

I hopefully that answered your questions, if not let me know.

Issue #1639, opened 2019-03-20
Sound Classifier need "Advanced Usage" user guide page


TobyRoseman:
With the 5.5 release the sounds classifier will have:
* A way of reusing deep feature across training (#1601)
* A way to control the custom neural network (#1477). Two layers of 100 units each will continue to be the default.

We need an advanced section of the user guide which explains each of these features in more detail and gives example code.

coolioxlr:
+1

TobyRoseman:
Done by #1722

PR #1662, opened 2019-03-25
Sound Classifier: allow user to control custom neural network structure

### LINKED ISSUE #1477: Sound Classifier Should Allow Tuning the Structure of the Custom Neural Network
Currently the custom neural network created by the sound classifier is always [two dense layer with 100 units each](https://github.com/apple/turicreate/blob/3236aa1e64ec76a4f8b775432ff46ba22b983115/src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py#L174). This is a sensible default but it should be configurable. The user should be able to specify how many dense layers they want and how many units should be in each layer.

TobyRoseman:
Fixes #1477

### DIFF
--- a/src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
+++ b/src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
@@ -76,7 +76,8 @@ def get_deep_features(audio_data, verbose=True):
     return feature_extractor.get_deep_features(audio_data, verbose=verbose)
 
 
-def create(dataset, target, feature, max_iterations=10, verbose=True,
+def create(dataset, target, feature, max_iterations=10,
+           custom_layer_sizes=[100, 100], verbose=True,
            validation_set='auto', batch_size=64):
     '''
     Creates a :class:`SoundClassifier` model.
@@ -104,6 +105,14 @@ def create(dataset, target, feature, max_iterations=10, verbose=True,
         the data can result in a more accurately trained model. Consider
         increasing this (the default value is 10) if the training accuracy is
         low.
+
+    custom_layer_sizes : list of ints
+        Specifies the architecture of the custom neural network. This neural
+        network is made up of a series of dense layers. This parameter allows
+        you to specify how many layers and the number of units in each layer.
+        The custom neural network will always have one more layer than the
+        length of this list. The last layer is always a soft max with units
+        equal to the number of classes.
 
     verbose : bool, optional
         If True, prints progress updates and model details.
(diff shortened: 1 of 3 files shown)


nickjong:
For this kind of thing, I'm most worried about how we communicate about this API in the user guide. To what degree are we encouraging developers to spend a lot of time grid searching parameter values here? What benefit are they likely to see for their investment in time and energy? How should they think about these parameters?

TobyRoseman:
> For this kind of thing, I'm most worried about how we communicate about this API in the user guide. To what degree are we encouraging developers to spend a lot of time grid searching parameter values here? What benefit are they likely to see for their investment in time and energy? How should they think about these parameters?

I agree this is something which will need to be carefully explained in our user guide. I created #1639 to make sure we explain this feature (as well as reusing deep feature and manually tweaking data augmentation) in our user guide. All three of theses features will be explained in an "Advanced Usage" section.

Here are my main thoughts about this parameter:
* I think for most users the default should work quite well. I see tuning this as more of an advanced features.
* If the user is training on a GPU and they are reusing deep features, then it should be very quick to try out a large number of different structures.
* If the number of classes is small, then they could use this feature to get a much smaller exported model with minimal affect on accuracy.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -135,6 +144,8 @@ def create(dataset, target, feature, max_iterations=10, verbose=True,
         raise _ToolkitError("'%s' column is not audio data." % feature)
     if target not in dataset.column_names():
         raise _ToolkitError("Target column '%s' does not exist" % target)
+    if not isinstance(custom_layer_sizes, list) or len(custom_layer_sizes) == 0 or not isinstance(custom_layer_sizes[0], int):
```
shantanuchhabra: Nit: What if someone passes in an array of the form [75, 200, None]? We probably want a check for all elements in the array?

We may want to add unit tests for corner cases (but maybe rarely encountered cases) like these?

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -135,6 +144,8 @@ def create(dataset, target, feature, max_iterations=10, verbose=True,
...
+        raise _ToolkitError("'custom_layer_sizes' must be a non-empty list of integers." % target)
```
shantanuchhabra: I think "% target" is a typo?

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -342,7 +362,14 @@ def _load_version(cls, state, version):
         # Load the custom nerual network
         num_classes = state['num_classes']
         num_inputs = state['_feature_extractor'].output_length
-        net = SoundClassifier._build_custom_neural_network(num_inputs, num_classes)
+        if 'custom_layer_sizes' in state:
+            # These are deserialized as floats
+            custom_layer_sizes = map(int, state['custom_layer_sizes'])
```
shantanuchhabra: Do we want to make it `list(map(...))` to not break on Python 3?

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: Good points. I added more checks here and manually tested the cases you mentioned.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: Yeah, typo. Fixed.

src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: Thanks. Done.

PR #1670, opened 2019-03-26
Created a Documented Section for Annotation

abhishekpratapa:


userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
 data.explore()
 ```
 
+#### Annotating Data
+
+If you only have images without corresponding annotations, you can 
+use the annoation utility built into the image_classifier. An example
+of it's usage is shown below with the **PetImages** subfolder above.
```
znation: `s/it's/its` -- "it's" is always a contraction for "it is" whereas "its" is the possessive.

userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
...
+
+```python
+import turicreate as tc
+
+# Load images (Note: you can ignore 'Not a JPEG file' errors)
```
znation: But should you? Why are we instructing users to ignore these errors?

userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
...
+data = tc.image_analysis.load_images('PetImages', with_path=True)
+
+# If you forget to assign the output to a variable
+tc.image_classifier.annotate(data)
+
+# recover your annotation with this method
+annotated_data = tc.image_classifier.recover_annotation()
+```
+
+The annotation utility supports label types of `string` and `integer`.
```
znation: I would either use the Python names (`str` and `int`), or make these regular text instead of code-quoted text.

userguide/image_classifier/README.md
nickjong: At the very least, you can describe them as warnings (since they're not fatal, right?) and comment that this will skip the files that cannot be loaded.

userguide/image_classifier/README.md
```diff
@@ -17,7 +17,7 @@ The [Kaggle Cats and Dogs Dataset](https://www.microsoft.com/en-us/download/deta
 ```python
 import turicreate as tc
 
-# Load images (Note: you can ignore 'Not a JPEG file' errors)
+# Load images (Note:'Not a JPEG file' errors are simply warnings, meaning those files will simply be skipped)
```
znation: The word `simply` appears twice in this sentence.

userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
 data.explore()
 ```
 
+#### Annotating Data
+
+If you only have images without corresponding annotations, you can 
+use the annoation utility built into the image_classifier. An example
```
znation: `s/annoation/annotation/g`

userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
 data.explore()
 ```
 
+#### Annotating Data
+
+If you only have images without corresponding annotations, you can
```
znation: I would suggest using the word `labels` rather than `annotations` in the first sentence, to help introduce the context.

userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
 data.explore()
 ```
 
+#### Annotating Data
```
srikris: Please don't add before the introductory example. Add this in the advance section or after the introductory example. Let's keep the first example really simple.

userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
...
+
+If you only have images without corresponding annotations, you can 
+use the annoation utility built into the image_classifier. An example
+of its usage is shown below with the **PetImages** subfolder above.
```
znation: Shown below with the subfolder above? Then the code starts with the code above, again? That's a bit confusing. Let's just say "shown below", then skip to the relevant parts of the code and assume the user has already typed in the above code.

userguide/image_classifier/README.md
```diff
@@ -30,6 +30,43 @@ data.save('cats-dogs.sframe')
...
+data = tc.image_analysis.load_images('PetImages', with_path=True)
+
+# Use the Annotation GUI to annotate your data.
+annotated_data = tc.image_classifier.annotate(data)
+
+```
+
+If you forget to assign the output of your annotation to a variable, 
+we've included a method to help you recover those annotations. The code
+for that is shown below:
```
znation: Again, let's skip the code that's duplicated from the above.

abhishekpratapa: Is it alright to duplicate if we move it to the advanced section, or do we assume they've already done it before?

userguide/image_classifier/README.md
abhishekpratapa: Updated the PR not to use a qualifier

PR #1672, opened 2019-03-26
Functionality for jump to image in Annotation 

### LINKED ISSUE #1521: Annotation UI should have functionality to jump to an image number.
At the moment, the arrows or infinite scroll are the only way to access images. Users cannot jump to an image number if they want.

shreyajain17:
fixes #1521


znation:
@shreyajain17 Is this still a WIP because of the layout issue (the bar appears below the fold)? If that's the case, @abhishekpratapa may be able to help with that. (I agree it should stay WIP until the bar appears above the fold).

shreyajain17:
Yeah that’s an issue and there is the state thing as well. I will just sit with him when he is free and settle this.

PR #1673, opened 2019-03-27
Sound classifier: bug fix and unit test fix

TobyRoseman:


TobyRoseman:
internal tests passed.

Issue #1567, opened 2019-03-10
Audit turicreate.utils


TobyRoseman:
There are several things under `turicreate.utils` that are longer used and should be deleted. Many of the things that remain should either be hidden or moved to `turicreate.test.util`.

TobyRoseman:
fixed by #1681

Issue #492, opened 2018-04-09
Top Level Python Name Space Clean Up


TobyRoseman:
The following module are exposed at the top level `turicreate` namespace and probably should be hidden:

- [x] `connect`
- [x] `cy_pylambda_workers`
- [x] `cython`
- [x] `dep`
- [x] `glconnect`
- [x] `Model`

hoytak:
util is used and documented; the rest should be moved to their _xxxxx equivalent.

TobyRoseman:
@hoytak thanks. I've removed that from the list. We're already tracking that issue in #1567.

TobyRoseman:
Fixed by #1681

PR #1681, opened 2019-03-30
Namespace cleanup

TobyRoseman:
Addresses #492 and #1567. Also removes some unused code.


TobyRoseman:
Internal tests passed.

PR #1689, opened 2019-04-01
Adding training accuracy to Activity Classifier

### LINKED ISSUE #1640: C-API Activity Classifier training should output accuracy
Example output for `tc.activity_classifier.create` for a small dataset:
```
+----------------+----------------+----------------+----------------+
| Iteration      | Train Accuracy | Train Loss     | Elapsed Time   |
+----------------+----------------+----------------+----------------+
| 1              | 0.687          | 0.913          | 1.4            |
| 2              | 0.853          | 0.447          | 2.9            |
| 3              | 0.880          | 0.342          | 4.3            |
| 4              | 0.893          | 0.292          | 5.6            |
| 5              | 0.907          | 0.249          | 7.0            |
| 6              | 0.917          | 0.229          | 8.4            |
| 7              | 0.925          | 0.207          | 9.7            |
| 8              | 0.933          | 0.186          | 10.9           |
| 9              | 0.937          | 0.168          | 12.2           |
| 10             | 0.934          | 0.171          | 13.5           |
+----------------+----------------+----------------+----------------+
```

Example output for `tc.extensions.activity_classifier().train`:
```
| Iteration    | Train Loss   | Elapsed Time |
| 1            | 0.864981     | 9.69s        |
| 2            | 0.448611     | 11.32s       |
| 3            | 0.346208     | 12.95s       |
| 4            | 0.300267     | 14.56s       |
| 5            | 0.261983     | 16.17s       |
| 6            | 0.237489     | 17.79s       |
| 10           | 0.176503     | 24.26s       |
```

We need to implement the accuracy computation for the C++ version

shreyajain17:
fixes #1640

### DIFF
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -65,6 +65,33 @@ std::vector<std::string> get_supported_metrics() {
           "confusion_matrix", "roc_curve"};
 }
 
+std::tuple<float,size_t> get_accuracy(size_t prediction_window, size_t num_classes, shared_float_array output_chunk, shared_float_array classes_chunk, data_iterator::batch::chunk_info info ){
+      
+
+    const float* output_ptr = output_chunk.data();
+    const float* truth_ptr = classes_chunk.data();
+    
+    size_t cumulative_samples = 0;
+    size_t seq_len = 0;
+    float cumulative_seq_accuracy = 0.f;
+
+
+      while (cumulative_samples < info.num_samples) {
+        cumulative_samples += prediction_window;
+       
+          if ((std::max_element(output_ptr, output_ptr + num_classes)- output_ptr == *truth_ptr)){
+            cumulative_seq_accuracy += 1.0;
+          };
+        seq_len++;
+        truth_ptr++;
+        output_ptr += num_classes;
+
+      }
+
+  return std::make_tuple(cumulative_seq_accuracy,seq_len);
+
+}
+
 }  // namespace
 
 void activity_classifier::init_options(
@@ -429,7 +456,7 @@ void activity_classifier::perform_training_iteration() {
   size_t num_batches = 0;
   size_t num_classes = read_state<size_t>("num_classes");
   size_t prediction_window = read_state<size_t>("prediction_window");
-  flex_vec preds(num_classes);
+  //flex_vec preds(num_classes);
   
   while (training_data_iterator_->has_next_batch()) {
     data_iterator::batch batch =
@@ -461,43 +488,24 @@ void activity_classifier::perform_training_iteration() {
       // Actual labels 
       shared_float_array classes_chunk = batch.labels[i];
       data_iterator::batch::chunk_info info = batch.batch_info[i];
-      
 
-      const float* output_ptr = output_chunk.data();
-      const float* truth_ptr = classes_chunk.data();
-      
-      size_t cumulative_samples = 0;
-      size_t seq_len =0;
+      size_t seq_len = 0;
       float cumulative_seq_accuracy = 0.f;
 
+      std::tie (cumulative_seq_accuracy, seq_len) = get_accuracy(prediction_window,num_classes, output_chunk, classes_chunk, info);
 
-      while (cumulative_samples < info.num_samples) {
 
-        // Copy the probability vector to get accuracy.
-        std::copy(output_ptr, output_ptr + num_classes, preds.begin());
-        output_ptr += num_classes;
-        
-        cumulative_samples += prediction_window;
-        seq_len++;
-        truth_ptr ++;
-        if ((std::max_element(preds.begin(), preds.end())-preds.begin() == *truth_ptr ) ){
-          cumulative_seq_accuracy+=1.0;
-
-        };
-
-      }
-
-      cumulative_per_batch_accuracy += cumulative_seq_accuracy/ (seq_len + 1e-5);
+      cumulative_per_batch_accuracy += cumulative_seq_accuracy / seq_len;
       
     };
 
-    cumulative_batch_accuracy += cumulative_per_batch_accuracy/ (batch.batch_info.size() + 1e-5);
+    cumulative_batch_accuracy += cumulative_per_batch_accuracy/ batch.batch_info.size();
     ++num_batches;
 
   }
 
   float average_batch_loss = cumulative_batch_loss / num_batches;
-  float average_batch_accuracy = cumulative_batch_accuracy / num_batches  ;
+  float average_batch_accuracy = cumulative_batch_accuracy / num_batches;
 
   // Report progress if we have an active table printer.
   // TODO: Report validation metrics.
@@ -509,7 +517,7 @@ void activity_classifier::perform_training_iteration() {
 
   add_or_update_state({
       { "training_iterations", iteration_idx + 1 },
-      {"training_accuracy" , average_batch_accuracy},
+      { "training_accuracy" , average_batch_accuracy },
       { "training_log_loss", average_batch_loss },
   });
 
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -499,9 +499,9 @@ void activity_classifier::perform_training_iteration() {
 
     
     const shared_float_array& output = results.at("output");
-    // std::cout << num_batches << '\t' << get_accuracy_per_batch(prediction_window, num_classes, output, batch) << '\t';
+    
     cumulative_batch_accuracy += get_accuracy_per_batch(prediction_window, num_classes, output, batch);
-    // std::cout << cumulative_batch_accuracy << '\n';
+    
     ++num_batches;
 
   }
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -65,34 +65,29 @@ std::vector<std::string> get_supported_metrics() {
           "confusion_matrix", "roc_curve"};
 }
 
-std::tuple<float,size_t> get_accuracy_per_sequence(size_t prediction_window, size_t num_classes, 
-  const shared_float_array& output_chunk, const shared_float_array& classes_chunk, data_iterator::batch::chunk_info info) {
+size_t count_correct_predictions(size_t prediction_window, size_t num_classes, const shared_float_array& output_chunk, 
+  const shared_float_array& label_chunk, data_iterator::batch::chunk_info info) {
       
-
   const float* output_ptr = output_chunk.data();
-  const float* truth_ptr = classes_chunk.data();
+  const float* truth_ptr = label_chunk.data();
     
   size_t cumulative_samples = 0;
-  size_t num_predictions = 0;
   size_t num_correct_predictions = 0.f;
-
-
+  
   while (cumulative_samples < info.num_samples) {
 
     size_t prediction = std::max_element(output_ptr, output_ptr + num_classes) - output_ptr;
-
     if (prediction == *truth_ptr) {
       num_correct_predictions += 1;
       };
 
     cumulative_samples += prediction_window;
-    num_predictions++;
     truth_ptr++;
     output_ptr += num_classes;
 
   }
 
-  return std::make_tuple(num_correct_predictions, num_predictions);
+  return num_correct_predictions;
 
 }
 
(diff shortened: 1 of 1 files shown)


src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -440,23 +446,70 @@ void activity_classifier::perform_training_iteration() {
...
+
+      while (cumulative_samples < info.num_samples) {
+
+        // Copy the probability vector to get accuracy.
+        std::copy(output_ptr, output_ptr + num_classes, preds.begin());
+        output_ptr += num_classes;
+        
+        cumulative_samples += prediction_window;
+        seq_len++;
+        truth_ptr ++;
```
nickjong: Hmm, I believe you need to increment `truth_ptr` after you use it in line 483. Note that as-is, you're never looking at the first ground-truth label.

In this case, it might help to follow the idiom of incrementing all the loop variables at the end of the loop body

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -440,23 +446,70 @@ void activity_classifier::perform_training_iteration() {
...
+        if ((std::max_element(preds.begin(), preds.end())-preds.begin() == *truth_ptr ) ){
+          cumulative_seq_accuracy+=1.0;
+
+        };
+
+      }
+
+      cumulative_per_batch_accuracy += cumulative_seq_accuracy/ (seq_len + 1e-5);
```
nickjong: Let's not add this `1e-5` bias that the Python implementation does. I believe we can be confident that `seq_len` is always non-zero. (If not, we should test for that explicitly and not touch `cumulative_per_batch_accuracy` at all.)

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -440,23 +446,70 @@ void activity_classifier::perform_training_iteration() {
...
+      
+    };
+
+    cumulative_batch_accuracy += cumulative_per_batch_accuracy/ (batch.batch_info.size() + 1e-5);
```
nickjong: Here too, let's not do the `+ 1e-5` bit. We already know that the batch should not be empty.

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -440,23 +446,70 @@ void activity_classifier::perform_training_iteration() {
...
     ++num_batches;
+
   }
 
   float average_batch_loss = cumulative_batch_loss / num_batches;
+  float average_batch_accuracy = cumulative_batch_accuracy / num_batches  ;
```
nickjong: Nit: let's not leave spaces before the trailing `;`

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -440,23 +446,70 @@ void activity_classifier::perform_training_iteration() {
...
+  
   if (training_table_printer_) {
     training_table_printer_->print_progress_row(
-        iteration_idx, iteration_idx + 1, average_batch_loss, progress_time());
+        iteration_idx, iteration_idx + 1, average_batch_accuracy, average_batch_loss, progress_time());
   }
 
   add_or_update_state({
       { "training_iterations", iteration_idx + 1 },
+      {"training_accuracy" , average_batch_accuracy},
```
nickjong: Nit: let's format this line like the one above and below it

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -440,23 +473,51 @@ void activity_classifier::perform_training_iteration() {
...
                                        0.f, std::plus<float>());
+
+    
     cumulative_batch_loss += batch_loss / batch.batch_info.size();
 
+    
+    const shared_float_array& output = results.at("output");
+    float cumulative_per_batch_accuracy = 0.f;
+    
+    for (size_t i = 0; i < batch.batch_info.size(); ++i) {
```
nickjong: Could you create a second helper function that computes the accuracy for a given training batch? This second helper function can call the first one, and help modularize the code further

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -65,33 +65,58 @@ std::vector<std::string> get_supported_metrics() {
           "confusion_matrix", "roc_curve"};
 }
 
-std::tuple<float,size_t> get_accuracy(size_t prediction_window, size_t num_classes, shared_float_array output_chunk, shared_float_array classes_chunk, data_iterator::batch::chunk_info info ){
+std::tuple<float,size_t> get_accuracy_per_sequence(size_t prediction_window, size_t num_classes,
```
nickjong: Sorry, as we iterate on this, I keep seeing new ways to simplify things. And I think we should err on the side of simplification!

Thinking more about this, I don't think we need to return a tuple. The `num_predictions` value we're returning can actually be computer (by the caller) relatively simply. (I think what I was doing in the prediction code was probably overkill.) Which means that this function can boil down just to counting the number of correct predictions in one sequence ("batch row").

I suspect you'll find that we can reduce this to e.g.
```
size_t count_correct_predictions(size_t num_classes, size_t num_predictions, const shared_float_array& output_chunk, const shared_float_array& label_chunk) {
```

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -65,33 +65,58 @@ std::vector<std::string> get_supported_metrics() {
...
-    const float* output_ptr = output_chunk.data();
-    const float* truth_ptr = classes_chunk.data();
+  const float* output_ptr = output_chunk.data();
+  const float* truth_ptr = classes_chunk.data();
     
-    size_t cumulative_samples = 0;
-    size_t seq_len = 0;
-    float cumulative_seq_accuracy = 0.f;
+  size_t cumulative_samples = 0;
+  size_t num_predictions = 0;
```
nickjong: Honestly, we can compute `num_predictions` ahead of time as e.g.
```
size_t num_predictions = (info.num_samples + prediction_window - 1) / prediction_window;
```
Note that the `+ prediction_window - 1` part is just to round up instead of round down.

But then this can just be computed by the caller of this function and passed in as a parameter

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -65,33 +65,58 @@ std::vector<std::string> get_supported_metrics() {
...
+  float cumulative_per_batch_accuracy = 0.f;
 
-  return std::make_tuple(cumulative_seq_accuracy,seq_len);
+  for (size_t i = 0; i < batch.batch_info.size(); ++i){
 
+    const shared_float_array& output_chunk = output[i];
+    const shared_float_array& classes_chunk = batch.labels[i];
+    data_iterator::batch::chunk_info info = batch.batch_info[i];
+    size_t num_correct_predictions;
+    size_t num_predictions;
```
nickjong: As commented previously, we can just compute `num_predictions = (info.num_samples + prediction_window - 1) / prediction_window` here

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -65,33 +65,58 @@ std::vector<std::string> get_supported_metrics() {
...
+
+    std::tie (num_correct_predictions, num_predictions) = get_accuracy_per_sequence(prediction_window, num_classes, output_chunk, classes_chunk, info);
```
nickjong: Nit: let's not leave a space between a function name `std::tie` and the opening `(` of its argument list

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -440,23 +497,32 @@ void activity_classifier::perform_training_iteration() {
...
                                        loss_batch.data() + loss_batch.size(),
                                        0.f, std::plus<float>());
+
+    
     cumulative_batch_loss += batch_loss / batch.batch_info.size();
 
+    
+    const shared_float_array& output = results.at("output");
+    float cumulative_per_batch_accuracy = 0.f;
+    cumulative_per_batch_accuracy = get_accuracy_per_batch(prediction_window, num_classes, output, batch);
```
nickjong: Is this right? I don't see `cumulative_per_batch_accuracy` being used anywhere. Do we still need this variable?

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -64,6 +65,54 @@ std::vector<std::string> get_supported_metrics() {
           "confusion_matrix", "roc_curve"};
 }
 
+size_t count_correct_predictions(size_t prediction_window, size_t num_classes, const shared_float_array& output_chunk, 
+  const shared_float_array& label_chunk, data_iterator::batch::chunk_info info) {
```
nickjong: We're only using `info` to get `info.num_samples`, so you just pass in `num_samples` instead. Furthermore, we're only using `num_samples` to count how many times to go through the inner `while` loop. That number must be equal to the `num_predictions` value you're computing anyway in `get_accuracy_per_batch`. So could we just pass in `num_predictions` instead?

Nit: let's indent this line with four spaces instead of two, to make it more visually obvious that this is a line continuation, and not e.g. the first line of the function body

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -64,6 +65,54 @@ std::vector<std::string> get_supported_metrics() {
...
+      
+  const float* output_ptr = output_chunk.data();
+  const float* truth_ptr = label_chunk.data();
+    
+  size_t cumulative_samples = 0;
+  size_t num_correct_predictions = 0.f;
+  
+  while (cumulative_samples < info.num_samples) {
```
nickjong: This could just be e.g.
```
for (size_t i = 0; i < num_predictions; ++i) {
```

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -64,6 +65,53 @@ std::vector<std::string> get_supported_metrics() {
...
 }
 
+size_t count_correct_predictions(size_t num_classes, const shared_float_array& output_chunk, 
+    const shared_float_array& label_chunk, size_t num_predictions) {
+      
+  const float* output_ptr = output_chunk.data();
+  const float* truth_ptr = label_chunk.data();
+    
+
+  size_t num_correct_predictions = 0.f;
```
nickjong: Should initialize a `size_t` variable with an integer literal `0`, not a floating-point literal `0.f`

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -64,6 +65,53 @@ std::vector<std::string> get_supported_metrics() {
...
+  
+  for (size_t i = 0; i < num_predictions; i++) {
+
+    size_t prediction = std::max_element(output_ptr, output_ptr + num_classes) - output_ptr;
+    if (prediction == *truth_ptr) {
+      num_correct_predictions += 1;
+      };
```
nickjong: No semicolon is needed after an if statement body. We usually put the closing `}` at the same indent level as the `if`

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -64,6 +65,53 @@ std::vector<std::string> get_supported_metrics() {
...
+
+  }
+
+  return num_correct_predictions;
+
+}
+
+
+float get_accuracy_per_batch(size_t prediction_window, size_t num_classes, 
+  const shared_float_array& output, const data_iterator::batch& batch) {
```
nickjong: Here too we should indent two additional spaces to visually set this apart from the body's indent level

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -64,6 +65,53 @@ std::vector<std::string> get_supported_metrics() {
...
+
+  for (size_t i = 0; i < batch.batch_info.size(); ++i){
+
+    const shared_float_array& output_chunk = output[i];
+    const shared_float_array& label_chunk = batch.labels[i];
+    data_iterator::batch::chunk_info info = batch.batch_info[i];
+    size_t num_predictions = (info.num_samples + prediction_window - 1) / prediction_window;
+    size_t num_correct_predictions;
+    
+    num_correct_predictions = count_correct_predictions(num_classes, output_chunk, label_chunk, num_predictions);
```
nickjong: We can combine the declaration of `num_correct_predictions` and its initialization into a single line:
```
size_t num_correct_predictions = ...;
```

PR #1700, opened 2019-04-04
Adding Validation Accuracy and Loss to C-API of Activity Classifier

### LINKED ISSUE #1641: C-API activity classifier training should support validation-set metrics
`activity_classifier::train` currently accepts an argument `validation_data`, but it's ignored. For example, in the Python version we can do:
```
>>> model = tc.activity_classifier.create(data1, validation_set=data2, session_id='session_id', target='target', prediction_window=50, max_iterations=10)
Pre-processing 500000 samples...
Using sequences of size 1000 for model creation.
Processed a total of 43 sessions.
Using GPU to create model (AMD Radeon Pro Vega 56)
+---------------------+---------------------+---------------------+---------------------+---------------------+---------------------+
| Iteration           | Train Accuracy      | Train Loss          | Validation Accuracy | Validation Loss     | Elapsed Time        |
+---------------------+---------------------+---------------------+---------------------+---------------------+---------------------+
| 1                   | 0.604               | 1.069               | 0.693               | 1.274               | 1.4                 | 
| 2                   | 0.714               | 0.730               | 0.801               | 0.780               | 2.9                 | 
| 3                   | 0.823               | 0.543               | 0.788               | 0.538               | 4.4                 | 
| 4                   | 0.845               | 0.438               | 0.871               | 0.366               | 5.9                 | 
| 5                   | 0.866               | 0.376               | 0.886               | 0.318               | 7.4                 | 
| 6                   | 0.888               | 0.315               | 0.893               | 0.294               | 8.8                 | 
| 7                   | 0.897               | 0.281               | 0.899               | 0.271               | 10.3                | 
| 8                   | 0.902               | 0.254               | 0.906               | 0.264               | 11.8                | 
| 9                   | 0.911               | 0.237               | 0.906               | 0.252               | 13.3                | 
| 10                  | 0.915               | 0.220               | 0.888               | 0.289               | 14.8                | 
+---------------------+---------------------+---------------------+---------------------+---------------------+---------------------+
Training complete
Total Time Spent: 14.7509s
```

shreyajain17:
fixes #1641

### DIFF
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -161,11 +161,6 @@ void activity_classifier::train(gl_sframe data, std::string target_column_name,
     { {"Iteration", 12}, {"Train Accuracy", 12}, {"Train Loss", 12}, {"Validation Accuracy", 12}, {"Validation Loss", 12}, {"Elapsed Time", 12} }));
 
   }
-  // else if (variant_is<flex_string>(validation_data) ){
-  //   if (variant_get_value<flex_string>(validation_data) == "auto"){
-  //     std::cout << "WIP";
-  //   }
-  // }
   else {
 
     training_table_printer_.reset(new table_printer(
@@ -576,7 +571,7 @@ void activity_classifier::perform_training_iteration() {
       }
       cumulative_val_accuracy += cumulative_per_seq_accuracy;
 
-      val_size = +val.batch_info.size();
+      val_size += val.batch_info.size();
     }
 
     float average_val_accuracy = cumulative_val_accuracy / val_size;
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -65,11 +65,9 @@ std::vector<std::string> get_supported_metrics() {
           "confusion_matrix", "roc_curve"};
 }
 
-
 size_t count_correct_predictions(size_t num_classes, const shared_float_array& output_chunk, 
     const shared_float_array& label_chunk, size_t num_predictions) {
 
-      
   const float* output_ptr = output_chunk.data();
   const float* truth_ptr = label_chunk.data();
     
@@ -470,6 +468,8 @@ void activity_classifier::init_train(
         read_state<flex_int>("prediction_window");
     validation_data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
     validation_data_iterator_ = create_iterator(validation_data_params);
+  } else {
+    validation_data_iterator_ = nullptr;
   }
   // Instantiate the compute context.
   training_compute_context_ = create_compute_context();
@@ -597,8 +597,6 @@ void activity_classifier::perform_training_iteration() {
                                    prediction_window, num_classes, batch_size);
 
     if (training_table_printer_) {
-
-      // training_table_printer_->reset();
 
       training_table_printer_->print_progress_row(
           iteration_idx, iteration_idx + 1, average_batch_accuracy,
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -65,8 +65,10 @@ std::vector<std::string> get_supported_metrics() {
           "confusion_matrix", "roc_curve"};
 }
 
+
 size_t count_correct_predictions(size_t num_classes, const shared_float_array& output_chunk, 
     const shared_float_array& label_chunk, size_t num_predictions) {
+
       
   const float* output_ptr = output_chunk.data();
   const float* truth_ptr = label_chunk.data();
@@ -152,18 +154,30 @@ void activity_classifier::train(gl_sframe data, std::string target_column_name,
 {
   // Begin printing progress.
   // TODO: Make progress printing optional.
-  
-  // TODO: Support validation set.
-  training_table_printer_.reset(new table_printer(
+
+  if (variant_is<gl_sframe>(validation_data) ){
+
+    training_table_printer_.reset(new table_printer(
+    { {"Iteration", 12}, {"Train Accuracy", 12}, {"Train Loss", 12}, {"Validation Accuracy", 12}, {"Validation Loss", 12}, {"Elapsed Time", 12} }));
+
+  }
+  // else if (variant_is<flex_string>(validation_data) ){
+  //   if (variant_get_value<flex_string>(validation_data) == "auto"){
+  //     std::cout << "WIP";
+  //   }
+  // }
+  else {
+
+    training_table_printer_.reset(new table_printer(
     { {"Iteration", 12}, {"Train Accuracy", 12}, {"Train Loss", 12}, {"Elapsed Time", 12} }));
-  
-  
-  // Instantiate the training dependencies: data iterator, compute context,
-  // backend NN model.
+  }
+
   init_train(std::move(data), std::move(target_column_name),
              std::move(session_id_column_name), std::move(validation_data),
              std::move(opts));
 
+  // Instantiate the training dependencies: data iterator, compute context,
+  // backend NN model.
   // Perform all the iterations at once.
   flex_int max_iterations = read_state<flex_int>("max_iterations");
   while (read_state<flex_int>("training_iterations") < max_iterations) {
@@ -404,6 +418,19 @@ void activity_classifier::init_train(
   data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
   training_data_iterator_ = create_iterator(data_params);
 
+  // Bind the validation data to a data iterator.
+  if (variant_is<gl_sframe>(validation_data)) {
+    gl_sframe validation_sf = variant_get_value<gl_sframe>(validation_data);
+    data_iterator::parameters validation_data_params;
+    validation_data_params.data = validation_sf;
+    validation_data_params.target_column_name = target_column_name;
+    validation_data_params.session_id_column_name = session_id_column_name;
+    validation_data_params.feature_column_names = feature_column_names;
+    validation_data_params.prediction_window =
+        read_state<flex_int>("prediction_window");
+    validation_data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
+    validation_data_iterator_ = create_iterator(validation_data_params);
+  }
   // Instantiate the compute context.
   training_compute_context_ = create_compute_context();
   if (training_compute_context_ == nullptr) {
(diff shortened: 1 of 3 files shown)


src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -106,17 +154,30 @@ void activity_classifier::train(gl_sframe data, std::string target_column_name,
...
 
-  // Instantiate the training dependencies: data iterator, compute context,
-  // backend NN model.
+  if (variant_is<gl_sframe>(validation_data) ){
+
+    training_table_printer_.reset(new table_printer(
+    { {"Iteration", 12}, {"Train Accuracy", 12}, {"Train Loss", 12}, {"Validation Accuracy", 12}, {"Validation Loss", 12}, {"Elapsed Time", 12} }));
+
+  }
+  // else if (variant_is<flex_string>(validation_data) ){
```
TobyRoseman: Let's remove commented out code.

shreyajain17: Yeah I am still working on this PR. I had uploaded it so that Nick can look at it because I was getting NAN values for accuracy and loss. I was not resetting the validation data iterator. Although that is fixed now but this code needs more work. 

The reason for keeping this code was that I want to eventually accept "auto" string as well. Then elaborate on splitting the dataset itself.

I will add you for request as soon as I am done.

shreyajain17:
> Would it be reasonably possible to write a unit test for this?

I am not sure if there are any unit tests for the C++ implementation of activity classifier.

nickjong:
It was on my to-do list for this sprint to add some of the infrastructure needed to make these tests less painful to write (I didn't think it was fair to ask Shreya to write this). It's something I'm currently planning to do for the next sprint. It's definitely something we need to have before we actually replace the shipping (Python-based) toolkit with the new C++ version.

src/unity/toolkits/activity_classification/activity_classifier.cpp
TobyRoseman: Sorry @shreyajain17 - I didn't realize it was WIP. I'll make sure to check for that in the future.

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -152,18 +154,25 @@ void activity_classifier::train(gl_sframe data, std::string target_column_name,
 {
   // Begin printing progress.
   // TODO: Make progress printing optional.
-  
-  // TODO: Support validation set.
-  training_table_printer_.reset(new table_printer(
+
+  if (variant_is<gl_sframe>(validation_data) ){
```
nickjong: It's looking like it could make sense to move the initialization of `training_table_printer_` into `init_train`, which is where we print the table header anyway.

(Also, let's make sure to run clang-format on this PR's commits too. We should aim for more consistent formatting than e.g. the parens spacing in this line.)

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -152,18 +154,25 @@ void activity_classifier::train(gl_sframe data, std::string target_column_name,
...
-  
-  // Instantiate the training dependencies: data iterator, compute context,
-  // backend NN model.
+  }
+
   init_train(std::move(data), std::move(target_column_name),
              std::move(session_id_column_name), std::move(validation_data),
              std::move(opts));
 
+  // Instantiate the training dependencies: data iterator, compute context,
```
nickjong: Why move this comment down here? It's meant to document the `init_train` invocation

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -507,18 +529,87 @@ void activity_classifier::perform_training_iteration() {
...
   // Report progress if we have an active table printer.
-  // TODO: Report validation metrics.
-  
-  if (training_table_printer_) {
-    training_table_printer_->print_progress_row(
-        iteration_idx, iteration_idx + 1, average_batch_accuracy, average_batch_loss, progress_time());
+  if (validation_data_iterator_ != nullptr) {
+
+    float cumulative_val_accuracy = 0.f;
+    float cumulative_val_loss = 0.f;
```
nickjong: Could we factor out the computation of `cumulative_val_accuracy` and `cumulative_val_loss` into a (single) separate (private) helper method?

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -507,18 +529,87 @@ void activity_classifier::perform_training_iteration() {
...
+      std::map<std::string, shared_float_array> results =
+          training_model_->predict({
+              {"input", val.features},
+              {"labels", val.labels},
+              {"weights", val.weights},
+          });
+
+      const shared_float_array &val_output = results.at("output");
+
+      shared_float_array loss = results.at("loss");
```
nickjong: Nit: `val_output` and `loss` should logically have the same type, either both a const reference or both just a `shared_float_array` value. May as well have them both be references.

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -507,18 +529,87 @@ void activity_classifier::perform_training_iteration() {
...
+        data_iterator::batch::chunk_info info = val.batch_info[i];
+        size_t num_predictions =
+            (info.num_samples + prediction_window - 1) / prediction_window;
+        size_t num_correct_predictions = count_correct_predictions(
+            num_classes, output_chunk, label_chunk, num_predictions);
+
+        cumulative_per_seq_accuracy +=
+            static_cast<float>(num_correct_predictions) / num_predictions;
+      }
+      cumulative_val_accuracy += cumulative_per_seq_accuracy;
```
nickjong: I don't think we need both `cumulative_val_accuracy` and `cumulative_seq_accuracy`. Can just add each sequence's accuracy directly into `cumulative_val_accuracy`

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -507,18 +529,87 @@ void activity_classifier::perform_training_iteration() {
...
+    }
+
   }
 
-  add_or_update_state({
-      { "training_iterations", iteration_idx + 1 },
-      { "training_accuracy" , average_batch_accuracy },
-      { "training_log_loss", average_batch_loss },
-  });
+  else {
```
nickjong: Right now the code is organized like:
```
if has_validation_data:
  compute validation metrics
  update progress table
  update state
else:
  update progress table
  update state
```

Could we instead do something more like:
```
  compute validation metrics if necessary
  update progress table
  update state
```
Where each step's behavior depends on whether validation data is available? To me, validation data availability should be the "lower order bit" from the point of view of code organization

shreyajain17:
I was thinking about it. I think I am not handling the case if the given Frame for validation is empty and also I don't know what it does if the data is less then one batch size.

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -106,7 +107,7 @@ float get_accuracy_per_batch(size_t prediction_window, size_t num_classes,
     
     cumulative_per_batch_accuracy += static_cast<float>(num_correct_predictions) / num_predictions;
   }
-  return cumulative_per_batch_accuracy / batch.batch_info.size();
+  return cumulative_per_batch_accuracy;
```
nickjong: Let's rename this helper function, since you've changed the semantics of the return value. 

I honestly don't know what this should really be called? Cumulative chunk accuracy?

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -145,19 +146,49 @@ void activity_classifier::init_options(
   add_or_update_state(flexmap_to_varmap(options.current_option_values()));
 }
 
+std::tuple<float, float> activity_classifier::compute_validation_metrics(
```
nickjong: Let's at least add a quick comment documenting what the semantics of the two tuple values are (accuracy and loss)

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -145,19 +146,49 @@ void activity_classifier::init_options(
...
+    data_iterator *validation_data_iterator_, size_t prediction_window,
```
nickjong: I don't think we need to pass the `data_iterator*` since we can just use the class data member, right?

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -145,19 +146,49 @@ void activity_classifier::init_options(
...
+        validation_data_iterator_->next_batch(batch_size);
+    std::map<std::string, shared_float_array> results =
+        training_model_->predict({
+            {"input", val.features},
+            {"labels", val.labels},
+            {"weights", val.weights},
+        });
+
+    const shared_float_array &output = results.at("output");
+    shared_float_array loss = results.at("loss");
```
nickjong: I feel like `output` and `loss` should either both be values (`shared_float_array`) or references (`const shared_float_array&`)

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -447,6 +496,19 @@ void activity_classifier::init_train(
...
+    data_iterator::parameters validation_data_params;
+    validation_data_params.data = validation_sf;
+    validation_data_params.target_column_name = target_column_name;
+    validation_data_params.session_id_column_name = session_id_column_name;
+    validation_data_params.feature_column_names = feature_column_names;
+    validation_data_params.prediction_window =
+        read_state<flex_int>("prediction_window");
+    validation_data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
+    validation_data_iterator_ = create_iterator(validation_data_params);
+  }
```
nickjong: Don't forget to add the else case here, to guard against multiple calls to `init_train`

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -539,33 +601,57 @@ void activity_classifier::perform_training_iteration() {
...
   float average_batch_accuracy = cumulative_batch_accuracy / num_batches;
 
-  // Report progress if we have an active table printer.
-  // TODO: Report validation metrics.
-  
-  if (training_table_printer_) {
-    training_table_printer_->print_progress_row(
-        iteration_idx, iteration_idx + 1, average_batch_accuracy, average_batch_loss, progress_time());
 
+  if (validation_data_iterator_ == nullptr) {
```
nickjong: The remainder of this function is still hard to follow for me

Could we do something more like:
```
float average_val_accuracy;
float average_val_loss;
if (validation_data_iterator) {
  // Compute average_val_accuracy and average_val_loss
}

std::map<std::string, variant_type> state_updates({
  // training metrics
});
if (validation_data_iterator) {
  state_updates["validation_accuracy"] = ...;
  ...
}

if (training_table_printer_) {
  if (validation_data_iterator) {
    // Print six columns
  } else {
    // Print four columns
  }
}
```

(FWIW, I think doing everything with a single unified `if (validation_data_iterator_)` statement could make sense, but only if we refactor things to the point where the entire if-else statement fits on "one screen", without scrolling)

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -447,6 +502,21 @@ void activity_classifier::init_train(
...
+    gl_sframe validation_sf = variant_get_value<gl_sframe>(validation_data);
+    data_iterator::parameters validation_data_params;
+    validation_data_params.data = validation_sf;
+    validation_data_params.target_column_name = target_column_name;
+    validation_data_params.session_id_column_name = session_id_column_name;
+    validation_data_params.feature_column_names = feature_column_names;
+    validation_data_params.prediction_window =
+        read_state<flex_int>("prediction_window");
+    validation_data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
+    validation_data_iterator_ = create_iterator(validation_data_params);
```
nickjong: It would be nice to change create_iterator to accept just the SFrame, since all the other `data_iterator::parameters` are the same every time. We can do this in the next PR, though

nickjong:
Don't forget to remove the "WIP" prefix before merging!

PR #1705, opened 2019-04-08
Adding per_window functionality to C-API Activity Classifier 

### LINKED ISSUE #1642: C-API activity classifier should support per-window output frequency
The Python version can yield predicts either per-row (the output SFrame contains one row per input row), or per-window (the output SFrame contains one row per `prediction_window` (size) rows. The C-API version should provide the same functionality.

shreyajain17:
fixes #1642

### DIFF
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -180,8 +180,7 @@ void activity_classifier::train(gl_sframe data, std::string target_column_name,
 }
 
 gl_sarray activity_classifier::predict(gl_sframe data,
-                                       std::string output_type)
-{
+                                       std::string output_type) {
   if (output_type.empty()) {
     output_type = "class";
   }
@@ -206,10 +205,16 @@ gl_sarray activity_classifier::predict(gl_sframe data,
   // Assume output_frequency is "per_row". Duplicate each probability vector a
   // number of times equal to the number of samples for that prediction.
   // TODO: Also support "per_window".
+  //std::cout << data_it.get();
+  
+  // gl_sframe preds_per_window = gl_sframe({{"session_id",data_it.session_id_column_name}});
+  //std::cout << raw_preds_per_window;
+
   size_t preds_column_index = raw_preds_per_window.column_index("preds");
+
   size_t num_samples_column_index =
       raw_preds_per_window.column_index("num_samples");
-  auto copy_per_row = [=](const sframe_rows::row& row) {
+    auto copy_per_row = [=](const sframe_rows::row& row) {
     return flex_list(row[num_samples_column_index], row[preds_column_index]);
   };
   gl_sarray duplicated_preds_per_window =
@@ -218,8 +223,7 @@ gl_sarray activity_classifier::predict(gl_sframe data,
       gl_sframe({{"temp", duplicated_preds_per_window}}).stack("temp", "preds");
 
   gl_sarray result = preds_per_row["preds"];
-
-  if (output_type == "class") {
+    if (output_type == "class") {
     flex_list class_labels = read_state<flex_list>("classes");
     auto max_prob_label = [=](const flexible_type& ft) {
       const flex_vec& prob_vec = ft.get<flex_vec>();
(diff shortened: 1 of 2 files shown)


src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -132,9 +132,10 @@ void activity_classifier::train(gl_sframe data, std::string target_column_name,
   nn_spec_->update_params(trained_weights);
 }
 
-gl_sarray activity_classifier::predict(gl_sframe data,
-                                       std::string output_type)
-{
+
+gl_sarray activity_classifier::predict(gl_sframe data, std::string output_type,
+                                       std::string output_frequency) {
```
nickjong: Looks like this is from an earlier version of the change. We don't need `predict` to take `output_frequency` any more, right?

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -154,15 +155,28 @@ gl_sarray activity_classifier::predict(gl_sframe data,
...
+  gl_sframe raw_preds_per_window =
+      perform_inference(data_it.get(), output_frequency);
+
 
   // Assume output_frequency is "per_row". Duplicate each probability vector a
   // number of times equal to the number of samples for that prediction.
   // TODO: Also support "per_window".
+  // std::cout << data_it.get();
+
+  // gl_sframe preds_per_window =
```
nickjong: Don't forget to remove this commented out development code!

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -181,6 +194,21 @@ gl_sarray activity_classifier::predict(gl_sframe data,
...
+  return result;
+}
+
+gl_sframe activity_classifier::predict_per_window(gl_sframe data,
+                                                  std::string output_type) {
+
+
+  // else if (output_frequency =="per_window"){
+  //   gl_sframe preds_per_window =
+  //   gl_sframe({{"session_id",data.session_id_column_name}});
```
nickjong: Let's clean up this commented-out development code

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -463,13 +492,23 @@ void activity_classifier::perform_training_iteration() {
...
+gl_sframe
+activity_classifier::perform_inference(data_iterator *data,
+                                       std::string output_frequency) const {
   // Open a new SFrame for writing.
-  gl_sframe_writer writer({ "session_id", "preds", "num_samples" },
-                          { data->session_id_type(),
-                            flex_type_enum::VECTOR,
-                            flex_type_enum::INTEGER  },
-                          /* num_segments */ 1);
+  if (output_frequency == "per_window") {
```
nickjong: Instead of making the behavior of `perform_inference` parameterized by `output_frequency`, could we just have it always return an SFrame that suffices for either case. It can include both `num_samples` and `prediction_id` every time. The inclusion of `prediction_id` should not affect the per-row case; for the per-window case, you can just drop the `num_samples` column. (Or heck, maybe it's useful to keep it in there. What do you think?)

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -523,8 +565,23 @@ gl_sframe activity_classifier::perform_inference(data_iterator* data) const {
         cumulative_samples += prediction_window;
 
         // Add a row to the output SFrame.
-        writer.write({ info.session_id, preds, num_samples },
-                     /* segment_id */ 0);
+
+        if (output_frequency == "per_window") {
+
+          if (info.session_id == previous_session_id) {
```
nickjong: What happens if the `session_id` of the first prediction happens to be 1? Wouldn't we get an erroneous `prediction_id` of 2 for the first prediction then?

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -205,11 +204,11 @@ gl_sarray activity_classifier::predict(gl_sframe data,
 
   // Assume output_frequency is "per_row". Duplicate each probability vector a
   // number of times equal to the number of samples for that prediction.
-  // TODO: Also support "per_window".
   size_t preds_column_index = raw_preds_per_window.column_index("preds");
+
   size_t num_samples_column_index =
       raw_preds_per_window.column_index("num_samples");
-  auto copy_per_row = [=](const sframe_rows::row& row) {
+    auto copy_per_row = [=](const sframe_rows::row& row) {
```
nickjong: This line shouldn't be indented any further

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -218,8 +217,7 @@ gl_sarray activity_classifier::predict(gl_sframe data,
       gl_sframe({{"temp", duplicated_preds_per_window}}).stack("temp", "preds");
 
   gl_sarray result = preds_per_row["preds"];
-
-  if (output_type == "class") {
+    if (output_type == "class") {
```
nickjong: This statement should also remain at its previous indent level, the same as the prior statement

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -581,8 +624,9 @@ gl_sframe activity_classifier::perform_inference(data_iterator* data) const {
         size_t num_samples = std::min(prediction_window,
                                       info.num_samples - cumulative_samples);
         cumulative_samples += prediction_window;
-
+
```
nickjong: Let's not add this trailing whitespace

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -581,8 +624,9 @@ gl_sframe activity_classifier::perform_inference(data_iterator* data) const {
         size_t num_samples = std::min(prediction_window,
                                       info.num_samples - cumulative_samples);
         cumulative_samples += prediction_window;
-
+        
         // Add a row to the output SFrame.
+
```
nickjong: Nit: why add a blank line here?

PR #1707, opened 2019-04-08
WIP: Towards elimination of mutable_get.

hoytak:
Eliminating mutable_get access to flexible_type will enable us to better
use memory pools and reference counting to reduce allocations and
copying.  In this PR, it's changed to a private method, and instead a
function, get_and_reset<...>(), returns the object using a move
operation if it's unique and a copy if it's not.

### DIFF
--- a/src/flexible_type/flexible_type.hpp
+++ b/src/flexible_type/flexible_type.hpp
@@ -9,6 +9,7 @@
 #include <string>
 #include <vector>
 #include <tuple>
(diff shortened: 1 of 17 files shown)


hoytak:
This requires some non-trivial work on the cython interface, which I did not realize ahead of time.  The logic there needs a bit of reworking.  Closing this for now; replaced by branch hoytak/WIP-mutable-get

PR #1715, opened 2019-04-09
Focus Function for the GUI of the Evaluation UI

### LINKED ISSUE #1548: Model Evaluation Doesn't Steal Focus
Calling `model.evaluate(test_set).explore()`does correctly open an explore window. However that window load in the background the user needs to alt-tab to the correct window to see it.

This is particularly confusing if it's the first time the user has called `explore` or if the user already as several explore windows open.

abhishekpratapa:
Fixes #1548
- Fixes the issue where the Window doesn't steal focus when it originally opens.

### DIFF
--- a/src/unity/python/turicreate/visualization/_plot.py
+++ b/src/unity/python/turicreate/visualization/_plot.py
@@ -29,6 +29,13 @@ def _get_client_app_path():
     if _sys.platform == 'linux2' or _sys.platform == 'linux':
         return _os.path.join(tcviz_dir, 'Turi Create Visualization', 'visualization_client')
 
+def _focus_client_app():
+    scpt = '''
+            delay .5
+            tell application \"Turi Create Visualization\" to activate
+            '''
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/visualization/_plot.py
```diff
@@ -29,6 +29,13 @@ def _get_client_app_path():
     if _sys.platform == 'linux2' or _sys.platform == 'linux':
         return _os.path.join(tcviz_dir, 'Turi Create Visualization', 'visualization_client')
 
+def _focus_client_app():
+    scpt = '''
+            delay .5
+            tell application \"Turi Create Visualization\" to activate
+            '''
+    focus = __subprocess._Popen(['osascript', '-'], stdout=_PIPE, stdin=_PIPE)
```
znation: I think we have the equivalent of this elsewhere (probably duplicated in the `show` and `explore` methods). It would be nice to refactor to share the implementation -- but not necessary for this PR.

Issue #1712, opened 2019-04-08
Sound Classifier predict for clips less than 975 milliseconds 


TobyRoseman:
Currently the sound classifier can only use clips which are at least 975 milliseconds. This means it's not possible to make predictions for clips which are shorter than that.

This enhancement was requested in #1702.

Off the top of my head I can think of three possible solutions shorter clips: 
1 - Pad the clip with zeros
2 - "stretch" out the clip. Basically upsampling with interpolation. Probably using [scipy.interpolate.interp1d](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.interpolate.interp1d.html)
3 - Loop the clip back on itself. So if we're missing 150 milliseconds of data we just tack the first 150 milliseconds on the end.  

Does anyone have an opinion about any of these three options? Does anyone have any other ideas?

TobyRoseman:
After discussing this issue with an audio expert, it seems option #1 is the only sensible approach.

rplom:
I can take this issue.  I have extensively tested it with padding with silence and have not seen any issues with the layers being confused.

TobyRoseman:
Thanks @rplom - the issue is yours. I'm happy to code review it when it's ready.

TobyRoseman:
This has now been fixed for deployment to CoreML. #2310 updated the user guide to pad partial windows with zeros/silence.

Let's leave this issue open until we have a similar fix for prediction on the Python side.

PR #1720, opened 2019-04-11
Sound clips less than 975ms should never cause sound classifier crash

### LINKED ISSUE #1702: Creating a Sound Classifier for Short Audio Clips
I'm trying to create a custom sound classifier and use a directory of folders for the labeling of each audio file. For example, all .wav files in the folder "Meow" should be labeled "Meow" in classifier and all .wav files in the folder "Woof" should be labeled "Woof". After modifying the [Sound Classifier Example](https://apple.github.io/turicreate/docs/userguide/sound_classifier/ "Sound Classifier"), I ended up with this code:

> \# Load the input data from the directories.
> data = tc.load_audio('./Audio/', with_path=True)
> data['label'] = data['path'].apply(lambda p: os.path.basename(os.path.dirname(p)))
> data.save('classifier.sframe')
> data = tc.SFrame('classifier.sframe')
> 
> \# Make a train-test split
> test_set, train_set = data.random_split(0.8)
>
> \# Create the model.
> model = tc.sound_classifier.create(train_set, target='label', feature='audio')
>
> \# Generate an SArray of predictions from the test set.
> data = model.predict(test_set)
>
> \# Evaluate the model and print the results
> metrics = model.evaluate(test_set)

However, when I try to execute 'model.predict', I receive an error message:
>"Input SArrays 'targets' and 'predictions' must be of the same length."

1. What causes this error message and how can I train a custom classifier using the approach above? I looked for the cause of the message online and it seems like it might have to do with a mismatch in the column lengths from the source data.
2. Do audio files have to be in a particular format for the classifier to work correctly? The audio files I'm working with have variable durations, bitrates, etc. I thought this might be causing the error above, but after converting the files to a standard format the issue persisted.

TobyRoseman:
Fixes #1702

### DIFF
--- a/src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
+++ b/src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
@@ -10,6 +10,7 @@ from __future__ import print_function as _
 from __future__ import division as _
 from __future__ import absolute_import as _
 
+import logging as _logging
 import numpy as _np
 
 from .. import _mxnet_utils
@@ -182,6 +183,10 @@ def create(dataset, target, feature, max_iterations=10,
 
     train_data = _tc.SFrame({'deep features': train_deep_features, 'labels': encoded_target})
     train_data = train_data.stack('deep features', new_column_name='deep features')
+    train_data, missing_ids = train_data.dropna_split(columns=['deep features'])
+
+    if len(missing_ids) > 0:
+        _logging.warning("Dropping %d examples which are less than 975ms in length." % len(missing_ids))
 
     if validation_set is not None:
         if verbose:
@@ -195,6 +200,7 @@ def create(dataset, target, feature, max_iterations=10,
 
         validation_data = _tc.SFrame({'deep features': validation_deep_features, 'labels': validation_encoded_target})
         validation_data = validation_data.stack('deep features', new_column_name='deep features')
+        validation_data = validation_data.dropna(columns=['deep features'])
 
         validation_batch_size = min(len(validation_data), batch_size)
         validation_data = mx.io.NDArrayIter(validation_data['deep features'].to_numpy(),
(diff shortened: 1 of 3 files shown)


src/unity/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -182,6 +183,10 @@ def create(dataset, target, feature, max_iterations=10,
 
     train_data = _tc.SFrame({'deep features': train_deep_features, 'labels': encoded_target})
     train_data = train_data.stack('deep features', new_column_name='deep features')
+    train_data, missing_ids = train_data.dropna_split(columns=['deep features'])
+
+    if len(missing_ids) > 0:
+        _logging.warning("Dropping %d examples which are less than 975ms in length." % len(missing_ids))
```
znation: In general this code looks good, but I'm wondering why we have this limitation -- is the underlying model not capable of handling smaller samples, or did we pick a hardcoded minimum sample size that seems to give good results?

TobyRoseman: The underlying model is only capable of handling 975ms chunks. VGGish (the second part of the model classification pipeline) only accepts a fixed sized input to get that sized input the preprocessing code must have 975ms of audio.

znation: I think it's fine for now to drop smaller samples and warn, but I wonder if it would make sense (as a future enhancement) to either stretch or pad smaller samples so they are >= 975ms. (I'm not sure whether this actually works or not in terms of getting good results from the model, though.)

TobyRoseman: I totally agree. I created #1712 to track this issue.

PR #1722, opened 2019-04-11
User Guide - Sound Classifier Advanced Usage 

TobyRoseman:
This covers both reusing deep features and tuning the custom neural network configuration. Both of these features will be a part of our next release. This pull request finishes up #1639.

The new Advanced Usage Section is staged here:
https://tobyroseman.github.io/turicreate/docs/userguide/sound_classifier/advanced-usage.html

@madebysasha @jamois @coolioxlr @rplom @yulingtianxia - Please let us know what you think of these new features. We would also be very interested in any other feedback you have about the Sound Classifier Toolkit. Thanks!

### DIFF
--- a/src/unity/python/doc/source/turicreate.toolkits.sound_classifier.rst
+++ b/src/unity/python/doc/source/turicreate.toolkits.sound_classifier.rst
@@ -15,4 +15,5 @@ sound classifier
 
   sound_classifier.create
   sound_classifier.SoundClassifier
+  sound_classifier.get_deep_features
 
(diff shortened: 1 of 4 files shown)


TobyRoseman:
> Do you think it's valuable to include known issues (such as the [975ms minimum clip size](https://github.com/apple/turicreate/issues/1712)) or should we leave that out?

In general I don't think it's worth it to mention known issues in the User Guide. We already have GitHub issues for known issues. If we add that information in another place, I think it's very easy for the two to get out of sync.

The 975ms issue might be worth mentioning but the experience [should be much improved](https://github.com/apple/turicreate/pull/1720) in our next release. Also there is an implicit mention of it in the [How it Works](https://apple.github.io/turicreate/docs/userguide/sound_classifier/how-it-works.html#signal-preprocessing-pipeline-stage) section.

PR #1731, opened 2019-04-16
Changes made in CoreML model exported by C-API activity classifier

### LINKED ISSUE #1703: Core ML model exported by C-API Activity Classifier should merge "cellIn" and "hiddenIn"
The CoreML model can provide a simpler API to the developer, since the cell activations and hidden layer activations can be treated as effectively one blob with the concatenation of the two components. The model itself can handle splitting apart the inputs and joining the outputs.

### LINKED ISSUE #1704: CoreML model exported by C-API activity classifier should provide named inputs
Currently the Core ML expects a single feature multi-array, whose outer dimension matches the number of feature columns used. The developer must parse the metadata to determine the correct order. We can instead provide a separate named input for each training column, making the integration of the Core ML model less error-prone

shreyajain17:
- Provides named inputs. Fixes #1704 

- Merges cellIn and hiddenIn. Fixes #1703

### DIFF
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -367,6 +367,11 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
   size_t num_classes = read_state<flex_int>("num_classes");
   size_t num_features = read_state<flex_int>("num_features");
   size_t prediction_window = read_state<flex_int>("prediction_window");
+  
+
+  for (size_t i=0; i<num_features; i++) {
+    
+  }
 
   result->add_permute("permute", "features", {{0, 3, 1, 2}});
   result->add_convolution(
@@ -384,6 +389,9 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
           NUM_CONV_FILTERS * prediction_window),
       /* bias_init_fn */        zero_weight_initializer());
   result->add_relu("relu1", "conv");
+
+  result->add_channel_slice("hiddenIn","stateIn",0,200,1);
+  result->add_channel_slice("cellIn","stateIn",200,400,1);
   result->add_lstm(
       /* name */                "lstm",
       /* input */               "relu1",
@@ -396,10 +404,10 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
       /* cell_clip_threshold */ LSTM_CELL_CLIP_THRESHOLD,
       /* initializers */  lstm_weight_initializers::create_with_xavier_method(
           NUM_CONV_FILTERS, LSTM_HIDDEN_SIZE));
-  // result->add_reshape("lstm1","lstm", {{0, 1, 3, 2}});
+  result->add_channel_concat("stateOut",{"hiddenOut","cellOut"});
   result->add_inner_product(
       /* name */ "dense0",
-      /* input */ "lstm1",
+      /* input */ "lstm",
       /* num_output_channels */ FULLY_CONNECTED_HIDDEN_SIZE,
       /* num_input_channels */ LSTM_HIDDEN_SIZE,
       /* weight_init_fn */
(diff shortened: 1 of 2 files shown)


src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -384,6 +389,9 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
           NUM_CONV_FILTERS * prediction_window),
       /* bias_init_fn */        zero_weight_initializer());
   result->add_relu("relu1", "conv");
+
+  result->add_channel_slice("hiddenIn","stateIn",0,200,1);
+  result->add_channel_slice("cellIn","stateIn",200,400,1);
```
nickjong: Instead of `200`, let's use the constant `LSTM_HIDDEN_SIZE` here

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -291,10 +291,12 @@ std::shared_ptr<MLModelWrapper> export_activity_classifier_model(
...
   // Write the primary input features.
-  set_array_feature(model_desc->add_input(), "features",
-                    get_activity_classifier_features_short_description(
-                        prediction_window, features),
-                    { 1, prediction_window, features.size() });
+  for (size_t i = 0; i < features.size(); i++) {
+    set_array_feature(model_desc->add_input(), features[i],
+                      get_activity_classifier_features_short_description(
+                          prediction_window, features[i]),
+                      {1, prediction_window, features[i].size()});
```
nickjong: This doesn't look right: if we have e.g. six features, you're saying that _each_ feature should have input shape [1, 50, 6] (with prediction_window=50).

We should figure out whether the right shape is [1, 50, 1] or just a 1-d array [50]. (If we do the latter, we'd need to add some reshape into the exported model)

nickjong: Note the interface that appears in the screenshot here: https://apple.github.io/turicreate/docs/userguide/activity_classifier/export_coreml.html

It looks like stateIn and stateOut should appear as `(Double 400)`, but we probably want each named feature to similarly appear as `(Double 50)`. Then the developer doesn't have to worry about multi-arrays at all

shreyajain17: That sounds reasonable, this is why we wanted it to be named to eliminate multi-arrays. I think I can reshape it after and it should not be a problem. I will make appropriate changes.

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -291,31 +291,29 @@ std::shared_ptr<MLModelWrapper> export_activity_classifier_model(
   ModelDescription* model_desc = model.mutable_description();
 
   // Write the primary input features.
-  set_array_feature(model_desc->add_input(), "features",
-                    get_activity_classifier_features_short_description(
-                        prediction_window, features),
-                    { 1, prediction_window, features.size() });
+  for (size_t i = 0; i < features.size(); i++) {
+    set_array_feature(model_desc->add_input(), features[i],
+                      get_activity_classifier_features_short_description(
```
nickjong: If I'm not mistaken, this will yield a short description like `Window × [acc_x]`

Maybe it should be something like `acc_x input window` or `acc_x window input`???

src/unity/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -291,31 +291,29 @@ std::shared_ptr<MLModelWrapper> export_activity_classifier_model(
...
+                          prediction_window, features[i]),
+                      {1, 1, prediction_window});
```
nickjong: Wait, could we make the shape just `{prediction_window}` here instead? Does that still work without adding in a reshape layer?

shreyajain17: I will try it and tell you.

PR #1302, opened 2019-01-10
Added dataset and edited code to use it (Text classification)

euwars:
[Reference](https://github.com/apple/turicreate/issues/1297)

PR #1733, opened 2019-04-16
Read CSV: quote_char value of None should be passed to C++ layer

TobyRoseman:
Without this change, calling `tc.SFrame.read_csv` with `quote_char=None` has no affect. The default value of `"` gets set in the C++ layer. 

This change is necessary to read non-multi line CSV files which have unmatched quotation marks.

### DIFF
--- a/src/unity/lib/unity_sframe.cpp
+++ b/src/unity/lib/unity_sframe.cpp
@@ -212,6 +212,8 @@ std::map<std::string, std::shared_ptr<unity_sarray_base>> unity_sframe::construc
   if (csv_parsing_config["quote_char"].get_type() == flex_type_enum::STRING) {
     std::string tmp = (flex_string)csv_parsing_config["quote_char"];
     if (tmp.length() > 0) tokenizer.quote_char = tmp[0];
+  } else if (csv_parsing_config["quote_char"].get_type() == flex_type_enum::UNDEFINED) {
+    tokenizer.quote_char = NULL;
   }
   if (csv_parsing_config.count("skip_initial_space")) {
     tokenizer.skip_initial_space = !csv_parsing_config["skip_initial_space"].is_zero();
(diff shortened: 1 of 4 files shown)


src/unity/lib/unity_sframe.cpp
```diff
@@ -212,6 +212,8 @@ std::map<std::string, std::shared_ptr<unity_sarray_base>> unity_sframe::construc
   if (csv_parsing_config["quote_char"].get_type() == flex_type_enum::STRING) {
     std::string tmp = (flex_string)csv_parsing_config["quote_char"];
     if (tmp.length() > 0) tokenizer.quote_char = tmp[0];
+  } else if (csv_parsing_config["quote_char"].get_type() == flex_type_enum::UNDEFINED) {
```
ylow: hmm. This is not the ideal fix. tokenizer.quote_char is a char, so this really has the effect is setting the quote char to '\0'. 

Technically the current csv parser implementation does not support the disabling of quote_char. In a way, setting to '\0' is kind of a workaround.

ylow: I will take a look at supporting the disabling of quote_char as an parsing option after I get #1737 in since that set of changes will most certainly conflict.

TobyRoseman: Ok, we can wait till after #1737 is merged. Setting `quote_char='\0'` in Python does solve the problem. However `quote_char=None` just makes a lot more sense in Python; that's how I tried to solve the problem and was surprised when it didn't work.

TobyRoseman:
This issues was originally discovered as part of this pull request #1302. I've updated that pull request to use `quote_char='\0'`. If we do end up making `quote_char=None` work, we should make sure to also update the user guide.

src/unity/lib/unity_sframe.cpp
TobyRoseman: #1734 has been merged. I've rebased this PR on top of master. Everything passes.

PR #1735, opened 2019-04-16
Remove sudo tag from travis config

TobyRoseman:
#1729


TobyRoseman:
> I thought we had it in there for a reason, but if the build passes without it, great!

When we set up Travis CI I believe there was a reason for it. Since then I think there has been changes on the Travis side.

PR #1750, opened 2019-04-19
Fully automated docker builds

### LINKED ISSUE #257: Wrong git SHA in egg/wheel
`tc.version_info.git_sha` seems wrong.

```python

In [2]: tc.version_info.version
Out[2]: '4.1'

In [3]: tc.version_info.git_sha
Out[3]: '54b25133c1d8b8d84e2e580f546a23348f3c6796'
```

However that doesn't seem to even be a commit in this repository:
https://github.com/apple/turicreate/commit/54b25133c1d8b8d84e2e580f546a23348f3c6796

I would expect it to match the SHA listed here (`39e7f27`): https://github.com/apple/turicreate/releases/tag/v4.1

### LINKED ISSUE #1441: GLIBCXX requirement is inconvenient on older Linux distributions
Running out of the box on Ubuntu 14.04 (for instance), without sudo permission to install a new libstdc++ package, users will get:

```
"/usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /app/.python/lib/python3.6/site-packages/turicreate/cython/cy_unity.so)"
```

znation:
This commit allows for fully automated Docker-based builds, using the
following scripts:

* `make_wheel.sh --docker-python${VERSION}` builds a wheel and automates
   the whole process within Docker including testing
* `test_wheel.sh` runs Python tests (optionally in Docker)
* `run_cpp_tests.py` now optionally runs C++ tests in Docker
* Travis CI now uses the above Docker config by default.
* Gitlab CI configuration provided; no server is hooked up. If you clone
  this repo to your own Gitlab repo, and provide runners, CI should work.

Fixes #257 -- with this change, we can now build release wheels directly
from this repo (thus picking up the correct git sha).

Fixes #1441 -- with this change, we are now compatible with libstdc++
from GCC 4.8 (compatible with Ubuntu 14.04).


scripts/test_wheel.sh
```diff
@@ -0,0 +1,92 @@
...
+
+  # Run the tests inside Docker
+  if [[ "${DOCKER_PYTHON}" == "2.7" ]]; then
+    docker run --rm \
+      --mount type=bind,source=$WORKSPACE,target=/build,consistency=delegated \
+      -e "VIRTUALENV=virtualenv --python=python${DOCKER_PYTHON}" \
+      turicreate/build-image-14.04:${TC_BUILD_IMAGE_VERSION} \
+      /build/scripts/test_wheel.sh
+  elif [[ "${DOCKER_PYTHON}" == "3.5" ]]; then
+    docker run --rm \
```
TobyRoseman: This look identical to the body of the previous if statement. Should we remove this `elif` and make the test on line 55 check for either 2.7 or 3.5?

src/unity/lib/annotation/utils.cpp
```diff
@@ -1,10 +1,11 @@
 #include "utils.hpp"
 
 bool turi::annotate::isInteger(std::string s) {
-	if(s.empty() || ((!isdigit(s[0])) && (s[0] != '-') && (s[0] != '+'))) return false;
```
TobyRoseman: Our these and the other annotator changes intensionally part of this PR?

.gitlab-ci.yml
```diff
@@ -0,0 +1,425 @@
...
+
+build_wheel_debug_linux_python27:
+  tags:
+    - docker-in-docker
+  services:
+    - docker:dind
+  image: docker:stable
+  stage: build
+  variables:
+    CI: 'false'
```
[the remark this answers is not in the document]
znation: Actually, we can get rid of that -- it was a copy/paste from a now obsolete rule. Thanks.

.gitlab-ci.yml
```diff
@@ -0,0 +1,425 @@
...
+  stage: build
+  script:
+    - source scripts/use_ccache.sh
+    - ./configure --no-visualization --no-python --release-opt-for-size --target-ios
+    - cd release/src/objc
+    - make -j8
+  artifacts:
+    expire_in: 2 weeks
+    paths:
+      - release/src/objc/libRecommender.dylib
```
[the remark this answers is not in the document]
znation: Thanks, I think I copy/pasted from an old version. I'll add that.

src/unity/lib/annotation/utils.cpp
znation: Yes (though only the non-whitespace changes were intended -- that was my editor trying to be "helpful"). There was a compiler (warning-as-)error on unused return value; I modified this code to suppress the error.

znation:
@TobyRoseman Feedback addressed, thanks!

scripts/test_wheel.sh
```diff
@@ -0,0 +1,86 @@
...
+    --docker-python3.6)     USE_DOCKER=1;DOCKER_PYTHON=3.6;;
+    --help)                 print_help ;;
+    *) unknown_option $1 ;;
+  esac
+  shift
+done
+
+set -x
+
+SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
hoytak: Are the " matched properly here?  I mean I trust you it works but this looks weird

src/external/coremltools_wrap/coremltools/mlmodel/build/format/OneHotEncoder_enums.h
```diff
@@ -21,6 +21,7 @@ static const char * MLOneHotEncoderCategoryType_Name(MLOneHotEncoderCategoryType
         case MLOneHotEncoderCategoryType_NOT_SET:
             return "INVALID";
     }
+    return "INVALID";
```
hoytak: If the compiler is generating a false negative on the return values being present, probably easier to just conditionally turn off that warning if the compiler in bad.  This directory was meant to be rsync clobbered by the coremltools directory.

scripts/test_wheel.sh
znation: They are; it's copied from https://github.com/apple/turicreate/blob/master/scripts/make_wheel.sh#L14.

src/external/coremltools_wrap/coremltools/mlmodel/build/format/OneHotEncoder_enums.h
znation: Good point. Since I already have tests passing on this one, let's put it in as is (assuming we don't need further commits), and open a GitHub issue to track the additional warnings (which will be fixed by the time we need to do another rsync).

scenario-tests/run_scenario_tests.sh
```diff
@@ -40,30 +51,57 @@ while [ $# -gt 0 ]
...
+      /build/scenario-tests/run_scenario_tests.sh $TC_WHEEL_UNDER_TEST
+  elif [[ "${DOCKER_PYTHON}" == "3.6" ]]; then
+    docker run --rm -m=4g \
+      --mount type=bind,source=$WORKSPACE,target=/build,consistency=delegated \
+      -e "VIRTUALENV=virtualenv --python=python${DOCKER_PYTHON}" \
+      turicreate/build-image-18.04:${TC_BUILD_IMAGE_VERSION} \
+      /build/scenario-tests/run_scenario_tests.sh $TC_WHEEL_UNDER_TEST
+  fi
+
+  exit 0
```
TobyRoseman: Seems like this should be an error. Maybe `exit 1` and an error message.

scenario-tests/run_scenario_tests.sh
```diff
@@ -40,30 +51,57 @@ while [ $# -gt 0 ]
...
+
+  # Run the tests inside Docker
+  if [[ "${DOCKER_PYTHON}" == "2.7" ]]; then
+    docker run --rm -m=4g \
+      --mount type=bind,source=$WORKSPACE,target=/build,consistency=delegated \
+      -e "VIRTUALENV=virtualenv --python=python${DOCKER_PYTHON}" \
+      turicreate/build-image-14.04:${TC_BUILD_IMAGE_VERSION} \
+      /build/scenario-tests/run_scenario_tests.sh $TC_WHEEL_UNDER_TEST
+  elif [[ "${DOCKER_PYTHON}" == "3.5" ]]; then
+    docker run --rm -m=4g \
```
TobyRoseman: The body of this `elif` block seem identical to the body of the `if` statement above it. If that is indeed the case, I think you should combine them, i.e the `if` test should check if it's 2.7 or 3.5.

LINUX_INSTALL.md
```diff
@@ -2,18 +2,18 @@
...
- - glibc >= 2.12 (Centos >= 6, Ubuntu >= 10.10)
+ - glibc >= 2.11 (Centos >= 6, Ubuntu >= 10.04)
  - For Neural Network support, glibc >= 2.17 is needed. Centos >= 7 and Ubuntu >= 13.04
- - libstdc++ >= 6.0.21
+ - libstdc++ >= 6.0.19
  - libgconf-2-4 (on Ubuntu 17.10 and later)
 
 ## Ubuntu
 
 ### Ubuntu >= 17.10
```
TobyRoseman: Do you think it would be a good idea to link to the relevant docker files?

scripts/test_wheel.sh
```diff
@@ -0,0 +1,86 @@
...
+    --docker-python2.7)     USE_DOCKER=1;DOCKER_PYTHON=2.7;;
+    --docker-python3.5)     USE_DOCKER=1;DOCKER_PYTHON=3.5;;
+    --docker-python3.6)     USE_DOCKER=1;DOCKER_PYTHON=3.6;;
+    --help)                 print_help ;;
+    *) unknown_option $1 ;;
+  esac
+  shift
+done
+
+set -x
```
TobyRoseman: Probably best to have this near the top of the file with the other `set`.

scripts/test_wheel.sh
```diff
@@ -0,0 +1,86 @@
...
+      /build/scripts/test_wheel.sh
+  elif [[ "${DOCKER_PYTHON}" == "3.6" ]]; then
+    docker run --rm -m=8g \
+      --mount type=bind,source=$WORKSPACE,target=/build,consistency=delegated \
+      -e "VIRTUALENV=virtualenv --python=python${DOCKER_PYTHON}" \
+      turicreate/build-image-18.04:${TC_BUILD_IMAGE_VERSION} \
+      /build/scripts/test_wheel.sh
+  fi
+
+  exit 0
```
TobyRoseman: Seems like this should be an error.

scenario-tests/run_scenario_tests.sh
znation: This is not an error; this is an early return for a happy path. If any of the above commands fails, the whole script will return a non-zero exit code (because of `set -e`).

scripts/test_wheel.sh
znation: This is not an error; this is an early return for a happy path. If any of the above commands fails, the whole script will return a non-zero exit code (because of `set -e`).

src/external/coremltools_wrap/coremltools/mlmodel/build/format/OneHotEncoder_enums.h
znation: This is already fixed in coremltools with https://github.com/apple/coremltools/pull/332.

scenario-tests/run_scenario_tests.sh
TobyRoseman: Sorry, I meant perhaps you should have a `else` which is an error.

LINUX_INSTALL.md
znation: Probably not under LINUX_INSTALL (the Docker files aren't really relevant to installation) but under BUILD I will.

scenario-tests/run_scenario_tests.sh
znation: Oh, that's a good point. I'll add that.

PR #1752, opened 2019-04-19
Added Evaluate Class Return to classifier models

### LINKED ISSUE #1352: Evaluation Objects Need to Be Returned From All Models
Since we are adding a new API method we should make all models return an Evaluation, and have explore throw NotImplementedException in the case of non-image-classifiers.

abhishekpratapa:
fixes #1352

### DIFF
--- a/src/unity/python/turicreate/toolkits/evaluation.py
+++ b/src/unity/python/turicreate/toolkits/evaluation.py
@@ -33,6 +33,19 @@ from turicreate.toolkits._internal_utils import _raise_error_if_not_sarray,\
                                               _check_categorical_option_type
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
 
+try:
+    from UserDict import UserDict
+except ImportError:
+    from collections import UserDict
+
+class _Evaluation(dict):
+    def __init__(self, obj = {}):
+        dict.__init__(self)
+        self.data = obj
+        self.update(obj)
+    def explore(self):
+        raise NotImplementedError("Explore is not implemented for this classifier type")
+
 def _check_prob_and_prob_vector(predictions):
     """
     Check that the predictionsa are either probabilities of prob-vectors.
--- a/src/unity/python/turicreate/toolkits/evaluation.py
+++ b/src/unity/python/turicreate/toolkits/evaluation.py
@@ -33,16 +33,12 @@ from turicreate.toolkits._internal_utils import _raise_error_if_not_sarray,\
                                               _check_categorical_option_type
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
 
-try:
-    from UserDict import UserDict
-except ImportError:
-    from collections import UserDict
-
 class _Evaluation(dict):
     def __init__(self, obj = {}):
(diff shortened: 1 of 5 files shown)


src/unity/python/turicreate/toolkits/evaluation.py
```diff
@@ -33,6 +33,19 @@
                                               _check_categorical_option_type
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
 
+try:
```
TobyRoseman: `UserDict` doesn't seem to be getting used anywhere.

znation:
> Please comment here or in the issue about motivation for this change.

The image classifier toolkit evaluate method already returns an Evaluation type that wraps dict. This PR is intended to fix the API inconsistency introduced by that change (the `evaluate` method on all models should return the same type). A future PR will address the functionality inconsistency (i.e. incrementally, these "not implemented" exceptions will get replaced with implementations).

src/unity/python/turicreate/toolkits/evaluation.py
```diff
@@ -33,6 +33,19 @@
...
+    from UserDict import UserDict
+except ImportError:
+    from collections import UserDict
+
+class _Evaluation(dict):
```
znation: The return value from image_classifier `evaluate` should probably inherit from this type, right? (The whole point of this PR is to fix API inconsistency, meaning all evaluate methods should return the same base type -- `isinstance` should behave the same on all of them.)

TobyRoseman:
> > Please comment here or in the issue about motivation for this change.
> 
> The image classifier toolkit evaluate method already returns an Evaluation type that wraps dict. This PR is intended to fix the API inconsistency introduced by that change (the `evaluate` method on all models should return the same type). A future PR will address the functionality inconsistency (i.e. incrementally, these "not implemented" exceptions will get replaced with implementations).

I get that currently in master things here are a bit inconsistent: you call `evaluate` on an image classifier and you get something which you can call `explore()` on, but you call `evaluate` on another model and you just get a `dict`. I don't think adding an `explore()` implementation which just raises an error makes this situation any better. I would recommend we just wait until we have the desired functionality before we add the `explore` methods.

znation:
@TobyRoseman I see your logic. I think I could go either way on this one. @abhishekpratapa, what are your thoughts on which option is better until we have a real `explore` method on all of them?

Or, is there anything else we can do in the meantime (like provide a default implementation for all evaluations that falls back to table view with a warning)?

src/unity/python/turicreate/toolkits/evaluation.py
abhishekpratapa: I changed it. As for what the behavior should be. It doesn't currently break anything I know of, it just wraps the dictionary object in python. That way the user still gets an explore warning in TuriCreate when they try to explore their evaluations.

src/unity/python/turicreate/toolkits/evaluation.py
abhishekpratapa: Thoughts? on the PR Toby?

PR #1762, opened 2019-04-22
Fix assertion error when evaluating C-API activity classifier trained with only two labels

nickjong:
The new evaluation code was always passing full probability vectors, even for binary-class evaluators that expected a single probability. It's simpler and safer just to use the multi-class versions at this point, which is what the supervised_learning toolkit does anyway

### DIFF
--- a/src/unity/toolkits/evaluation/metrics.cpp
+++ b/src/unity/toolkits/evaluation/metrics.cpp
@@ -175,11 +175,13 @@ variant_map_type compute_classifier_metrics_from_probability_vectors(
     class_to_index[class_labels[i]] = i;
   }
 
-  // Initialize the evaluators.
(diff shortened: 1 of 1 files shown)


nickjong:
In creating (for the C-API activity classifier) a fusion of the Python evaluation API and what supervised learning does in C++, I missed that supervised_learning now treats all classifiers as multi-class for the purposes of evaluation (https://github.com/apple/turicreate/blob/d2b900bb1864555b058ba5f3b339fe49abb086ea/src/unity/toolkits/supervised_learning/supervised_learning.cpp#L752). Instead, I was passing probability vectors into binary evaluators that expected single probabilities. Adopting the supervised_learning toolkit's approach simplifies things.

This will also make this code more straightforward to wrap and expose as a standalone evaluation function for the C-API

PR #1767, opened 2019-04-23
Refactoring C-API for Activity Classifier 

### LINKED ISSUE #1766: Refactoring needed for data iterator for C-API of activity classifier
The refactor is needed as the data iterator takes data parameters as argument now but can be made simpler if it takes the data itself and the class labels.

shreyajain17:
- fixes #1766 
- fixes the double print statements

### DIFF
--- a/src/unity/toolkits/activity_classification/ac_data_iterator.hpp
+++ b/src/unity/toolkits/activity_classification/ac_data_iterator.hpp
@@ -60,6 +60,8 @@ public:
      * an exception will be thrown upon encountering an unexpected label.
      */
     flex_list class_labels;
+
+    bool is_train;
   };
 
   /** Defines the output of a data_iterator. */
--- a/src/unity/toolkits/activity_classification/ac_data_iterator.hpp
+++ b/src/unity/toolkits/activity_classification/ac_data_iterator.hpp
@@ -61,7 +61,8 @@ public:
      */
     flex_list class_labels;
 
-    bool is_train;
+    /**  Generates verbose output when set to true. */
+    bool verbose;
   };
 
   /** Defines the output of a data_iterator. */
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -515,7 +515,6 @@ void activity_classifier::init_train(
   training_data_iterator_ = create_iterator(data, true);
 
   add_or_update_state({{"classes", training_data_iterator_->class_labels()}});
-  flex_list class_labels = read_state<flex_list>("classes");
 
   // Bind the validation data to a data iterator.
   if (variant_is<gl_sframe>(validation_data)) {
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -223,7 +223,7 @@ gl_sarray activity_classifier::predict(gl_sframe data,
   flex_list class_labels = read_state<flex_list>("classes");
 
   std::unique_ptr<data_iterator> data_it =
-      create_iterator(data, class_labels, false);
+      create_iterator(data, false);
 
   // Accumulate the class probabilities for each prediction window.
   gl_sframe raw_preds_per_window = perform_inference(data_it.get());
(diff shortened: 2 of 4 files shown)


src/unity/toolkits/activity_classification/ac_data_iterator.hpp
```diff
@@ -60,6 +60,8 @@ class data_iterator {
      * an exception will be thrown upon encountering an unexpected label.
      */
     flex_list class_labels;
+
+    bool is_train;
```
nickjong: Could we instead call this parameter `verbose`? Then the name better suggests what impact the value has. We can also set a default here.

Either way, let's add a comment documenting the parameter. (Follow the example of the other struct members here, which use Doxygen-style comments that can be automatically compiled into our C++ API docs.)

src/unity/toolkits/activity_classification/activity_classifier.hpp
```diff
@@ -171,43 +171,42 @@ class EXPORT activity_classifier: public ml_model_base {
...
-  // prediction window size).
-  virtual gl_sframe perform_inference(data_iterator *data) const;
-
-  // Utility code
-
-  template <typename T>
-  T read_state(const std::string& key) const {
-    return variant_get_value<T>(get_state().at(key));
-  }
+   virtual std::unique_ptr<data_iterator>
```
nickjong: I think you accidentally indented this declaration (and all the ones below) with 3 spaces instead of 2!

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -220,15 +220,10 @@ gl_sarray activity_classifier::predict(gl_sframe data,
...
-  data_params.session_id_column_name = read_state<flex_string>("session_id");
-  data_params.feature_column_names =
-      std::vector<std::string>(features.begin(), features.end());
-  data_params.prediction_window = read_state<flex_int>("prediction_window");
-  data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
-  std::unique_ptr<data_iterator> data_it = create_iterator(data_params);
+  flex_list class_labels = read_state<flex_list>("classes");
+
+  std::unique_ptr<data_iterator> data_it =
+      create_iterator(data, class_labels, false);
```
nickjong: Huh, I guess we weren't passing the expected class labels before. Good catch!

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -492,28 +493,22 @@ void activity_classifier::init_train(
...
   // Bind the data to a data iterator.
-  data_iterator::parameters data_params;
-  data_params.data = data;
-  data_params.target_column_name = target_column_name;
-  data_params.session_id_column_name = session_id_column_name;
-  data_params.feature_column_names = feature_column_names;
-  data_params.prediction_window = read_state<flex_int>("prediction_window");
-  data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
-  training_data_iterator_ = create_iterator(data_params);
+  training_data_iterator_ = create_iterator(data, {}, true);
```
nickjong: Hmm, I notice this call to `create_iterator` is the only one that provides an empty list of class labels, and it's the only one that passes `true` for `is_train`. Here's an idea: what if `create_iterator` just took the `data` and `is_train` arguments, and its implementation looks as `is_train` to determine whether to populate the `class_labels` parameter from the model state. (It can also then just set the param struct's `verbose` flag with the value of `is_train`.)

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -220,15 +220,9 @@ gl_sarray activity_classifier::predict(gl_sframe data,
...
-  flex_list features = read_state<flex_list>("features");
-  data_iterator::parameters data_params;
-  data_params.data = data;
-  data_params.session_id_column_name = read_state<flex_string>("session_id");
-  data_params.feature_column_names =
-      std::vector<std::string>(features.begin(), features.end());
-  data_params.prediction_window = read_state<flex_int>("prediction_window");
-  data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
-  std::unique_ptr<data_iterator> data_it = create_iterator(data_params);
+  flex_list class_labels = read_state<flex_list>("classes");
```
nickjong: I don't think we need this anymore

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -271,16 +265,9 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
                                 "Expected one of: probability_vector, class");
   }
 
+  flex_list class_labels = read_state<flex_list>("classes");
```
nickjong: Or this one

PR #1770, opened 2019-04-24
Remove jinja2 dependency

znation:
Turns out we (apparently) aren't using it anywhere. The only place it
might be used is by sphinx, in the optional `make jinja2` for sphinx
docs, but I don't think we normally run this as part of our
documentation generation process.


znation:
See also https://github.com/apple/coremltools/pull/355

PR #1787, opened 2019-04-25
Add C-API 'compute_classifier_metrics'

### LINKED ISSUE #1782: C-API for evaluating classifier metrics directly from predictions
The older existing functions in this space only compute one metric at a time. It should be possible to compute all the default "auto" metrics for a typical classifier simultaneously.

nickjong:
Fixes #1782

### DIFF
--- a/src/unity/toolkits/evaluation/metrics.cpp
+++ b/src/unity/toolkits/evaluation/metrics.cpp
@@ -25,6 +25,75 @@
 namespace turi {
 namespace evaluation {
 
+namespace {
+
+std::vector<std::string> get_default_classifier_metrics() {
+  return {"accuracy", "auc", "precision", "recall", "f1_score", "log_loss",
+          "confusion_matrix", "roc_curve"};
+}
+
+std::vector<std::string> get_classifier_metrics(const std::string& metric) {
+  // Validate metric and determine list of metrics to compute.
+  std::vector<std::string> metrics = get_default_classifier_metrics();
+  if (metric != "auto") {
+    // If the caller didn't request "auto", then adjust the list of metrics.
+    if (metric == "report") {
+      // Add the per-class report to the standard list of metrics.
+      metrics.push_back("report_by_class");
+    } else {
+      // Just compute the requested metric, if valid.
+      if (std::find(metrics.begin(), metrics.end(), metric) == metrics.end()) {
+        log_and_throw("Unsupported metric " + metric);
+      } else {
+        metrics = {metric};
+      }
+    }
+  }
+  return metrics;
+}
+
+gl_sarray get_prediction_probability_vectors(const gl_sarray& predictions,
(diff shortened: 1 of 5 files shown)


src/unity/toolkits/evaluation/metrics.cpp
```diff
@@ -25,6 +25,75 @@
...
+      for (const auto& kv : dict) {
+        vec[class_to_index->at(kv.first)] = kv.second;
+      }
+      return vec;
+    };
+    result = predictions.apply(flatten, flex_type_enum::VECTOR);
+    break;
+  }
+  default:
+    log_and_throw("Could not convert predictions to probability vectors for "
```
shantanuchhabra: I think we're missing a `+` at the end of this line?

src/unity/toolkits/evaluation/metrics.cpp
```diff
@@ -145,7 +214,7 @@ variant_map_type compute_classifier_metrics_from_probability_vectors(
   auto new_metrics_end = std::remove(metrics.begin(), metrics.end(),
                                      "confusion_matrix");
   if (new_metrics_end != metrics.end()) {
-    // Don't expose get_evaluator_metric() to "report_by_class".
+    // Don't expose get_evaluator_metric() to "confusion_matrix".
```
nickjong: This is just opportunistically cleaning up a copy-and-paste error in the comments

src/unity/toolkits/evaluation/metrics.cpp
```diff
@@ -227,7 +296,7 @@ variant_map_type compute_classifier_metrics_from_probability_vectors(
     }
     return false;  // Never stop before all rows have been tallied.
   };
-  data.materialize_to_callback(callback);
+  data.materialize_to_callback(callback, turi::thread::cpu_count());
```
nickjong: This resolves a scary bug that somehow hadn't cropped up before: the evaluators default to allocating buffers to support `turi::thread::cpu_count` parallelism (== 8 on my MBP), but `gl_sframe::materialize_to_callback` defaults to `SFRAME_DEFAULT_NUM_SEGMENTS`, which for some reason was 16 during testing.

src/unity/toolkits/evaluation/metrics.cpp
nickjong: Actually, I don't believe `+` is defined for C string literals. However, if the compiler sees two C string literals separated only by whitespace, it will simply concatenate them.

See also: the third paragraph under "Notes" at https://en.cppreference.com/w/cpp/language/string_literal

shantanuchhabra: Got it, thanks!

berrylaw:
「Shreya Jain <notifications@github.com>」在 2019年4月26日 週五，下午2:02 寫道：

Issue #1785, opened 2019-04-25
Evaluating Model Takes a Long Time 


nseidl:
Hi,

**Why does it take almost twice as long to evaluate the model on 2k samples than it does to train on 17k samples?**

I'm using TuriCreate 5.4 (Python 3.6) on Ubuntu16.04 with Cuda8.0 (410.79). I have access to a machine with 4 GPUs and 80GB of RAM.

Here's my setup:

    pip install turicreate
    pip uninstall -y mxnet
    pip install mxnet-cu80==1.1.0
    echo "export LD_LIBRARY_PATH=/usr/local/cuda/lib64:$LD_LIBRARY_PATH" >> ~/.bash_profile
    source ~/.bash_profile

Here are config settings I've overriden:

    tc.config.set_runtime_config('TURI_FILEIO_MAXIMUM_CACHE_CAPACITY', 70*1024*1024*1024)
	tc.config.set_runtime_config('TURI_FILEIO_MAXIMUM_CACHE_CAPACITY_PER_FILE', 70*1024*1024*1024)
	tc.config.set_runtime_config('TURI_NUM_GPUS', 4)
	tc.config.set_runtime_config('TURI_DEFAULT_NUM_PYLAMBDA_WORKERS', 64)
	tc.config.set_num_gpus(4)

I train on ~17K .jpgs (4000x3000, in SFrame), validate on ~2k .jpgs, and evaluate on ~2k .jpgs.

	new_model = tc.image_classifier.create(
		train_sframe, # ~44GB
		'label',
		batch_size=128,
		max_iterations=num_iterations,
		model=base,
		validation_set=validation_sframe, # ~4.5GB
		verbose=True
	)

I then evaluate the model with the handy `.evaluate(...)`:

    evaluation = new_model.evaluate(test_sframe) # ~4.5GB

Here is a log of the progress. As you can see, it takes ~55min to train the model on 17k, and ~95min to evaluate the model on 2k.

    05:57:09 creating and training model
    06:53:25 done training model, took 3375.67470098s
    06:53:26 saving model
    06:53:39 evaluating model
    08:28:18 done evaluating model, took 5679.14030695s
    08:29:27 saving evaluation

**Why does it take almost twice as long to evaluate the model on 2k samples than it does to train on 17k samples?**

(I'd like to add that I'm not hitting any resource caps, as far as I know. See below screenshot of CPU and RAM usage (don't have GPU usage yet))

[...]

PR #1793, opened 2019-04-26
A Fix for the Evaluate Error in Image Classifier.

abhishekpratapa:
- partially fixes the #1785 Issue of `KeyError: 'predictions'`
- Still have the Issues around the user wanting multiple metrics and not wanting the output of the Evaluation Object.

- Maybe offload this work in a materialize function present on the returned evaluation object? This way unnecessary computation won't occur when the user only wants select metrics.


srikris:
I think the right fix is to move what you need to move to the evaluation code in C++. It should be single pass that way.

PR #1821, opened 2019-04-30
flex_image constructor from Boost GIL Image Types

### LINKED ISSUE #1828: Constructor in image_type from boost::gil::rgb8_image_t

shantanuchhabra:
This will be useful for the One Shot Object Detection functionality when we want to construct a `flex_image` from Boost GIL Image Types.

Only fixes #1828

### DIFF
--- a/src/image/image_type.cpp
+++ b/src/image/image_type.cpp
@@ -20,16 +20,16 @@ image_type::image_type(const char* image_data, size_t height, size_t width, size
   std::copy(image_data, image_data + image_data_size, &m_image_data[0]);
 }
 
-template<typename gil_image_type>
-image_type::image_type(const gil_image_type &gil_image)
+// template<typename gil_image_type>
+image_type::image_type(const boost::gil::rgb8_image_t &gil_image)
 : m_height(gil_image.height())
 , m_width(gil_image.width())
-, m_channels(boost::gil::num_channels<gil_image_type>())
-, m_image_data_size(gil_image.height() * gil_image.width() * boost::gil::num_channels<gil_image_type>())
+, m_channels(boost::gil::num_channels<boost::gil::rgb8_image_t>())
+, m_image_data_size(gil_image.height() * gil_image.width() * boost::gil::num_channels<boost::gil::rgb8_image_t>())
 , m_version(IMAGE_TYPE_CURRENT_VERSION)
 , m_format(Format::RAW_ARRAY)
 {
-  size_t image_data_size = gil_image.height() * gil_image.width() * boost::gil::num_channels<gil_image_type>();
+  size_t image_data_size = gil_image.height() * gil_image.width() * boost::gil::num_channels<boost::gil::rgb8_image_t>();
   auto it = const_view(gil_image).begin();
   const char* data = reinterpret_cast<const char*>(&boost::gil::at_c<0>(*it));
   m_image_data.reset(new char[image_data_size]);
--- a/src/image/image_type.cpp
+++ b/src/image/image_type.cpp
@@ -17,12 +17,12 @@ image_type::image_type(const boost::gil::rgb8_image_t &gil_image) {
   auto it = view(gil_image).begin();
   const uint8_t* data = reinterpret_cast<const uint8_t*>(&boost::gil::at_c<0>(*it));
   m_image_data.reset(new char[m_image_data_size]);
-  memcpy(&m_image_data[0], data, m_image_data_size);
+  std::copy(data, data + m_image_data_size, &m_image_data[0]);
 }
 
 image_type::image_type(const char* image_data, size_t height, size_t width, size_t channels, size_t image_data_size, int version, int format){
   m_image_data.reset(new char[image_data_size]);
-  memcpy(&m_image_data[0], image_data, image_data_size);
+  std::copy(data, data + image_data_size, &m_image_data[0]);
   m_height = height;
   m_width = width;
   m_channels = channels;
--- a/src/image/image_type.cpp
+++ b/src/image/image_type.cpp
@@ -20,15 +20,16 @@ image_type::image_type(const char* image_data, size_t height, size_t width, size
   std::copy(image_data, image_data + image_data_size, &m_image_data[0]);
 }
 
-image_type::image_type(const boost::gil::rgb8_image_t &gil_image)
+template<typename image_type>
+image_type::image_type(const image_type &gil_image)
 : m_height(gil_image.height())
 , m_width(gil_image.width())
-, m_channels(boost::gil::num_channels<boost::gil::rgb8_image_t>())
-, m_image_data_size(gil_image.height() * gil_image.width() * boost::gil::num_channels<boost::gil::rgb8_image_t>())
+, m_channels(boost::gil::num_channels<image_type>())
+, m_image_data_size(gil_image.height() * gil_image.width() * boost::gil::num_channels<image_type>())
 , m_version(IMAGE_TYPE_CURRENT_VERSION)
 , m_format(Format::RAW_ARRAY)
 {
-  size_t image_data_size = gil_image.height() * gil_image.width() * boost::gil::num_channels<boost::gil::rgb8_image_t>();
+  size_t image_data_size = gil_image.height() * gil_image.width() * boost::gil::num_channels<image_type>();
   auto it = const_view(gil_image).begin();
   const char* data = reinterpret_cast<const char*>(&boost::gil::at_c<0>(*it));
   m_image_data.reset(new char[image_data_size]);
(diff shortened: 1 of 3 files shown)


src/image/image_type.cpp
```diff
@@ -7,6 +7,19 @@
 
 namespace turi{
 
+image_type::image_type(boost::gil::rgb8_image_t gil_image) {
```
TobyRoseman: Can/should this parameter be const?

src/image/image_type.cpp
```diff
@@ -7,6 +7,19 @@
...
+  auto it = view(gil_image).begin();
+  auto data = (uint8_t*) &boost::gil::at_c<0>(*it);
```
hoytak: We should get in the habit of using the more proper and less error-prone C++ cast operators.  In this case, you want to use reinterpret_cast<const uint8_t*>(&boost::gil::at_c<0>(*it));

src/image/image_type.cpp
hoytak: I think this can be const boost::gil::rgb8_image_t& gil_image to avoid a copy.

src/image/image_type.cpp
hoytak: Also, move this down to where data is used.

src/image/image_type.cpp
```diff
@@ -7,6 +7,19 @@
 
 namespace turi{
 
+image_type::image_type(boost::gil::rgb8_image_t gil_image) {
+  auto it = view(gil_image).begin();
```
TobyRoseman: I really dislike the use of `auto` as a convenience for the writer. Code is read so much more often than it is written. I think usually its' usage doesn't make sense, there are of course exceptions.

You should probably get use rid of the usage of `auto` on this line and the next.

src/image/image_type.cpp
```diff
@@ -7,6 +7,19 @@
...
+  auto data = (uint8_t*) &boost::gil::at_c<0>(*it);
+  m_height = gil_image.height();
+  m_width = gil_image.width();
+  m_channels = boost::gil::num_channels<boost::gil::rgb8_image_t>();
+  m_image_data_size = m_height * m_width * m_channels;
+  m_version = IMAGE_TYPE_CURRENT_VERSION;
+  m_format = Format::RAW_ARRAY;
+  m_image_data.reset(new char[m_image_data_size]);
```
TobyRoseman: How is this memory ever released?

src/image/image_type.cpp
```diff
@@ -7,6 +7,19 @@
...
+  memcpy(&m_image_data[0], data, m_image_data_size);
```
hoytak: When possible, use the C++ std::copy(...) function instead of memcpy -- it's safer to use and does the correct thing for all types (memcpy may not).

src/image/image_type.hpp
```diff
@@ -50,6 +51,8 @@ class image_type {
   Format m_format = Format::UNDEFINED; 
   /// Constructor
   image_type() = default;
+  /// Construct from a Boost GIL Image
+  explicit image_type(boost::gil::rgb8_image_t boost_image);
```
hoytak: Yes, use const boost::gil::rgb8_image_t& boost_image

src/image/image_type.cpp
```diff
@@ -7,6 +7,19 @@
 
 namespace turi{
 
+image_type::image_type(boost::gil::rgb8_image_t gil_image) {
+  auto it = view(gil_image).begin();
+  auto data = (uint8_t*) &boost::gil::at_c<0>(*it);
+  m_height = gil_image.height();
```
hoytak: These can be moved to the initializers where possible -- do 
```
image_type::image_type(boost::gil::rgb8_image_t gil_image) 
: m_height(gil_image.height())
, m_width(gil_image.width())
...
{
/// Just the data copy here
```

src/image/image_type.hpp
```diff
@@ -9,6 +9,7 @@
 #include <string>
 #include <serialization/serialization_includes.hpp>
 #include <boost/shared_ptr.hpp>
+#include <boost/gil/gil_all.hpp>
```
hoytak: Make sure you add boost to the CMakeLists.txt requires list

src/image/image_type.cpp
hoytak: Eh, this auto is hard with iterator types.

src/image/image_type.cpp
hoytak: It's in a std::unique_ptr

TobyRoseman: Never mind: `m_image_data` is a `boost::shared_ptr`.

src/image/image_type.hpp
shantanuchhabra: Since we were already using <boost/shared_ptr.hpp>, wouldn't it already be there?

src/image/image_type.cpp
shantanuchhabra: Done, and changed the traditional flex_image constructor to use std::copy as well. Let me know if my usage is okay!

src/image/image_type.cpp
shantanuchhabra: I rearchitected the traditional image_type constructor to use initializers, and made the constructor from the boost::gil image just call that. How does that sound?

src/image/image_type.cpp
shantanuchhabra: Got rid of the auto!

TobyRoseman: Ok, that sounds like an exception. I was mainly talking about the one that was on the next line.

src/image/image_type.hpp
hoytak: Eh, check -- I don't think it's there now. 

With header only types like shared pointer, the consequence of not having it there is pretty minimal (essentially, running make in this directory after a clean configure will fail).  But if it's not header only (as I think the GIL is), there can be annoying linker errors down the road.

ylow: Can we avoid including boost gil in a header? This will slow down compilation massively. image_type.hpp is included in quite a lot of places.

ylow: Given the code we can just forward declare boost::gil::rgb8_image_t

shantanuchhabra: I also call `const_view` on the `rgb8_image_t` as well as `boost::gil::num_channels`. I'm assuming I forward declare those too?

znation: Boost is automatically added as a dependency to all other (non-external) targets. The `make_library` and `make_executable` macros do this.

shantanuchhabra: @hoytak @ylow @znation Got rid of the `boost/gil/gil_all` import by making the constructor a template function that can take multiple Boost GIL image types. Please let me know if that looks good!

shantanuchhabra: Hmm, looks like it fails the CI: https://travis-ci.com/apple/turicreate/jobs/197093292

Not really sure why...

shantanuchhabra: Pushed a fix for the build failure.

src/image/image_type.hpp
```diff
@@ -53,6 +53,10 @@ class image_type {
   /// Construct from existing data
   image_type(const char* image_data, size_t height, size_t width,
              size_t channels, size_t image_data_size, int version, int format);
+  /// Construct from a Boost GIL Image
+  /// Note: image_type can be any of the Boost GIL image types.
+  template<typename image_type>
+  explicit image_type(const image_type &gil_image);
```
hoytak: Use a different name for the template typename parameter from the class name, E.g. BoostImage.  That's the reason for your compile failure. 

However, the problem is that this will catch anything, including non-boost types, and then fail later.  The syntax for restricting this is a little convoluted, but here it would be: 

```
template<typename T>
   explicit image_type(const T &gil_image, 
typename std::enable_if<
std::is_same<T, boost::gil::image_type_1>::value || std::is_same<T, boost::gil::image_type_2>::value || (more in here)>::type* = 0) { 
```

shantanuchhabra: That would require forward declaring the boost::gil image types, but turns out that was more complicated than we thought because as @znation pointed out, the boost includes don't seem resilient to having their own types forward declared.

hoytak: What's the error when that happens?  That's odd and should not be the case.  Something else is going on -- likely that the header isn't included where it's needed.

shantanuchhabra: Sent the error to you on a private thread. What if I gracefully handle passing in a non Boost GIL image type in the CPP?

shantanuchhabra: @ylow and I just discussed this and he recommended I include the boost/gil/typedefs.hpp header which is basically like a forward declaration. Also got rid of the template since @ylow said it would be cleaner to just have overloaded constructors. Let me know if this looks good!

PR #1824, opened 2019-05-01
Improved the Image Classifier Evaluation Speed

abhishekpratapa:
- Improved the Image Classification Evaluation Speed

Screen Shot 2019-05-01 at 1 10 17 PM

Histogram of the time to evaluate 1000 images.


TobyRoseman:
What is the y-axis in the graph? Is it the number of classes?

I think removing `evaluation_result['hesitant_conf_mat']` and `evaluation_result['confidently_wrong_conf_mat']` makes sense.

I think we also need to refactor `get_confusion_matrix(...)` since it does two passes over the entire dataset for each class.

abhishekpratapa:
@TobyRoseman The y axis is the binned number of evaluations. I ran 100 evaluations per trial. Each color indicated 100 evaluations. In total I ran evaluation 300 times. The histogrammed values are shown above.

TobyRoseman:
> @TobyRoseman The y axis is the binned number of evaluations. I ran 100 evaluations per trial. Each color indicated 100 evaluations. In total I ran evaluation 300 times. The histogrammed values are shown above.

Good to know. How many classes where in the dataset?

The performance issue here is caused by the number of classes. If you run your experiment with a large number of classes, I would expect the difference before and after the regression to be greater. I would also expect this fix not to be as good as before the regression.

abhishekpratapa:
There are 10~ classes. I'll test with more in the meantime.

PR #1827, opened 2019-05-02
Constructing annotations based on the Parametrized Transform

### LINKED ISSUE #1775: Construct synthetic annotations using the parametrized transform matrix

shantanuchhabra:
Fixes #1775


src/unity/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -32,9 +32,11 @@ namespace one_shot_object_detection {
  */
 class ParameterSampler {
```
abhishekpratapa: Really depends on how picky you want me to be. I think it looks fine personally. Wondering is this only going to be used in this file? Is that the reason for not having this class in a header file?

shantanuchhabra: Fixed!

PR #1836, opened 2019-05-06
Always throw an exception on bad type conversion

### LINKED ISSUE #1835: Process crash with wrong flexible_type::get<>
Repro steps:

* Use the C API to call a function that expects (i.e.) an SArray of `dict`s, and pass in an SArray of `string`.

Expected:

An exception is thrown from C++, goes through the C API mechanism of error handling, and the function populates the tc_error pointer and returns.

Actual:

In debug:

The assert(false) is hit, and takes down the process immediately.

In release:

Weird and bad things happen because of an unsafe reinterpret_cast. Likely to crash in the immediate future after the assertion would've happened in debug.

Discussed with @hoytak, and we think the perf hit from potentially throwing on .get won't be too bad (and in the places where profiling says it's a perf hit, we should either plumb through an `unsafe_get` separately, or refactor to avoid calling `get` in a tight loop).

znation:
Instead of throwing in debug and crashing in release, let's always
throw. That way we won't take down the user's process when we are called
through the C or C++ API.

Fixes #1835

### DIFF
--- a/src/flexible_type/flexible_type.hpp
+++ b/src/flexible_type/flexible_type.hpp
@@ -14,18 +14,18 @@
 #include <typeindex>
 #include <parallel/atomic.hpp>
 #include <util/int128_types.hpp>
-//#define FLEX_TYPE_ASSERT DASSERT_TRUE
 #include <flexible_type/flexible_type_base_types.hpp>
 namespace turi {
 void flexible_type_fail(bool);
 }
 
+// Note: Uses the same assert behavior in both release and debug,
+// to ensure we never crash (always throw instead).
+// See https://github.com/apple/turicreate/issues/1835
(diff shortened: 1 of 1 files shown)


nickjong:
The typical motivation to make checks like these debug-only is to avoid the performance cost to these conditional checks. These queries are pretty low-level, so I worry that we're adding branch statements at the innermost loops in our code. Do we have any performance benchmarks to evaluate this kind of change, @hoytak?

znation:
@nickjong Yes, I spoke to @hoytak about this -- see the description on #1835. We have a plan to address perf degradations if they happen, but we don't expect the perf changes to be noticeable.

hoytak:
Let's get this in.

PR #1841, opened 2019-05-06
Fixing updating evaluation metrics after train in C-API Activity Classifier 

shreyajain17:
The accuracy and loss marginally drops after the last iteration because during training the accuracy is calculated "per window" while after training is done, it is calculated using "per row".

### DIFF
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -238,9 +238,8 @@ void activity_classifier::train(gl_sframe data, const std::string& target_column
 
   // Instantiate the training dependencies: data iterator, compute context,
   // backend NN model.
-  init_train(std::move(train_data), std::move(target_column_name),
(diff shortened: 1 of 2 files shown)


src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -225,10 +225,10 @@ void activity_classifier::init_table_printer(bool has_validation) {
   }
 }
 
-void activity_classifier::train(gl_sframe data, std::string target_column_name,
-                                std::string session_id_column_name,
+void activity_classifier::train(gl_sframe data, const std::string& target_column_name,
+                                const std::string& session_id_column_name,
                                 variant_type validation_data,
-                                std::map<std::string, flexible_type> opts)
+                                const std::map<std::string, flexible_type>& opts)
```
nickjong: Let's remove the `std::move` usage from these variables. Although they are now harmless, they are misleading.

PR #1843, opened 2019-05-06
Minor bugfixes and code cleanup

shantanuchhabra:
RGB images work end-to-end.


src/unity/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -87,7 +87,7 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
   
   Eigen::Matrix<float, 3, 3> mat = parameter_sampler.get_transform();
 
-  std::vector<Eigen::Vector3f> warped_corners = {
+  const std::vector<Eigen::Vector3f> &warped_corners = {
```
znation: Since this is stored on the stack, I think you actually want a concrete instance here (not a const reference). You could mark it const anyway (to prevent modification) but the storage should be explicit, I think.

(My thoughts -- I'm not sure what happens if you take a reference to a literal. Does it initialize it somewhere in memory (neither the stack nor heap) and point to it? Does it put it on the stack? Does it point to garbage? Does it depend on what compiler you're using?)

shantanuchhabra: Fixed! Thanks for pointing it out -- I had a feeling it would be undefined behavior, which would be dependent on the compiler then.

An interesting thread on a related note: https://softwareengineering.stackexchange.com/questions/224501/why-are-pointers-to-literals-not-possible

PR #1844, opened 2019-05-07
Sanitize recommender errors

shreyajain17:


src/unity/toolkits/recsys/recsys_model_base.cpp
```diff
@@ -163,7 +163,7 @@ void recsys_model_base::setup_and_train(
               << "Warning: Ignoring columns " << columns_ss.str() << ";"
               << std::endl;
           logprogress_stream
-              << "    To use one of these as a target column, set target = <column_name> "
+              << " Specify the column name to be used as target "
```
hoytak: This is conditional, so we should still say "To use one of these as a rating column, ..."

src/unity/toolkits/recsys/recsys_model_base.cpp
```diff
@@ -145,7 +145,7 @@ void recsys_model_base::setup_and_train(
         logprogress_stream << "Warning: " << "Column '"
                            << data.column_name(additional_columns[0])
                            << "' ignored." << std::endl;
-        logprogress_stream << "    To use this column as the target, set target = \""
+        logprogress_stream << " Specify the column name to be used as target "
```
hoytak: Let's still say "   To use this column as a rating column, ..."

PR #1846, opened 2019-05-07
Reporting row number on a draw_bounding_boxes crash

### LINKED ISSUE #1838: draw_bounding_boxes should provide debugging information when it fails.
tc.object_detector.util.draw_bounding_boxes, when run on an SFrame, should provide feedback on which image it failed on, if it did.

Inspired from #1837

shantanuchhabra:
Fixes #1838

### DIFF
--- a/src/unity/python/turicreate/toolkits/object_detector/util/_visualization.py
+++ b/src/unity/python/turicreate/toolkits/object_detector/util/_visualization.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import as _
 import turicreate as _tc
 import numpy as _np
 from turicreate.toolkits._internal_utils import _numeric_param_check_range
+from turicreate.toolkits._main import ToolkitError as _ToolkitError
 
 
 def _string_hash(s):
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/toolkits/object_detector/util/_visualization.py
```diff
@@ -128,24 +129,33 @@ def draw_bounding_boxes(images, annotations, confidence_threshold=0):
...
+                                        _width=image.shape[1],
+                                        _height=image.shape[0],
+                                        _channels=image.shape[2],
+                                        _format_enum=FORMAT_RAW,
+                                        _image_data_size=image.size)
+        except Exception as e:
+            if row_number == -1:
+                # indication that it was a single image and not an SFrame
+                raise _ToolkitError(e)
+            raise _ToolkitError("Received exception at the " + str(row_number) + "th row: " + e)
```
znation: I think it's better to say "at row n" than "at the nth row". Namely because it's not always `th`: the `2th row` sounds weird.

shantanuchhabra: Fixed!

PR #1855, opened 2019-05-08
Added constructor for flex_image from boost::gil::rgba8

### LINKED ISSUE #1829: Constructor in image_type from boost::gil::rgba8_image_t

shantanuchhabra:
Fixes #1829

### DIFF
--- a/src/image/image_type.cpp
+++ b/src/image/image_type.cpp
@@ -35,6 +35,27 @@ image_type::image_type(const boost::gil::rgb8_image_t &gil_image)
   std::copy(data, data + image_data_size, &m_image_data[0]);
 }
 
+image_type::image_type(const boost::gil::rgba8_image_t &gil_image)
+: m_height(gil_image.height())
+, m_width(gil_image.width())
+, m_channels(boost::gil::num_channels<boost::gil::rgba8_image_t>())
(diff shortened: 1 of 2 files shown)


src/image/image_type.cpp
```diff
@@ -35,6 +35,20 @@ image_type::image_type(const boost::gil::rgb8_image_t &gil_image)
   std::copy(data, data + image_data_size, &m_image_data[0]);
 }
 
+image_type::image_type(const boost::gil::rgba8_image_t &gil_image)
+: m_height(gil_image.height())
+, m_width(gil_image.width())
+, m_channels(boost::gil::num_channels<boost::gil::rgba8_image_t>())
```
znation: I think we are stretching the limits of our image type here. Perhaps we need to start keeping track of the (expected/assumed/explicit) byte layout, in addition to the number of channels? (I.e. "rgba8" vs. "4"). Otherwise, how are users of the image type supposed to know how to interpret the bytes? Right now we make some assumptions (1 channel = 8 bit grayscale, 3 channels = rgb8) but we might soon have some ambiguity here (if we don't already).

znation: For example, what happens if we add an `argb8` constructor? How will we differentiate the two `image_type` instances? Interpreting one as the other could give incorrect results.

PR #1861, opened 2019-05-10
Fix state updates after training in C-API Activity Classification

shreyajain17:


src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -256,38 +256,35 @@ void activity_classifier::train(gl_sframe data, const std::string& target_column
...
-      {{"training_precision", train_metric["precision"]},
-       {"training_recall", train_metric["recall"]},
-       {"training_accuracy", train_metric["accuracy"]},
-       {"training_log_loss", train_metric["log_loss"]},
-       {"training_confusion_matrix", train_metric["confusion_matrix"]},
-       {"training_number_of_examples", train_data.size()}});
+  for (auto &p : train_metric) {
+    state_update["training_" + p.first] = p.second;
+  }
+  state_update["training_number_of_examples"] = train_data.size();
```
nickjong: Let's skip `number_of_examples` for now, so we're more consistent with the `supervised_learning` toolkit. We could potentially add this metric in both places later

PR #1869, opened 2019-05-13
WIP: Docker TuriCreate

abhishekpratapa:
- Add Docker TuriCreate here.

### DIFF
--- a/deployment/app.py
+++ b/deployment/app.py
@@ -1,0 +1,312 @@
+# -*- coding: utf-8 -*-
+from flask import Flask, jsonify, send_file, request
+import turicreate as tc
+from enum import Enum
+from PIL import Image
+import numpy as np
+import argparse
+import datetime
+import zipfile
+import json
+
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+class ModelType(Enum):
+    recommender = 0
+    image_classifier = 1 #
+    drawing_classification = 2 #
+    sound_classification = 3
+    object_detection = 4 # -
+    style_transfer = 5 #
+    activity_classification = 6 #
+    image_similarity = 7
+    classifiers = 8
+    regression = 9
+    clustering = 10
+    text_classification = 11 #-
+
+def return_success(messages, success):
+    messages['success'] = success
+    return jsonify(messages)
+
+def serve_pil_image(pil_img):
+    img_io = StringIO()
+    pil_img.save(img_io, 'JPEG', quality=70)
+    img_io.seek(0)
+    return send_file(img_io, mimetype='image/jpeg')
+
+def ModelType2String(info):
+    info = info.copy()
+    if(info["type"] == ModelType.recommender):
+        info["type"] = "Recommender"
+        return info
+    if(info["type"] == ModelType.image_classifier):
+        info["type"] = "Image Classifier"
+        return info
+    if(info["type"] == ModelType.drawing_classification):
+        info["type"] = "Drawing Classifier"
+        return info
+    if(info["type"] == ModelType.sound_classification):
+        info["type"] = "Sound Classifier"
+        return info
+    if(info["type"] == ModelType.object_detection):
+        info["type"] = "Object Detection"
+        return info
+    if(info["type"] == ModelType.style_transfer):
+        info["type"] = "Style Transfer"
+        return info
+    if(info["type"] == ModelType.activity_classification):
+        info["type"] = "Activity Classifier"
+        return info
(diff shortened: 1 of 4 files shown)


deployment/app.py
```diff
@@ -0,0 +1,476 @@
...
+import datetime
+import zipfile
+import json
+
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+class ModelType(Enum):
```
znation: I suggest removing the ModelType enum, and the info dict. You should be able to streamline a lot of this logic by using the `model` object directly (which comes with all the same info as is in the `info` dict, including its type information).

deployment/app.py
```diff
@@ -0,0 +1,476 @@
...
+             info["type"] == ModelType.item_content_recommender or \
+             info["type"] == ModelType.popularity_recommender or \
+             info["type"] == ModelType.factorization_recommender:
+                return return_success({"Recommendation" : str(prediction[0])})
+            if info["type"] == ModelType.kmeans:
+                return return_success({"Cluster" : str(prediction[0])})
+            return return_success({"Label": str(prediction[0])}, True)
+
+        elif info["type"] == ModelType.sound_classification:
+            # TODO
```
znation: You could let it fall into the else case below for now (an "Unimplemented Toolkit" error seems about as good a TODO as any).

deployment/app.py
```diff
@@ -0,0 +1,476 @@
...
+        elif info["type"] == ModelType.sound_classification:
+            # TODO 
+
+
+        else:
+            return return_success({"error": "Unimplemented Toolkit"}, False)
+    return return_success({}, True)
+
+if __name__ == '__main__':
+    app.run(debug=True, host='0.0.0.0')
```
znation: You should take `--debug` as a command line flag to the app, and pass it through here.

deployment/app.py
```diff
@@ -0,0 +1,476 @@
...
+             info["type"] == ModelType.boosted_tree_regression:
+                return return_success({"Estimated Value" : str(prediction[0])})
+            if info["type"] == ModelType.item_similarity_recommender or \
+             info["type"] == ModelType.item_content_recommender or \
+             info["type"] == ModelType.popularity_recommender or \
+             info["type"] == ModelType.factorization_recommender:
+                return return_success({"Recommendation" : str(prediction[0])})
+            if info["type"] == ModelType.kmeans:
+                return return_success({"Cluster" : str(prediction[0])})
+            return return_success({"Label": str(prediction[0])}, True)
```
znation: I think you might want to return more info to the caller (i.e. the full output from the model, including probabilities if possible, not just the top prediction). Various apps will have various uses for that info, and it isn't prohibitively costly to send all of it.

znation: Or, make it a parameter to the API.

deployment/app.py
```diff
@@ -0,0 +1,389 @@
+# -*- coding: utf-8 -*-
```
TobyRoseman: copy right header

deployment/app.py
```diff
@@ -0,0 +1,389 @@
...
+@app.route('/info')
+def model_info():
+    result = dict()
+    result["model"] = str(type(model))
+    for field in model._list_fields():
+        result[field] = model._get(field) if json_serializable(model._get(field)) else str(model._get(field))
+    return return_success(result, True)
+
+@app.route('/predict', methods=['GET', 'POST'])
+def model_predict():
```
TobyRoseman: I'm having a really hard time understanding this function. It's too long, but more importantly there are too many levels of nesting. I think it will be difficult for other developers to make changes here without introducing bugs.

At a high level what you're trying to do here is map toolkits to result processing logic. I would recommend creating a separate function for each top level code block. 

Also rather than lots of complicated top level if/elif-statements, it's probably cleaner to do that mapping with a dictionary. 

To make it cleaner, I would recommend restructuring the top level function to look something like this:
```python
@app.route('/predict', methods=['GET', 'POST'])
def model_predict():
    TOOLKIT_TO_PROCESSING_FUNCTION = {
        tc.toolkits.image_classifier.image_classifier.ImageClassifier: processing1,
        tc.toolkits.style_transfer.style_transfer.StyleTransfer: processing2,
        ......
        ......
    }

    model_type = type(model)
    if model_type not in TOOLKIT_TO_PROCESSING_FUNCTION:
        return_success({"error": "Unimplemented Toolkit"}, False)

    result = TOOLKIT_TO_PROCESSING_FUNCTION[model_type]()
    return_success(result, True)
}

PR #1872, opened 2019-05-14
Masks for RGB images are now encoded as the alpha channel.

### LINKED ISSUE #1859: More efficient Mask for superposition
For RGB images, the mask used need not be an RGB image. The mask can be a more space and time-efficient data structure. Maybe an unordered set of coordinates that are in the quadrilateral? Maybe a 0/1 bitmap?

shantanuchhabra:
Fixes #1859


src/unity/toolkits/object_detection/one_shot_object_detection/util/quadrilateral_geometry.cpp
```diff
@@ -6,16 +6,13 @@
 
 #include <unity/toolkits/object_detection/one_shot_object_detection/util/quadrilateral_geometry.hpp>
 
-static const boost::gil::rgb8_pixel_t BLACK(0, 0, 0);
-static const boost::gil::rgb8_pixel_t WHITE(255, 255, 255);
+static const size_t TRANSPARENT(0);
+static const size_t OPAQUE(255);
```
nickjong: Nit: for simple types like size_t, you should be able to use `constexpr` (making these essentially compile-time constants with no allocated memory at runtime), and to use the clearer `=` syntax, e.g.
```
static constexpr size_t TRANSPARENT = 0;
```

src/unity/toolkits/object_detection/one_shot_object_detection/util/superposition.cpp
```diff
@@ -3,119 +3,99 @@
...
-  boost::gil::rgb8_image_t masked(boost::gil::rgb8_image_t::point_t(background_view.dimensions()));
-  fill_pixels(view(masked), WHITE);
-  superimpose_image_rgba(view(masked), view(transformed), 
-                         view(background_rgba));
+  boost::gil::rgb8_image_t masked(boost::gil::rgba8_image_t::point_t(background_view.dimensions()));
+  fill_pixels(view(masked), RGB_WHITE);
+  if (object.m_channels == 4) {
+    superimpose_rgba_object_image(parameter_sampler, object, view(masked), view(transformed), view(background_rgba));
+  } else {
+    superimpose_rgb_object_image(parameter_sampler, object, view(masked), view(transformed), view(background_rgba));
```
nickjong: Could we combine `superimpose_rgb_object_image` and `superimpose_rgba_object_image` into one function that introspects the format of the `flex_image` internally?

shantanuchhabra: That was my thought behind `create_synthetic_image` which is the only function exposed in `superposition.hpp`. `superimpose_rgb_object_image` and `superimpose_rgba_object_image` are meant to be helper functions for `create_synthetic_image`.

src/unity/toolkits/object_detection/one_shot_object_detection/util/quadrilateral_geometry.cpp
shantanuchhabra: Fixed!

src/unity/toolkits/object_detection/one_shot_object_detection/util/superposition.cpp
nickjong: The two functions seem to share much of their logic. I guess what I don't understand is why the RGB case (for the user's object image) can't be handled just by converting from RGB to RGBA at the very beginning, and then having everything else be the same code path.

Boost GIL might even have some way of producing an RGBA view over an RGB image

shantanuchhabra: Fixed.

src/unity/toolkits/object_detection/one_shot_object_detection/util/superposition.cpp
```diff
@@ -39,44 +41,43 @@ void superimpose_image(const boost::gil::rgb8_image_t::view_t &masked,
...
-    object.m_width,
-    object.m_height,
-    (boost::gil::rgba8_pixel_t*) (object.get_image_data()),
-    object.m_channels * object.m_width // row length in bytes
-  );
-  resample_pixels(starter_image_view, transformed, M, boost::gil::bilinear_sampler());
-  superimpose_image(masked, transformed, background);
+  resample_pixels(view(starter_image), transformed, M, boost::gil::bilinear_sampler());
+  if (object.m_channels == 3) {
+    color_quadrilateral(transformed, parameter_sampler.get_warped_corners());
```
nickjong: What does this do, and why is it only needed for RGB input?

src/unity/toolkits/object_detection/one_shot_object_detection/util/mapping_function.hpp
```diff
@@ -44,7 +44,7 @@ boost::gil::point2<F> transform(const Eigen::Matrix<F, 3, 3>& mat, const boost::
 
 template <typename F>
 struct mapping_traits<Eigen::Matrix<F, 3, 3> >{
-    typedef boost::gil::point2<F> result_type;
+  typedef boost::gil::point2<F> result_type;
```
nickjong: In new code, let's use using declarations instead of typedef.

See also: https://google.github.io/styleguide/cppguide.html#Aliases

src/unity/toolkits/object_detection/one_shot_object_detection/util/superposition.cpp
shantanuchhabra: Fixed after talking in person. Summarizing here. Turns out we need color_quadrilateral with bilinear sampling for RGBA images as well. After timing the nearest_neighbor sampling against bilinear sampling along with color_quadrilateral, the latter approach was faster and had better quality.

src/unity/toolkits/object_detection/one_shot_object_detection/util/mapping_function.hpp
shantanuchhabra: Fixed.

src/unity/toolkits/object_detection/one_shot_object_detection/util/superposition.cpp
```diff
@@ -42,41 +42,27 @@ void superimpose_image(const boost::gil::rgb8_image_t::view_t &superimposed,
...
+    (boost::gil::rgba8_pixel_t*) (object.get_image_data()),
+    object.m_channels * object.m_width // row length in bytes
+  );
   Eigen::Matrix<float, 3, 3> M = parameter_sampler.get_transform().inverse();
-  resample_pixels(view(starter_image), transformed, M, boost::gil::bilinear_sampler());
-  if (object.m_channels == 3) {
-    color_quadrilateral(transformed, parameter_sampler.get_warped_corners());
-  }
+  resample_pixels(starter_image_view, transformed, M, boost::gil::bilinear_sampler());
+  color_quadrilateral(transformed, parameter_sampler.get_warped_corners());
```
nickjong: Might be worth a comment as to why this is needed (to guard against artifacts in the alpha channel due to transformations?)

shantanuchhabra: Done!

PR #1877, opened 2019-05-15
Update Linux build requirements

### LINKED ISSUE #1669: Build dependencies on Centos 6 are partial
The build instructions for Linux: https://github.com/apple/turicreate/blob/master/BUILD.md#build-dependencies

mention some,  but not all of the dependencies.

I build on a Docker image of `centos:6.9`, thus I begin with a very clean system, which I believe is the reason why I uncovered some unmentioned dependencies.

Since I'm on centos, I use the `scl` collections `devtoolset-7` and `python27` to get the build tools.

Apart from that, I installed the dependencies from the `BUILD.md` page - and still failed. It seems the following were missing:

* `git`  # as the build process git clones things,
* `patch`
* `rsync`
* `libpng` and `libpng-devel`

I use this script for my build

```bash
#!/bin/bash
# This was done in a docker container of centos 6.9 :
# First, click on the Docker icon in the top-right corner and choose Preferences.
# In the Advanced tab, raise the CPUs to 4, Memory to 8GB, and the Swap to 2GB.
# Then:
# docker run --init --rm -it centos:6.9
# In general I followed the official instructions from
# https://github.com/apple/turicreate/blob/master/LINUX_INSTALL.md#centos-6

yum install -y centos-release-scl
yum install -y python27
# https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7/
yum install -y devtoolset-7

# https://stackoverflow.com/a/30365810/2064085
# Don't use `scl enable` which stops running the next commands in scripts where it's used.
for env in $(scl --list); do source "/opt/rh/${env}/enable"; done

python2.7 -c "import sys;assert sys.maxunicode == 1114111"
pip2.7 install --upgrade pip  # or just pip?

yum install -y blas lapack blas-devel lapack-devel wget nano git cmake patch

cd /
wget https://www.samba.org/ftp/ccache/ccache-3.5.1a.tar.gz
tar -xvf ccache-3.5.1a.tar.gz
cd ccache-3.5.1
./configure && make && make install
cd /

cd /
git clone https://github.com/ninja-build/ninja.git
cd ninja
./configure.py --bootstrap
export PATH=$(pwd):$PATH
export GENERATOR="-G Ninja"
cd /

# Assert pyfpe.h exists
file /opt/rh/python27/root/usr/include/python2.7/pyfpe.h

# gif_lib.h
yum install -y giflib-devel

# xxd
yum install -y vim-common

# zlib
yum install -y zlib zlib-devel

# visualization
yum install -y libpng libpng-devel
yum install -y rsync
yum install -y libX11-devel

REPO_DIR=/turicreate

if [[ ! -d "${REPO_DIR}" ]]; then

    mkdir -v -p "${REPO_DIR}"
    cd "${REPO_DIR}"
    git init

    git fetch --depth 1 https://github.com/apple/turicreate.git master:refs/remotes/origin/master
    git checkout master
    cd /
fi

cd "${REPO_DIR}"

# https://stackoverflow.com/a/48241814/2064085
# export PYTHON_INCLUDE_DIR=/usr/include/python2.7
# export PYTHON_LIBRARY=/usr/lib/python2.7/config/libpython2.7.so
./configure \
    -D PYTHON_INCLUDE_DIR=$(python -c "from distutils.sysconfig import get_python_inc; print(get_python_inc())")  \
    -D PYTHON_LIBRARY=$(python -c "import distutils.sysconfig as sysconfig; print(sysconfig.get_config_var('LIBDIR'))") \
    -D PYTHON_EXECUTABLE:FILEPATH=$(which python)

cd debug
ninja -j4
```

znation:
* Clarifies required compiler toolchain minimum versions.
* Adds some dependencies that aren't installed by default on some Linux
  distributions (but are required).

Fixes #1669

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -39,7 +39,10 @@ You will need:
 
 * On macOS, [Xcode](https://itunes.apple.com/us/app/xcode/id497799835) with command line tools (tested with Xcode 9 and later)
 * On Linux:
-  * A C++ compiler toolchain with C++11 support (provided by `build-essential` on Ubuntu)
+  * A C++ compiler toolchain with C++11 support (`gcc` >= 4.8 or `clang` >= 3.3)
+  * `git` (typically provided by the `git` package)
+  * `patch` (typically provided by `patch` package)
+  * `rsync` (typically provided by `rsync` package)
   * blas and lapack development libraries (typically provided by the `liblapack-dev` package)
   * `pyfpe.h` (typically provided by the `libpython2.7-dev` package for Python 2.7, or `libpython3.6-dev` for Python 3.6)
   * `gif_lib.h` (typically provided by the `libgif-dev` package)


BUILD.md
```diff
@@ -39,7 +39,10 @@ You will need:
 
 * On macOS, [Xcode](https://itunes.apple.com/us/app/xcode/id497799835) with command line tools (tested with Xcode 9 and later)
 * On Linux:
-  * A C++ compiler toolchain with C++11 support (provided by `build-essential` on Ubuntu)
+  * A C++ compiler toolchain with C++11 support (`gcc` >= 4.8 or `clang` >= 3.3)
+  * `git` (typically provided by the `git` package)
+  * `patch` (typically provided by `patch` package)
```
TobyRoseman: Just curious, do you know what this is used for?

znation: It's used to apply [this patch](https://github.com/apple/turicreate/blob/master/deps/patches/boost_date_time_parsers3.patch) to Boost to fix date/time parsing. There's another .patch in that directory, but I don't think that one is actually used.

PR #1881, opened 2019-05-15
C-API of Activity Classifier fixes error in prediction

### LINKED ISSUE #1880: C-API of Activity Classifier gives error when predict is called without giving class labels column
model2.predict(sf)

<ipython-input-8-fe6e1ea9e4d1> in <module>()
----> 1 model2.predict(sf)

/Users/shreyajain/turicreate/debug/src/unity/python/turicreate/extensions.pyc in <lambda>(*args, **kwargs)
    288         elif name in self._functions:
    289             # is it a function?
--> 290             ret = lambda *args, **kwargs: self.__run_class_function(name, args, kwargs)
    291             ret.__doc__ = "Name: " + name + "\nParameters: " + str(self._functions[name]) + "\n"
    292             try:

/Users/shreyajain/turicreate/debug/src/unity/python/turicreate/extensions.pyc in __run_class_function(self, fnname, args, kwargs)
    275         except RuntimeError as exc:
    276             # Expose C++ exceptions using ToolkitError.
--> 277             raise _ToolkitError(exc)
    278         ret = _wrap_function_return(ret)
    279         return ret

ToolkitError: Column name activity does not exist.

shreyajain17:
fixes #1880

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,1 @@
-
-
 BasedOnStyle: Google
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+
+
+BasedOnStyle: Google
--- a/src/unity/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/unity/toolkits/activity_classification/activity_classifier.cpp
@@ -332,7 +332,8 @@ gl_sarray activity_classifier::predict(gl_sframe data,
   }
 
   // Bind the data to a data iterator.
-  std::unique_ptr<data_iterator> data_it = create_iterator(data, false, false);
+  std::unique_ptr<data_iterator> data_it =
+      create_iterator(data, /* requires_labels */ false, /* is_train */ false);
 
   // Accumulate the class probabilities for each prediction window.
   gl_sframe raw_preds_per_window = perform_inference(data_it.get());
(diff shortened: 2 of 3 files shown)


nickjong:
@hoytak @znation Note that with this PR, we're introducing a `.clang-format` file. For now, we're just starting with "Google" style, since I think all of us have indicated being okay with that one

.clang-format
```diff
@@ -0,0 +1,3 @@
+
+
+BasedOnStyle: Google
```
nickjong: Super nit: we don't need the two blank lines before this, do we?

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -332,7 +332,7 @@ gl_sarray activity_classifier::predict(gl_sframe data,
   }
 
   // Bind the data to a data iterator.
-  std::unique_ptr<data_iterator> data_it = create_iterator(data, false);
+  std::unique_ptr<data_iterator> data_it = create_iterator(data, false, false);
```
nickjong: Nit: for boolean args, I find it really helps readability to add an in-line comment indicating what the argument name is. For example:
```
std::unique_ptr<data_iterator> data_it = create_iterator(data, /* requires_labels */ false, /* is_train */ false);
```

PR #1887, opened 2019-05-16
GPU Setup for MPS

abhishekpratapa:
- The GPU Setup for MPS using the drawing classifier as an example.

### DIFF
--- a/src/unity/python/turicreate/test/test_drawing_classifier.py
+++ b/src/unity/python/turicreate/test/test_drawing_classifier.py
@@ -323,3 +323,7 @@ class DrawingClassifierUsingQuickdraw245(DrawingClassifierTest):
         super(DrawingClassifierUsingQuickdraw245, self).setUpClass(
             warm_start="quickdraw_245_v0")
 
+class DrawingClassifierMPS(unittest.TestCase):
+    def test_cpp(self):
+        
+
--- a/src/unity/toolkits/drawing_classifier/drawing_classifier.mm
+++ b/src/unity/toolkits/drawing_classifier/drawing_classifier.mm
@@ -5,7 +5,7 @@ namespace turi {
 namespace drawing_classifier {
 
 void init_model(std::map<std::string, std::vector<float>> weights) {
-
+	
 }
 
 } // namespace drawing_classifier
--- a/src/unity/toolkits/drawing_classifier/drawing_classifier.mm
+++ b/src/unity/toolkits/drawing_classifier/drawing_classifier.mm
@@ -1,0 +1,12 @@
+#include <unity/toolkits/drawing_classifier/drawing_classifier.hpp>
+#include <unity/toolkits/drawing_classifier/mps_drawing_classifier.hpp>
+
+namespace turi {
+namespace drawing_classifier {
+
+void init_model(std::map<std::string, std::vector<float>> weights) {
+
+}
+
+} // namespace drawing_classifier
+} // namespace turi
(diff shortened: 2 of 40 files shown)


abhishekpratapa:
#1888

src/unity/python/turicreate/test/test_drawing_classifier.py
```diff
@@ -323,3 +323,7 @@ def setUpClass(self):
         super(DrawingClassifierUsingQuickdraw245, self).setUpClass(
             warm_start="quickdraw_245_v0")
 
+class DrawingClassifierMPS(unittest.TestCase):
+    def test_cpp(self):
```
TobyRoseman: So this is a syntax error.

Also please add unit tests.

src/unity/toolkits/sound_classifier/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+project(unity_toolkits)
```
TobyRoseman: Let's not worry about adding anything for the sound classifier to this review.

src/unity/toolkits/sound_classifier/sound_classifier.cpp
```diff
@@ -0,0 +1 @@
+/* TODO: sound classifier class */
```
TobyRoseman: Also don't worry about this empty file.

src/unity/toolkits/drawing_classifier/drawing_classifier.mm
```diff
@@ -0,0 +1,12 @@
+#include <unity/toolkits/drawing_classifier/drawing_classifier.hpp>
+#include <unity/toolkits/drawing_classifier/mps_drawing_classifier.hpp>
+
+namespace turi {
+namespace drawing_classifier {
+
+void init_model(std::map<std::string, std::vector<float>> weights) {
+
```
TobyRoseman: I'm confused. Seems like there should be a lot more here before it's even code complete.

src/unity/toolkits/drawing_classifier/mps_drawing_classifier.mm
```diff
@@ -0,0 +1,16 @@
...
+- (id _Nullable)initWithParameters:(NSString *_Nonnull)name
+                         inputNode:(MPSNNImageNode *_Nonnull)inputNode
+                            device:(id<MTLDevice> _Nonnull)dev
+                         cmd_queue:(id<MTLCommandQueue> _Nonnull)cmd_q
+                       initWeights:(std::map<std::string, std::vector<float>>)weights {
+  @autoreleasepool {
+    self = [super init];
+    return self;
+  }
+}
```
TobyRoseman: Also seems like this is missing a lot. I would expect to see the actual network definition here, but I'm not seeing that any where.

abhishekpratapa:
> We talked about removing the classes which just wrap MPS layers, as these don't really add much. Are you still planning to remove these? (To be clear: I do agree we need the padding and data loading classes.)

Yes, I hadn't got to the refactor!

src/unity/toolkits/drawing_classifier/mps_drawing_classifier.mm
abhishekpratapa: I forgot to push a commit.

abhishekpratapa: I pushed it's under the commit history but it's not showing up here. Hmm.

nickjong:
After syncing with Abhishek offline, he'll take a pass at revising/refactoring:

1. (Minor) run `clang-format` to make some of the ObjC code stuff more readable
2. Make `mps_drawing_classifier`'s interface close to the `model_backend` interface (but in ObjC)
3. Revisit the question of wrapping each MPS node in its own layer class, with an eye towards making the actual graph definition as readable as possible

PR #1891, opened 2019-05-17
Bug fixes

### LINKED ISSUE #1889: Regression in TC image classifier evaluate
The evaluation from 5.4 regressed on `confusion_matrix` and `roc_curve` which are missing from the evaluation dictionary.

The evaluation code path is also missing from the unit tests
https://github.com/apple/turicreate/blob/master/src/unity/python/turicreate/test/test_image_classifier.py

### LINKED ISSUE #1890: Import broken in 5.4 on linux GPUs for OD
Due to the lazy import of MXNet, on GPUs, object detection is broken with 5.4.

https://github.com/apple/turicreate/search?q=_print_neural_compute_device&unscoped_q=_print_neural_compute_device

```
PIP: package/turicreate/toolkits/_internal_utils.py, line 724 should be:
 from . Import _mxnet
And line 728 should be:
 _mxnet._mxnet_utils...
```

TobyRoseman:
fixes #1890
fixes #1889


abhishekpratapa:
Yes, that's the fix! Looks good!

PR #1895, opened 2019-05-19
Python API rough draft + minor input type checks

### LINKED ISSUE #1799: Finalize API design for One Shot Object Detection

shantanuchhabra:
Fixes #1799


src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+        The number of training iterations. If 0, then it will be automatically
+        be determined based on the amount of data you provide.
+
+    verbose : bool optional
+        If True, print progress updates and model details.
+
+    **kwargs : dict optional
+        A dictionary of advanced parameters.
+        Here are the parameters currently supported:
+        - yaw  : rotation of the logo (or any 2-D starter image) along Y axis
```
TobyRoseman: I think you want to say 'image' here not 'logo'.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+            ##############################      ^
+            ########## ________ ##########      |
+            ##########|  logo  |##########      |
+            ##########|________|##########      |-------> x (pitch)
+            ##############################     /
+            ##############################    /
+            ##############################   /
+            ##############################  z (roll)
+
+        - blur: A boolean flag indicating whether augmented images should be
```
TobyRoseman: Maybe `blur` and `lighting_perturbation` should be an ints rather than a bools. Does it really have to be all or nothing?

src/unity/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
```diff
@@ -0,0 +1,21 @@
...
+def preview_augmented_images(data, target, n=1):
+    """
+    A utility function that would allow the user to visualize 
+    `n` augmented images per row of the data if data is an SFrame,
+    and just one augmented image if the data is an Image.
+
+    Parameters
+    ----------
+
+    dataset : SFrame | tc.Image
```
TobyRoseman: You're missing some pretty standard text here.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+        any padding.
+        RGB and RGBA images allowed.
+
+    target : string
+        The target column name in the SFrame that contains all the labels. 
+        If the dataset is a single tc.Image, this parameter is just a label for
+        that image.
+
+    augmentation_mode : string optional
+        An augmentation mode is a shortcut to setting the 'flip_horizontal',
```
TobyRoseman: So is this parameter ignored if you pass in values for `kwargs`? Do `kwargs` take precedence over this parameter?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+        An SFrame that contains all the starter images along with their
+        corresponding labels.
+        If the dataset is a single tc.Image, this parameter is just that image.
+        Every starter image should entirely be just the starter image without
+        any padding.
+        RGB and RGBA images allowed.
+
+    target : string
+        The target column name in the SFrame that contains all the labels. 
+        If the dataset is a single tc.Image, this parameter is just a label for
```
nickjong: It seems strange to me to allow this parameter to be either a column name or a class label. Is there precedent for this elsewhere in Turi Create?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
-    model = _extensions.one_shot_object_detector()
+           verbose=True,
+           **kwargs):
+    """
+    Create a :class:`ObjectDetector` model.
+
+    Parameters
+    ----------
+
+    dataset : SFrame | tc.Image
```
nickjong: If an SFrame is provided, how does the toolkit know which column contains the image to use?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+        An SFrame that contains all the starter images along with their
+        corresponding labels.
+        If the dataset is a single tc.Image, this parameter is just that image.
+        Every starter image should entirely be just the starter image without
+        any padding.
+        RGB and RGBA images allowed.
```
nickjong: Why not support grayscale images too?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+        - 'auto'
+        - 'logo'
+        - 'road_sign'
+        - 'playing_card'
+        - 'chart'
+        TBD/TODO: Specify the different pre-set parameter values each of these
+        modes corresponds to, probably as a table.
+        If 'dataset' is an SFrame, the augmentation_mode specified as this
+        parameter will be applied to all the images provided in the SFrame.
+        If you'd like different images in the 'dataset' SFrame to have different
```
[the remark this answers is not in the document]
nickjong: Yes, please do not make the semantics of parameter names so context dependent. Pythonic code, as I understand it, attempts to be type-agnostic, not semantics-agnostic.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+        augmentation_mode's, those modes can be specified as a column in the
+        SFrame, and that column name should be passed in to the
+        'augmentation_mode' parameter.
+
+    flip_horizontal : bool optional
+        Boolean flag that indicates whether the object can be flipped
+        horizontally,
+        i.e. whether a water image of the object should appear can be in the
```
nickjong: What does "water image" mean?

Note that the OD toolkit has an advanced parameter `aug_horizontal_flip`, which controls whether data augmentation can reflect the image horizontally, across the vertical line passing through the image's center. Are you giving this parameter the opposite meaning?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
nickjong: I think we should think twice before publishing API for these kinds of augmentations that occur after the superimposition of the user image over the background. What happens when we expose the augmentation options in OD? How would these interact?

Note that OD users can already manipulate e.g. jitter in hue, brightness, and saturation, as well as noise, via the `kwargs` argument of `turicreate.object_detector.create`.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -6,64 +6,196 @@
...
+                                        batch_size=batch_size,
+                                        max_iterations=max_iterations,
+                                        verbose=verbose)
+    return OneShotObjectDetector(model.__proxy__)
 
 
-class OneShotObjectDetector(_CustomModel):
-    _PYTHON_ONE_SHOT_OBJECT_DETECTOR_VERSION = 1
-    
+class OneShotObjectDetector(_ObjectDetector):
```
nickjong: I thought we decided that the new class should _not_ inherit from the existing `ObjectDetector` type. We don't want to blindly forward the entire API surface.

src/unity/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Apple Inc. All rights reserved.
+#
+# Use of this source code is governed by a BSD-3-clause license that can
+# be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+
+def preview_augmented_images(data, target, n=1):
```
nickjong: Probably this should share a lot of the same interface with the `create` method.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
TobyRoseman: I didn't know those things could be manipulated via the `kwargs` of argument of `turicreate.object_detector.create`. Looks like we haven't documented that: https://apple.github.io/turicreate/docs/api/generated/turicreate.object_detector.create.html#turicreate.object_detector.create

I created #1901 to track this issue.

nickjong: Yeah, I suspect the intent was for it to be "hidden" API. Before we promoted the `batch_size` parameter to the documented API, for example, we did tell some users how to manipulate the batch size in this manner.

Ultimately, we should decide on some consistent API for data augmentation that fits well with the relationship between OD and one-shot OD. In the absence of that design, I think we should err on the side of publishing the minimal API that serves users

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
shantanuchhabra: Sorry, I was under the impression that the new class should inherit from the existing ObjectDetector, so that the model we get behaves like an Object Detector. How would we ensure that the model we get out of `.create` is an `ObjectDetector` and not a `OneShotObjectDetector`?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
shantanuchhabra: I meant `flip_horizontal` to mean whether data augmentation can reflect the image vertically, along the horizontal line passing through the image's center. In other words, whether the user would like to see a _"water image" (reflection in water)_  of the image passed in, in the augmented data.

         o 
          |
         / \
----------------- water
         \ /
          |
         o

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
shantanuchhabra: Anything passed into kwargs takes precedence. However, it's slightly different from completely ignoring the parameter. Consider a scenario where the user passes in `augmentation_mode="logo"`, but also passes in `flip_horizontal=True`. All the pre-set values that come from `augmentation_mode="logo"` still stand, except for `flip_horizontal`.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
shantanuchhabra: Let's table exposing blur, color_perturbation, and noise until we have a story around how they interact with OD as well? Does that sound good?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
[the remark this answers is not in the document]
shantanuchhabra: Okay, to not make the semantics of parameter names context-dependent, we'll make the parameter names column names entirely. How does that sound? It seems more generic. If someone wants to use the same value for all images in their SFrame, they can just create a column with that value repeated.

shantanuchhabra:
@nickjong @TobyRoseman Open for review again!

PR #1905, opened 2019-05-21
WIP: Automatically splitting data into train and validation for C-API Object Detector

shreyajain17:


shreyajain17:
@nickjong I cannot figure out how to write init_train_impl to do export weights and then update params. That gives an error and I wasn't able to figure it out.

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -511,7 +565,81 @@ BOOST_AUTO_TEST_CASE(test_object_detector_train) {
...
+                             model_spec::padding_type::SAME,
+                             /* weight_init_fn */ [](float*w , float* w_end) {
+                               for (int i = 0; i < w_end - w; ++i) {
+                                 w[i] = static_cast<float>(i);
+                               }
+                             });
+
+    float_array_map trained_weights = model.export_weights();
     
+    nn_spec->update_params(trained_weights);
```
nickjong: It looks like you're trying to address the fact that `object_detector::train` depends on `nn_spec_` being non-`NULL` (as well as `training_model_`), after the call to `init_train`. So we just need to set these data members manually in `init_train_impl`. Unfortunately, to this end we should probably add protected setters to the base class, e.g. `set_nn_spec` and `set_training_model`... for now. (In the longer term, I think the class can be refactored a bit, so that e.g. `init_train` returns all the things that `init_train` actually initializes.)

Then the body of `init_train_impl` looks something more like:
```
model.set_training_model(std::unique_ptr<model_backend>(new mock_model_backend));
model.set_nn_spec(std::unique_ptr<nn_spec>(new nn_spec));
```

Note that all `object_detector::train` does with `training_model_` is call `export_weights`, and a default `mock_model_backend` already returns an empty `float_array_map`, which is fine.

Then `object_detector::train` will call `nn_spec_->update_params` with an empty `float_array_map`, which should also be fine.

shreyajain17:
Captured this in PR #2101

PR #1922, opened 2019-05-23
Use JSON.stringify to display arrays in explore

### LINKED ISSUE #1921: explore() incorrectly concatenates array elements
Create an SArray with a dtype of `array` that contain ints or floats.
```python

In [2]: sa = tc.SArray([[1,2], [3,4]])

In [3]: print(sa)
[array('d', [1.0, 2.0]), array('d', [3.0, 4.0])]

In [4]: sa.explore()
Materializing SFrame
```
This is what pops up:
Screen Shot 2019-05-22 at 5 57 33 PM
Note: the element values for each row have been concatenated.

znation:
* Adds JSON.stringify to fix the display of arrays.
* Since the above was already used for ndarrays and lists, this commit
  combines the logic of all 3 types.

Fixes #1921.


abhishekpratapa:
Awesome thanks!

PR #1924, opened 2019-05-23
Added logic to download background files from docs-assets.developer.apple.com

### LINKED ISSUE #1820: Upload Background dataset and load it internally before training

shantanuchhabra:
Fixes #1820

### DIFF
--- a/src/unity/python/turicreate/toolkits/_data_zoo.py
+++ b/src/unity/python/turicreate/toolkits/_data_zoo.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright © 2017 Apple Inc. All rights reserved.
+# Copyright © 2019 Apple Inc. All rights reserved.
 #
 # Use of this source code is governed by a BSD-3-clause license that can
 # be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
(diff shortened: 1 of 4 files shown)


src/unity/python/turicreate/toolkits/_data_zoo.py
```diff
@@ -0,0 +1,49 @@
...
+import sys as _sys
+import requests as _requests
+import turicreate as _tc
+import hashlib as _hashlib
+from six.moves.urllib import parse as _urlparse
+from ._pre_trained_models import _download_and_checksum_files
+
+DATA_URL_ROOT = 'https://docs-assets.developer.apple.com/turicreate/data/'
+
+def _get_data_cache_dir():
```
TobyRoseman: Almost this exact same function already exists in `toolkits/_pre_trained_models.py`. I don't think copying and pasting it then change one line is a great idea. Why not add a parameter to the existing function for the one line change, then import it and use it here?

shantanuchhabra: Done!

src/unity/python/turicreate/toolkits/_data_zoo.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2017 Apple Inc. All rights reserved.
```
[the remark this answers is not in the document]
shantanuchhabra: Done.

PR #1925, opened 2019-05-23
Clarifying warm_start in Drawing Classifier user guide

### LINKED ISSUE #1892: Unable to access pre-trained drawing model
When I click the "[here](https://docs-assets.developer.apple.com/turicreate/)" link at the bottom of [this page](https://apple.github.io/turicreate/docs/userguide/drawing_classifier/advanced-usage.html) I get an Access Denied error.

### LINKED ISSUE #1893: Clarify default warm start model behavior in the drawing documentation
Reading through [this page](https://apple.github.io/turicreate/docs/userguide/drawing_classifier/advanced-usage.html) at the bottom it says

> we provide the option of loading in a pretrained model for a warm start ... we have published our pre-trained model here

So I was under the impression if I wanted a warm start I have to download this model and specify the path to it in the parameters to `drawing_classifier.create`, but the [API docs](https://apple.github.io/turicreate/docs/api/generated/turicreate.drawing_classifier.create.html?highlight=drawing) indicate otherwise

> warm_start : string optional
> A string to denote which pretrained model to use. Set to “auto” by default which uses a model trained on 245 of the 345 classes in the Quick, Draw! dataset. To disable warm start, pass in None to this argument.

shreyajain17:
fixes #1893 
fixes #1892

### DIFF
--- a/userguide/drawing_classifier/advanced-usage.md
+++ b/userguide/drawing_classifier/advanced-usage.md
@@ -31,4 +31,5 @@ in the network to something better than random, thereby improving accuracy
 and also helping the model converge faster.
 
 We have published a pre-trained model that automatically gets downloaded when you pass in 
-`auto` to the `warm_start` parameter. 
+`auto` to the `warm_start` parameter. This model is trained on millions of drawings from 
+the “Quick,Draw!” dataset
--- a/userguide/drawing_classifier/advanced-usage.md
+++ b/userguide/drawing_classifier/advanced-usage.md
@@ -31,8 +31,4 @@ in the network to something better than random, thereby improving accuracy
 and also helping the model converge faster.
 
 We have published a pre-trained model that automatically gets downloaded when you pass in 
-`auto` to the `warm_start` parameter. This model is trained on millions of drawings spread 
-over the first 245 classes of the “Quick,Draw!” dataset.
-
-We trained this model on over 60,000 examples each of the first 245 of the
-345 classes in the Quick, Draw! dataset.
+`auto` to the `warm_start` parameter. 
--- a/userguide/drawing_classifier/advanced-usage.md
+++ b/userguide/drawing_classifier/advanced-usage.md
@@ -22,7 +22,8 @@ provided value of `num_epochs` would be ignored.
 
 To boost the accuracy of the Drawing Classifier models you train, and to help
 those models converge faster, we provide the option of loading in a 
-pretrained model for a warm start. 
+pretrained model for a warm start. This can be done by passing `auto` in the `warm_start` 
+parameter. 
 
 A cold start to training would be when the weights are randomly initialized. 
 A warm start is when the weights in the neural network are loaded from those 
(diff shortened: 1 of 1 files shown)


userguide/drawing_classifier/advanced-usage.md
```diff
@@ -22,15 +22,14 @@ provided value of `num_epochs` would be ignored.
...
 A cold start to training would be when the weights are randomly initialized. 
 A warm start is when the weights in the neural network are loaded from those 
 of an already trained model. This improves the initial values of the weights 
 in the network to something better than random, thereby improving accuracy 
 and also helping the model converge faster.
 
-We have published our pre-trained model 
-[here](https://docs-assets.developer.apple.com/turicreate/). 
-We trained this published model on 1,000 examples each of the first 245 of the
+We trained this model on 1,000 examples each of the first 245 of the
```
shantanuchhabra: How about “We have published a pre-trained model that automatically gets downloaded when you pass in “auto” to the ‘warm_start’ parameter. This model is trained on millions of drawings spread over the first 245 classes of the “Quick,Draw!” dataset”

1000 drawings/class is actually a typo. It’s more like 68167 drawings/class.

shreyajain17: That sounds good! I will change it to that with the correction.

shreyajain17: Should I write 68167 drawings/class or more like over 60000 drawings/class?

srikris: I don't think we need to mention so many details. This model will change over time and it's going to be hard to maintain the user guide will all the details. Err on the side of permanent and persistent content that conveys the same message.

shreyajain17:
Hey 
@srikris @shantanuchhabra Is this fine? I removed details about the training to keep it generic?

userguide/drawing_classifier/advanced-usage.md
shantanuchhabra: Yeah I think "This model is trained on millions of drawings from the “Quick,Draw!” dataset" should do.

userguide/drawing_classifier/advanced-usage.md
```diff
@@ -22,15 +22,13 @@ provided value of `num_epochs` would be ignored.
...
 
 A cold start to training would be when the weights are randomly initialized. 
 A warm start is when the weights in the neural network are loaded from those 
 of an already trained model. This improves the initial values of the weights 
 in the network to something better than random, thereby improving accuracy 
 and also helping the model converge faster.
 
-We have published our pre-trained model 
-[here](https://docs-assets.developer.apple.com/turicreate/). 
-We trained this published model on 1,000 examples each of the first 245 of the
```
TobyRoseman: I think removing the link is a good idea.

This is an advanced section of the user guide. So I think it's good to mention that it was trained on 1,000 example of 245 out 345 classes of the Quick, Draw! dataset. I'm not sure about the phrase "the first 245". Is there some natural ordering of these classes that I'm missing?

shantanuchhabra: The natural ordering of these classes on the Quick, Draw! website and in the dataset is lexicographic. I agree that we shouldn't mention "the first 245". The model is trained on 68167 images/class and not 1000 so we should either fix that typo or get rid of it entirely. Maybe something like "This model is trained on millions of drawings from the “Quick,Draw!” dataset"? Sounds general enough that it gives us the opportunity to modify it when we ship a better model down the road, but also gives advanced users a ballpark of how many images the pre-trained model is trained on.

PR #1928, opened 2019-05-23
Gives error when output_type is probability for multi-class classification

### LINKED ISSUE #1920: Multi-class Drawing Classifier model accepts probability output type for predict(..)
A drawing classifier model, trained with more than two classes, still allows `output_type='probability'` when calling `predict(...)`. It seems to return the probability for the most likely class without any indication of what class.

Calling `predict(...)` on a multi-class model should throw an exception. The doc strings already imply this is what should happen.

shreyajain17:
fixes #1920


src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -872,10 +872,13 @@ class as a vector. Label ordering is dictated by the ``classes``
...
-            target = self.target
-            return predicted.apply(
-                lambda row: row["probability"][_class_to_index[row[target]]])
+            if len(self.classes) == 2:
+                _class_to_index = self._class_to_index
+                target = self.target
+                return predicted.apply(
+                    lambda row: row["probability"][_class_to_index[row[target]]])
+            else:
+                raise RuntimeError("Use probability vector in case of multi-class classification")
```
shantanuchhabra: raise _ToolkitError

PR #1959, opened 2019-05-28
Added extract_label option to load_images

shantanuchhabra:
Extracts the label from the path (assuming that the path is of the form ..././../../../label.png) if `extract_label=True` is passed into `load_images`.

This should be very a very useful feature for One Shot OD.

### DIFF
--- a/src/unity/python/turicreate/toolkits/image_analysis/image_analysis.py
+++ b/src/unity/python/turicreate/toolkits/image_analysis/image_analysis.py
@@ -9,7 +9,7 @@ from __future__ import absolute_import as _
 from ...data_structures.image import Image as _Image
 
 
-def load_images(url, format='auto', with_path=True, recursive=True, ignore_failure=True, random_order=False):
+def load_images(url, format='auto', with_path=True, recursive=True, ignore_failure=True, random_order=False, extract_label=False):
     """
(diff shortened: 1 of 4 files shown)


src/unity/python/turicreate/toolkits/image_analysis/image_analysis.py
```diff
@@ -39,13 +39,19 @@ def load_images(url, format='auto', with_path=True, recursive=True, ignore_failu
...
     out : SFrame
         Returns an SFrame with either an 'image' column or both an 'image' and
-        a 'path' column. The 'image' column is a column of Image objects. If
-        with_path is True, there is also a 'path' column which contains the image
-        path for each of each corresponding Image object.
+        a 'path' column. The 'image' column is a column of Image objects.
+        If 'with_path' is True, there is also a 'path' column which contains
+        the image path for each of each corresponding Image object.
+        If 'extract_label' is True, there is also a 'label' column which
+        contains the 'label' column for each corresponding Image object.
```
srikris: which contains the 'label' for each corresponding image.

srikris:
@TobyRoseman @shantanuchhabra I think this seems like adding this would make it easier for one-shot but it's not quite right for image classification since there, the labels are often in folders. I think it's probably best to not add this API to the function.

PR #1963, opened 2019-05-28
Fixed model save and load

shantanuchhabra:


src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -95,12 +95,12 @@ def _get_native_state(self):
...
+        state['detector'] = state['detector']._get_native_state()
         return state
 
     @classmethod
     def _load_version(cls, state, version):
-        assert(version == _PYTHON_ONE_SHOT_OBJECT_DETECTOR_VERSION)
+        assert(version == cls._PYTHON_ONE_SHOT_OBJECT_DETECTOR_VERSION)
         # we need to undo what we did at save and turn the proxy object
         # back into a Python class
         state['detector'] = _ObjectDetector(state['detector'])
```
nickjong: Have you tested this? Doesn't this need to decode the native state back into an OD model first?

shantanuchhabra: Yeah, I tested `get_augmented_images` and checked for the `detector` key in it and it seemed to work!

nickjong: It doesn't suffice for the `detector` key to exist. The value must be a (functioning) object detector.

nickjong: If `state['detector']` is the result of `_get_native_state()`, how does `_ObjectDetector` correctly parse that?

shantanuchhabra: Fixed.

shantanuchhabra: Fixed!

PR #1967, opened 2019-05-28
Fixes evaluation console, drops missing values before training and fixes time elapsed in training

### LINKED ISSUE #1918: Confusing console output on drawing_classifier.evaluate(...)
```python
In [17]: len(test_data)
Out[17]: 143

In [18]: _ = model.evaluate(test_data)
Predicting 144/143

In [19]: len(longer_test_data)
Out[19]: 429

In [20]: _ = model.evaluate(longer_test_data)
Predicting 257/429
```
The output from the first call to `evaluate(...)` is really the most confusing. How is it predicting one more examples than was passed in? This seems like an off by one bug.

The output from the second call to `evaluate(...)` makes it seem like it never finishes. The last/only output makes it seem like it's only about half done. The final output should probably be `Predicting X/X` where `X` is the number of examples passed in.

Both calls to evaluate finish quickly. I would recommend only outputting progress info if it's been at least 30 seconds.

### LINKED ISSUE #1936: Bad Error Message when training with None Types
If some of the images are `None`, the drawing classifier give a poor error message:
```

In [2]: from random import randint

In [3]: data = tc.load_sframe('./quickdraw/sframes/bitmap.sframe/')

In [4]: data['drawing'] = data['drawing'].apply(lambda x: None if randint(1, 20) == 10
                                                                                                              else x)

In [5]: assert len(data['drawing'].dropna()) < len(data['drawing'])

In [6]: tc.drawing_classifier.create(data, feature='drawing', target='label')
PROGRESS: Creating a validation set from 5 percent of training data. This may take a while.
          You can set ``validation_set=None`` to disable validation tracking.

+------------------------+------------------------+------------------------+------------------------+------------------------+
| Iteration              | Training Loss          | Training Accuracy      | Validation Accuracy    | Elapsed Time (seconds) |
+------------------------+------------------------+------------------------+------------------------+------------------------+
AttributeError                            Traceback (most recent call last)
<ipython-input-6-da952d119295> in <module>()
----> 1 tc.drawing_classifier.create(data, feature='drawing', target='label')

/Users/tobyroseman/virts/rcTesting/lib/python2.7/site-packages/turicreate/toolkits/drawing_classifier/drawing_classifier.pyc in create(input_dataset, target, feature, validation_set, warm_start, batch_size, max_iterations, verbose)
    276             accuracy_metric.update(batch_label, outputs)
    277 
--> 278     for train_batch in train_loader:
    279         train_batch_data, train_batch_label = get_data_and_label_from_batch(train_batch)
    280         with _autograd.record():

/Users/tobyroseman/virts/rcTesting/lib/python2.7/site-packages/turicreate/toolkits/drawing_classifier/_sframe_loader.pyc in next(self)
    137 
    138     def next(self):
--> 139         return self._next()
    140 
    141     @property

/Users/tobyroseman/virts/rcTesting/lib/python2.7/site-packages/turicreate/toolkits/drawing_classifier/_sframe_loader.pyc in _next(self)
    157         for b in range(self.batch_size):
    158             try:
--> 159                 row = next(self.data_source)
    160             except StopIteration:
    161                 if b == 0:

/Users/tobyroseman/virts/rcTesting/lib/python2.7/site-packages/turicreate/toolkits/drawing_classifier/_sframe_loader.pyc in next(self)
     38 
     39     def next(self):
---> 40         return self._next()
     41 
     42     def _next(self):

/Users/tobyroseman/virts/rcTesting/lib/python2.7/site-packages/turicreate/toolkits/drawing_classifier/_sframe_loader.pyc in _next(self)
     59         is_stroke_input = (not isinstance(drawing_feature, _tc.Image))
     60         if is_stroke_input:
---> 61             pixels_01 = drawing_feature.pixel_data.reshape(
     62                 1, bitmap_width, bitmap_height) / 255.
     63         else:

AttributeError: 'NoneType' object has no attribute 'pixel_data'
```
I would never guess that this mean my data has `None`s in it. 

At the very least we should give a clear error message. It would be better to drop the rows with `None` in them and give a warning about it.

shreyajain17:
fixes #1918
fixes #1936

### DIFF
--- a/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -44,14 +44,11 @@ def _raise_error_if_not_drawing_classifier_input_sframe(
     if len(dataset) == 0:
         raise _ToolkitError("Input Dataset is empty!")
 
-def drop_missing_values(dataset, feature, is_train):
+def _drop_missing_values(dataset, feature, is_train):
     original_dataset_size = len(dataset)
     dataset = dataset.dropna(columns=feature)
-    if is_train:
-        dataset_type = "train"
-    else:
-        dataset_type = "validation"
     if original_dataset_size != len(dataset):
+        dataset_type = "train" if is_train else "validation"
         print("Dropping " + str(original_dataset_size - len(dataset)) + " rows for missing features in the " + dataset_type + " dataset.")
     return dataset
 
@@ -208,10 +205,10 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         raise TypeError("Unrecognized type for 'validation_set'."
             + validation_set_corrective_string)
 
-    dataset = drop_missing_values(dataset, feature, is_train =True)
+    dataset = _drop_missing_values(dataset, feature, is_train =True)
 
     if validation_set is not None:
-        validation_dataset = drop_missing_values(validation_dataset, feature, is_train=False)
+        validation_dataset = _drop_missing_values(validation_dataset, feature, is_train=False)
 
     train_loader = _SFrameClassifierIter(dataset, batch_size,
                  feature_column=feature,
@@ -574,7 +571,7 @@ class DrawingClassifier(_CustomModel):
         dataset_size = len(dataset)
         ctx = _mxnet_utils.get_mxnet_context()
 
-        index = 0
+        num_processed = 0
         last_time = 0
 
         from turicreate import SArrayBuilder
@@ -607,14 +604,14 @@ class DrawingClassifier(_CustomModel):
                 split_length = z.shape[0]
                 all_predicted_builder.append_multiple(predicted)
                 all_probabilities_builder.append_multiple(z.tolist())
-                index += split_length 
+                num_processed += split_length 
 
                 cur_time = _time.time()
                 # Do not print progress if only a few samples are predicted
                 if verbose and (dataset_size >= 5
-                    and cur_time > last_time + 30 or index >= dataset_size ):
-                    print('Predicting {num_processed:{width}d}/{max_n:{width}d}'.format(
-                        num_processed = index,
+                    and cur_time > last_time + 30 or num_processed >= dataset_size ):
+                    print('Predicting {cur_n:{width}d}/{max_n:{width}d}'.format(
+                        cur_n = num_processed,
                         max_n = dataset_size,
                         width = len(str(dataset_size))))
                     last_time = cur_time
--- a/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -44,6 +44,17 @@ def _raise_error_if_not_drawing_classifier_input_sframe(
     if len(dataset) == 0:
         raise _ToolkitError("Input Dataset is empty!")
 
+def drop_missing_values(dataset, feature, is_train):
+    original_dataset_size = len(dataset)
+    dataset = dataset.dropna(columns=feature)
+    if is_train:
+        dataset_type = "train"
+    else:
+        dataset_type = "validation"
+    if original_dataset_size != len(dataset):
+        print("Dropping " + str(original_dataset_size - len(dataset)) + " rows for missing features in the " + dataset_type + " dataset.")
+    return dataset
+
 def create(input_dataset, target, feature=None, validation_set='auto',
             warm_start='auto', batch_size=256, 
             max_iterations=100, verbose=True):
@@ -197,17 +208,10 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         raise TypeError("Unrecognized type for 'validation_set'."
             + validation_set_corrective_string)
 
-
-    original_dataset_size = len(dataset)
-    dataset = dataset.dropna(columns=feature)
-    if original_dataset_size != len(dataset):
-        print("Dropping " + str(original_dataset_size - len(dataset)) + " rows from the training dataset.")
+    dataset = drop_missing_values(dataset, feature, is_train =True)
 
     if validation_set is not None:
-        original_validation_size = len(validation_dataset)
-        validation_dataset = validation_dataset.dropna(columns=feature)
-        if original_validation_size != len(validation_dataset):
-            print("Dropping " + str(original_validation_size - len(validation_dataset)) + " rows from the validation dataset.")
+        validation_dataset = drop_missing_values(validation_dataset, feature, is_train=False)
 
     train_loader = _SFrameClassifierIter(dataset, batch_size,
                  feature_column=feature,
@@ -309,7 +313,6 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         trainer.step(train_batch.data[0].shape[0])
         # calculate training metrics
         train_loss = loss.mean().asscalar()
-        train_time = _time.time() - start_time
 
         if train_batch.iteration > iteration:
 
@@ -323,7 +326,7 @@ def create(input_dataset, target, feature=None, validation_set='auto',
                 kwargs = {  "iteration": iteration + 1,
                             "train_loss": float(train_loss),
                             "train_accuracy": train_accuracy.get()[1],
-                            "time": train_time}
+                            "time": _time.time() - start_time}
                 if validation_set is not None:
                     kwargs["validation_accuracy"] = validation_accuracy.get()[1]
                 table_printer.print_row(**kwargs)
@@ -336,7 +339,7 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         'input_image_shape': (1, BITMAP_WIDTH, BITMAP_HEIGHT),
         'training_loss': train_loss,
         'training_accuracy': train_accuracy.get()[1],
-        'training_time': train_time,
+        'training_time': _time.time() - start_time,
         'validation_accuracy': validation_accuracy.get()[1] if validation_set else None, 
         # None if validation_set=None
         'max_iterations': max_iterations,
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -602,7 +614,7 @@ def _predict_with_probabilities(self, input_dataset, batch_size=64,
                 if verbose and (dataset_size >= 5
                     and cur_time > last_time + 30 or done):
                     print('Predicting {cur_n:{width}d}/{max_n:{width}d}'.format(
-                        cur_n = index + 1 if not done else index,
+                        cur_n = index + 1 if index + 1 > dataset_size - 1 else index,
```
TobyRoseman: I'm having a hard time understanding this line change.

Is `done` not actually getting set to `True` when things are done?

shreyajain17: So I had made this change in my earlier PR with not done in my earlier PR but that hadn't fixed this issue. If you see index is being incremented by split_length and then checks if it is equal to dataset_size - 1. So basically index skips over the dataset_size - 1 value for done to be true. But I am not fixing it correctly I should have just make checked if index > dateset_size -1 and kept this code as it is. Does this make sense?

TobyRoseman: That sounds good. Let's not change this line and let's just make sure that `done` is getting set correctly.

shreyajain17: Yes, got it. I will change it.

nickjong: This code reads weirdly to me. AFAICT, `index` exists only to track how many instances have been processed. I would be in favor of just renaming it `cur_n` directly. I also think we can just remove the variable `done` entirely. It looks to me like the current computation of `done` is incorrect, it should be:
```
if cur_n >= dataset_size:
  cur_n = dataset_size
```
And then instead of testing `done` you can just test `cur_n >= dataset_size`.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -197,6 +197,18 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         raise TypeError("Unrecognized type for 'validation_set'."
             + validation_set_corrective_string)
 
+
+    original_dataset_size = len(dataset)
+    dataset = dataset.dropna(columns=feature)
+    if original_dataset_size != len(dataset):
+        print("Dropping " + str(original_dataset_size - len(dataset)) + " rows from the training dataset.")
```
nickjong: Could we indicate in the message why we're dropping the user's data?

shreyajain17: Dropping x rows for containing None values in the training dataset?

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -197,6 +197,18 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         raise TypeError("Unrecognized type for 'validation_set'."
             + validation_set_corrective_string)
 
+
+    original_dataset_size = len(dataset)
+    dataset = dataset.dropna(columns=feature)
```
srikris: Won't this mutate the original dataset? I think we should not change the user data.

TobyRoseman: I don't see how. `dropna` returns a new SFrame, doesn't change things in place.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -197,6 +197,18 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         raise TypeError("Unrecognized type for 'validation_set'."
             + validation_set_corrective_string)
 
+
+    original_dataset_size = len(dataset)
```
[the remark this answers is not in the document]
shreyajain17: I can do that

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shreyajain17: It doesn't mutate the data

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -44,6 +44,17 @@ def _raise_error_if_not_drawing_classifier_input_sframe(
     if len(dataset) == 0:
         raise _ToolkitError("Input Dataset is empty!")
 
+def drop_missing_values(dataset, feature, is_train):
```
TobyRoseman: This method should be hidden. Please prefix method name with a `_`.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -44,6 +44,17 @@ def _raise_error_if_not_drawing_classifier_input_sframe(
...
+    original_dataset_size = len(dataset)
+    dataset = dataset.dropna(columns=feature)
+    if is_train:
```
nickjong: Nit: this if/else should probably go under the if statement below, only if we're actually going to print something. Then it can also be a one liner:
```
if original_dataset_size != len(dataset):
  dataset_type = "train" if is_train else "validation"
  ...
```

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -594,15 +608,13 @@ def _predict_with_probabilities(self, input_dataset, batch_size=64,
...
 
                 cur_time = _time.time()
                 # Do not print progress if only a few samples are predicted
                 if verbose and (dataset_size >= 5
-                    and cur_time > last_time + 30 or done):
-                    print('Predicting {cur_n:{width}d}/{max_n:{width}d}'.format(
-                        cur_n = index + 1 if not done else index,
+                    and cur_time > last_time + 30 or index >= dataset_size ):
+                    print('Predicting {num_processed:{width}d}/{max_n:{width}d}'.format(
+                        num_processed = index,
```
TobyRoseman: Has this change been testing? I'm not certain but if it's possible for this to be true `index > dataset_size - 1`, then we could be reporting that we processed more example than there are in the dataset.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -561,7 +576,6 @@ def _predict_with_probabilities(self, input_dataset, batch_size=64,
 
         index = 0
```
nickjong: Can we just rename `index` to `num_processed`?

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -297,7 +313,6 @@ def compute_accuracy(accuracy_metric, batch_loader):
         trainer.step(train_batch.data[0].shape[0])
         # calculate training metrics
         train_loss = loss.mean().asscalar()
-        train_time = _time.time() - start_time
```
TobyRoseman: Why this change and the next two changes? Are you fixing another issue?

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -594,15 +608,13 @@ def _predict_with_probabilities(self, input_dataset, batch_size=64,
...
                 cur_time = _time.time()
                 # Do not print progress if only a few samples are predicted
                 if verbose and (dataset_size >= 5
-                    and cur_time > last_time + 30 or done):
-                    print('Predicting {cur_n:{width}d}/{max_n:{width}d}'.format(
-                        cur_n = index + 1 if not done else index,
+                    and cur_time > last_time + 30 or index >= dataset_size ):
+                    print('Predicting {num_processed:{width}d}/{max_n:{width}d}'.format(
+                        num_processed = index,
                         max_n = dataset_size,
```
nickjong: In this print statement, let's either use "num_processed" and "max_processed" or "cur_n" and "max_n"

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shreyajain17: This works. 

```
model.evaluate(data)
Predicting 256/256
```

```
In [13]: model.evaluate(data)
Predicting  256/7237
Predicting 7237/7237
```

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shreyajain17: Okay

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shreyajain17: In one of the previous PRs, I was suspicious the time elapsed is incorrectly calculated. The time printing earlier was for one mini batch of the epoch, this change correctly calculates time taken per epoch.

src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -197,6 +205,11 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         raise TypeError("Unrecognized type for 'validation_set'."
             + validation_set_corrective_string)
 
+    dataset = _drop_missing_values(dataset, feature, is_train =True)
+
+    if validation_set is not None:
```
nickjong: Nit: `validation_dataset` is always an (possibly empty) `SFrame` at this point, right? Could we just unconditionally do line 211?

PR #1968, opened 2019-05-28
Fix some broken Python3 str/bytes encoding in vis

znation:


znation:
@TobyRoseman I'm not sure about that. [This doc](https://docs.python.org/2.7/library/stdtypes.html#str.encode) says:
```
Return an encoded version of the string. Default encoding is the current default string encoding.
```

TobyRoseman:
> @TobyRoseman I'm not sure about that. [This doc](https://docs.python.org/2.7/library/stdtypes.html#str.encode) says:
> 
> ```
> Return an encoded version of the string. Default encoding is the current default string encoding.
> ```

@znation - You'e right. Looks like what I said is only true for Python 3.

PR #1970, opened 2019-05-28
Fix Linux build break in IC Annotate

znation:


znation:
@TobyRoseman Thanks. I'm hoping for a bigger refactor in a future sprint that will disentangle all of the `ifdef` stuff more cleanly. So for now I'll keep it as is.

abhishekpratapa:
Unused Variables got it.

PR #1971, opened 2019-05-29
Fix one more linux build break in IC Annotation

znation:


TobyRoseman:
I also can't find anywhere that `featurizer_thread` is used (inside of a `#ifdef __APPLE__` or otherwise).

TobyRoseman:
Looks like `m_image_feature_extraction_sarray` and `m_writer` should also be moved inside the `#ifdef` in the hpp file.

znation:
Thanks @TobyRoseman -- for now I'm going to take this fix (since it gives us a passing Linux build) and plan to refactor this whole class soon.

Issue #1985, opened 2019-05-29
One Shot - Missing Docstrings


TobyRoseman:
The following need docstrings:
- [ ] `tc.one_shot_object_detector.OneShotObjectDetector`
- [ ] `tc.one_shot_object_detector.OneShotObjectDetector.predict`
- [ ] `tc.one_shot_object_detector.OneShotObjectDetector.evaluate`
- [ ] `tc.one_shot_object_detector.util`

Issue #1986, opened 2019-05-29
One shot detector remove nested namespace


TobyRoseman:
We should remove `tc.one_shot_object_detector.one_shot_object_detector` (note the two `one_shot_object_detector`). Everything there is already either in `tc.one_shot_object_detector` or `tc.one_shot_object_detector.utils`.

srikris:
Unfortunately this is an issue for not just one-shot but many more of our toolkits including

- [ ] Drawing classifier
- [ ] Object detector
- [ ] Image classifier
- [ ] Image similarity
- [ ] One shot object detector.

srikris:
On further investigation, this is a deeper issue which would require more changes that we should pick up for 5.6.

PR #1993, opened 2019-05-29
Adds api docs for one shot object detector

shreyajain17:


src/unity/python/doc/source/turicreate.toolkits.one_shot_object_detector.rst
```diff
@@ -0,0 +1,24 @@
...
+  create
+  OneShotObjectDetector
+
+utilities
+---------
+.. autosummary::
+  :toctree: generated/
+  :nosignatures:
+
+  util.preview_augmented_images
```
TobyRoseman: Do we also want to add `util.check_one_shot_input`?

shreyajain17: We should ask @shantanuchhabra. It had nothing in comments so I didn't add it yet.

srikris: I don't think we should add that. That;s probably an underscore error.

srikris:
Fixes #1985

src/unity/python/doc/source/turicreate.toolkits.one_shot_object_detector.rst
srikris: @TobyRoseman The _ error will be fixed in #1986

shreyajain17:
@srikris Sorry I missed that. Will do it!

shreyajain17:
> We should also add the appropriate API docs for the predict, evaluate, and export to CoreML methods.

API Docs for predict, evaluate and export_coreml are in my PR #1997

PR #1998, opened 2019-05-30
Added repr and summary.

### LINKED ISSUE #1990: One Shot create(...) should take a single image.
The doc string for `tc.one_shot_object_detector.create(...)` claim you can pass in a single `tc.Image` for the  `data` parameter. However that doesn't work. It gives the error:
```
TypeError: object of type 'Image' has no len()
```

### LINKED ISSUE #1991: One Shot - has no __str__ or summary methods

srikris:
Fixes #1990 
Fixes #1991

### DIFF
--- a/src/unity/python/turicreate/toolkits/_internal_utils.py
+++ b/src/unity/python/turicreate/toolkits/_internal_utils.py
@@ -359,7 +359,7 @@ def _make_repr_table_from_sframe(X):
     return ['  '.join(row) for row in out_data]
 
 
-def _toolkit_repr_print(model, fields, section_titles, width = None):
+def _toolkit_repr_print(model, fields, section_titles, width = None, class_name = None):
     """
     Display a toolkit repr according to some simple rules.
 
@@ -406,7 +406,10 @@ def _toolkit_repr_print(model, fields, section_titles, width = None):
         "The number of section titles ({0}) ".format(len(section_titles)) +\
         "doesn't match the number of groups of fields, {0}.".format(len(fields))
 
-    out_fields = [ ("Class", model.__class__.__name__), ""]
+    if class_name is not None:
+        out_fields = [ ("Class", class_name), ""]
+    else:
+        out_fields = [ ("Class", model.__class__.__name__), ""]
 
     # Record the max_width so that if width is not provided, we calculate it.
     max_width = len("Class")
--- a/src/unity/python/turicreate/toolkits/_internal_utils.py
+++ b/src/unity/python/turicreate/toolkits/_internal_utils.py
@@ -359,7 +359,7 @@ def _make_repr_table_from_sframe(X):
     return ['  '.join(row) for row in out_data]
 
 
-def _toolkit_repr_print(model, fields, section_titles, width = None, class_name = None):
+def _toolkit_repr_print(model, fields, section_titles, width = None, class_name = 'auto'):
     """
     Display a toolkit repr according to some simple rules.
 
@@ -406,7 +406,7 @@ def _toolkit_repr_print(model, fields, section_titles, width = None, class_name 
         "The number of section titles ({0}) ".format(len(section_titles)) +\
         "doesn't match the number of groups of fields, {0}.".format(len(fields))
 
-    if class_name is not None:
+    if class_name == 'auto':
         out_fields = [ ("Class", class_name), ""]
     else:
         out_fields = [ ("Class", model.__class__.__name__), ""]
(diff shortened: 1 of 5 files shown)


src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -105,6 +116,63 @@ def _load_version(cls, state, version):
...
+            A list of section titles.
+              The order matches that of the 'sections' object.
+        """
+        model_fields = [
+            ('Number of classes', 'num_classes'),
+            ('Input image shape', 'input_image_shape'),
+        ]
+        data_fields = [
+            ('Number of synthetically generated examples', 'num_examples'),
+            ('Number of synthetically generated bounding boxes', 'num_bounding_boxes'),
```
shantanuchhabra: Where are the num_examples and num_bounding_boxes fields getting set?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -105,6 +116,63 @@ def _load_version(cls, state, version):
...
+        ]
+        training_fields = [
+            ('Training time', '_training_time_as_string'),
+            ('Training iterations', 'training_iterations'),
+            ('Training epochs', 'training_epochs'),
+            ('Final loss (specific to model)', 'training_loss'),
```
shantanuchhabra: These are all fields on the state of the underlying `ObjectDetector` model. We might want to save them in the `OneShotObjectDetector` state  in `.create` by doing something like `od_model.__proxy__["training_iterations"]`?

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
srikris: They are from the detector model.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
srikris: I erred on the side of not changing surface area.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
shantanuchhabra: Do we need to save the 'num_examples' and 'num_bounding_boxes' fields for the OneShotObjectDetector? Currently, they are fields on the 'detector' field of the OneShotObjectDetector.

srikris: I haven't added anything to the model. Just showing things in the __repr__

src/unity/python/turicreate/toolkits/_internal_utils.py
```diff
@@ -359,7 +359,7 @@ def _make_repr_table_from_sframe(X):
     return ['  '.join(row) for row in out_data]
 
 
-def _toolkit_repr_print(model, fields, section_titles, width = None):
+def _toolkit_repr_print(model, fields, section_titles, width = None, class_name = None):
```
TobyRoseman: `class_name = 'auto'` probably makes more sense.

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -19,29 +20,22 @@ def create(data,
...
-
     data : SFrame | tc.Image
-        An SFrame that contains all the starter images along with their
-        corresponding labels.
-        If the dataset is a single tc.Image, this parameter is just that image.
-        Every starter image should entirely be just the starter image without
-        any padding.
-        RGB and RGBA images allowed.
+        A single starter image or an SFrame that contains the starter images
+        along with their corresponding labels. These images can be of in RGB
```
TobyRoseman: You want to get rid of the `of`. On the start of the next line I think `or` makes more since than `and`. Also the padding statement should probably be it's own sentence. 

So I would replace the last sentence of this paragraph with `These image(s) can be in either RGB or RGBA format. They should not be padded.`

src/unity/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
```diff
@@ -11,41 +11,33 @@
...
-
     data : SFrame | tc.Image
-        An SFrame that contains all the starter images along with their
-        corresponding labels.
-        If the dataset is a single tc.Image, this parameter is just that image.
-        Every starter image should entirely be just the starter image without
-        any padding.
-        RGB and RGBA images allowed.
+        A single starter image or an SFrame that contains the starter images
+        along with their corresponding labels. These images can be of in RGB
```
TobyRoseman: I would make the same changes here too.

PR #2000, opened 2019-05-30
Fixed tests on master.

srikris:
Unit tests on master should now be passing.

### DIFF
--- a/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -208,7 +208,8 @@ def create(input_dataset, target, feature=None, validation_set='auto',
             + validation_set_corrective_string)
 
     dataset = _drop_missing_values(dataset, feature, is_train =True)
-    validation_dataset = _drop_missing_values(validation_dataset, feature, is_train=False)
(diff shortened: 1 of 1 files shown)


src/unity/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -208,7 +208,8 @@ def create(input_dataset, target, feature=None, validation_set='auto',
             + validation_set_corrective_string)
 
     dataset = _drop_missing_values(dataset, feature, is_train =True)
-    validation_dataset = _drop_missing_values(validation_dataset, feature, is_train=False)
+    if len(validation_set) > 0:
```
nickjong: Note: Travis is failing with
```
>       if len(validation_set) > 0:
E       TypeError: object of type 'NoneType' has no len()
```

nickjong: Per Shreya, this should check `len(validation_dataset)` not `len(validation_set)`. The former is guaranteed to be an SFrame. The latter can be a string or `None`.

srikris: Looks like I forgot to push my last commit. Pushing now. Thanks for the catch!

PR #2008, opened 2019-05-31
Ignore existing undefined labels when annotating

### LINKED ISSUE #1979: Annotations when column 'annotations' already exists
```
data_h = data.head()

# Annotate 2 images
data_a = tc.image_classifier.annotate(data_h)

# Call annotation again
data_new_a = tc.image_classifier.annotate(data_a)
```

image

znation:
Fixes #1979

### DIFF
--- a/src/unity/lib/annotation/image_classification.cpp
+++ b/src/unity/lib/annotation/image_classification.cpp
@@ -90,22 +90,23 @@ annotate_spec::Annotations ImageClassification::getAnnotations(size_t start,
   std::vector<flexible_type> flex_data = filtered_data->to_vector();
 
   for (size_t i = 0; i < flex_data.size(); i++) {
-    annotate_spec::Annotation *annotation = annotations.add_annotation();
-    annotate_spec::Label *label = annotation->add_labels();
-
-    label->imageclassificationlabel();
-
     const flexible_type& flex_label = flex_data.at(i);
     if (flex_label.get_type() == flex_type_enum::UNDEFINED) {
       // skip unlabeled items
       continue;
     }
 
+    annotate_spec::Annotation *annotation = annotations.add_annotation();
+    annotate_spec::Label *label = annotation->add_labels();
+
+    // initialize the label as an image classification label
+    label->mutable_imageclassificationlabel();
+
     if (flex_label.get_type() == flex_type_enum::STRING) {
-      std::string label_value = flex_label.get<flex_string>();
+      const flex_string& label_value = flex_label.get<flex_string>();
       label->set_stringlabel(label_value);
     } else if (flex_label.get_type() == flex_type_enum::INTEGER) {
-      int label_value = flex_label;
+      flex_int label_value = flex_label.get<flex_int>();
       label->set_intlabel(label_value);
     }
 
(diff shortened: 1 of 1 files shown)


src/unity/lib/annotation/image_classification.cpp
```diff
@@ -95,11 +95,17 @@ annotate_spec::Annotations ImageClassification::getAnnotations(size_t start,
 
     label->imageclassificationlabel();
```
nickjong: Aside: isn't this a no-op?

src/unity/lib/annotation/image_classification.cpp
```diff
@@ -95,11 +95,17 @@ annotate_spec::Annotations ImageClassification::getAnnotations(size_t start,
...
 
-    if (flex_data.at(i).get_type() == flex_type_enum::STRING) {
-      std::string label_value = flex_data.at(i).get<flex_string>();
+    const flexible_type& flex_label = flex_data.at(i);
+    if (flex_label.get_type() == flex_type_enum::UNDEFINED) {
+      // skip unlabeled items
+      continue;
```
nickjong: To confirm: it's intentional that we added a new (default-valued) `annotate_spec::Label` to the `annotate_spec::Annotation`, without adding a row index?

src/unity/lib/annotation/image_classification.cpp
```diff
@@ -95,11 +95,17 @@ annotate_spec::Annotations ImageClassification::getAnnotations(size_t start,
...
+    }
+
+    if (flex_label.get_type() == flex_type_enum::STRING) {
+      std::string label_value = flex_label.get<flex_string>();
```
nickjong: Nit: can avoid a potential copy by using `const std::string&` here instead of `std::string`

src/unity/lib/annotation/image_classification.cpp
```diff
@@ -434,17 +440,24 @@ annotate_spec::MetaData ImageClassification::metaData() {
...
+    if (label.get_type() == flex_type_enum::UNDEFINED) {
+      // skip unlabeled items
+      continue;
+    }
+
     if (array_type == flex_type_enum::STRING) {
       annotate_spec::MetaLabel *labels_meta =
           image_classification_meta->add_label();
-      labels_meta->set_stringlabel(label_vector.at(x).to<std::string>());
+      labels_meta->set_stringlabel(label.to<std::string>());
```
nickjong: I think `get` here could avoid constructing a new `std::string` object, too

src/unity/lib/annotation/image_classification.cpp
znation: Not if it's in a `oneof` that hasn't yet had any of its members selected. `oneofs` are really zero-of-until-accessed... accessing one member makes it the one. By design in proto3.

src/unity/lib/annotation/image_classification.cpp
znation: I don't think that was my intention. I'll revisit this logic.

src/unity/lib/annotation/image_classification.cpp
nickjong: Strange... the implementation appears to be a const method that I don't believe sets the one-of state: https://github.com/apple/turicreate/blob/5e960c084f236195a55b60bb7ba3ab0e650c87d5/src/unity/lib/annotation/build/format/cpp/annotate.pb.cc#L1984

If the intent is to set the one of, could we use the more idiomatic `mutable_imageclassificationlabel` instead? We can see that that method sets the field explicitly:
https://github.com/apple/turicreate/blob/5e960c084f236195a55b60bb7ba3ab0e650c87d5/src/unity/lib/annotation/build/format/cpp/annotate.pb.cc#L1993

znation: Actually, that's probably what was intended. I'll change it.

znation:
@nickjong Feedback addressed; please re-review, thanks!

PR #2016, opened 2019-05-31
Bump version to 5.6

znation:


scripts/update_version_number.sh
```diff
@@ -45,10 +45,10 @@ set_version() {
   # set the engine version string
   cd ${WORKSPACE}/src/unity/lib
   # C++, replace "1.8"//{{VERSION_STRING}} with "new_version_string"//{{VERSION_STRING}}"
-  sed -i -e "s/\".*\"\/\/#{{VERSION_STRING}}/\"${VERSION_NUMBER}\"\/\/#{{VERSION_STRING}}/g" version_number.hpp
+  sed -i '' -e "s/\".*\"\/\/#{{VERSION_STRING}}/\"${VERSION_NUMBER}\"\/\/#{{VERSION_STRING}}/g" version_number.hpp
```
TobyRoseman: This and the other change in this file prevents backup files from getting created?

znation: Yes - there is really no reason to have separate backup files since we're editing files that are already tracked by git. The backup is the repo history (and/or the git staging mechanism). I always end up deleting the backup files that get created without looking at them, so I'm adding this change to prevent them from getting created.

PR #2020, opened 2019-05-31
Added draw_bounding_boxes for one_shot_object_detector

shantanuchhabra:


src/unity/python/turicreate/toolkits/one_shot_object_detector/util/_visualization.py
```diff
@@ -0,0 +1,32 @@
...
+        Confidence threshold can limit the number of boxes to draw. By
+        default, this is set to 0, since the prediction may have already pruned
+        with an appropriate confidence threshold.
+
+    Returns
+    -------
+    annotated_images: SArray or Image
+        Similar to the input `images`, except the images are decorated with
+        boxes to visualize the object instances.
+
```
TobyRoseman: I think it would be very helpful to have a few lines of example code showing how to take a one shot model and generate predictions with bounding boxes. Something like:
```
pred = model.predict(data)
predictions_with_bounding_boxes = tc.one_shot_object_detector.util.draw_bounding_boxes(data['images'], pred)
predictions_with_bounding_boxes.explore()
```

shreyajain17:
Can you also add it to the API docs by adding it in .rst file

src/unity/python/turicreate/toolkits/one_shot_object_detector/util/_visualization.py
shantanuchhabra: Done!

shantanuchhabra:
> Can you also add it to the API docs by adding it in .rst file

Done.

PR #1999, opened 2019-05-30
[WIP] Added unit tests for one-shot object detection

srikris:
Fixes #1788

PR #2021, opened 2019-05-31
Unit tests for one shot object detector

### LINKED ISSUE #1788: Unit tests for One Shot Object Detection
Assign to Project 5.6 Sprint 2 when board becomes available.

shreyajain17:
fixes #1788

### DIFF
--- a/src/unity/python/turicreate/test/test_one_shot_object_detector.py
+++ b/src/unity/python/turicreate/test/test_one_shot_object_detector.py
@@ -45,7 +45,7 @@ def _get_data(feature, target):
     classes = _CLASSES
     images = []
     FORMATS = ['png', 'jpeg', 'raw']
-    for i in range(num_examples):
+    for _ in range(num_examples):
         # Randomly determine image size (should handle large and small)
         img_shape = tuple(rs.randint(100, 1000, size=2)) + (3,)
         img = rs.randint(255, size=img_shape)
@@ -101,13 +101,9 @@ class OneObjectDetectorSmokeTest(unittest.TestCase):
                                                         batch_size=2,
                                                         max_iterations=1)
 
-        ## Answers
-        self.opts = self.def_opts.copy()
-        self.opts['max_iterations'] = 1
-
         self.get_ans = {
            'target': lambda x: x == self.target,
-           'num_starter_images': lambda x: x > 0,
+           'num_starter_images': lambda x: len(self.train),
            'num_classes': lambda x: x == len(_CLASSES),
            'detector': lambda x: isinstance(x, tc.object_detector.object_detector.ObjectDetector),
            '_detector_version': lambda x: x==1
@@ -125,8 +121,6 @@ class OneObjectDetectorSmokeTest(unittest.TestCase):
 
     def test_predict(self):
         sf = self.test.head()
-        # Make sure this does not need the annotations column to work
-
         pred = self.model.predict(sf.head())
 
         # Check the structure of the output
@@ -182,8 +176,6 @@ class OneObjectDetectorSmokeTest(unittest.TestCase):
             self.assertEqual(ret['confidence'].shape[1], len(_CLASSES))
             self.assertEqual(ret['coordinates'].shape[0], 
                 ret['confidence'].shape[0])
-            # A numeric comparison of the resulting of top bounding boxes is
-            # not that meaningful unless the model has converged
 
         # Also check if we can train a second model and export it (there could
         # be naming issues in mxnet)
@@ -192,7 +184,7 @@ class OneObjectDetectorSmokeTest(unittest.TestCase):
         # class
         sf = tc.SFrame({'image': tc.SArray([self.train[self.feature][0]]),
                         'label': tc.SArray([self.train[self.target][0]])})
-        model2 = tc.object_detector.create(sf, max_iterations=1)
+        model2 = tc.one_shot_object_detector.create(sf, 'label', max_iterations=1)
         model2.export_coreml(filename2, 
             include_non_maximum_suppression=False)
 
@@ -222,12 +214,7 @@ class OneObjectDetectorSmokeTest(unittest.TestCase):
         with test_util.TempDirectory() as filename:
             self.model.save(filename)
             self.model = tc.load_model(filename)
-
             self.test_predict()
-            print("Predict passed")
             self.test_get()
-            print("Get passed")
             self.test_summary()
-            print("Summary passed")
             self.test__list_fields()
-            print("List fields passed")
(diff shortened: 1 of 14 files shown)


TobyRoseman:
Are we abandoning #1999?

TobyRoseman:
How long does this test take to run?

shreyajain17:
@TobyRoseman  12 passed in 61.26 seconds

src/unity/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -0,0 +1,233 @@
...
+                pil_img.save(f, format=image_format)
+                return tc.Image(f.name)
+
+    num_examples = 100
+    num_starter_images = 5
+    max_num_boxes_per_image = 10
+    classes = _CLASSES
+    images = []
+    FORMATS = ['png', 'jpeg', 'raw']
+    for i in range(num_examples):
```
TobyRoseman: nit: in Python when a variable is never intended to actually be used it's standard to name the variable `_`.

src/unity/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -0,0 +1,233 @@
...
+        # Model
+        self.train, self.test, self.backgrounds = _get_data(feature=self.feature, target=self.target)
+        self.model = tc.one_shot_object_detector.create(self.train,
+                                                        target=self.target,
+                                                        backgrounds=self.backgrounds,
+                                                        batch_size=2,
+                                                        max_iterations=1)
+
+        ## Answers
+        self.opts = self.def_opts.copy()
```
TobyRoseman: It looks like `self.opts` is never getting used. . . . which means you don't need this line or the next line or the declaration of `self.def_opts`.

src/unity/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -0,0 +1,233 @@
...
+        self.opts['max_iterations'] = 1
+
+        self.get_ans = {
+           'target': lambda x: x == self.target,
+           'num_starter_images': lambda x: x > 0,
```
TobyRoseman: Why not `lambda x: x == len(self.train)`?

src/unity/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -0,0 +1,233 @@
...
+            tc.one_shot_object_detector.create(self.train, target='wrong_feature', 
+                backgrounds=self.backgrounds, max_iterations=1)
+
+    def test_create_with_empty_dataset(self):
+        with self.assertRaises(_ToolkitError):
+            tc.one_shot_object_detector.create(self.train[:0], target=self.target)
+
+    def test_predict(self):
+        sf = self.test.head()
+        # Make sure this does not need the annotations column to work
```
TobyRoseman: I don't understand this comment. The object detector's prediction function doesn't even take an annotation column.

src/unity/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -0,0 +1,233 @@
...
+        img_fixed = tc.image_analysis.resize(img, 416, 416, 3)
+        pil_img = Image.fromarray(img_fixed.pixel_data)
+        if _mac_ver() >= (10, 13):
+            ret = coreml_model.predict({self.feature: pil_img}, 
+                usesCPUOnly = True)
+            self.assertEqual(ret['coordinates'].shape[1], 4)
+            self.assertEqual(ret['confidence'].shape[1], len(_CLASSES))
+            self.assertEqual(ret['coordinates'].shape[0], 
+                ret['confidence'].shape[0])
+            # A numeric comparison of the resulting of top bounding boxes is
```
TobyRoseman: I would be very curious to understand why.

src/unity/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -0,0 +1,233 @@
...
+        self.assertEqual(type(str(model)), str)
+        self.assertEqual(type(model.__repr__()), str)
+
+    def test_save_and_load(self):
+        with test_util.TempDirectory() as filename:
+            self.model.save(filename)
+            self.model = tc.load_model(filename)
+
+            self.test_predict()
+            print("Predict passed")
```
TobyRoseman: I would remove these print statements

src/unity/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -59,6 +59,7 @@ def create(data,
         # Make predictions on the training set and as column to the SFrame
         >>> test_data['predictions'] = model.predict(test_data)
     """
+    _tkutl._raise_error_if_column_exists(data, target, "dataset", target)
```
nickjong: The variable name is `data` so the third argument, `dataset_variable_name`, should be `"data"`

src/unity/python/turicreate/test/test_one_shot_object_detector.py
shreyajain17: Yeah I deleted the code but forgot to delete the comment.

src/unity/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -0,0 +1,220 @@
...
+    @classmethod
+    def setUpClass(self):
+        """
+        The setup class method for the basic test case with all default values.
+        """
+        self.feature = 'myimage'
+        self.target = 'mytarget'
+
+        ## Create the model
+        self.def_opts = {
```
TobyRoseman: This isn't being used.

nickjong:
> I think something might be messed up with the history here; I don't think you want the changes outside of `test_one_shot_object_detector.py`

Note that the changes in `one_shot_object_detector.py` and `one_shot_object_detector.cpp` fix actual bugs revealed by the new tests.

nickjong:
It looks like `test_export_coreml` is failing in Travis (with `std::bad_alloc`) or simply not returning (after ten minutes). Does this test succeed locally?

znation:
From an in-person discussion with @shreyajain17, merging this now so we can get the toolkit fixes in, and we'll fix unit tests in master if they are still failing there. It's likely bad_alloc in Travis is a result of Travis being RAM-constrained, and not a real bug (we already disable other tests on Travis for this reason).

shreyajain17:
> It looks like `test_export_coreml` is failing in Travis (with `std::bad_alloc`) or simply not returning (after ten minutes). Does this test succeed locally (on Mac and Linux)?

Yes locally passes everything on Mac

Issue #2005, opened 2019-05-31
One Shot - needs better tar file management for SArray background images


TobyRoseman:
There are two related issues here:

1 - Each time `tc.one_shot_object_detector.create(...)` is called the tar file is always extracted, even if those files have already been extracted. The extraction takes a little over 30 seconds. So we shouldn't be doing it if we don't need to do it.

2 - The files are extracted to the **user's current directory** (in an SArray named `one_shot_backgrounds.sarray`). We should be extracting these files to somewhere in the Turi Create's data cache (i.e. somewhere under the value return by `tc.toolkits._pre_trained_models._get_cache_dir(_type="data")`).

PR #2022, opened 2019-06-01
Add progress printing for one-shot augmentation

### LINKED ISSUE #2019: One-shot OD hangs for 5+ minutes while preparing augmented data
The one-shot OD toolkit gives no user output while doing augmentation; this can take at least 5 minutes, and up to about an hour. During this process, there is no indication anything is happening and the Python prompt appears to have hung.

We should use a `table_printer` with progress to give progress output here.

znation:
This will give the user some feedback while augmentation is happening.
There is still quite a delay (30s to 1m in debug) before the table even
shows, which I am not sure how to account for, but it's a lot better
than without any table printing.

Fixes #2019

### DIFF
--- a/src/unity/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
+++ b/src/unity/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
@@ -127,11 +127,22 @@ gl_sframe augment_data(const gl_sframe &data,
                        const std::string& image_column_name,
                        const std::string& target_column_name,
                        const gl_sarray &backgrounds,
-                       long long seed) {
+                       long long seed,
+                       bool verbose) {
+  size_t backgrounds_size = backgrounds.size();
+  size_t total_augmented_rows = data.size() * backgrounds_size;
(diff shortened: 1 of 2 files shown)


TobyRoseman:
> There is still quite a delay (30s to 1m in debug) before the table even
> shows, which I am not sure how to account for, but it's a lot better
> than without any table printing.

I believe this can be explained by the first issue mentioned in #2005.

src/unity/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -127,11 +127,22 @@ gl_sframe augment_data(const gl_sframe &data,
...
+  if (verbose) {
+    logprogress_stream << "Augmenting input images using " << backgrounds.size() << " background images." << std::endl;
+    table.print_header();
+  }
+
   auto column_index_map = generate_column_index_map(data.column_names());
   std::vector<flexible_type> annotations;
   std::vector<flexible_type> images;
+  int input_row_index = -1;
   for (const auto& row: data.range_iterator()) {
```
nickjong: We’re iterating over all the rows in a single for loop? This really should use `gl_sframe::apply`. Would be 8-16 times faster on most machines.

Probably too late for 5.6 though. Filed #2023

PR #2028, opened 2019-06-03
Added drawing classifier annotate

abhishekpratapa:
- added drawing classifier annotate


znation:
@abhishekpratapa Does this need to go into 5.6?

Issue #1817, opened 2019-04-30
object detection : pls support higher image resolutions


romanoss:
as I found out the tiny yolo network has a quite poor detection performance for my project  compared to the much bigger full yolo model. but when you train with a param setting of 608x608 for image size, its gets quite better.
For me the tiny yolo model is rather useless except when trained with higher image resolutions (I tried this with darknet from AlexeyAB with images size 608 and parameter random=1. which leaded to a 3 times higher training time but far better results).
So I would say you really should provide this and other settings as a param for training.

srikris:
We are exploring the use or larger networks for users who have more memory and computational budget. We'd love to hear your use case and deployment scenario more.

romanoss:
I just try to build a small app that can recognize tropical fish.
When visiting an aquarium or when people are snorkeling, one is usually overwhelmed by the richness and diversity of the species one is watching. So giving the people an app which can recognize a species and provide further information about it, could be a nice additional experience.

pusateri:
Does this apply to the image similarity model as well? I am noticing poor matching for trained images and wonder if I can force it to use higher quality images.

romanoss:
I did not try - sorry

lp55:
I'm also have the same problem. For small object detection, the resolution really affects precision. The object_detector.create should really have a resolution parameter. I'm gonna take a look at the code and see if this is easy to implement. Any hints?

Random would be nice, but that could be done when building the sframe, so it's not critical.

romanoss:
probably in object_detector.cpp: GRID_SIZE*SPATIAL_REDUCTION results in  416

lp55:
Yes, I did some searching / reading and found out about the grid_size variable on turicreate\src\unity\toolkits\object_detection\object_detector.cpp and on turicreate\src\unity\python\turicreate\toolkits\object_detector\object_detector.py. Seems pretty trivial to make that a param in the create function. Any catches? Anything else I have to change?

I'll fork this, change the code and create a pull request :)

romanoss:
I dont know about memory implications

did u manage to use a retrained yolo model in an android app and if yes how :) ?

romanoss:
I am currently trying to train a mobilenet with tf and going from there to an ios and android model

nickjong:
[...]

PR #2053, opened 2019-06-13
Higher resolution support

lp55:
Hi,

   As mentioned on [#1817](https://github.com/apple/turicreate/issues/1817) I changed the python code to have an option to enable higher resolution inputs. I only altered the python code, as mentioned that the toolkit is in a bit of a transient state.
   The C++ code would need deeper changes from what I checked. The static grid shape is deeper in the code. But for now this works.


src/unity/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -96,8 +96,8 @@ def _raise_error_if_not_detection_sframe(dataset, feature, annotations, require_
 
 
 def create(dataset, annotations=None, feature=None, model='darknet-yolo',
-           classes=None, batch_size=0, max_iterations=0, verbose=True,
-           **kwargs):
+           grid_shape=[13, 13], classes=None, batch_size=0,
```
znation: To avoid breaking existing API calls that rely on argument ordering, please add `grid_shape` last after `verbose` (and before `**kwargs`). Python allows explicit kwargs to be passed either by keyword or argument ordering, so existing calls may be assuming that the 5th argument is `classes` (and so on).

src/unity/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -221,7 +226,6 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
             (8.0, 16.0), (8.0, 8.0), (16.0, 8.0),
             (16.0, 32.0), (16.0, 16.0), (32.0, 16.0),
         ],
-        'grid_shape': [13, 13],
```
znation: In case this parameter is needed elsewhere outside this function, it's probably safer to keep it in the `params` dictionary (which gets passed around to various other places). You could keep it as  `'grid_shape': grid_shape` to use the user-provided value.

lp55:
Hi,

   I made the requested changes. Is it ok now?

PR #2055, opened 2019-06-13
Code re-organization for better usability.

hoytak:
Our source tree is difficult to navigate for new developers and reflects a very evolutionary process.  I propose reworking our directory structure in order to better accommodate the functionality so that common features are easily found.

No current functionality change is proposed.

> High level: divide by functionality.
src/core                     ← All of the lower level data structures the rest are built on.
src/python               ← Now at the top level.  Doesn't change outside import directories.
src/capi                     ← C API.
src/ml                     <-- ML tools to be consumed by the toolkits. 
src/toolkits             ← All the ml tasks.
src/model_server         ← Was unity_server.    Also contains all the model base classes.
src/visualization        ← Visualization
src/external             ← External dependencies
src/interfaces           <- tools for interfacing to other languages, e.g. ObjC.

> src/core → Move core functionality into src/core subdirectories.
src/core/storage
src/core/json
src/core/ml
src/core/generics
src/core/globals
src/core/parallel
src/core/logging
src/core/util
src/core/system
src/core/random
src/core/data
src/core/config
src/core/export.hpp
src/core/data -> Core data parts (flexible type, image, gl_s*)
src/core/storage → All of the internal disk storage associated with the lower level sframe stuff.  Different from the high-level sframe, sarray, and sgraph APIs above.

> src/core/system →. Really low-level stuff. Subsumes platform subdirectories.
src/core/system/cppipc
src/core/system/fault
src/core/system/exceptions
src/core/system/crash_handler
src/core/system/network
src/core/system/perf
src/core/system/shmipc
src/core/system/platform
src/core/system/minipsutil
src/core/system/process
src/core/system/so_utils

> src/core/ml → Move foundational ML tools needed as components of many different toolkits, etc.
src/core/ml/optimization
src/core/ml/ml_data
src/core/ml/neural_net
src/core/ml/sketches


znation:
> If we get these in separately, then it will make collisions less likely.

I am not following the logic here, and I don't think we should break master. Can we put two separate PRs into a branch (i.e. review them separately), then merge that whole branch into master with 2 commits in it?

rename_script.py
```diff
@@ -0,0 +1,82 @@
+import re
```
znation: I assume this file shouldn't be committed? This rename should only happen once.

znation:
> src/interfaces <- tools for interfacing to other languages, e.g. ObjC.

Not sure this one makes sense to me. Most of the directories are assumed to contain C++ code. Other language bindings like `python` and `capi` are still top level. Why wouldn't we want `src/objc` at the top level, or `src/xyz` for language `xyz`?

nickjong:
> > If we get these in separately, then it will make collisions less likely.
> 
> I am not following the logic here, and I don't think we should break master. Can we put two separate PRs into a branch (i.e. review them separately), then merge that whole branch into master with 2 commits in it?

Despite what the PR description says, I believe that's what we have here. There are two commits, and the first appears to be pure renames ("File renamed without changes."). The second commit gets things to recompile.

The only sane way to review this PR is to look at the two diffs separately. This can be done in the Files Changed tab by clicking where it says "Changes from all commits"

nickjong:
> > src/interfaces <- tools for interfacing to other languages, e.g. ObjC.
> 
> Not sure this one makes sense to me. Most of the directories are assumed to contain C++ code. Other language bindings like `python` and `capi` are still top level. Why wouldn't we want `src/objc` at the top level, or `src/xyz` for language `xyz`?

I think it's confusing to equate our ObjC code with the Python / C-API stuff. The latter is about bindings exposing the C++ functionality. The latter is code specific to Core ML integration, that the user integrates into their app.

znation:
> I think it's confusing to equate our ObjC code with the Python / C-API stuff. The latter is about bindings exposing the C++ functionality. The latter is code specific to Core ML integration, that the user integrates into their app.

Fair enough, but I think that's an argument in favor of ObjC code in each toolkit, instead of consolidated into `src/interfaces`.

scripts/repo
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python
```
znation: Should this file have a .py extension?

hoytak:
> > src/interfaces <- tools for interfacing to other languages, e.g. ObjC.
> 
> Not sure this one makes sense to me. Most of the directories are assumed to contain C++ code. Other language bindings like `python` and `capi` are still top level. Why wouldn't we want `src/objc` at the top level, or `src/xyz` for language `xyz`?

This one is mostly about our code calling into objc, not calling our code from objc.

znation:
> > > src/interfaces <- tools for interfacing to other languages, e.g. ObjC.
> > 
> > 
> > Not sure this one makes sense to me. Most of the directories are assumed to contain C++ code. Other language bindings like `python` and `capi` are still top level. Why wouldn't we want `src/objc` at the top level, or `src/xyz` for language `xyz`?
> 
> This one is mostly about our code calling into objc, not calling our code from objc.

I see, in that case, I think this organization makes sense.

znation:
> > > > src/interfaces <- tools for interfacing to other languages, e.g. ObjC.
> > > 
> > > 
> > > Not sure this one makes sense to me. Most of the directories are assumed to contain C++ code. Other language bindings like `python` and `capi` are still top level. Why wouldn't we want `src/objc` at the top level, or `src/xyz` for language `xyz`?
> > 
> > 
> > This one is mostly about our code calling into objc, not calling our code from objc.
> 
> I see, in that case, I think this organization makes sense.

Actually, looking at the code here, I don't think this is the case -- `src/interfaces/objc` builds two dylibs, one of which is completely standalone, the other of which calls into Turi Create C/C++ API from ObjC. In neither case is it our code calling into ObjC... in both cases, it's our code providing an ObjC interface externally.

nickjong:
> > > > > src/interfaces <- tools for interfacing to other languages, e.g. ObjC.
> > > > 
> > > > 
> > > > Not sure this one makes sense to me. Most of the directories are assumed to contain C++ code. Other language bindings like `python` and `capi` are still top level. Why wouldn't we want `src/objc` at the top level, or `src/xyz` for language `xyz`?
> > > 
> > > 
> > > This one is mostly about our code calling into objc, not calling our code from objc.
> > 
> > 
> > I see, in that case, I think this organization makes sense.
> 
> Actually, looking at the code here, I don't think this is the case -- `src/interfaces/objc` builds two dylibs, one of which is completely standalone, the other of which calls into Turi Create C/C++ API from ObjC. In neither case is it our code calling into ObjC... in both cases, it's our code providing an ObjC interface externally.

Yes, as I said above, Objective C is an implementation detail of some existing libraries already (that were not initially moved into the directory in question in this PR). It doesn't make sense to name this directory by language. These delis are for integration into developers' app. The language they use is incidental.

hoytak:
> Please rename src/interfaces, since this directory does not contain interfaces

Fixed! Now in src/objc.   Also, it passes all the internal tests -- can we merge it?

nickjong:
> > Please rename src/interfaces, since this directory does not contain interfaces
> 
> Fixed! Now in src/objc. Also, it passes all the internal tests -- can we merge it?

That's no worse than what we had before, although honestly most of our ObjC code lies outside this directory. Really this should be called something like `deployment_support` or `coreml_support` or `coreml_custom_models`, since all the code in here is for defining classes that implement Core ML's `MLCustomModel` protocol.

scripts/repo.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python
```
TobyRoseman: Did you mean to include this file? I'm not sure it make sense to include this in our repository.

hoytak: Yes.  It's a script that's a collection of tools to make managing files in the repo easier.   E.g. moving headers.

PR #2059, opened 2019-06-17
Expose random split in C-API of activity classifier

shreyajain17:
It works as the following now -
```
train, test = tc.extensions._activity_classifier_random_split_by_session(data, 'session_id', 0.9, 1)
```


src/toolkits/activity_classification/class_registrations.cpp
```diff
@@ -17,6 +17,25 @@ namespace activity_classification {
 BEGIN_FUNCTION_REGISTRATION
 REGISTER_FUNCTION(_activity_classifier_prepare_data , "data" , "features" , "session_id" , "prediction_window" , "predictions_in_chunk" , "target")
 REGISTER_FUNCTION(_activity_classifier_prepare_data_verbose , "data" , "features" , "session_id" , "prediction_window" , "predictions_in_chunk" , "target")
+REGISTER_FUNCTION(activity_classifier::random_split_by_session, "data",
```
nickjong: I think `tc.extensions.random_split_by_session` might be a bit too generic. Could we try `tc.extensions.activity_classifier.random_split_by_session`? I believe you should be able to accomplish that using a macro called something like REGISTER_NAMED_FUNCTION, and the name "activity_classifier.random_split_by_session".

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -204,6 +171,42 @@ void activity_classifier::init_options(
   add_or_update_state(flexmap_to_varmap(options.current_option_values()));
 }
 
+// Randomly split an SFrame into two SFrames based on the `session_id` such
```
nickjong: Nit: In this case, I think we may as well have the static function come first in this file, just after the anonymous namespace. This way, it doesn't break up the sequence of instance methods.

src/toolkits/activity_classification/class_registrations.cpp
shreyajain17: I changed it to a REGISTER_NAMED_FUNCTION. With this change it compiles but does not work. I looked up the definition of the macro and I think it is suppose to work this way. Can you look at the change and let me know if there is something I am missing.

src/toolkits/activity_classification/class_registrations.cpp
```diff
@@ -17,6 +17,24 @@ namespace activity_classification {
 BEGIN_FUNCTION_REGISTRATION
 REGISTER_FUNCTION(_activity_classifier_prepare_data , "data" , "features" , "session_id" , "prediction_window" , "predictions_in_chunk" , "target")
 REGISTER_FUNCTION(_activity_classifier_prepare_data_verbose , "data" , "features" , "session_id" , "prediction_window" , "predictions_in_chunk" , "target")
+REGISTER_NAMED_FUNCTION("_activity_classifier_random_split_by_session", activity_classifier::random_split_by_session, "data","session_id_column_name", "fraction", "seed");
```
nickjong: Let's use "session_id" instead of "session_id_column_name", for consistency with the `train` method (C API naming) on `activity_classifier`.

shreyajain17: train method uses ```session_id_column_name```

shreyajain17: Got it

PR #2060, opened 2019-06-17
Data Augmentation for C-API activity classifier

shreyajain17:
In activity classification, say prediction window is 100 and lstm sequence length is 20, so the batch row is 20 x 100 = 2000 and there are 32 batch rows in a batch (if the batch size is 32). So per batch we have 32 x 100 x 20 x #features for features and 32 x 20 labels (majority voting done in every prediction window). Originally (without data augmentation), we would preprocess it in chunks of 100 x 20 x #features features and 20 labels. Now, the number of rows (=batch size) were picked up and passed as a batch by the iterator. So here in every epoch we were going to pick up the same chunked up data 0-1999, 2000-3999, 4000-5999 and so on.

The goal of data augmentation was to have different prediction windows in each iteration. To implement data augmentation, we have chunked up data to save it as the whole session features and the whole session labels. Let's assume number of data points in session 1 are n and in session 2 are m. Then, the converted sframe would look like this -

| features                |      labels   |
| n x #features.       |       n         |
| m x #features       |        m       |
and so on..

For example let's say prediction window = 100, then we would chunk this session row once with a random offset between 0-99. In each iteration, each session is picked up with a different random offset. Assuming we picked up 23, then the data picked up is 23-2022, 2023-4022, and so on. For next session, a new random offset is picked.

### DIFF
--- a/src/toolkits/activity_classification/ac_data_iterator.cpp
+++ b/src/toolkits/activity_classification/ac_data_iterator.cpp
@@ -281,6 +281,143 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
     return result_dict;
 }
 
+variant_map_type _activity_classifier_prepare_data_aug_impl(
+    const gl_sframe &data, const std::vector<std::string> &features,
+    const std::string &session_id, int prediction_window,
+    int predictions_in_chunk, const std::string &target, bool verbose) {
+
+#ifndef NDEBUG
+  DASSERT_GT(features.size(), 0);
+  DASSERT_TRUE(prediction_window > 0);
+  DASSERT_TRUE(predictions_in_chunk > 0);
+  DASSERT_TRUE(data.contains_column(session_id));
+  for (auto &feat : features) {
+    DASSERT_TRUE(data.contains_column(feat));
+  }
+#endif
+
+  bool use_target = (target != "");
+  DASSERT_TRUE(!use_target || data.contains_column(target));
+
+  if (verbose) {
+    logprogress_stream << "Pre-processing " << data.size() << " samples..."
+                       << std::endl;
+  }
+
+  // Build a dict of the column order by column name, to later access within
+  // the iterator
+  auto column_index_map = generate_column_index_map(data.column_names());
+  size_t chunk_length = prediction_window * predictions_in_chunk;
+
+  flex_vec curr_chunk_targets;
+  flex_vec curr_chunk_features;
+
+  flexible_type last_session_id = data[session_id][0];
+  size_t number_of_sessions = 0;
+
+  // Prepare an output SFrame writer, that will write a new SFrame in the
+  // converted batch-processing ready format.
+  std::vector<std::string> output_column_names = {"features", "chunk_len",
+                                                  "session_id"};
+  std::vector<flex_type_enum> output_column_types = {flex_type_enum::VECTOR,
+                                                     flex_type_enum::INTEGER,
+                                                     data[session_id].dtype()};
+  if (use_target) {
+    output_column_names.push_back("target");
+    output_column_types.push_back(flex_type_enum::VECTOR);
+  }
+
+  gl_sframe_writer output_writer(output_column_names, output_column_types, 1);
+
+  if (verbose) {
+    logprogress_stream << "Using sequences of size " << chunk_length
+                       << " for model creation." << std::endl;
+  }
+
+  time_t last_print_time = time(0);
+  size_t processed_lines = 0;
+
+  // Iterate over the user data. The features and targets are aggregated, and
+  // handled whenever at the ending of a session is reached.
+  size_t chunk_size = 0;
+  for (const auto &line : data.range_iterator()) {
+    flexible_type curr_session_id = line[column_index_map[session_id]];
+
+    if (curr_session_id != last_session_id) {
+
+      // Write the aggregated data of the current chunk (which includes all
+      // the samples in the current session) as a single new vector in the
+      // converted SFrame, and init all aggregation vectors to begin a new
+      // chunk.
+      if (curr_chunk_features.size() > 0) {
+        if (use_target) {
+          output_writer.write({curr_chunk_features, chunk_size, last_session_id,
+                               curr_chunk_targets},
+                              0);
+        } else {
+          output_writer.write(
+              {curr_chunk_features, chunk_size, last_session_id}, 0);
+        }
+        curr_chunk_features.clear();
+        curr_chunk_targets.clear();
+      }
+
+      chunk_size = 0;
+      last_session_id = curr_session_id;
+      number_of_sessions++;
+    }
+
+    chunk_size += 1;
+
+    for (const std::string &feature_name : features) {
+      curr_chunk_features.push_back(line[column_index_map[feature_name]]);
+    }
+
+    if (use_target) {
+      curr_chunk_targets.push_back(line[column_index_map[target]]);
+    }
+
+    time_t now = time(0);
+    if (verbose && difftime(now, last_print_time) > 10) {
+      logprogress_stream << "Pre-processing: " << std::setw(3)
+                         << (100 * processed_lines / data.size())
+                         << "% complete" << std::endl;
+      last_print_time = now;
+    }
+    processed_lines += 1;
+  }
+
+
+  if (curr_chunk_features.size() > 0) {
+    if (use_target) {
+      output_writer.write({curr_chunk_features, chunk_size, last_session_id,
+                           curr_chunk_targets},
+                          0);
+    } else {
+      output_writer.write({curr_chunk_features, chunk_size, last_session_id},
+                          0);
+    }
+    curr_chunk_features.clear();
+    curr_chunk_targets.clear();
+  }
+
+  // Update the count of the last session in the dataset
+  number_of_sessions++;
+
+  if (verbose) {
+    logprogress_stream << "Processed a total of " << number_of_sessions
+                       << " sessions." << std::endl;
+  }
+  gl_sframe converted_sframe = output_writer.close();
+  converted_sframe.materialize();
+
+  variant_map_type result_dict;
+  result_dict["converted_data"] = converted_sframe;
+  result_dict["num_of_sessions"] = number_of_sessions;
+
+  return result_dict;
+}
+
 }  // namespace
 
 variant_map_type _activity_classifier_prepare_data( const gl_sframe &data,
@@ -352,14 +489,12 @@ simple_data_iterator::preprocessed_data simple_data_iterator::preprocess_data(
   }
 
   // Chunk the data, so that each row of the resulting SFrame corresponds to a
-  // sequence of up to predictions_in_chunk prediction windows (from the same
-  // session), each comprising up to prediction_window rows from the original
-  // SFrame.
-  variant_map_type result_map = _activity_classifier_prepare_data_impl(
+  // all samples in a session from the original SFrame.
+  variant_map_type result_map = _activity_classifier_prepare_data_aug_impl(
       data, feature_column_names, params.session_id_column_name,
       static_cast<int>(params.prediction_window),
       static_cast<int>(params.predictions_in_chunk), params.target_column_name,
-      /* verbose */ params.verbose);
+      /* verbose */ params.is_train);
 
   preprocessed_data result;
   result.chunks = variant_get_value<gl_sframe>(result_map.at("converted_data"));
@@ -373,14 +508,16 @@ simple_data_iterator::preprocessed_data simple_data_iterator::preprocess_data(
   return result;
 }
 
-simple_data_iterator::simple_data_iterator(const parameters& params)
-  : data_(preprocess_data(params)),
-    num_samples_per_prediction_(params.prediction_window),
-    num_predictions_per_chunk_(params.predictions_in_chunk),
-    range_iterator_(data_.chunks.range_iterator()),
-    next_row_(range_iterator_.begin()),
-    end_of_rows_(range_iterator_.end())
-{}
+simple_data_iterator::simple_data_iterator(const parameters &params)
+    : data_(preprocess_data(params)),
+      num_samples_per_prediction_(params.prediction_window),
+      num_predictions_per_chunk_(params.predictions_in_chunk),
+      range_iterator_(data_.chunks.range_iterator()),
+      next_row_(range_iterator_.begin()),
+      end_of_rows_(range_iterator_.end()),
+      sample_in_row_(0),
+      is_train_(params.is_train),
+      use_data_augmentation_(params.use_data_augmentation) {}
 
 const flex_list& simple_data_iterator::feature_names() const {
   return data_.feature_names;
@@ -400,6 +537,7 @@ bool simple_data_iterator::has_next_batch() const {
 
 data_iterator::batch simple_data_iterator::next_batch(size_t batch_size) {
 
+
   size_t num_samples_per_chunk =
       num_samples_per_prediction_ * num_predictions_per_chunk_;
   size_t num_features = data_.feature_names.size();
@@ -411,10 +549,9 @@ data_iterator::batch simple_data_iterator::next_batch(size_t batch_size) {
   size_t chunk_len_column_index = data_.chunks.column_index("chunk_len");
   size_t session_id_column_index = data_.chunks.column_index("session_id");
   size_t labels_column_index = 0;
-  size_t weights_column_index = 0;
+
   if (data_.has_target) {
     labels_column_index = data_.chunks.column_index("target");
-    weights_column_index = data_.chunks.column_index("weights");
   }
 
   // Allocate buffers for the resulting batch data.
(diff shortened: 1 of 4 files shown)


src/unity/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -36,95 +36,32 @@ static std::map<std::string,size_t> generate_column_index_map(const std::vector<
...
-
-        if(histogram.size() < (value + 1)){
-          histogram.resize(value + 1);
-        }
-
-        histogram[value]++;
-    }
+static double vec_mode(flex_vec::const_iterator first,
+                       flex_vec::const_iterator last) {
+  std::vector<int> histogram;
```
TobyRoseman: Looks like you're putting in `size_t` in the vector. It would probably be best if it was an `std::vector` of `size_t`.

src/unity/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -726,7 +731,11 @@ void activity_classifier::perform_training_iteration() {
     pop_until_size(1);
 
     result result_batch;
-    result_batch.data_info = training_data_iterator_->next_batch(batch_size);;
+    result_batch.data_info = training_data_iterator_->next_batch(batch_size);
+
+    // std::cout<< result_batch.data_info.features << "\n";
```
TobyRoseman: Let's remove commented out code.

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -1008,4 +1008,4 @@ gl_sframe activity_classifier::perform_inference(data_iterator *data) const {
   return writer.close();
 }
 }  // namespace activity_classification
-}  // namespace turi
+}  // namespace turi
```
TobyRoseman: Probably want to revert this change.

src/toolkits/activity_classification/ac_data_iterator.hpp
```diff
@@ -44,7 +44,7 @@ class data_iterator {
 
     /**
      * Each group of this many consecutive samples from the same session are
-     * assumed to have the same class label.
+     * assumed to have the same class label.
```
TobyRoseman: Nit: let's not add a space here or on line 82

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -487,4 +664,4 @@ void simple_data_iterator::reset() {
 }
 
 }  //activity_classification
-}  //turi
+}  //turi
```
TobyRoseman: Nit: generally we want files to end in a new line.

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -280,6 +280,153 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+
+        // If target column exists, the targets are aggregated for the duration
+        // of prediction_window.
+        // Each prediction_window subsampled into a single target value, by selecting
+        // the most frequent value (statistical mode) within the window.
+        if (use_target) {
+          curr_chunk_targets.push_back(line[column_index_map[target]]);
+        }
+
+        time_t now = time(0);
```
TobyRoseman: I think this line should be moved into the body of the if-statement.

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -280,6 +280,153 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+    }
+
+    if (verbose) {
+        logprogress_stream << "Pre-processing " << data.size() << " samples..." << std::endl;
+    }
+
+    // Build a dict of the column order by column name, to later access within
+    // the iterator
+    auto column_index_map = generate_column_index_map(data.column_names());
+    int chunk_length = prediction_window * predictions_in_chunk;
```
TobyRoseman: Would this make more sense as a `size_t` rather than an `int`?

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -280,6 +280,153 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+                                                        const int &prediction_window,
+                                                        const int &predictions_in_chunk,
+                                                        const std::string &target,
+                                                        bool verbose) {
+
+    DASSERT_TRUE(features.size() > 0);
+    DASSERT_TRUE(prediction_window > 0);
+    DASSERT_TRUE(predictions_in_chunk > 0);
+    DASSERT_TRUE(data.contains_column(session_id));
+    for (TURI_ATTRIBUTE_UNUSED_NDEBUG auto &feat : features) {
```
TobyRoseman: I would probably wraps this for-loop in a 
```
#ifndef NDEBUG
....
#endif
```
You don't want to be iterating over the features for no reason in release mode. Although perhaps the compiler would be smart enough not to let that happen.

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -280,6 +280,153 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+        DASSERT_TRUE(data.contains_column(feat));
+    }
+
+    bool use_target = (target != "");
+    if (use_target) {
```
TobyRoseman: Similar to the previous point, I would probably replace this entire if-statement with:
`DASSERT_TRUE(use_target && data.contains_column(target));`

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -280,6 +280,153 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+
+    gl_sframe_writer output_writer(output_column_names, output_column_types, 1);
+
+    if (verbose) {
+      logprogress_stream << "Using sequences of size " << chunk_length << " for model creation." << std::endl;
+    }
+
+    time_t last_print_time = time(0);
+    size_t processed_lines = 0;
+    int feature_size;
```
TobyRoseman: `size_t` here and for `chunk_size`?

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -372,14 +517,15 @@ simple_data_iterator::preprocessed_data simple_data_iterator::preprocess_data(
...
-{}
+size_t sample = 0;
+
+simple_data_iterator::simple_data_iterator(const parameters &params)
+    : data_(preprocess_data(params)),
+      num_samples_per_prediction_(params.prediction_window),
+      num_predictions_per_chunk_(params.predictions_in_chunk),
+      range_iterator_(data_.chunks.range_iterator()),
+      next_row_(range_iterator_.begin()), end_of_rows_(range_iterator_.end()),
+      sample_in_row_(0), training(params.verbose) {}
```
TobyRoseman: It seems strange to me that you're initializing `training` to the values of `params.verbose`.

src/toolkits/activity_classification/ac_data_iterator.cpp
shreyajain17: I think being outside is correct. Can you look at it once more? We want print progress if its been more than 10 seconds since the last time we printed.

src/toolkits/activity_classification/ac_data_iterator.cpp
shreyajain17: I renamed the parameter to ```is_train```. It does both but it makes sense to called ```is_train``` rather than ```verbose```.

src/toolkits/activity_classification/activity_classifier.hpp
```diff
@@ -32,7 +32,7 @@ class EXPORT activity_classifier: public ml_model_base {
 
   void train(gl_sframe data, const std::string& target_column_name,
              const std::string& session_id_column_name,
-             variant_type validation_data,
+             variant_type validation_data, bool has_data_augmentation,
```
nickjong: Could we move the parameter into the options dictionary? Let's also use "use_data_augmentation", since we're giving an instruction, not describing the current state of something

src/toolkits/activity_classification/ac_data_iterator.hpp
```diff
@@ -62,7 +62,10 @@ class data_iterator {
     flex_list class_labels;
 
     /**  Generates verbose output when set to true. */
-    bool verbose;
+    bool is_train;
```
nickjong: Do we need to update the comment?

src/toolkits/activity_classification/ac_data_iterator.hpp
```diff
@@ -61,8 +61,11 @@ class data_iterator {
      */
     flex_list class_labels;
 
-    /**  Generates verbose output when set to true. */
-    bool verbose;
+    /**  Set to true, when the data is used for training. */
+    bool is_train;
+
+    /** Augments training data when set to true*/
+    bool use_data_augmentation;
```
nickjong: For data members with primitive types, such as int, bool, and float, it’s often a good idea to give them a default value (such as for `predictions_in_chunk` above). This way, forgetting to set them later will yield a deterministic result, instead of getting some non deterministic, uninitialized value.

src/toolkits/activity_classification/ac_data_iterator.hpp
```diff
@@ -179,6 +182,9 @@ class simple_data_iterator: public data_iterator {
   gl_sframe_range range_iterator_;
   gl_sframe_range::iterator next_row_;
   gl_sframe_range::iterator end_of_rows_;
+  size_t sample_in_row_;
+  bool is_train_;
+  bool use_data_augmentation_;
```
nickjong: Let’s initialize these too. Unless another value makes sense, you can use `false` or 0 as the default initializer.

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -372,14 +509,18 @@ simple_data_iterator::preprocessed_data simple_data_iterator::preprocess_data(
...
 
-simple_data_iterator::simple_data_iterator(const parameters& params)
-  : data_(preprocess_data(params)),
-    num_samples_per_prediction_(params.prediction_window),
-    num_predictions_per_chunk_(params.predictions_in_chunk),
-    range_iterator_(data_.chunks.range_iterator()),
-    next_row_(range_iterator_.begin()),
-    end_of_rows_(range_iterator_.end())
-{}
+size_t sample = 0;
```
nickjong: Debugging code? No longer needed?

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -425,39 +566,70 @@ data_iterator::batch simple_data_iterator::next_batch(size_t batch_size) {
...
   float* labels_out = labels.data();
   float* weights_out = weights.data();
-  std::vector<batch::chunk_info> batch_info;
-  batch_info.reserve(batch_size);
+
   while (batch_info.size() < batch_size && next_row_ != end_of_rows_) {
 
     const sframe_rows::row& row = *next_row_;
 
+    const flex_int &chunk_length = row[chunk_len_column_index].get<flex_int>();
```
nickjong: Nit: there’s no reason to store a const reference to a `flex_int`. Unlike an object, it’s cheaper to copy than to use a reference

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,291 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
                                                         const int &predictions_in_chunk,
                                                         const std::string &target,
                                                         bool verbose) {
+#ifndef NDEBUG
```
nickjong: Sorry, but given the instability in the Python-side test for the data iteration (the crash we saw in Travis, I think it’s safer not to touch the original _activity_classifier_prepare_data_impl.

We can apply the previous review feedback to the partially copied new version, since that’s the one that will live on after we remove the Python toolkit.

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,291 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
-                               use_target);
-            }
-
-            last_session_id = curr_session_id;
-            number_of_sessions++;
-        }
+variant_map_type _activity_classifier_prepare_data_aug_impl(const gl_sframe &data,
+                                                        const std::vector<std::string> &features,
+                                                        const std::string &session_id,
+                                                        const int &prediction_window,
```
nickjong: There’s little reason to pass integers by const reference. May as well pass by value and have simpler, more readable code

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,291 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+                                                        const int &predictions_in_chunk,
+                                                        const std::string &target,
+                                                        bool verbose) {
+#ifndef NDEBUG
+  DASSERT_TRUE(features.size() > 0);
```
nickjong: Nit: this can be written as `DASSERT_GT(features.size(), 0)`, which should give clearer error messages when the assertion fails.

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,291 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+
+  // Prepare an output SFrame writer, that will write a new SFrame in the
+  // converted batch-processing ready format.
+  std::vector<std::string> output_column_names = {"features", "chunk_len",
+                                                  "session_id"};
+  std::vector<flex_type_enum> output_column_types = {flex_type_enum::VECTOR,
+                                                     flex_type_enum::INTEGER,
+                                                     data[session_id].dtype()};
+  if (use_target) {
+    output_column_names.insert(output_column_names.end(), {"target"});
```
nickjong: Can be simplified to
```
output_column_names.push_back("target");
```

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,291 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+  }
+
+  time_t last_print_time = time(0);
+  size_t processed_lines = 0;
+
+  // Iterate over the user data. The features and targets are aggregated, and
+  // handled whenever a the ending of a session is reached.
+  size_t chunk_size = 0;
+  for (const auto &line : data.range_iterator()) {
+    auto curr_session_id = line[column_index_map[session_id]];
```
nickjong: May as well use `flexible_type` instead of `auto`
It’s not much longer and makes it clearer to the reader what the type is

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,291 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
-                       curr_window_targets,
-                       last_session_id,
-                       output_writer,
-                       chunk_size,
-                       feature_size,
-                       predictions_in_chunk,
-                       use_target);
+    chunk_size += 1;
+
+    for (const auto feature_name : features) {
```
nickjong: Use `const std::string&` instead of `const auto`

Using a reference prevents a copy in some cases. `std::string` is not much longer than `auto` and provides more clarity

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,291 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
...
+      curr_chunk_features.push_back(line[column_index_map[feature_name]]);
     }
 
-    // Update the count of the last session in the dataset
-    number_of_sessions++;
+    // If target column exists, the targets are aggregated for the duration
+    // of prediction_window.
+    // Each prediction_window subsampled into a single target value, by
+    // selecting the most frequent value (statistical mode) within the window.
```
nickjong: Does this comment still apply? The mode isn’t computed during pre-processing any more, right?

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -742,16 +745,19 @@ void activity_classifier::init_train(
                        {"features", flex_list(feature_column_names.begin(),
                                               feature_column_names.end())}});
 
+  bool use_data_augmentation = read_state<bool>("use_data_augmentation");
```
nickjong: Could we put this line after the comment, which applies to both this line and the actual call to `create_iterator`?

src/toolkits/activity_classification/ac_data_iterator.cpp
shreyajain17: This is what we do in ```_activity_classifier_prepare_data_impl``` also.

nickjong: You shouldn't take that function as an ideal example of readable C++ code.

But for now, let's leave it alone, since we expect to deprecate/delete it later this year. Then we don't have to worry about additional testing that the Python side still works after your changes.

src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -132,152 +132,284 @@ variant_map_type _activity_classifier_prepare_data_impl(const gl_sframe &data,
                                                         const int &predictions_in_chunk,
                                                         const std::string &target,
                                                         bool verbose) {
+  DASSERT_TRUE(features.size() > 0);
```
nickjong: Let's just revert these changes. Don't run clang-format on this bit. In the final squashed commit, let's not touch the history for these lines, and then it will be easier to review the overall change. And we don't have to worry about the flaky unit test for the Python path.

PR #2062, opened 2019-06-17
Fixes a hang in CSV parsing when parsing a token of a single '['

ylow:
The lost lasting issue is that there doesn't seem to be an easy way to
disble the skip parser in boost spirit. A hack is to use qi::eoi as the
skip parser, but that is not ideal; under certain circumstances when
the parser calls the skip parser at the end of the line, this can cause a hang
since the eoi parser does not consume any characters.

What seems to work is to use !eps as the parser. eps just returns true.
!eps thus returns false. And that works to fully disable the skip
parser.


hoytak:
Passed in internal gitlab tests.

PR #2064, opened 2019-06-18
add more info in terms of building source with Ninja

Jarvi-Izana:
Adding extra information to guide people choosing to use Ninja to build the source.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -102,12 +102,12 @@ If `Ninja` is chosen as the generator in the previous step, you should use comma
 
     cd debug/
     # check ninja -h for more info
-    ninja -j 4
+    ninja
 
 will perform up to 4 build tasks in parallel. When building in release mode,
 Turi Create does require a large amount of memory to compile with the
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -98,10 +98,16 @@ process. For instance:
     cd debug/src/unity
     make -j 4
 
+If `Ninja` is chosen as the generator in the previous step, you should use command listed below instead:
+
+    cd debug/
+    # check ninja -h for more info
+    ninja -j 4
```
znation: We should recommend running `ninja` with no arguments, since it automatically parallelizes based on the number of CPU cores you have. I only use `-j` if I want to artificially limit it.

Jarvi-Izana: Just want to align with the `make` example. I can remove it if it's redundant.

znation: I think we could make it `ninja` (without `-j 4`), and add a comment that says ninja will automatically parallelize by default.

Jarvi-Izana: I just notice you have the notice to recommend user use default parallelism. A new commit is made.

PR #2068, opened 2019-06-18
Add browser target for .show

znation:
* Adds a visualization HTTP server based on boost::beast.
  * Automatically determines and uses an available port on localhost.
  * Keeps a reference by unique ID to each plot registered with the
    server, and makes spec/data available by HTTP request.
* Adds a "browser" target to `tc.visualization.set_target` in Python.
* Adds a C API for getting a plot URL, and a unit test for the C API.

This introduces a new user-facing target for Python `tc.Plot` objects,
and adds infrastructure for future visualization improvements based on
this web server approach. Some future improvements to follow:

* Simplifying the client-side logic around loading data and state; all
  frontends should in theory share this logic. In the long term, we
  could eliminate the pipe approach (fast, but the complexity of which did
  not scale well) and replace it with a client/server architecture
  everywhere.

* Performance improvements for image loading; instead of sending images
  one at a time over the pipe, we should be able to load them a real web
  browser would (by making multiple parallel HTTP requests, and expecting
  a web server to handle them in parallel). This also allows us to use the
  `<img src=` attribute on the client side, simplifying the loading logic.


src/unity/lib/visualization/server.cpp
```diff
@@ -0,0 +1,591 @@
...
+        key   =  qi::char_("a-zA-Z_\\-") >> *qi::char_("a-zA-Z_0-9\\-");
+        value = +qi::char_("a-zA-Z_0-9\\-");
+    }
+
+    qi::rule<Iterator, pairs_type()> query;
+    qi::rule<Iterator, std::pair<std::string, std::string>()> pair;
+    qi::rule<Iterator, std::string()> key, value;
+};
+
+// Mime type from file extension
```
TobyRoseman: Minor note: "Mime" -> "MIME"

src/unity/lib/visualization/server.cpp
```diff
@@ -0,0 +1,591 @@
...
+            boost::asio::socket_base::max_listen_connections, ec);
+        if(ec)
+        {
+            fail(ec, "listen");
+            return;
+        }
+    }
+
+    // Start accepting incoming connections
+    void
```
TobyRoseman: It seems odd that `void` is on its own line for this function and the next three.

PR #2075, opened 2019-06-20
fix for range sequence operation bug

Jarvi-Izana:
Big thanks to @hoytak and @ylow for helping me fixing this bug.

[start_val, ....   end_val]
    ^                      ^
 begin_idx.       end_idx

The is issue is caused by the query optimizer will treat two different sequences as the same by only using begin and end index to identify sequence identity. All binary operations are affected by this.

As @ylow points out, head value of a sequence should also be involved to identify whether two sequence are the same and should be fused during the optimization phase.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -102,12 +102,12 @@ If `Ninja` is chosen as the generator in the previous step, you should use comma
 
     cd debug/
     # check ninja -h for more info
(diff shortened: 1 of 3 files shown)


Jarvi-Izana:
passed the internal test pipeline. good to go.

PR #2077, opened 2019-06-21
Convolutional and Fully Connected Helper Functions

abhishekpratapa:


src/ml/neural_net/mps_layer_conv_padding.h
```diff
@@ -0,0 +1,36 @@
+#ifndef mps_layer_conv_padding_h
+#define mps_layer_conv_padding_h
```
nickjong: We don't need these preprocessor directives in Objective-C headers.

src/ml/neural_net/mps_layer_conv_padding.h
```diff
@@ -0,0 +1,36 @@
...
+
+#import <Accelerate/Accelerate.h>
+#import <Foundation/Foundation.h>
+#import <Metal/Metal.h>
+#import <MetalPerformanceShaders/MetalPerformanceShaders.h>
+
+API_AVAILABLE(macos(10.14))
+@interface ConvolutionPadding : NSObject <MPSNNPadding>
+{
```
nickjong: Let's try to keep the header files focused on the interface as much as possible, and leave implementation details to the .m files. This makes headers easier to read, and these are the files that clients must read most often.

In Objective C, we can omit the instance variables entirely. I would suggest putting this code in the .m file:
```
@interface ConvolutionPadding ()

@property (nonatomic) NSUInteger paddingWidth;
@property (nonatomic) NSUInteger paddingHeight;
@property (nonatomic) NSUInteger strideWidth;
@property (nonatomic) NSUInteger strideHeight;

@end
```

This will auto-synthesize instance variables with the standard naming convention, e.g. `_paddingWidth`.

src/ml/neural_net/mps_layer_conv_padding.h
```diff
@@ -0,0 +1,36 @@
...
+  int mPaddingWidth;
+  int mPaddingHeight;
+  int mStrideWidth;
+  int mStrideHeight;
+}
+
+@property (class, readonly) BOOL supportsSecureCoding;
+
+- (id) initWithParams:(int)paddingWidth
```
nickjong: Initializers in modern Objective C code should return `instancetype`. Also, let's use idiomatic Objective C integer types, like NSUInteger, or whatever the underlying MPS API expects.

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
+#include <ml/neural_net/mps_layer_conv_padding.h>
+
+@implementation ConvolutionPadding
+
++ (BOOL) supportsSecureCoding {
+  return true;
```
nickjong: Nit: for Objective C code, use YES or NO instead of true or false.

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
...
+}
+
+- (id) initWithParams:(int)paddingWidth
+        paddingHeight:(int)paddingHeight
+          strideWidth:(int)strideWidth
+         strideHeight:(int)strideHeight {
+  @autoreleasepool {
```
nickjong: I don't see a need for an autorelease pool here

nickjong: Typically it's a good idea to put use these in C++ functions whose implementations contained Objective C calls, but otherwise it should only be used to avoid long-lived allocations when needed

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
...
+    self = [self init];
+    mPaddingWidth = paddingWidth;
```
nickjong: The idiomatic implementation of this initializer would be something like:
```
self = [super init];
if (self) {
  _paddingHeight = paddingHeight;
  ....
}
return self;
```

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
...
+
+- (MPSNNPaddingMethod) paddingMethod {
+  return MPSNNPaddingMethodCustom;
+}
+
+- (MPSImageDescriptor *)destinationImageDescriptorForSourceImages:(NSArray<MPSImage *> *)sourceImages
+                                                     sourceStates:(NSArray<MPSState *> *)sourceStates
+                                                        forKernel:(MPSKernel *)kernel
+                                              suggestedDescriptor:(MPSImageDescriptor *)inDescriptor {
+  MPSCNNConvolution *layer = (MPSCNNConvolution *) kernel;
```
nickjong: If we assume that the kernel is a `MPSCNNConvolution`, can we just use that type in the function signature?

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
...
+  
+  MPSOffset pad;
+  pad.x = (int) ((layer.kernelWidth/2) - mPaddingWidth);
+  pad.y = (int) ((layer.kernelHeight/2) - mPaddingHeight);
+  pad.z = 0;
+  
+  layer.offset = pad;
```
nickjong: It is surprising to me that this function modifies the `kernel` input. I think that would be very confusing for potential future users of this code

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
...
+  layer.edgeMode = MPSImageEdgeModeZero;
+  
+  inDescriptor.width += (mPaddingWidth * 2)/mStrideWidth;
```
nickjong: Same here. How is this function expected to be used?

nickjong: In this case should we copy the suggested descriptor first?

src/ml/neural_net/mps_layer_helper.h
```diff
@@ -0,0 +1,41 @@
...
+
++ (MPSCNNFullyConnectedNode *) createFullyConnected:(MPSNNImageNode *)inputNode
+                               inputFeatureChannels:(NSUInteger)inputFeatureChannels
+                              outputFeatureChannels:(NSUInteger)outputFeatureChannels
+                                        inputHeight:(NSUInteger)inputHeight
+                                         inputWidth:(NSUInteger)inputWidth
+                                            weights:(float *)weights
+                                             biases:(float *)biases
+                                              label:(NSString *)label
+                                      updateWeights:(bool)updateWeights
```
nickjong: Nit: for Obj-C is more idiomatic to use BOOL

src/ml/neural_net/mps_layer_helper.h
```diff
@@ -0,0 +1,41 @@
+#import <Accelerate/Accelerate.h>
+#import <Foundation/Foundation.h>
+#import <MetalPerformanceShaders/MetalPerformanceShaders.h>
+#import <Metal/Metal.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+API_AVAILABLE(macos(10.14))
+@interface TCMPSLayerHelper: NSObject
```
nickjong: It looks like you're creating a class just to define some free functions in Objective C. The more typical way of doing this might be to add a category to e.g MPSCNNFullyConnectedNode, something like

@interface MPSCNNFullyConnectNode (TCMPSLayerHelper)

+ (instancetype)fullyConnectNodeWithInput:...

@end

src/ml/neural_net/mps_layer_helper.mm
```diff
@@ -0,0 +1,90 @@
...
+                                            weights:(float *)weights
+                                             biases:(float *)biases
+                                              label:(NSString *)label
+                                      updateWeights:(bool)updateWeights
+                                             device:(id<MTLDevice>)dev
+                                          cmd_queue:(id<MTLCommandQueue>) cmd_q {
+  
+  turi::neural_net::OptimizerOptions optimizerOptions;
+
+  TCMPSConvolutionWeights *FullyConnectedDataLoad = [[TCMPSConvolutionWeights alloc] initWithKernelWidth:inputWidth
```
nickjong: Do me a favor and just start the entire RHS of this assignment on the next line (indented 4 over), e.g.
```
TCMPSConvolutionWeights *FullyConnectedDataLoad =
    [[TCMPSConvolutionWeights alloc] initWithKernelWidth:inputWidth
...
```

src/ml/neural_net/mps_layer_helper.mm
```diff
@@ -0,0 +1,90 @@
...
+                                                                                       neuronA:0.0f
+                                                                                       neuronB:0.0f
+                                                                        kernelParamsBinaryName:[label UTF8String]
+                                                                                        device:dev
+                                                                                     cmd_queue:cmd_q
+                                                                               init_weight_ptr:weights
+                                                                                 init_bias_ptr:biases
+                                                                              optimizerOptions:optimizerOptions];
+        
+  MPSCNNFullyConnectedNode* FullyConnectedNode = [MPSCNNFullyConnectedNode nodeWithSource:inputNode
```
nickjong: Nit: variable names should not start with a capital letter

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: If I that's the specific function provided by MPS for me to override don't I have to follow their function signature to the T?

src/ml/neural_net/mps_layer_conv_padding.h
abhishekpratapa: Changed!

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: Changed!

src/ml/neural_net/mps_layer_conv_padding.h
abhishekpratapa: done!

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: done!

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: done!

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
...
+    mPaddingWidth = paddingWidth;
+    mPaddingHeight = paddingHeight;
+    mStrideWidth = strideWidth;
+    mStrideHeight = strideHeight;
+    return self;
+  }
+}
+
+- (nullable instancetype)initWithCoder:(NSCoder *)coder {
+  @autoreleasepool {
```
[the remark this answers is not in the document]
abhishekpratapa: done!

src/ml/neural_net/mps_layer_helper.h
abhishekpratapa: updated

src/ml/neural_net/mps_layer_helper.h
abhishekpratapa: done!

src/ml/neural_net/mps_layer_helper.mm
abhishekpratapa: changed to lowercase.

src/ml/neural_net/mps_layer_helper.mm
abhishekpratapa: alright, done!

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: Should I copy the kernel and modify it instead?

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: Same we can copy it!

src/ml/neural_net/mps_layer_conv_padding.h
```diff
@@ -0,0 +1,30 @@
+#import <Accelerate/Accelerate.h>
+#import <Foundation/Foundation.h>
+#import <Metal/Metal.h>
+#import <MetalPerformanceShaders/MetalPerformanceShaders.h>
+
+API_AVAILABLE(macos(10.14))
+@interface ConvolutionPadding : NSObject <MPSNNPadding>
```
nickjong: Objective C doesn't have namespaces, so it's incumbent on library code to use prefixes to reduce the likelihood of namespace collisions. I suggest `TCMPS`, so we can rename this class `TCMPSConvolutionPadding`

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,57 @@
+#include <ml/neural_net/mps_layer_conv_padding.h>
```
nickjong: Use `#import` in pure Objective-C code. More generally, whenever you import and ObjC header

src/ml/neural_net/mps_layer_conv_padding.h
abhishekpratapa: Got it. Updated the library name to avoid confusion.

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: Updated to use import vs include.

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,65 @@
+#import <ml/neural_net/mps_layer_conv_padding.h>
+
+@implementation TCMPSConvolutionPadding
+
++ (BOOL) supportsSecureCoding {
```
znation: Based on my reading [here](https://developer.apple.com/documentation/foundation/nssecurecoding), this should probably return `NO`. It doesn't look like key/value coding is properly implemented in this interface.

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,65 @@
...
+    _paddingWidth = paddingWidth;
+    _paddingHeight = paddingHeight;
+    _strideWidth = strideWidth;
+    _strideHeight = strideHeight;
+  }
+
+  return self;
+}
+
+- (nullable instancetype)initWithCoder:(NSCoder *)coder {
```
znation: Does this need to be overridden? Can we just allow the superclass `NSObject` to handle it?

src/ml/neural_net/mps_layer_conv_padding.m
znation: Or let the superclass `NSObject` handle it?

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,65 @@
...
+  return self;
+}
+
+- (MPSNNPaddingMethod)paddingMethod {
+  return MPSNNPaddingMethodCustom;
+}
+
+/*
+*
+* Due to the specifities of the MPS api, we modify the kernel and imageDescriptor directly
```
znation: `s/specifities/specificities/`
`s/api/API/`

src/ml/neural_net/mps_layer_conv_padding.m
```diff
@@ -0,0 +1,65 @@
...
+  layer.offset = pad;
+  layer.edgeMode = MPSImageEdgeModeZero;
+  
+  inDescriptor.width += (_paddingWidth * 2)/_strideWidth;
+  inDescriptor.height += (_paddingHeight * 2)/_strideHeight;
+  
+  return inDescriptor;
+}
+
+- (void)encodeWithCoder:(NSCoder *)coder {}
```
znation: Can this be omitted and left to the superclass `NSObject` to handle?

src/ml/neural_net/mps_layer_helper.h
```diff
@@ -0,0 +1,44 @@
...
+                              outputFeatureChannels:(NSUInteger)outputFeatureChannels
+                                        inputHeight:(NSUInteger)inputHeight
+                                         inputWidth:(NSUInteger)inputWidth
+                                            weights:(float *)weights
+                                             biases:(float *)biases
+                                              label:(NSString *)label
+                                      updateWeights:(BOOL)updateWeights
+                                             device:(id<MTLDevice>)dev
+                                          cmd_queue:(id<MTLCommandQueue>)cmd_q
+                                          API_AVAILABLE(macosx(10.14));
```
znation: Are these `API_AVAILABLE(macosx(10.14))` required on each method, even when the overall interface has an `API_AVAILABLE` on it?

abhishekpratapa: Good catch! We don't need it in both places.

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: unfortunately I don't believe so: 

```
/Users/abhishekpratapa/Desktop/turicreate/src/ml/neural_net/mps_layer_conv_padding.m:66:10: error: no visible @interface for 'NSObject' declares the selector 'encodeWithCoder:'
```

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: Unfortunately the same issue:

```
/Users/abhishekpratapa/Desktop/turicreate/src/ml/neural_net/mps_layer_conv_padding.m:27:10: error: no visible @interface for 'NSObject' declares the selector 'initWithCoder:'
```

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: Thanks!

src/ml/neural_net/mps_layer_conv_padding.m
abhishekpratapa: Unfortunately we can't let NSObject handle it. But you're right that we should return `NO`. I've updated the values.

PR #2082, opened 2019-06-24
Added Instance Norm Data Loader

abhishekpratapa:
- added an Objective-C instance norm data loader.


nickjong:
It looks like this overlaps with #2077, so I’ll wait until that is merged before reviewing this one.

Generally it’s preferable either to wait until dependent PRs have merged before putting up the next one, or including all the changes in one PR...

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,92 @@
+#ifndef mps_layer_instance_norm_data_loader_h
+#define mps_layer_instance_norm_data_loader_h
+
+#import <Accelerate/Accelerate.h>
+#import <Foundation/Foundation.h>
+#import <Metal/Metal.h>
+#import <MetalPerformanceShaders/MetalPerformanceShaders.h> 
+
+typedef struct{
```
nickjong: We should avoid defining types in the global namespace with generic names. I don't see this being used anywhere anyway, so can we just delete this one?

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,92 @@
...
+  int size;
+  float beta1;
+  float beta2;
+  float epsilon;
+  float lr_t;
+} params_t;
+
+API_AVAILABLE(macos(10.14))
+@interface InstanceNormDataLoader: NSObject <MPSCNNInstanceNormalizationDataSource> {
```
nickjong: Objective C class names must be prefixed to avoid collisions, e.g. with something like `TCMPS`. Please apply this rule to all future PRs.

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,92 @@
...
+  NSString *mName;
```
nickjong: These instance variables are largely implementation details. Let's move them to the .mm file. In addition, we use properties and let ObjC auto-synthesize the instance variables themselves.

Look at the code in https://github.com/apple/turicreate/blob/master/src/ml/neural_net/TCMPSImageAugmenting.h for an example of this pattern

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
nickjong: Some documentation of what this class does/represents is needed!

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,92 @@
...
+
+  MPSNNOptimizerAdam *adamGamma;
+  MPSNNOptimizerAdam *adamBeta;
+
+  id<MTLComputePipelineState> runningUpdatePipelineState;
+
+  id<MTLCommandQueue> mCq;
+}
+
+- (id) initWithParams:(NSString *)name
```
nickjong: initializers should return `instancetype`. Please apply this rule to future PRs moving forward

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,92 @@
...
+     gammaWeights:(float **)gammaWeights
+      betaWeights:(float **)betaWeights
+numberFeatureChannels:(int)numberFeatureChannels
+         styles:(int)styles
```
nickjong: For ObjC interfaces, let's use NSInteger and NSUInteger, as appropriate

abhishekpratapa: Fixed!

src/ml/neural_net/mps_layer_helper.h
```diff
@@ -38,6 +38,16 @@ NS_ASSUME_NONNULL_BEGIN
...
+
++ (MPSCNNInstanceNormalizationNode *) createInstanceNormalization:(MPSNNImageNode *)inputNode
+                                                         channels:(NSUInteger)channels
+                                                           styles:(NSUInteger)styles
+                                                            gamma:(float * _Nonnull * _Nonnull)gamma
+                                                             beta:(float * _Nonnull * _Nonnull)beta
+                                                            label:(NSString *)label
+                                                           device:(id<MTLDevice>)dev
+                                                        cmd_queue:(id<MTLCommandQueue>)cmd_q
+                                                        API_AVAILABLE(macosx(10.14));
```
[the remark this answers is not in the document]
abhishekpratapa: I tried it and it gave me compilation errors about the api not being available when pertaining to that specific method.

The change:

Screen Shot 2019-07-03 at 11 33 00 AM

The result in compilation:

Screen Shot 2019-07-03 at 11 34 56 AM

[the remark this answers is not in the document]
abhishekpratapa: It was cleaner to do it on the category, but resulted in compilation error which I go around by applying it to the methods individually.

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
abhishekpratapa: Added documentation!

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,104 @@
...
+ *        Given a command_queue on a Metal Device we can synchronize resources
+ *        from that command_queue. For beta and gamma those might be their
+ *        respective values on the GPU, but for training we would also need
+ *        momentum and velocity vectors. We therefore synchronize those as well.
+ *
+ * - updateGammaAndBetaWithCommandBuffer
+ *        MPS calls this method and we need to update the gamma and the beta
+ *        weights ourselves. To do this we have an adams optimizer for both the
+ *        gamma and the beta values. For each value in the batch we don't want
+ *        to increment the adam optimizer;s time step as the learning rate would
```
nickjong: s/;/'/

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,104 @@
...
+ *        be affected. So we store the timestep at the beginning of each batch
+ *        and use it across the batch. This `Hack` makes the behavior of the
+ *        weight updates, very similar to the behavior we see in MxNet.
+ */
+API_AVAILABLE(macos(10.14))
+@interface TCMPSInstanceNormDataLoader: NSObject <MPSCNNInstanceNormalizationDataSource>
+
+@property (nonatomic) float **gamma_weights;
```
nickjong: It looks like these properties are all implementation details of the class, not intended to be part of the public interface. Can we move them all to the .m file? 

You should be able to do this by creating a class extension. See https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW3

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,104 @@
...
+
+@property (nonatomic) MPSVectorDescriptor *vDesc;
+@property (nonatomic) MPSCNNNormalizationGammaAndBetaState *state;
+@property (nonatomic) id<MTLCommandQueue> cq;
+
+@property (nonatomic) MPSNNOptimizerAdam *adamGamma;
+@property (nonatomic) MPSNNOptimizerAdam *adamBeta;
+
+- (instancetype) initWithParams:(NSString *)name
+                   gammaWeights:(float **)gammaWeights
```
nickjong: What is the format/semantics of this `float**`?

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
```diff
@@ -0,0 +1,104 @@
...
+                    betaWeights:(float **)betaWeights
+          numberFeatureChannels:(NSUInteger)numberFeatureChannels
+                         styles:(NSUInteger)styles
+                         device:(id<MTLDevice> _Nonnull)dev
```
nickjong: Nit: we don't need any of the _Nonnull declarations in this file

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -0,0 +1,202 @@
...
+    
+  if (self) {
+    _name = name;
+    _numberOfFeatureChannels = numberFeatureChannels;
+
+    _styles = styles;
+    
+    _currentStyle = 0;  
+  
+    _gamma_weights = gammaWeights;
```
nickjong: How should we reason about the ownership of these weights? Where is the actual memory?

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -0,0 +1,202 @@
...
+    _beta_weights = betaWeights;
+
+    _cq = cmd_q;
+
+    float *zeros_ptr = (float*) calloc(_numberOfFeatureChannels, sizeof(float));
```
nickjong: Where are these allocations de-allocated?

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -0,0 +1,202 @@
+#include <ml/neural_net/mps_layer_instance_norm_data_loader.h>
+#include <ml/neural_net/mps_weight.h>
+#include <stdio.h>
+#include <stdlib.h>
```
nickjong: Where are these two C headers used?

I also don't see any C++ in this file. Could we leave a TODO here that this file can become a .m file once `mps_weight.h` stops using C++?

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -0,0 +1,202 @@
...
+                                                 descriptor:_vDesc];
+
+
+    _state = [[MPSCNNNormalizationGammaAndBetaState alloc] initWithGamma:_gammaBuffer
+                                                                    beta:_betaBuffer];
+  }
+  return self;
+}
+
+- (void) updateNumberOfStyles:(NSUInteger)styles {
```
nickjong: Hmm, looks like this just be exposed as a public property. So for this one, in the header:
```
@property (nonatomic) NSUInteger stylesCount;
```

Then in the .m file
```
- (void)setStylesCount:(NSUInteger)count {
  _stylesCount = count;
}
```

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -0,0 +1,202 @@
...
+  _currentStyle = 0;
+  _styles = styles;
+}
+
+- (void) updateCurrentStyle:(NSUInteger)style {
+  _currentStyle = style;
+}
+
+- (NSUInteger) getCurrentStyle {
```
nickjong: In idiomatic obj-c code, we should just use the `currentStyle` property. ObjC auto-generates this getter and setter for you.

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
abhishekpratapa: changed to a single float pointer

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: solution, copy the weights

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: solution called free

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: removed the c++ header files.

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: changed.

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: done

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
abhishekpratapa: Still working on the implementation with multiple styles.

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -97,24 +132,23 @@ - (instancetype) initWithParams:(NSString *)name
...
   _currentStyle = style;
 }
 
-- (NSUInteger) getCurrentStyle {
-  return _currentStyle;
-}
-
-- (void) loadBeta:(float **)beta {
+- (void) loadBeta:(float *)beta {
   _beta_weights = beta;
```
nickjong: Should this be a copy?

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -21,9 +53,12 @@ - (instancetype) initWithParams:(NSString *)name
     _styles = styles;
     
     _currentStyle = 0;  
-  
-    _gamma_weights = gammaWeights;
-    _beta_weights = betaWeights;
+    
+    _gamma_weights = (float *) malloc(numberFeatureChannels * styles * sizeof(float));
```
nickjong: Hmm, seeing more of this in context, I think it's probably clearest and safest just to use `NSMutableData` to allocate and own this memory. So you could make `_gamma_weights` an `NSMutableData *` and initialize as e.g.:
```
_gamma_weights = [NSMutableData dataWithLength:numberFeatureChannels * styles * sizeof(float)];
```
Then you can get to the memory using `_gamma_weights.mutableBytes`.

See https://developer.apple.com/documentation/foundation/nsmutabledata?language=objc 

With NSData, you don't have to worry about implementing `dealloc`, and then you can just pass `mutableBytes` into the MPS API

src/ml/neural_net/mps_layer_instance_norm_data_loader.h
abhishekpratapa: Updated

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: Fixed!

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: Perfect, updated the implementation to use NSMutableData

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -2,31 +2,31 @@
...
   NSUInteger _styles;
 
-  id<MTLBuffer> _gammaBuffer;
-  id<MTLBuffer> _betaBuffer;
+  NSMutableArray<MPSVector *> *_gammaVectorArray;
+  NSMutableArray<MPSVector *> *_betaVectorArray;
+  NSMutableArray<MPSVector *> *_gammaMomentumVectorArray;
+  NSMutableArray<MPSVector *> *_gammaVelocityVectorArray;
+  NSMutableArray<MPSVector *> *_betaMomentumVectorArray;
+  NSMutableArray<MPSVector *> *_betaVelocityVectorArray;
```
nickjong: Instead of making all of these instance variables arrays, could we instead define one Objective C class to hold all of these things, and then create a single array of that?

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -69,101 +69,143 @@ - (instancetype) initWithParams:(NSString *)name
...
-                                           options:MTLResourceStorageModeManaged];
+    for (size_t index = 0; index < styles; index ++){
+      id<MTLBuffer> gammaBuffer = [dev newBufferWithBytes:_gamma_weights.mutableBytes
+                                                   length:sizeof(float) * _numberOfFeatureChannels
+                                                  options:MTLResourceStorageModeManaged];
 
-    _gammaVelocityBuffer = [dev newBufferWithBytes:zeros_ptr
-                                            length:sizeof(float) * _numberOfFeatureChannels
-                                           options:MTLResourceStorageModeManaged];
+      [_gammaBufferArray addObject:gammaBuffer];
```
nickjong: Where is this array used/consumed?

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -54,11 +54,11 @@ - (instancetype) initWithParams:(NSString *)name
...
     _currentStyle = 0;  
     
-    _gamma_weights = (float *) malloc(numberFeatureChannels * styles * sizeof(float));
-    _beta_weights = (float *) malloc(numberFeatureChannels * styles * sizeof(float));
+    _gamma_weights = [NSMutableData dataWithLength:numberFeatureChannels * styles * sizeof(float)];
+    _beta_weights = [NSMutableData dataWithLength:numberFeatureChannels * styles * sizeof(float)];
 
-    memcpy(_gamma_weights, gammaWeights, numberFeatureChannels * styles * sizeof(float));
-    memcpy(_beta_weights, betaWeights, numberFeatureChannels * styles * sizeof(float));
+    memcpy(_gamma_weights.mutableBytes, gammaWeights, numberFeatureChannels * styles * sizeof(float));
```
nickjong: Hmm, instead of creating an empty NSMutableData and then filling it, you should be able to do something like
```
_gamma_weights = [NSMutableData dataWithBytes:gammaWeights length:numberFeatureChannels * styles * sizeof(float)];
```

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -69,101 +69,143 @@ - (instancetype) initWithParams:(NSString *)name
...
+    _movingVarianceBufferArray = [[NSMutableArray alloc] init];
+    _gammaMomentumBufferArray = [[NSMutableArray alloc] init];
+    _gammaVelocityBufferArray = [[NSMutableArray alloc] init];
+    _betaMomentumBufferArray = [[NSMutableArray alloc] init];
+    _betaVelocityBufferArray = [[NSMutableArray alloc] init];
 
-    _gammaMomentumBuffer = [dev newBufferWithBytes:zeros_ptr
-                                            length:sizeof(float) * _numberOfFeatureChannels
-                                           options:MTLResourceStorageModeManaged];
+    for (size_t index = 0; index < styles; index ++){
```
nickjong: Let's use NSUInteger instead of size_t here, since this is principally a Obj C file

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
nickjong: Note that it should suffice to just to do something like, in this .mm file:
```
@interface TCMPSMyClassNameGoesHere

@property (nonatomic) MPSVector *gammaMomentumVector;
@property (nonatomic) MPSVector *gammaVelocityVector;
...

@end

@implementation TCMPSMyClassNameGoesHere
@end  // All properties will be auto-synthesized
```

Then you can just use this like:
```
NSMutableArray<TCMPSMyClassNameGoesHere *> *styles;
for (NSUInteger index = 0; index < self.stylesCount; ++index) {
  TCMPSMyClassNameGoesHere *style = [[TCMPSMyClassNameGoesHere alloc] init];
  style.gammaMomentumVector = ...;
  ...;
}
```

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: It was used in a previous version when testing as a potential avenue to get the updated the weights from the GPU but we now get those values directly from the state. I'll remove it

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: Got it!

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: I should have looked at the api a bit closer. That seems to work!

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
abhishekpratapa: Ahh, that's a good idea!

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -237,30 +199,30 @@ - (MPSCNNNormalizationGammaAndBetaState *)updateGammaAndBetaWithCommandBuffer:(i
...
 - (void)checkpointWithCommandQueue:(nonnull id<MTLCommandQueue>)commandQueue {
   id<MTLCommandBuffer> commandBuffer = [commandQueue commandBuffer];
   id<MTLBlitCommandEncoder> blit = commandBuffer.blitCommandEncoder;
 
-  for (size_t index = 0; index < _styles; index ++){
-    [blit synchronizeResource:_betaMomentumBufferArray[index]];
-    [blit synchronizeResource:_betaVelocityBufferArray[index]];
-    [blit synchronizeResource:nonnull_cast(stateArray[index].beta)];
+  for (size_t index = 0; index < _styles; index ++) {
+    [blit synchronizeResource:[[style_props objectAtIndex: index] betaMomentumBuffer]];
```
nickjong: Let's factor out the repeated array indexing to get the props for this style once, at the top of the loop!

abhishekpratapa: Looks good?!

Issue #2088, opened 2019-06-25
SAarray fails to construct from data when dtype is explicitly specified


Jarvi-Izana:
The test failed when passing a map type.

```
        # map input type
        toy_data = SFrame({'a': range(100)})
        map_result = map(lambda x: x+1, [1, 30])
        result = toy_data.filter_by(map_result, 'a')
>       self.assertEqual(len(result), 2)
E       AssertionError: 1 != 2
```

minimal repro:
```
map_result = map(lambda x: x+1, [1, 30])
g = tc.SArray(map(lambda x: x+1, [1, 30]), int)
TypeError                                 Traceback (most recent call last)
<ipython-input-40-04fb1819e4fc> in <module>
----> 1 g = tc.SArray(map(lambda x: x+1, [1, 30]), int)

~/Work/guicreate/debug/src/python/turicreate/data_structures/sarray.py in __init__(self, data, dtype, ignore_cast_failure, _proxy)
    424                     self.__proxy__.load_from_iterable(data, dtype, ignore_cast_failure)
    425             else:
--> 426                 raise TypeError("Unexpected data source. " \
    427                                 "Possible data source types are: list, " \
    428                                 "numpy.ndarray, pandas.Series, and string(url)")

TypeError: Unexpected data source. Possible data source types are: list, numpy.ndarray, pandas.Series, and string(url)
```

TobyRoseman:
@Jarvi-Izana - I'm assuming this is a Python 3 issues. Can you confirm what version of Python you are using?

Jarvi-Izana:
passed the internal test.

Jarvi-Izana:
I'm using python 3.6.

Jarvi-Izana:
Solved by #2089

Issue #1716, opened 2019-04-09
filter_by should always accept None value


TobyRoseman:
Regardless of the column type `filter_by` should accept `None` as a value. 

```python
import turicreate as tc
sf = tc.SFrame({'a': [1,2,None,3]})
assert sf[2]['a'] is None    # this passes
sf.filter_by(None, 'a')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-18cfe1b718b0> in <module>()
      2 sf = tc.SFrame({'a': [1,2,None,3]})
      3 assert sf[2]['a'] is None    # this passes
----> 4 sf.filter_by(None, 'a')

/Users/tobyroseman/Documents/turicreate/debug/src/unity/python/turicreate/data_structures/sframe.pyc in filter_by(self, values, column_name, exclude)
   4414         if given_type != existing_type:
   4415             raise TypeError("Type of given values does not match type of column '" +
-> 4416                 column_name + "' in SFrame.")
   4417 
   4418         # Make sure the values list has unique values, or else join will not

TypeError: Type of given values does not match type of column 'a' in SFrame.
```

Using `sf.dropna(columns='a')` works fine.

PR #2089, opened 2019-06-26
bug fix for sarray constructor not accepting map/filter if dtype is explicitly set

Jarvi-Izana:
bug fix for issue #2088

this pr also supports generator types; before any type inference and passing data to the cpp proxy, transform the data from generator/map/filter to a valid list first.

This issue is caused by a patch for #1716. The internal test is triggered.

### DIFF
--- a/src/python/turicreate/data_structures/sarray.py
+++ b/src/python/turicreate/data_structures/sarray.py
@@ -355,6 +355,11 @@ class SArray(object):
                 self.__proxy__ = data.astype(dtype).__proxy__
         else:
             self.__proxy__ = UnitySArrayProxy()
+
+            ## data transfromation from generator to list
+            if sys.version_info.major >= 3 and isinstance(data, (filter, map)):
+                data = list(data)
+
             # we need to perform type inference
             if dtype is None:
                 if HAS_PANDAS and isinstance(data, pandas.Series):
@@ -398,8 +403,6 @@ class SArray(object):
                       (sys.version_info.major < 3 and isinstance(data, unicode))):
                     # if it is a file, we default to string
                     dtype = str
-                elif sys.version_info.major >= 3 and isinstance(data, (filter, map)):
-                    data = list(data)
                 elif isinstance(data, array.array):
                     dtype = pytype_from_array_typecode(data.typecode)
                 elif isinstance(data, collections.Sequence):
--- a/src/python/turicreate/data_structures/sarray.py
+++ b/src/python/turicreate/data_structures/sarray.py
@@ -34,6 +34,7 @@ import datetime
 import warnings
 import numbers
 import six
+import types
 
 __all__ = ['SArray']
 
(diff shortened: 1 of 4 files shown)


src/python/turicreate/test/test_sarray.py
```diff
@@ -73,26 +73,38 @@ def __test_almost_equal(self, _sarray, _data, _type):
             else:
                 self.assertAlmostEqual(l[i], _data[i])
 
-    def __test_creation(self, data, dtype, expected):
```
Jarvi-Izana: split this helper is due to the fact that generators, map, and filter can be used in only one pass.

src/python/turicreate/test/test_sarray.py
```diff
@@ -117,6 +129,37 @@ def test_creation(self):
...
+                                 int,
+                                 [x + 10 for x in self.int_data])
+        self.__test_creation_raw(map(lambda x: x * 10, self.int_data),
+                                 float,
+                                 [float(x) * 10 for x in self.int_data])
+        self.__test_creation_raw(map(lambda x: x * 10, self.string_data),
+                                 str,
+                                 [x * 10 for x in self.string_data])
+
+        self.__test_creation_raw(filter(lambda x: x < 5, self.int_data),
```
Jarvi-Izana: pd.Series doesn't support filter. I will create pr for them.

src/python/turicreate/data_structures/sarray.py
```diff
@@ -355,6 +356,11 @@ def __init__(self, data=[], dtype=None, ignore_cast_failure=False, _proxy=None):
                 self.__proxy__ = data.astype(dtype).__proxy__
         else:
             self.__proxy__ = UnitySArrayProxy()
+
+            ## data transfromation from generator to list
+            if (isinstance(data, types.GeneratorType)) or (sys.version_info.major >= 3 and isinstance(data, (filter, map))):
```
TobyRoseman: Should we add `range` to the `(filter, map)` tuple?

Jarvi-Izana: surprisingly the range is not a generator. if we want to support range, then we should also support xrange in py2? Besides, [sarrary.from_sequence](https://apple.github.io/turicreate/docs/api/generated/turicreate.SArray.from_sequence.html#turicreate.SArray.from_sequence) have a more efficient representation of the data structure as it doesn't need to materialize the range into a list until evaluation time.

TobyRoseman: Supporting `xrange` in py2.7 is a great idea. 

I agree it would be better for our users to use `sarray.from_sequence(...)` but I think we should also support range here. Also `sarray.from_sequence(...)` doesn't have a `step` parameter like the built in `range` so it may not work for their use case.

Jarvi-Izana: yeah! Good point! I will add this on!

Jarvi-Izana:
passed the internal tests.

PR #2090, opened 2019-06-26
Boost Coroutine Removal

ylow:
This patch removes the use of boost coroutines from everywhere, instead
using a simple stackless coroutine implementation in src/util/coro.hpp
which uses a switch statement (like the Duff's device) to resume function
execution.

The 2 main places where coroutines are used are in the query execution
engine and in the sarray decoder.

The query execution required minimal changes to switch to the new
mechanism since the operators are already cleanly isolated in classes.
Simply moving stack state to class members suffices (as well as some
slightly more interesting changes in the driver code).

The sarray decoder required significant surgery since there coroutines
are used as a convenience mechanic to make a regular stateful decoder
(for loop which decodes all values in a block) act in a streaming
fashion. The changes here are more involving requiring wrapping each
type decoder in a struct; correctly handling "recursive" cases (where
one stream decoder calls another stream decoder), and the various
decoding paths we can go through based on various flags in the block.

However, switching to a "proper" use of a consumer/producer coroutine
mechanic allowed value skipping to be pushed deeper into the decoder
speeding up random access significantly (up to 4x for integers).
There are certainly more optimization that can be done here.

Since this change impacts the sarray decoder at a deep level, exhaustive
testing is recommended; in particular on old previously saved sframes to
ensure that the "legacy" decoding paths are implemented correctly.

### DIFF
--- a/src/core/storage/query_engine/execution/execution_node.cpp
+++ b/src/core/storage/query_engine/execution/execution_node.cpp
@@ -50,76 +50,6 @@ void execution_node::reset() {
   m_output_queue.reset();
 }
 
-struct source_executor {
-  bool supports_skipping;
-  bool is_linear_operator;
-  bool done = false;
-  execution_node* owner;
-  query_context context;
-  emit_state initial_operator_state;
-
-  DECL_CORO_STATE(emit_callback);
-  DECL_CORO_STATE(execute);
-
-  emit_state emit_callback(const std::shared_ptr<sframe_rows>& rows) {
-    owner->add_operator_output(rows);
-    // we are supposed to skip the next block
-    if (owner->m_skip_next_block) {
-      if (supports_skipping) {
-        // operator supports skipping. tell it
-        // we are skipping
-        return emit_state::SKIP_NEXT_BLOCK;
-      } else if (is_linear_operator) {
-        // make it look like the input is shorter
-        // just consume the inputs
-        for (size_t i = 0;i < owner->num_inputs(); ++i) {
-          owner->get_next_from_input(i, true);
-        }
-        // write a fake output, this is the skipped
-        // block. And sink again.
-        return emit_state::NONE;
-      } else {
-        // operator does not support skipping.
-        // read it as usual
-        return emit_state::NONE;
-      }
-    } else {
-      return emit_state::NONE;
-    }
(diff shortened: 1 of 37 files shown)


src/core/storage/sframe_data/sarray_v2_type_encoding.cpp
```diff
@@ -597,7 +577,601 @@ bool typed_decode(const block_info& info,
...
+  if (skip == 0) {
+    ASSERT_TRUE(decodebuffer.first != nullptr && decodebuffer.second > 0);
+  } else {
+    ASSERT_TRUE(decodebuffer.first == nullptr && decodebuffer.second == 0);
+  }
+  size_t decode_bufpos = 0;
+  if (perform_type_decoding) {
+    if (num_types == 1) {
+      if (decodebuffer.first && column_type == flex_type_enum::UNDEFINED) {
+        for (size_t i = 0;i < decodebuffer.second; ++i) {
```
TobyRoseman: Do you want to create a local variable that shadows a class variable, on this line and line 1040, but not on line 1048?

src/util/coro.hpp
```diff
@@ -0,0 +1,135 @@
+#ifndef RAMPUDDLE_PUDDLE_CORO_HPP
```
TobyRoseman: Probably want to `#ifndef RAMPUDDLE_PUDDLE_CORO_HPP` -> `#ifndef CORO_HPP`

src/core/storage/query_engine/execution/execution_node.cpp
```diff
@@ -9,23 +9,10 @@
 #include <core/storage/query_engine/execution/query_context.hpp>
 #include <core/storage/query_engine/execution/execution_node.hpp>
 #include <core/system/cppipc/cppipc.hpp>
+#include <util/coro.hpp>
```
hoytak: We should probably put this in core/system/util/.  

To move that file easily, just `git mv` this there, then run `./scripts/repo.py --fix-header coro.hpp` and all the include paths will get corrected.

hoytak:
This passes the tests on internal CI with the no-python PR here enabled.  Going to merge this one to keep atomic commits.

PR #2091, opened 2019-06-26
Get Travis CI Working Again

TobyRoseman:


.travis.yml
```diff
@@ -32,8 +32,10 @@ env:
     - CXX=g++-5
 
 script:
-  # Set cache size to 4G (defaults to 512M)
-  - ccache -M 4G
+  - free -m
```
znation: Can probably remove this line.

TobyRoseman: I'd like to keep it in. It only adds a few more lines of output to our already very large amount of output.

PR #2097, opened 2019-06-28
Upgrade for Curl

hoytak:
As part of my attempt to get Turi Compiling on MacOS 10.15 beta, I upgraded curl
and moved bz2 out of the deps and into src/external.
These weren't the problems, but these upgrades are useful (the new curl contains a
number of security bugfixes), so there's no reason not to adopt them.


hoytak:
This passes all internal tests.

PR #2098, opened 2019-06-28
Implemented memory pools for shared flexible type datatypes.

hoytak:
A profiling of a recent benchmark case involving groupby operations
showed that 30-50% of the time was spend on allocations and
deallocations of flexible type std::pair<ref_count, string> objects.
This PR attempts to alleviate that pressure by using the boost memory pools
for these intermediate types.


hoytak:
Turns out that this is likely not the best way to do this, nor is it likely that fast.

Issue #1804, opened 2019-04-29
Feature request: support for dropna() on vectors


dbickson:
The current implementation supports dropping NaNs on floats.
It would be nice to support dropping Nans inside float vectors and shown below:
```
import turicreate as tc
import numpy as np
data = tc.SFrame({'a':[[0, np.nan],[1,2]]})
Columns:
	a	array

Rows: 2

Data:
+------------+
|     a      |
+------------+
| [0.0, nan] |
| [1.0, 2.0] |
+------------+
[2 rows x 1 columns]

data.dropna()
```

=> Currently nothing is dropped, it would be nice if the first row would be dropped.

Jarvi-Izana:
for a flex_vec, if it contains nan, should it be treated as nan?

Jarvi-Izana:
track by #2102

PR #2102, opened 2019-06-28
enable recursive dropna for elements in SFrame

Jarvi-Izana:
feature request #1804 

per user request:
```
import turicreate as tc
import numpy as np
data = tc.SFrame({'a':[[0, np.nan],[1,2]]})
data.dropna()
Columns:
	a	array

Rows: 1

Data:
+------------+
|     a      |
+------------+
| [1.0, 2.0] |
+------------+
[1 rows x 1 columns]

data = tc.SFrame({'a':[[0, None, "sd"],[1,2]]})
data.dropna() # same result
```

when dropna is called, the program will recursively check each cell in sframe whether it contains **nan** for array type, **None** for other types. if it contains, it will be dropped.

==========update on July 11, 2019=============
user can opt in dropna with recursive check on whether nan values are contained in each cell.
```
data.dropna(recursive=True)
```

### DIFF
--- a/src/core/data/sframe/gl_sframe.cpp
+++ b/src/core/data/sframe/gl_sframe.cpp
@@ -922,14 +922,14 @@ gl_sframe gl_sframe::sort(const std::vector<std::pair<std::string, bool>>& colum
   return get_proxy()->sort(keys, order);
 }
 gl_sframe gl_sframe::dropna(const std::vector<std::string>& columns,
-                            std::string how) const {
-  auto ret = get_proxy()->drop_missing_values(columns, how == "all", false, false);
+                            std::string how, bool recursive) const {
+  auto ret = get_proxy()->drop_missing_values(columns, how == "all", false, recursive);
   ASSERT_EQ(ret.size(), 2);
   return *(ret.begin());
 }
 std::pair<gl_sframe, gl_sframe> gl_sframe::dropna_split(const std::vector<std::string>& columns,
-                                                        std::string how) const {
-  auto ret = get_proxy()->drop_missing_values(columns, how == "all", false, false);
+                                                        std::string how, bool recursive) const {
+  auto ret = get_proxy()->drop_missing_values(columns, how == "all", false, recursive);
   ASSERT_EQ(ret.size(), 2);
   return {*(ret.begin()), *(ret.rbegin())};
 }
--- a/src/core/data/sframe/gl_sframe.cpp
+++ b/src/core/data/sframe/gl_sframe.cpp
@@ -923,13 +923,13 @@ gl_sframe gl_sframe::sort(const std::vector<std::pair<std::string, bool>>& colum
 }
 gl_sframe gl_sframe::dropna(const std::vector<std::string>& columns,
                             std::string how) const {
-  auto ret = get_proxy()->drop_missing_values(columns, how == "all", false);
+  auto ret = get_proxy()->drop_missing_values(columns, how == "all", false, false);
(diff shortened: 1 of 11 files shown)


ylow:
This is a change of behavior. Should we put it behind an option?

Jarvi-Izana:
Thanks to @znation for pointing out the `nan == nan` will always yield false. That saves my day by not comparing the nan part after a \_split call.

Jarvi-Izana:
yeah, it's a change of behavior. We can provide one extra flag to let the user use it explicitly.

znation:
I agree with Yucheng. Let's keep the existing behavior as the default, and make this new behavior available under an optional parameter.

Jarvi-Izana:
sure. the code change is its the way.

Jarvi-Izana:
```
std::list<std::shared_ptr<unity_sframe_base>> unity_sframe::drop_missing_values(
    const std::vector<std::string> &column_names, bool all, bool split, bool recursive=false)
```
even though adding default argument can be easier but it's not a good practice to provide a default argument for the virtual function which is intended to be derived later. The child class may override the base class implementation with default args, which may cause a surprising issue that default args are chosen by its declared type instead of its real client type.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1542,23 +1585,49 @@ std::list<std::shared_ptr<unity_sframe_base>> unity_sframe::drop_missing_values(
...
-                                  size_t num_missing_values = 0;
-                                  for(const auto &i : column_indices) {
-                                    if(row[i].is_na()) {
-                                      ++num_missing_values;
-                                    }
-                                  }
-                                  return (num_missing_values != column_indices.size());
-                                };
+    // for perf, I choose not to use std::function to wrap _contains_nan or is_nan
+    if (recursive) {
```
Jarvi-Izana: After talking with @ylow, I hesitate to use std::function to wrap is_na and _contains_nan. I dup the code to avoid extra virtual calls on each cell.

ylow: I wouldn't worry about it unless a perf profile suggests otherwise. Premature optimization is not a good thing.

Jarvi-Izana: I think it's ok to dup the code for this simple case (if we'd like to provide more options in the future, we can do refactor later). The optimization is to conduct if-else check once to register the right func for the filter before the filter is invoked for thousands of rows from a sframe, without paying the overhead of virtual function call for each row.

src/core/data/sframe/gl_sframe.cpp
```diff
@@ -923,13 +923,13 @@ gl_sframe gl_sframe::sort(const std::vector<std::pair<std::string, bool>>& colum
 }
 gl_sframe gl_sframe::dropna(const std::vector<std::string>& columns,
                             std::string how) const {
```
hoytak: This API is meant to mimic the python one, so let's go ahead and add in the parameter here as well to mimic the python API, with a default of false, along with a test around line 276 of test/unity/gl_sframe.cxx. After that, this PR should be ready to go.

hoytak:
Looking excellent.  One small thing, then let's get this in.

src/core/data/sframe/gl_sframe.cpp
Jarvi-Izana: Thanks for the catch!

Jarvi-Izana:
@hoytak checked in. Good to go.

Issue #1912, opened 2019-05-22
SFrameBuilder Segmentation Fault on appending incorrect number of columns


guyinbar:
```
In [2]: import turicreate as tc
In [3]: sb = tc.SFrameBuilder([int,str])
In [4]: sb.append([1,'foo', 1.1])

Segmentation fault: 11

Stack trace returned 10 entries:
[bt] (0) 0   libmxnet.so                         0x000000011e7b07cf libmxnet.so + 59343
[bt] (1) 1   libmxnet.so                         0x000000011fa9b3c6 MXNDListFree + 3450630
[bt] (2) 2   libsystem_platform.dylib            0x0000000101c22b5d _sigtramp + 29
[bt] (3) 3   ???                                 0x0000000102d517d0 0x0 + 4342486992
[bt] (4) 4   libunity_shared.dylib               0x0000000114d5b322 turi::sarray_group_format_writer_v2<turi::flexible_type>::write_segment(unsigned long, std::__1::vector<turi::flexible_type, std::__1::allocator<turi::flexible_type> >&&) + 66
[bt] (5) 5   libunity_shared.dylib               0x0000000114e54d8e std::__1::__function::__func<turi::sframe::get_output_iterator(unsigned long)::$_32, std::__1::allocator<turi::sframe::get_output_iterator(unsigned long)::$_32>, void (std::__1::vector<turi::flexible_type, std::__1::allocator<turi::flexible_type> > const&)>::operator()(std::__1::vector<turi::flexible_type, std::__1::allocator<turi::flexible_type> > const&) + 446
[bt] (6) 6   cy_sframe_builder.so                0x000000011dbafaba __pyx_f_10turicreate_6cython_17cy_sframe_builder_23UnitySFrameBuilderProxy_append(__pyx_obj_10turicreate_6cython_17cy_sframe_builder_UnitySFrameBuilderProxy*, _object*, unsigned long, int) + 138
[bt] (7) 7   cy_sframe_builder.so                0x000000011dbb5052 __pyx_pw_10turicreate_6cython_17cy_sframe_builder_23UnitySFrameBuilderProxy_5append(_object*, _object*, _object*) + 146
[bt] (8) 8   Python                              0x0000000100749545 _PyCFunction_FastCallDict + 245
[bt] (9) 9   Python                              0x00000001007d23e4 call_function + 612

```

Jarvi-Izana:
resolved in pr #2103.

PR #2103, opened 2019-07-01
fix sigfault when appending size mismatched row into SFrameBuilder

Jarvi-Izana:
resolved issue #1912. adding error checking logic when writing from const ref object into sframe.

### DIFF
--- a/src/core/storage/sframe_data/sframe.cpp
+++ b/src/core/storage/sframe_data/sframe.cpp
@@ -539,6 +539,12 @@ sframe::iterator sframe::get_output_iterator(size_t segmentid) {
   return sframe::iterator(
       [=](const std::vector<flexible_type>& val)->void{
         // check that all the types match up
+        if (UNLIKELY(val.size() != ctypes.size())) {
+          std::stringstream ss;
+          ss << "Write to sframe with row size mismatch. "
+             << "Expected: " << ctypes.size() << " Actual: " << val.size();
+          log_and_throw(ss.str());
+        }
         bool badtype = false;
         for (size_t i = 0;i < val.size(); ++i) {
           if (val[i].get_type() != flex_type_enum::UNDEFINED &&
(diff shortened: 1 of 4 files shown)


src/python/turicreate/data_structures/sframe_builder.py
```diff
@@ -58,7 +58,7 @@ class SFrameBuilder(object):
 
     >>> sb.append_multiple([[2,2.0,"2"],[3,3.0,"3"]])
 
-    >>> sb.close()
+    >>> new_sf = sb.close()
```
TobyRoseman: The output shown here doesn't match what the code will produce. I like the idea of using a temporary variable. I would add `print(new_sf)` so the code will produce the output shown.

src/core/storage/sframe_data/sframe.cpp
```diff
@@ -539,6 +539,12 @@ sframe::iterator sframe::get_output_iterator(size_t segmentid) {
   return sframe::iterator(
       [=](const std::vector<flexible_type>& val)->void{
         // check that all the types match up
+        if (UNLIKELY(val.size() != ctypes.size())) {
+          std::stringstream ss;
+          ss << "Write to sframe with row size mismatch. "
```
TobyRoseman: I think this would be clearer:

```
          ss << "Can not write to SFrame, got the wrong number of columns. "
	     << "Expected: " << ctypes.size() << " columns. Got: " << val.size() << " columns.";
```

src/python/turicreate/data_structures/sframe_builder.py
Jarvi-Izana: yeah! That's a good idea! Thanks!

src/core/storage/sframe_data/sframe.cpp
Jarvi-Izana: will add more detailed info!

PR #2106, opened 2019-07-02
Fixing training/validation accuracy in progress table in C-API Activity Classifier

shreyajain17:
The training/validation accuracies in the training progress table are different from the ones stored in the state at the end. This was because the progress table was using 'per_window' logic while the final metrics stored were calculated using 'per_row'. Now, both would consistently use 'per_row'.


src/toolkits/activity_classification/ac_data_iterator.hpp
```diff
@@ -108,6 +109,15 @@ class data_iterator {
...
     neural_net::shared_float_array weights;
 
+
+    /**
+     * An array with shape: (requested_batch_size, 
+     * 1, prediction_window * predictions_in_chunk, 1)
+     *
+     * Each row is the sequence of raw class labels (indices) corresponding every feature.
+     */
+    neural_net::shared_float_array ground_truth;
```
nickjong: Let's call this something like `labels_per_row`, so it's more obvious that this is really the same thing as `labels` but more granular

src/toolkits/activity_classification/ac_data_iterator.hpp
```diff
@@ -108,6 +109,15 @@ class data_iterator {
      */
     neural_net::shared_float_array weights;
 
+
+    /**
+     * An array with shape: (requested_batch_size, 
+     * 1, prediction_window * predictions_in_chunk, 1)
+     *
+     * Each row is the sequence of raw class labels (indices) corresponding every feature.
```
nickjong: I'm not sure I follow what "corresponding every feature" means. Maybe "sequence of raw class labels (indices) for each individual sample"? What do you think?

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -94,13 +96,11 @@ float cumulative_chunk_accuracy(size_t prediction_window, size_t num_classes,
...
     const shared_float_array& output_chunk = output[i];
-    const shared_float_array& label_chunk = batch.labels[i];
+    const shared_float_array& label_chunk = batch.ground_truth[i];
     data_iterator::batch::chunk_info info = batch.batch_info[i];
-    size_t num_predictions = (info.num_samples + prediction_window - 1) / prediction_window;
-    size_t num_correct_predictions = count_correct_predictions(num_classes, output_chunk, label_chunk, num_predictions);
-
-
-    cumulative_per_batch_accuracy += static_cast<float>(num_correct_predictions) / num_predictions;
+    //size_t num_predictions = (info.num_samples + prediction_window - 1) / prediction_window;
```
nickjong: Let's remove this debugging code

src/toolkits/activity_classification/ac_data_iterator.hpp
shreyajain17: That makes more sense!

src/toolkits/activity_classification/ac_data_iterator.hpp
shreyajain17: Yeah I think that would be more intuitive!

PR #2108, opened 2019-07-02
Updated syntax for Python 2.7 and 3.6 Build Breaks

abhishekpratapa:
- Updated the testing syntax for Python 2.7 and 3.6 Build Breaks in master.

### DIFF
--- a/src/python/turicreate/test/test_sframe_builder.py
+++ b/src/python/turicreate/test/test_sframe_builder.py
@@ -62,7 +62,7 @@ class SFrameBuilderTest(unittest.TestCase):
             sb.append(extra_data)
 
         sb = SFrameBuilder(self.all_types)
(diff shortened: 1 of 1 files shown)


abhishekpratapa:
Fixes #2109

Jarvi-Izana:
Thanks for the catch! That's weird because I ran the local test with 3.6 successfully.

abhishekpratapa:
> Thanks for the catch! That's weird because I ran the local test with 3.6 successfully.

Normally this wouldn't happen, we have something inconsistent in our Travis Builds right now. We test the wheel that we've built with Python 3.6 on a Python 2.7 machine.

That's the reason for the build breaks on both Python 2.7 and Python 3.6 in travis. The tests are running in 2.7.

Jarvi-Izana:
glad to know the context! That's super helpful!

Issue #2110, opened 2019-07-02
Travis 3.6 builds are tested on a 2.7 machine


abhishekpratapa:
- Builds that are created on a Python 3.6 machine are tested with a Python 2.7 machine.

PR #2111, opened 2019-07-02
Fixed travis Python 3.6 build

abhishekpratapa:
Install virtualenv before running build-script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,12 @@ script:
   - tar xf cmake-3.13.4-Linux-x86_64.tar.gz
   - export PATH=${PWD}/cmake-3.13.4-Linux-x86_64/bin:${PATH}
 
-  # Use the correct virtualenv
+  # On certain Travis builds the machines contain the correct `python` and `pip`
+  # versions. Running `which python` and `which pip` points to the right path
+  # to both the `python` and `pip` programs. Running `which virtualenv` uncovers
+  # that some machines on travis have `virtualenv` installed in their current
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -34,6 +34,9 @@ env:
 script:
   - free -m
 
+  # Hacky af: But to temporarily fix the travis build breaks
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then alias python=python3; alias pip=pip3; fi
```
TobyRoseman: I would recommend doing `python=python3.6` and `pip=pip3.6`

abhishekpratapa: Curious. I'll try that see if the build fails.

abhishekpratapa: Updated the builds should work.

abhishekpratapa: Updated. The builds should work.

abhishekpratapa:
#2110

.travis.yml
```diff
@@ -43,6 +43,9 @@ script:
   - tar xf cmake-3.13.4-Linux-x86_64.tar.gz
   - export PATH=${PWD}/cmake-3.13.4-Linux-x86_64/bin:${PATH}
 
+  # Use the correct virtualenv
```
znation: Why does installing virtualenv suddenly cause it to use the correct virtualenv? A bit more of a comment here explaining the fix would be great.

abhishekpratapa: Updated comment.

PR #2114, opened 2019-07-03
C++ object detector should allow user to specify detection grid size

nickjong:
Brings the C++ implementation to parity with the Python changes for #2053

### DIFF
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -53,10 +53,6 @@ constexpr size_t MEMORY_REQUIRED_FOR_DEFAULT_BATCH_SIZE = 4294967296;
 // We assume RGB input.
 constexpr int NUM_INPUT_CHANNELS = 3;
 
-// Annotated and predicted bounding boxes are defined relative to a
-// GRID_SIZE x GRID_SIZE grid laid over the image.
-constexpr int GRID_SIZE = 13;
-
 // The spatial reduction depends on the input size of the pre-trained model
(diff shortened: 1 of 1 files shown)


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -205,6 +203,20 @@ void object_detector::init_options(
       FLEX_UNDEFINED,
       1,
       std::numeric_limits<int>::max());
+  options.create_integer_option(
+      /* name          */ "grid_height",
```
nickjong: n.b. I added the comments here in part to prevent clang-format from condensing these into one or two lines. The comments do make things a bit more readable, I think

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -375,15 +389,15 @@ variant_map_type object_detector::perform_evaluation(gl_sframe data,
...
-      /* n */     options.value("batch_size"),
-      /* c_in */  NUM_INPUT_CHANNELS,
-      /* h_in */  GRID_SIZE * SPATIAL_REDUCTION,
-      /* w_in */  GRID_SIZE * SPATIAL_REDUCTION,
-      /* c_out */ num_output_channels,
-      /* h_out */ GRID_SIZE,
-      /* w_out */ GRID_SIZE,
-      get_prediction_config(),
-      get_model_params());
+      /* n       */ options.value("batch_size"),
```
nickjong: clang-format wanted to remove the extra white space between the comment and the code, so I just moved the formatting white space into the comment....

Issue #1621, opened 2019-03-18
SFrame.join() should enable customized suffix for overlapping columns


guyinbar:
`SFrame.join()` should enable customized suffix for overlapping columns. Currently, a `.1` suffix is added to the right SFrame.

Jarvi-Izana:
please follow up with #2121.

PR #2121, opened 2019-07-09
set customized column name when joining two sframes

Jarvi-Izana:
#1621

User provided conflict resolution should not introduce a new name collision.

Two import restriction here:
1. names in sframe are unique
2. names in sframe is non-empty.

We can conclude that if we merge 2 sframe, the same name can at most appear twice before a merge (join keys, for instance).

The rule is simple, providing a resolution when a collision happens when merging columns. 

Example:
2 sframes, left and right  join on key 'A'.
<A, B, C, D>, <A, C, F> represents their column names, respectively.

if we do` left.join(right, on='A')`, 'C' from the right will be renamed to 'C.1'

if we provide our customized solution: 
`left.join(right, on='A', alter_name={'C':'X'})`, the right 'C' will output as 'X'. So in the end we get <A,B,C,D,X,F>

The name resolution operation should be **order-independent**. For instance,
`alter_name={'C':'X', 'D':'Y'}` has 2 operations, it should not affect the final result due to whether `{'C':'X'}` is applied before `{'D' :'Y'}` or not.

Under the hood, alter_name is a dictionary where `key` and `value` is restricted by rules:
1. `key` must be from right sframe and cannot be one of join keys (because if it's join key, it will not be triggered since join keys are meant to be duplicated).
2. `value`s should be unique
3. `value`s should not be any names from left or right column names. 
rule 3 also means `value`s should not be equal to `key`s because of rule 1.

~2. `value` cannot be equal to any names from left (also it's true that `value` cannot be join keys because join keys are part of left sframe)~

~The resolution is sequential. That means, the resolution is not commutative.~
~`left.join(right, on='A', alter_name={'C':'F', 'F': 'G'})~
~out put will be <A,B,C,D,F,G>. This is useful to swap names **iff** collision is detected.~

~The way how it works is simple, the solver first loads all names from left sframe.~
~1. A,B,C,D~
~2. A,B,C,D <- when adding 'C' into table name list, a collision is detected, it just replace 'C' to 'F', regardless the fact 'F' will be added later. As long as for all names visited, there's no collision, the resolution attempt will proceed.~
~3. A,B,C,D,F <- 'F', trivial.~


src/python/turicreate/data_structures/sframe.py
```diff
@@ -4349,7 +4362,18 @@ def join(self, right, on=None, how='inner'):
...
+            if alter_name is None:
+                return SFrame(_proxy=self.__proxy__.join(right.__proxy__, how, join_keys))
+            if type(alter_name) is dict:
+                left_names = self.column_names()
+                right_names = right.column_names()
+                for (k, v) in alter_name.items():
+                    if k not in right_names or k in join_keys:
+                        raise KeyError("Redundant key %s for collision resolution" % k)
+                    if k != v and v in left_names:
+                        # if v in right_names, things are more complicated and will be handled by proxy
```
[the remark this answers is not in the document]
Jarvi-Izana: the logic will be handled by Cpp backend. A rudimentary error check is provided here.

Jarvi-Izana:
@TobyRoseman the result actually is deterministic. Because the resolution is done through the ordinal order in which names appear from left to right.

src/core/storage/sframe_data/join_impl.cpp
```diff
@@ -156,67 +161,154 @@ hash_join_executor::hash_join_executor(const sframe &left,
     _right_join = true;
   }
 
+  ASSERT_EQ(_left_join_positions.size(), _right_join_positions.size());
+
+  /* handy script for reverse lookup */
+  for(size_t i = 0; i < _left_join_positions.size(); ++i) {
+    auto ret = _right_to_left_join_positions.emplace(_right_join_positions[i],
```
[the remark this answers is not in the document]
Jarvi-Izana: I think `auto` can avoid some surprising implicit type conversions. Also, it makes code clear, brief and loosely coupled with a specific type, which can make refactoring code more easily.

For readability, the nowadays IDE can take advantage of type inference for `auto` to have a quick glance at the object's type.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -4264,6 +4264,19 @@ def join(self, right, on=None, how='inner'):
...
 
+        alter_name : None | dict
+            user provided names to resolve column name conflict when merging two sframe.
+
+            * 'None', then default conflict resolution will be used. For example, if 'X' is
+            defined in the sframe on the left side of join, and there's an column also called
+            'X' in the sframe on the right, 'X.1' will be used as the new column name when
+            appending the column 'X' from the right sframe, in order to avoid column name collision.
+
+            * if a dict is given, the dict key should be obtained from column names from the right
```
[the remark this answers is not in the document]
Jarvi-Izana: I think if a user provides a resolution, we should make sure it won't lead to further conflict. An exception will be thrown if their solution won't work.

src/python/turicreate/data_structures/sframe.py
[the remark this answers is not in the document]
Jarvi-Izana:
```
The resolution is sequential. That means, the resolution is not commutative.
`left.join(right, on='A', alter_name={'C':'F', 'F': 'G'})
out put will be <A,B,C,D,F,G>. This is useful to swap names iff collision is detected.

The way how it works is simple, the solver first loads all names from left sframe.

A,B,C,D
A,B,C,D <- when adding 'C' into table name list, a collision is detected, it just replace 'C' to 'F', regardless the fact 'F' will be added later. As long as for all names visited, there's no collision, the resolution attempt will proceed.
A,B,C,D,F <- 'F', trivial.
```
the logic of further resolution is kind of tricky.

src/core/storage/sframe_data/join_impl.cpp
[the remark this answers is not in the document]
ylow: auto is good for readability for interesting derived types like iterators and stuff where the actual type of the object is "far away", or types may change due to overloading.
```
void do_something(std::map<std::string, size_t>& idx) {
   ...
    auto iter = idx.find(...)
}
```
where the alternative is kind of inconvenient
```
void do_something(std::map<std::string, size_t>& idx) {
    ...
    std::map<std::string, size_t>::iterator iter = idx.find(...) 
}
```
And also requires a lot of code changes if the argument types change in anyway.
```
void do_something(const std::map<std::string, size_t>& idx) {
    ...
    std::map<std::string, size_t>::iterator iter = idx.find(...)  // compile error. should be const_iterator
}
```

Also useful for things like range based for loops over complicated types (really just as above the "correct" type for a range based for loop is std::map<std::string, size_t>::value_type, that gets a little annoying.

But I will prefer that auto is not used for simple types, or potentially platform dependent types. 
```
uint64_t i = 0; // good. I can tell immediately what width this integer is and if it is signed.
auto i = 0; // bad. Hard to tell immediately.
float j = 0.0; //good. I know that this is a 32-bit float
auto j = 0.0; //bad. How wide is j?
```
This is crucial. For instance when we were porting from Linux/Mac to Windows, it was far easier to search for the "long" keyword to replace all uses of "long" with int64_t or "long long" (long is 32-bit on Windows, 64-bit elsewhere). Whereas if we had used "auto" this would have been significantly harder to fix.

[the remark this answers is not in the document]
ylow: There are some uses of auto here for instance in integer for loops, which would be better to replace with a standard integer type like size_t, ssize_t, int64_t, or uint64_t.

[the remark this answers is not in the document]
Jarvi-Izana: that's a good point. I will refrain from using auto for POD types.

TobyRoseman:
> @TobyRoseman the result actually is deterministic. Because the resolution is done through the ordinal order in which names appear from left to right.

@Jarvi-Izana - are you certain the names will always appear in the same order? I believe the order of columns in an SFrame is not deterministic.

src/core/storage/sframe_data/join_impl.cpp
```diff
@@ -156,67 +161,154 @@ hash_join_executor::hash_join_executor(const sframe &left,
...
+          ss << "user provided column name { " << itr->second << " } conflicts with table name used in SFrame";
+          log_and_throw(ss.str());
+        }
+        _output_column_types.push_back(right.column_type(right.column_index(name)));
+        _output_column_names.push_back(itr->second);
+        new_table.insert(itr->second);
+      } else {
+        _output_column_types.push_back(right.column_type(right.column_index(name)));
+        // default collision resolv. see SFrame::generate_valid_column_name.
+        // if SFrame::generate_valid_column_name changes, this will break.
```
Jarvi-Izana: I'm pretty sure the order is applied as the same as what SFrame does. Under the hood, first in and first serves. @TobyRoseman you can actually verify it by reading `SFrame::generate_valid_column_name`.

ylow: Column order is deterministic.

TobyRoseman: @ylow - I'm talking about column order across sessions. The column order can't be deterministic if a dictionary is used to create the SFrame.

Here is the use case I’m concerned about: suppose a user has a script that they run regularly. The script creates some Sframes. Then does a merge using this new feature. Then does some more stuff and makes assumptions about the names of columns. This script could potentially break because the column names across session is not deterministic.

ylow: It is true that SFrame order will not be deterministic if I do
```
tc.SFrame({'a':[1,2,3],'b':[2,3,4]...})
```
But since an SFrame cannot contain two columns of the same name, it is unclear to me that this causes issues? Can you come up with an example?

Jarvi-Izana: @ylow https://github.com/apple/turicreate/issues/2129

the name resolution {A->B, B->C} assumes A appears ahead of B in SFrame. But if you do
```
tc.SFrame({'A':[1,2,3],'B':[2,3,4]...})
```
Then it's non-deterministic that A appears before B.

Jarvi-Izana: @TobyRoseman @ylow 
SFrame left: key, A
SFrame right: Key, A, B
if we resolve in an order A->B, then B->C, the output of join on `Key` is Key, A, B, C.

but somehow we have the right frame built from a dict to be
SFrame right: Key, B, A.
Then it will raise exception:
> Key, A
> Key, A, B
> Key, A, B, <- adding A, conflict found, replace A to B, conflict found again, raise an exception. 

So the resolution depends on the order in SFrame. One way to resolve this is to make the operation order independent. The other is to make sfame construction deterministic.

Jarvi-Izana:
As #2129 is closed, I will make this operation order-independent instead of requiring sframe to provide a deterministic column sequence when constructed from the associative data structures.

src/core/storage/sframe_data/join_impl.cpp
```diff
@@ -156,67 +161,188 @@ hash_join_executor::hash_join_executor(const sframe &left,
...
+  _output_column_types = left.column_types();
+
+  std::set<std::string> new_table = {
+      std::make_move_iterator(col_names_original_order.begin()),
+      std::make_move_iterator(col_names_original_order.end())};
+
+  // get orginal right sframe's column names
+  col_names_original_order = right.column_names();
+
+  /* check the name resolution is order dependent or not.
```
Jarvi-Izana: make the resolution ops order-independent.

Jarvi-Izana: @TobyRoseman

src/core/storage/sframe_data/join_impl.cpp
```diff
@@ -156,67 +161,188 @@ hash_join_executor::hash_join_executor(const sframe &left,
...
+    if (unique_names_from_right.count(entry.second)) {
+          std::stringstream ss;
+          ss << "user provided resolution name { " << entry.second
+             << " } is not allowed to be same with any name in right SFrame.";
+          log_and_throw(ss.str());
+    }
+  }
+
+  // check the original right sframe columns
+  for (auto& name : col_names_original_order) {
```
hoytak: Minor: Let's use const auto& where appropriate -- use auto& only if you explicitly want to change the contents present.

src/core/storage/sframe_data/join_impl.cpp
```diff
@@ -156,67 +161,188 @@ hash_join_executor::hash_join_executor(const sframe &left,
...
+   * resolution {A -> B} is order dependent iff
+   * 1. A belongs to right.column_names()
+   * 2. B is not equal to keys from right.column_names()
+   * 3. B should be unique to other B'
+  */
+  std::set<std::string> unique_names_from_right = {
+      std::begin(col_names_original_order), std::end(col_names_original_order)};
+
+  std::set<std::string> unique_names_from_user;
+  for (auto& entry : _alter_names_right) {
```
hoytak: Minor, below too: Let's use const auto& where appropriate -- use auto& only if you explicitly want to change the contents present.

src/core/storage/sframe_data/join_impl.cpp
```diff
@@ -382,51 +508,14 @@ sframe hash_join_executor::grace_hash_join() {
...
-      // wasn't renamed.  This is an extra step we take to keep the result
-      // looking normal.
-      auto orig_name_entry = _changed_dup_names.find(i);
-      if(orig_name_entry == _changed_dup_names.end()) {
-        swapped_names[cur_column_idx] = result_frame.column_name(i);
-      } else {
-        swapped_names[cur_column_idx] = orig_name_entry->second;
-      }
-      ++cur_column_idx;
+    for (auto& entry : _reverse_to_original) {
```
hoytak: Same, let's use const auto& here.

src/core/storage/sframe_data/join_impl.cpp
Jarvi-Izana: Got it. For this one, name is actually used and will be moved in some conditions.

src/core/storage/sframe_data/join_impl.cpp
Jarvi-Izana: This feature is no longer used due to making the conflict resolution independent.

PR #2125, opened 2019-07-12
fix for turicreate cannot access private enterprise aws s3

Jarvi-Izana:
Private S3 users should explicitly set their region name by using env var `TURI_S3_REGION` in order to access the private cloud.


znation:
@Jarvi-Izana Please add this new config variable to the documentation in `src/python/turicreate/config/__init__.py` as well.

Jarvi-Izana:
Will do.

PR #2127, opened 2019-07-13
Allow the creation of Plot objects from Vega spec

znation:
* Adds a new parameter to the Plot constructor in Python and C++,
  `vega_spec`, which can take either a Vega or Vega-Lite spec.
* Plots created by vega spec do not have a data transformer or data
  source; the include_data parameter on get_spec is ignored.
* Exposes the new plot constructor through C API.


src/visualization/server/plot.cpp
```diff
@@ -21,14 +22,32 @@
...
+      m_vega_spec(vega_spec),
+      m_size_array(size_array),
+      m_transformer(transformer) {
+        // This constructor expects builtin TC format strings
+        // that already have their newlines removed.
+        DASSERT_EQ(m_vega_spec.find("\n"), std::string::npos);
+      }
+
     const std::string& Plot::get_id() const {
+      static auto uuid_generator = boost::uuids::random_generator();
```
Saba9: Why not put this in the if?

znation: Since it's static, it's only going to run once (on the first call to this function). I think the semantics of static initialization are clearer if it's not inside an `if`, and we're going to use it on first call to this function anyway.

PR #2130, opened 2019-07-15
Quick fix to get compilation to work on linux with GCC 7.4.

hoytak:


src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -2046,6 +2051,7 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
         ASSERT_UNREACHABLE();
       }
     }
+#pragma GCC diagnostic pop
```
znation: Should the pop be in the same scope as the push? Feels weird to me to have the push inside a method, and pop outside.

hoytak: Good catch!  It was meant to be.

hoytak: Good catch -- it was meant to be!

PR #2131, opened 2019-07-15
VegaRenderer no longer holds reference to CGImageRef

Saba9:
Fixes bug where `_image` was being released twice; once by the caller of `CGImage` and once in `VegaRenderer`'s destructor.


znation:
Thanks for fixing that!

PR #2079, opened 2019-06-21
Python 3.7 macOS Support

TobyRoseman:
I believe this is almost all of the changes we need (in this repository) to support Python 3.7. The only other changes that I know will be needed are in`.gitlab-ci.yml`.

Currently calling `predict(...)` on a coremltools model is broken in Python 3. This is being tracked in [an issue](https://github.com/apple/coremltools/issues/372) in the coremltools repository. We'll need to wait for that to get fixed and for a new version to be released.

All non-CoreML related unit tests pass for Python 3.7 on macOS.

This also fixes #1455

PR #783, opened 2018-06-28
Clean up Python version handling in configure

znation:
Virtualenv always implicitly provided the Python version. Prior to
this commit, that version had to match the user's chosen Python
version in ./configure. A mismatch would silently lead to other issues
further in execution.

With this commit, we no longer accept --pythonxx flags on ./configure,
since the Python version should explicitly come from virtualenv.
Instead, virtualenv is now documented in BUILD.md as the way to build
for different Python versions. Pointing to a different virtualenv binary
seems to work across OSes and Python distributions.

* Fixes #494 by using `python -m pip` instead of a pip binary name.
* Fixes #553 by removing the two conflicting sources of truth.
* Fixes #562 (same root cause as #553).

PR #2132, opened 2019-07-15
Added option in configure script to specify python executable.

hoytak:
Now, to set up everything with, e.g. python3.7, you can just use:

```
./configure --python=`which python3.7` 
```

This allows one to easily compile for certain python versions.

### DIFF
--- a/configure
+++ b/configure
@@ -33,7 +33,8 @@ function print_help {
   echo "  --no-capi                         Skip building the C API."
   echo
   echo "  --with-python (default)           Build python components."
-  echo "  --no-python"
+  echo "  --python PYTHON_EXE               Manually set the python executable to use."
+  echo "  --no-python                       Build only C++ / C-API components."
   echo
   echo "  --with-visualization (default)    Build Turi Create visualization client."
   echo "  --no-visualization"
@@ -103,7 +104,7 @@ function install_python_toolchain {
   echo "Using virtualenv at $VIRTUALENV."
 
   pushd ${TURI_HOME}
-  VIRTUALENV=${VIRTUALENV} ./scripts/install_python_toolchain.sh
+  VIRTUALENV=${VIRTUALENV} PYTHON_EXECUTABLE=${python_exe} ./scripts/install_python_toolchain.sh
   popd
 }
 
@@ -113,6 +114,7 @@ python_only=0
 default_yes=0
 
 with_python=1
+python_exe=`which python`
 
 with_system_cmake=1
 with_ccache=1
@@ -155,6 +157,8 @@ while [ $# -gt 0 ]
     --no-capi)              with_capi=0;;
 
     --with-python)          with_python=1;;
+    --python=*)             python_exe=${1##--python=} ;;
+    --python)               shift; python_exe="$1";;
     --no-python)            with_python=0;;
 
     --with-extra-opt-flags) with_extra_opt_flags=1;;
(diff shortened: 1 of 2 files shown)


nickjong:
In https://github.com/apple/turicreate/pull/783 we explicitly removed the command line option to specify Python version, replacing it with `--virtualenv`. Does that not satisfy your use case?

TobyRoseman:
@hoytak - take a look at #2079 I believe this is the correct way to add python 3.7 support

hoytak:
> @hoytak - take a look at #2079 I believe this is the correct way to add python 3.7 support

@TobyRoseman -- How does that add support for using a system python 3.7?  

Here's my use case.  On linux, with no docker.  The python 3.7 executable is available only with python3.7 explicitly.  virtualenv is installed through python 2.7, with no virtualenv that explicitly defaults to the python I want.  

With this change, it just pushes the python version support to the virtualenv executable, which works across python arguments.  In addition, it mimics the command line option for virtualenv.

hoytak:
> In #783 we explicitly removed the command line option to specify Python version, replacing it with `--virtualenv`. Does that not satisfy your use case?

Hey @nickjong, this way does not work if there is no virtualenv executable installed that defaults to the correct python environment.  Virtualenv is designed to work independent of it's base python version through use of the --python argument, so there isn't really a need to have multiple virtualenv versions installed.  I could use the existing command through setting `./configure --virtualenv "virtualenv --python=python3.7"`, but I don't think that's very intuitive (or should be documented). 

(Also, I like this way as it mimics how virtualenv handles it as well.)

TobyRoseman:
@hoytak - To get python 3.7 working on Linux without docker just makes sure that the virtualenv in your path is for python 3.7. Or pass in that virtual environment using the `--virtualenv` parameter.

hoytak:
> @hoytak - To get python 3.7 working on Linux without docker just makes sure that the virtualenv in your path is for python 3.7. Or pass in that virtual environment using the `--virtualenv` parameter.

virtualenv is python version independent, so there isn't a virtualenv for a particular python version.  Yes, it defaults to the one it was installed with, but it works with any version.  See my comments above.

znation:
> > @hoytak - To get python 3.7 working on Linux without docker just makes sure that the virtualenv in your path is for python 3.7. Or pass in that virtual environment using the `--virtualenv` parameter.
> 
> virtualenv is python version independent, so there isn't a virtualenv for a particular python version. Yes, it defaults to the one it was installed with, but it works with any version. See my comments above.

And that's exactly why we don't need a code change to specify a Python version.

`./configure --virtualenv="virtualenv --python=\`which python3.7\`` should work in current master.

znation:
Or

```
export VIRTUALENV="virtualenv --python=`which python3.7`"
./configure
```

See https://github.com/apple/turicreate/blob/master/BUILD.md#compiling for other variations. We can list these variations there if it helps.

hoytak:
Yes, we can do that; it would probably be good to document it at least.  I just thought this would make the flow much easier and more intuitive; I don't have to read the docs or understand more advanced details of virtualenv to select the python version I want to work with.  As it is, I don't think it's intuitive or obvious how to do it (which I like).

znation:
> Yes, we can do that; it would probably be good to document it at least. I just thought this would make the flow much easier and more intuitive; I don't have to read the docs or understand more advanced details of virtualenv to select the python version I want to work with. As it is, I don't think it's intuitive or obvious how to do it (which I like).

I was going more for robust and consistent than intuitive or obvious. For instance, what happens if you specify:

```
./configure --virtualenv="virtualenv --python=python2" --python=python3.6
```

The reason I omitted `--python` in #783 was so that there would be exactly one way to specify Python version (through virtualenv command), and no ambiguity about the behavior. For most users, most of the time, the default `virtualenv` creates a Python environment in the desired Python version, and in my experience, when users want to override that, they usually expect to specify a different virtualenv vs. using their existing virtualenv (installed in the site-packages of a particular Python version) with a different Python version.

hoytak:
With discussion, it seems we can just do ./configure --virtualenv="virtualenv --python=python3.7"` to do this, so not doing this enhancement.

PR #2133, opened 2019-07-15
JSCanvas now falls back to system font when the requested font is unavailable

Saba9:
Fixes bug where calling `setFont` with an unavailable font would cause an exception.

### DIFF
--- a/src/visualization/vega_renderer/JSCanvas.m
+++ b/src/visualization/vega_renderer/JSCanvas.m
@@ -60,6 +60,69 @@
 @end
 
 @implementation VegaCGFontProperties
+- (instancetype)initWithString:(NSString*)fontStr {
+    self = [super init];
+    _cssFontString = fontStr;
+    _fontFamily = nil;
+    _fontSize = nil;
+    _fontVariant = nil;
+    _fontWeight = nil;
+    _fontStyle = nil;
+    _lineHeight = nil;
+
+    NSArray<NSString *> *elements = [fontStr componentsSeparatedByCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+    for (size_t i=0; i<elements.count; i++) {
+        NSString *element = elements[i];
+        if ([element isEqualToString:@"normal"]) {
+            continue;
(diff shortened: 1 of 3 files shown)


test/vega_renderer/examples/turicreate/simple_themed_bar_chart.vl.json
```diff
@@ -169,6 +169,7 @@
         },
         "guide-title": {
           "fill": "#fff",
+          "font": "not-a-font",
```
Saba9: I'm not a huge fan of having this in the spec but I wasn't sure of a better way to test behavior when an unavailable font is specified.

znation: I think this is fine. I agree there isn't really a better way to test the behavior (at the spec level) when an invalid or missing font is specified.

src/visualization/vega_renderer/JSCanvas.m
```diff
@@ -268,32 +267,41 @@ - (void)setFont:(NSString *)fontStr {
...
     }
-    assert(font != nil);
-    _nsFont = font;
+
+    if(newFont == nil) {
+        newFont = [NSFont systemFontOfSize:fontSize];
+        // TODO should we be updating _font to reflect the system font we've fallen back to?
+        NSLog(@"The specified font: '%@' is unavailable. Falling back to '%@'.", fontStr, [newFont displayName]);
+    } else {
+        _font = fontStr;
```
znation: Do we still need to store both `_font` and `_nsFont` on the instance of this class? Or should we get rid of `_font`?

Saba9: I'm not sure how else we would meet the requirements for the canvas `font` property otherwise https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/font

znation: Oh, right. I keep forgetting that's why it's in there.

PR #2134, opened 2019-07-15
add the doc TURI_S3_REGION

Jarvi-Izana:
add the doc for `TURI_S3_REGION`.

### DIFF
--- a/src/python/turicreate/config/__init__.py
+++ b/src/python/turicreate/config/__init__.py
@@ -250,11 +250,11 @@ def set_runtime_config(name, value):
     - *TURI_S3_ENDPOINT*: The S3 Endpoint to connect to. If not specified AWS
       S3 is assumed.
 
-    - *TURI_S3_REGION*: The S3 Region to connect to. If not specified AWS
-      S3 region is first inferred by looking up the url from a rudimentary
-      static table of commonly used url-region mappings on our side. If no
-      url is matched, no region information is set and AWS will do the
-      best guess for the region.
+    - *TURI_S3_REGION*: The S3 Region to connect to. If this environment variable
+      if not set, `AWS_DEFAULT_REGION` will be loaded if available. Otherwise,
+      S3 region is then inferred by looking up the commonly used url-to-region mappings
+      in our codebase. If no url is matched, no region information is set and AWS will
+      do the best guess for the region.
 
     **SSL Configuration**
 
--- a/src/python/turicreate/config/__init__.py
+++ b/src/python/turicreate/config/__init__.py
@@ -250,6 +250,9 @@ def set_runtime_config(name, value):
     - *TURI_S3_ENDPOINT*: The S3 Endpoint to connect to. If not specified AWS
       S3 is assumed.
 
+    - *TURI_S3_REGION*: The S3 Region to connect to. If not specified AWS
+      S3 region is assumed.
+
     **SSL Configuration**
 
     - *TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE*: The location of an SSL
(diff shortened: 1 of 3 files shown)


src/python/turicreate/config/__init__.py
```diff
@@ -250,6 +250,9 @@ def set_runtime_config(name, value):
     - *TURI_S3_ENDPOINT*: The S3 Endpoint to connect to. If not specified AWS
       S3 is assumed.
 
+    - *TURI_S3_REGION*: The S3 Region to connect to. If not specified AWS
```
znation: > `If not specified AWS S3 region is assumed.`

I am not sure this is correct -- specifically, there is more than one region in AWS, and I thought we try all of them sequentially (perhaps starting with the default, `us-east-1`). Though I can't find the code to do that -- the only code I can find seems to default to (and only try) `us-east-1`. @ylow, can you confirm which is the intended behavior here?

Even if we are only defaulting to a single region, we should specify which one it will default to (`us-east-1`, I think) instead of leaving it ambiguous here.

Jarvi-Izana: first to look up regions through end point
```
const std::map<std::string, std::string> AWS_S3_ENDPOINT_TO_REGION {
  {"s3.amazonaws.com","us-east-1"},
  {"s3-us-west-1.amazonaws.com","us-west-1"},
  {"s3-us-west-2.amazonaws.com","us-west-2"},
  {"s3-eu-west-1.amazonaws.com","eu-west-1"},
  {"s3-eu-central-1.amazonaws.com","eu-central-1"},
  {"s3-ap-southeast-1.amazonaws.com","ap-southeast-1"},
  {"s3-ap-southeast-2.amazonaws.com","ap-southeast-2"},
  {"s3-ap-northeast-1.amazonaws.com","ap-northeast-1"},
  {"s3-ap-northeast-2.amazonaws.com","ap-northeast-2"},
  {"s3-sa-east-1.amazonaws.com","sa-east-1"},
  {"s3-ap-south-1.amazonaws.com","ap-south-1"}}
```
if no association is found, will let `AWS` decide from the endpoint the user provide. On my end is us-west-1 if not specified.

znation: I see -- from this, it does look to me like the default is `us-east-1`, since most users would have `s3.amazonaws.com` as their endpoint. Or if the user doesn't set endpoint explicitly, does it bypass this lookup?

Jarvi-Izana: nah. I'm using a private cloud endpoint which is not in the list above. The region for s3 client is set to empty string. I happen to know the value is `us-west-1` by reading the response from aws afterwards.

znation: I see. In that case, we can probably say something like  "If not specified, the region is determined from the endpoint URL, and/or determined automatically on the server." in the doc string.

Jarvi-Izana: agreed. I shouldn't be this lazy next time. Thanks for poking into this pr.

hoytak: IMO, the workflow should be something like this: 

If TURI_S3_REGION is set, then use that.  Error out if it doesn't work. 
If TURI_S3_REGION is not set, then check the AWS_S3_REGION variable.  If set, then use that.  
If neither TURI_S3_REGION nor AWS_S3_REGION is set, then rely on the aws-cli library to make the decision.  I don't think it makes sense to hardcode aws regions if there's a way to get aws-cli to make that decision for us.

Jarvi-Izana: Good point! I will check what options aws client offers as we discussed yesterday.

hoytak: Sounds great!  Thanks for working on this!

Jarvi-Izana:
related issue #2125, where `AWS_DEFAULT_REGION` is loaded.

src/python/turicreate/config/__init__.py
Jarvi-Izana: Now AWS_DEDAULT_REGION will be loaded. I think we are good to go.

hoytak:
Looks good; let's get this in.

PR #2135, opened 2019-07-16
Adding support for annotations coordinate transform in OD

fareeha:
TuriCore expects each OD annotation to specify a bounding box as a dictionary with four keys: x, y, width, height. Currently, the code assumes annotations to be in pixel scale, centered at origin, with top-left image origin. We want to support conversion from any given coordinate system to one which the code consumes.

Annotation options added:
* `annotation_scale` with values`pixel` or `normalized`
* `image_origin` with values `top_left` or `bottom_left`
* `annotation_position` with values `center`, `top_left` or `bottom_left`

Default values are `annotation_scale`:`pixel`, `image_origin`:`top_left` and `annotation_position`:`center`, which is the currently assumed annotations input format. 

These options can be added in the `opts` dictionary passed in OD `train()`. At the backend, `od_data_iterator` parses these options and updates the annotation values in each batch.

### DIFF
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -224,19 +224,19 @@ void object_detector::init_options(
       1,
       std::numeric_limits<int>::max());
   options.create_integer_option(
-      /* name          */ "grid_height",
-      /* description   */
+      /* name             */ "grid_height",
+      /* description      */
       "Height of the grid of features computed for each image",
-      /* default_value */ 13,
-      /* lower_bound   */ 1,
-      /* upper_bound   */ std::numeric_limits<int>::max());
+      /* default_value    */ 13,
+      /* lower_bound      */ 1,
+      /* upper_bound      */ std::numeric_limits<int>::max());
   options.create_integer_option(
-      /* name          */ "grid_width",
-      /* description   */
+      /* name             */ "grid_width",
+      /* description      */
       "Width of the grid of features computed for each image",
-      /* default_value */ 13,
-      /* lower_bound   */ 1,
-      /* upper_bound   */ std::numeric_limits<int>::max());
+      /* default_value    */ 13,
+      /* lower_bound      */ 1,
+      /* upper_bound      */ std::numeric_limits<int>::max());
   options.create_integer_option(
       "random_seed",
       "Seed for random weight initialization and sampling during training",
@@ -244,26 +244,26 @@ void object_detector::init_options(
       std::numeric_limits<int>::min(),
       std::numeric_limits<int>::max());
   options.create_categorical_option(
-      /* name          */"annotation_scale",
-      /* description   */
+      /* name              */"annotation_scale",
+      /* description       */
       "Defines annotations scale: pixel or normalized",
-      /* default_value */ "pixel",
-      /* allowed_values*/ {flexible_type("pixel"), flexible_type("normalized")},
-      /*allowed_overwrite*/ false);
+      /* default_value     */ "pixel",
+      /* allowed_values    */ {flexible_type("pixel"), flexible_type("normalized")},
+      /* allowed_overwrite */ false);
   options.create_categorical_option(
-      /* name          */"image_origin",
-      /* description   */
+      /* name              */"image_origin",
+      /* description       */
       "Defines image origin: top_left or bottom_left",
-      /* default_value */ "top_left",
-      /* allowed_values*/ {flexible_type("top_left"), flexible_type("bottom_left")},
-      /*allowed_overwrite*/ false);
+      /* default_value     */ "top_left",
+      /* allowed_values    */ {flexible_type("top_left"), flexible_type("bottom_left")},
+      /* allowed_overwrite */ false);
   options.create_categorical_option(
-      /* name          */"annotation_position",
-      /* description   */
+      /* name              */"annotation_position",
+      /* description       */
       "Defines annotations position: center, top_left or bottom_left",
-      /* default_value */ "center",
-      /* allowed_values*/ {flexible_type("center"), flexible_type("top_left"), flexible_type("bottom_left")},
-      /*allowed_overwrite*/ false);
+      /* default_value     */ "center",
+      /* allowed_values    */ {flexible_type("center"), flexible_type("top_left"), flexible_type("bottom_left")},
+      /* allowed_overwrite */ false);
  
   // Validate user-provided options.
   options.set_options(opts);
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -251,7 +251,7 @@ void object_detector::init_options(
       /* allowed_values    */ {flexible_type("pixel"), flexible_type("normalized")},
       /* allowed_overwrite */ false);
   options.create_categorical_option(
-      /* name              */ "image_origin",
+      /* name              */ "annotation_origin",
       /* description       */
       "Defines image origin: top_left or bottom_left",
       /* default_value     */ "top_left",
@@ -264,7 +264,7 @@ void object_detector::init_options(
       /* default_value     */ "center",
       /* allowed_values    */ {flexible_type("center"), flexible_type("top_left"), flexible_type("bottom_left")},
       /* allowed_overwrite */ false);
- 
+
   // Validate user-provided options.
   options.set_options(opts);
 
@@ -715,23 +715,23 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
   iterator_params.class_labels = std::move(class_labels);
   iterator_params.repeat = repeat;
 
-  std::string image_origin = read_state<flex_string>("image_origin");
+  std::string annotation_origin = read_state<flex_string>("annotation_origin");
   std::string annotation_scale = read_state<flex_string>("annotation_scale");
   std::string annotation_position = read_state<flex_string>("annotation_position");
 
   // Setting input for Image Origin
-  if (image_origin == "top_left") {
-      iterator_params.image_origin = data_iterator::image_origin_enum::TOP_LEFT;
+  if (annotation_origin == "top_left") {
+      iterator_params.annotation_origin = data_iterator::annotation_origin_enum::TOP_LEFT;
   }
-  if (image_origin == "bottom_left") {
-      iterator_params.image_origin = data_iterator::image_origin_enum::BOTTOM_LEFT;
+  else if (annotation_origin == "bottom_left") {
+      iterator_params.annotation_origin = data_iterator::annotation_origin_enum::BOTTOM_LEFT;
   }
 
   // Setting input for Annotation Scale
   if (annotation_scale == "pixel") {
       iterator_params.annotation_scale = data_iterator::annotation_scale_enum::PIXEL;
   }
-  if (annotation_scale == "normalized") {
+  else if (annotation_scale == "normalized") {
       iterator_params.annotation_scale = data_iterator::annotation_scale_enum::NORMALIZED;
   }
 
@@ -739,10 +739,10 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
   if (annotation_position == "center") {
       iterator_params.annotation_position = data_iterator::annotation_position_enum::CENTER;
   }
-  if (annotation_position == "top_left") {
+  else if (annotation_position == "top_left") {
       iterator_params.annotation_position = data_iterator::annotation_position_enum::TOP_LEFT;
   }
-  if (annotation_position == "bottom_left") {
+  else if (annotation_position == "bottom_left") {
       iterator_params.annotation_position = data_iterator::annotation_position_enum::BOTTOM_LEFT;
   }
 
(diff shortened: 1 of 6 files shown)


test/unity/toolkits/object_detection/test_od_data_iterator.cxx
```diff
@@ -119,6 +123,146 @@ BOOST_AUTO_TEST_CASE(test_simple_data_iterator) {
...
+  // Case 1: Input: ('top_left', 'pixel', 'center') → ('top_left', 'normalized', 'top_left')
+  const labeled_image& case1_example = create_data_opts("top_left", "pixel", "center", 28, 18, 16, 16);
+  TS_ASSERT_EQUALS(case1_example.annotations[0].bounding_box,
+                      image_box(static_cast<float>(20) / IMAGE_WIDTH,
+                                 static_cast<float>(10) / IMAGE_HEIGHT,
+                                 16.f / IMAGE_WIDTH,
+                                 16.f / IMAGE_HEIGHT
+                                 ));
+
+  // Case 2: ('bottom_left', 'normalized', 'bottom_left') → ('top_left', 'normalized', 'top_left')
```
nickjong: Looks like this case is ‘pixel’ not ‘normalized’?

nickjong:
Note that to avoid the merge conflict identified by GitHub, I would recommend rebasing from master and omitting the commits from my original PR branch. So
```
git rebase -i master nick_od_seed_branch
```
And in the resulting editor, delete the two commits of mine. Then git will give you the result of applying the remaining commit (yours) to master.

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -397,7 +396,7 @@ BOOST_AUTO_TEST_CASE(test_object_detector_train) {
 
   // The following callbacks capture by reference so that they can transfer
   // ownership of the mocks created above.
-
+  /*
```
TobyRoseman: Should this code just be deleted rather than commenting it out? If it should just be commented out, please add a comment about why.

test/unity/toolkits/object_detection/test_od_data_iterator.cxx
```diff
@@ -119,6 +123,146 @@ BOOST_AUTO_TEST_CASE(test_simple_data_iterator) {
...
+    if (image_origin == "top_left") {
+        std::cout << "image : top_left" <<std::endl;
+        result.image_origin = data_iterator::image_origin_enum::TOP_LEFT;
+    }
+    else if (image_origin == "bottom_left") {
+        std::cout << "image : bottom_left" <<std::endl;
+        result.image_origin = data_iterator::image_origin_enum::BOTTOM_LEFT;
+    }
+    else {
+        // Add error logging
```
TobyRoseman: Do you want to do that as part of this PR? Or should this be a todo? Same question for the other two similar spots in this function.

test/unity/toolkits/object_detection/test_object_detector.cxx
fareeha: This is a work-in-progress PR. Not ready for code clean-up review yet. I wanted Nick's feedback on the logic.

test/unity/toolkits/object_detection/test_od_data_iterator.cxx
fareeha: Same as above

test/unity/toolkits/object_detection/test_object_detector.cxx
fareeha: When code is ready for a full review, I'll remove the 'WIP' tag.

TobyRoseman: Sounds good.

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -110,14 +113,67 @@ std::vector<image_annotation> parse_annotations(
     // instances we report will actually equal the number of image_annotation
     // values.
     if (has_label && has_x && has_y && annotation.bounding_box.area() > 0.f) {
+      
+      if (annotation_scale_ == data_iterator::annotation_scale_enum::NORMALIZED) {
+        image_height = 1;
```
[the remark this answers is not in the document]
fareeha: Yes it wasn't obvious to me either, until I broke Case 2. I'll add a comment.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -243,8 +243,26 @@ void object_detector::init_options(
...
+      /* default_value */ "pixel",
+      /*allowed_overwrite*/ true);
+  options.create_string_option(
+      /* name          */"image_origin",
+      /* description   */
+      "Defines image origin: top_left or bottom_left",
+      /* default_value */ "top_left",
+      /*allowed_overwrite*/ true);
+  options.create_string_option(
+      /* name          */"annotation_origin",
```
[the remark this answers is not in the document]
fareeha: That's a good idea! I wanted your opinion on the method and variable naming as well. Feel free to suggest any other name edits as well.

test/unity/toolkits/object_detection/test_od_data_iterator.cxx
fareeha: Yes, typo. Thank you!

src/toolkits/object_detection/object_detector.cpp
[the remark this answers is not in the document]
nickjong: FWIW, I like the names you had in the PR description: annotation_scale, annotation_origin, and annotation_position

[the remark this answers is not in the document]
fareeha: .. I just realized I used 'annotation_position' in the description 🙈

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -224,26 +224,47 @@ void object_detector::init_options(
...
-
+  options.create_categorical_option(
+      /* name              */ "annotation_scale",
+      /* description       */
+      "Defines annotations scale: pixel or normalized",
+      /* default_value     */ "pixel",
+      /* allowed_values    */ {flexible_type("pixel"), flexible_type("normalized")},
+      /* allowed_overwrite */ false);
+  options.create_categorical_option(
+      /* name              */ "image_origin",
```
nickjong: Could we make this "annotation_origin" instead?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -694,8 +715,45 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
...
-      new simple_data_iterator(iterator_params));
+  std::string image_origin = read_state<flex_string>("image_origin");
+  std::string annotation_scale = read_state<flex_string>("annotation_scale");
+  std::string annotation_position = read_state<flex_string>("annotation_position");
+
+  // Setting input for Image Origin
+  if (image_origin == "top_left") {
+      iterator_params.image_origin = data_iterator::image_origin_enum::TOP_LEFT;
+  }
+  if (image_origin == "bottom_left") {
```
nickjong: Nit: let's just use `else if` here, so
```
} else if (image_origin == "bottom_left") {
```

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -50,7 +53,10 @@ gl_sframe get_data(const data_iterator::parameters& params) {
 std::vector<image_annotation> parse_annotations(
     const flex_list& flex_annotations,
     size_t image_width, size_t image_height,
-    const std::unordered_map<std::string, int>& class_to_index_map) {
+    const std::unordered_map<std::string, int>& class_to_index_map,
+    image_origin_enum image_origin_,
```
nickjong: Nit: for function arguments (and other local variables), let's use names that do not end in underscore, so `annotation_scale` instead of `annotation_scale_`

The underscore should be a signal that the variable is a data member (instance variable)

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -110,14 +116,68 @@ std::vector<image_annotation> parse_annotations(
     // instances we report will actually equal the number of image_annotation
     // values.
     if (has_label && has_x && has_y && annotation.bounding_box.area() > 0.f) {
+      
+      // If the annotations are normalised, they will range between 0 and 1  
+      if (annotation_scale_ == annotation_scale_enum::NORMALIZED) {
+        image_height = 1;
```
nickjong: Upon further thought, I think this would be clearest if we defined new local variables, e.g.
```
float annotation_image_height;
float annotation_image_width;
switch (annotation_scale_) {
case annotation_scale_enum::PIXEL:
  annotation_image_height = static_cast<float>(image_height);
  annotation_image_width = static_cast<float>(image_width);
case annotation_scale_enum::NORMALIZED:
  annotation_image_height = 1.f;
  annotation_image_width = 1.f;
  break;
}
```

src/toolkits/object_detection/od_data_iterator.hpp
```diff
@@ -25,6 +25,26 @@ namespace object_detection {
  */
 class data_iterator {
 public:
+  
+  /* Enumerate possible values for image origin */
+  enum class image_origin_enum {
```
nickjong: Could we name this `annotation_origin_enum`?

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -212,7 +212,7 @@ class test_object_detector: public object_detector {
 public:
   using create_iterator_call =
       std::function<std::unique_ptr<data_iterator>(
-          gl_sframe data, std::vector<std::string> class_labels, bool repeat)>;
+        data_iterator::parameters iterator_params)>;
```
nickjong: Nit: could we avoid adding trailing whitespace, like the space after the semi-colon here? Might be useful to run git-clang-format

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -719,6 +713,9 @@ BOOST_AUTO_TEST_CASE(test_object_detector_auto_split) {
   // used as a normalizer.
   gl_sframe data({{"ignored", gl_sarray::from_sequence(0, test_num_examples)}});
 
+  data_iterator::parameters iterator_params; 
+  iterator_params.data = std::move(data);
```
nickjong: Don't use std::move here, since `data` is still used below.

nickjong: Wait, where is `iterator_params` used? Do we still need it?

src/toolkits/object_detection/object_detector.cpp
fareeha: `image_origin` denotes the origin of the image (whether it is placed top-left or bottom-left), not the origin for annotation. So I'm thinking it may be unclear? What do you think

test/unity/toolkits/object_detection/test_object_detector.cxx
fareeha: Good catch, we don't need this anymore

src/toolkits/object_detection/od_data_iterator.cpp
fareeha: They are data members of the `simple_data_iterator` class and I'm passing them in `parse_annotations`. Maybe I don't need to pass them to use them..

src/toolkits/object_detection/object_detector.cpp
nickjong: I think it’s more clear to give all three options a common prefix, since all three define (collectively) the semantics of the annotations. Here, the “annotation_origin” defines the origin of the coordinate system used to specify the annotations. I don’t think it should be confused with “annotation_position”. Note that any single annotation is defined by a position and a size (like CGRect); I don’t think of the corner of a bounding box as the box’s origin.

src/toolkits/object_detection/od_data_iterator.cpp
nickjong: Yes, the arguments you pass to `parse_annotations` when you call it may have underscores at the end, because the arguments are data members of `simple_data_iterator`, but the named arguments for the function itself are not data members. They are local variables that happen to be initialized (at the call site) by those data members.

Suppose I have a function `add_one` that adds one to its argument, and I use it to increment a counter (that is a data member):
```
  counter_ = add_one(counter_);
```

Does that mean that the name of the argument in the definition of `add_one` should be `counter_`? What makes more sense:
```
  int add_one(int x) { return x + 1; }
```
or
```
int add_one(int counter_) { return counter_ + 1; }
```
?

src/toolkits/object_detection/object_detector.cpp
fareeha: Hmmm, that's a good point. All three options collectively define the semantics of the annotations so it makes sense to give them the same prefix.

nickjong: FWIW, I could see "annotations_origin" being more clear than "annotation_origin". Then it's more obvious that "origin" is a property shared across all annotations. (Of course, then we would change all the prefixes to "annotations_".) Up to you if you think that's worth changing

src/toolkits/object_detection/od_data_iterator.cpp
fareeha: Ohhh - definitely the first one. Thank you for explaining this.

PR #2141, opened 2019-07-18
Option to disable post-evaluation metrics.

hoytak:
This PR disables the automatic computation of post-train evaluation metrics in the supervised learning toolkit, as these can be expensive at times.  All other features should still work fine, including manual evaluation calls.


src/python/turicreate/test/test_boosted_trees.py
```diff
@@ -56,7 +56,8 @@ def setUpClass(self):
                       'step_size': 1,
                       'min_loss_reduction': 1,
                       'max_iterations': 10,
-                      'min_child_weight': 1}
+                      'min_child_weight': 1,
+                      "default_evaluation_metric" : "none"}
```
znation: Won't this change the now-current default behavior? Is that intentional?

src/toolkits/supervised_learning/boosted_trees.cpp
```diff
@@ -82,6 +82,12 @@ void init_boosted_trees_common_options(option_manager &options){
       std::numeric_limits<int>::max(),
       false);
 
+  options.create_flexible_type_option(
+      "default_evaluation_metric",
```
srikris: Why is this called default_evaluation_metric? Can the user change this? Can we still compute multiple metrics?

PR #2142, opened 2019-07-18
Style Transfer Wrapper Class (Objective-C)

abhishekpratapa:


src/ml/neural_net/mps_layer_helper.h
```diff
@@ -11,20 +11,30 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class TCMPSConvolutionWeights;
+@class TCMPSInstanceNormDataLoader;
+
+@interface MPSCNNFullyConnectedNode(TCMPSLayerProperty)
+@property (nonatomic, retain) TCMPSConvolutionWeights *weights API_AVAILABLE(macosx(10.14));
```
nickjong: Nit: in ARC projects like ours, use `strong` instead of `retain`. Although I think they are functionally equivalent

src/ml/neural_net/mps_layer_helper.h
```diff
@@ -11,20 +11,30 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class TCMPSConvolutionWeights;
+@class TCMPSInstanceNormDataLoader;
+
+@interface MPSCNNFullyConnectedNode(TCMPSLayerProperty)
```
nickjong: Just curious, why have a separate `TCMPSLayerProperty` category, separate from `TCMPSLayerHelper`?

src/ml/neural_net/mps_layer_helper.mm
```diff
@@ -5,10 +5,9 @@
  */
 
 #include <ml/neural_net/mps_layer_helper.h>
-#include <ml/neural_net/mps_weight.h>
-
 #import <ml/neural_net/mps_layer_conv_padding.h>
 #import <ml/neural_net/mps_layer_instance_norm_data_loader.h>
+#include <ml/neural_net/mps_weight.h>
```
nickjong: This looks like this should be (always have been) a #import, because it's an ObjC header. Since we're touching this line anyway, let's make it right!

src/ml/neural_net/mps_layer_helper.h
nickjong: Huh, where is this property implemented? And where is it used?

src/ml/neural_net/style_transfer/mps_style_transfer.h
```diff
@@ -0,0 +1,34 @@
+/* Copyright © 2018 Apple Inc. All rights reserved.
```
nickjong: May as well use the current date in new headers

src/ml/neural_net/style_transfer/mps_style_transfer.h
```diff
@@ -0,0 +1,34 @@
...
+
+#import <ml/neural_net/mps_descriptor_utils.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+API_AVAILABLE(macos(10.15))
+@interface TCMPSStyleTransfer : NSObject
+
+@property (nonatomic) NSUInteger batchSize;
+@property (nonatomic) NSNumber *contentLossMultiplier;
```
nickjong: Any reason to prefer `NSNumber` over `float`?

src/ml/neural_net/style_transfer/mps_style_transfer.h
```diff
@@ -0,0 +1,34 @@
...
+@property (nonatomic) NSNumber *styleLossMultiplier;
+@property (nonatomic) BOOL finetuneAllParams;
```
nickjong: Nit: how about `updateAllParams`?

src/ml/neural_net/style_transfer/mps_style_transfer_decoding_node.mm
```diff
@@ -71,4 +74,34 @@ - (MPSNNImageNode *) backwardPass:(MPSNNImageNode *) inputNode {
...
+  [[_conv weights] setLearningRate:lr];
+  [[_instNorm weights] setLearningRate:lr];
+}
+
+- (NSDictionary<NSString *, NSData *> *)exportWeights {
+  NSMutableDictionary<NSString *, NSData *> *weights;
+
+  NSUInteger convWeightSize = (NSUInteger)([[_conv weights] weight_size] * sizeof(float));
+
+  NSMutableData* convDataWeight = [NSMutableData dataWithCapacity:convWeightSize];
```
nickjong: Use `dataWithLength:` to ensure the buffer is actually allocated

src/ml/neural_net/style_transfer/mps_style_transfer_encoding_node.mm
```diff
@@ -65,4 +68,34 @@ - (MPSNNImageNode *) backwardPass:(MPSNNImageNode *) inputNode {
   return [convGrad resultImage];
 }
 
+- (void)setLearningRate:(float)lr {
+  [[_conv weights] setLearningRate:lr];
+  [[_instNorm weights] setLearningRate:lr];
+}
+
+- (NSDictionary<NSString *, NSData *> *)exportWeights {
+  NSMutableDictionary<NSString *, NSData *> *weights;
```
nickjong: Need to initialize this variable (by allocating and initializing a `NSMutableDictionary`)

src/ml/neural_net/style_transfer/mps_style_transfer_pre_processing.h
```diff
@@ -0,0 +1,30 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+API_AVAILABLE(macos(10.14))
+@interface TCMPSStyleTransferPreProcessing : NSObject 
+
+@property (nonatomic) MPSNNImageNode *output;
+
+- (instancetype) initWithParameters:(NSString *)name
+                          inputNode:(MPSNNImageNode *)inputNode
+                         scaleNode:(MPSNNImageNode *)scaleNode
```
nickjong: Mega-nit: Align parentheses? Maybe just use git-clang-format? (Shreya has started doing this in some of her PRs...)

src/ml/neural_net/style_transfer/mps_style_transfer_transformer_network.mm
```diff
@@ -165,4 +168,119 @@ - (MPSNNImageNode *) backwardPass:(MPSNNImageNode *) inputNode {
...
+  [[_conv weights] setLearningRate:lr];
+  [[_instNorm weights] setLearningRate:lr];
+}
+
+- (NSDictionary<NSString *, NSData *> *)exportWeights {
+  NSMutableDictionary<NSString *, NSData *> *weights;
+
+  NSDictionary<NSString *, NSData *> * encode1Weights = [_encoding1 exportWeights];
+
+  weights[@"transformer_conv0_weight"] = encode1Weights[@"conv_weights"];
```
nickjong: Abhishek, meet `-[NSMutableDictionary addEntriesFromDictionary:`. `-[NSMutableDictionary addEntriesFromDictionary:`, meet Abhishek. 😄 

https://developer.apple.com/documentation/foundation/nsmutabledictionary/1411035-addentriesfromdictionary?language=objc

nickjong: Oh wait, the keys are changing. Is there some way to do this more systematically?

For example, maybe some of the `exportWeights` methods can take a "prefix" argument?

src/ml/neural_net/style_transfer/mps_style_transfer_utils.h
```diff
@@ -0,0 +1,24 @@
...
+
+#import <ml/neural_net/style_transfer/mps_style_transfer.h>
+#import <ml/neural_net/mps_descriptor_utils.h>
+
+@interface TCMPSStyleTransfer (TCMPSStyleTransferUtils)
++ (TCMPSTransformerDescriptor *)defineTransformerDescriptor:(NSUInteger)numStyles
+                                             tuneAllWeights:(BOOL)tuneAllWeights;
++ (TCMPSVgg16Descriptor *)defineVGG16Descriptor:(NSUInteger)numStyles;
+
++ (NSDictionary<NSString *, NSDictionary *> *)defineTransformerWeights:(NSDictionary<NSString *, NSData *> *)weights;
```
nickjong: Why do we need nested dictionaries for these? It would be helpful to document somewhere the structure of this value, since it's not obvious (to me)

src/ml/neural_net/style_transfer/mps_style_transfer_utils.h
```diff
@@ -0,0 +1,24 @@
...
++ (NSDictionary<NSString *, NSDictionary *> *)defineVGG16:(NSDictionary<NSString *, NSData *> *)weights;
++ (void)populateMean:(NSMutableData *)data;
++ (void)populateMultiplication:(NSMutableData *)data;
```
nickjong: How should someone reading this code understand what `populateMean` and `populateMultiplication` do?

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
```diff
@@ -0,0 +1,657 @@
...
+  self.conv2_2 = _nn.Conv2D(in_channels=128, channels=128, kernel_size=3, padding=1)
+
+  self.conv3_1 = _nn.Conv2D(in_channels=128, channels=256, kernel_size=3, padding=1)
+  self.conv3_2 = _nn.Conv2D(in_channels=256, channels=256, kernel_size=3, padding=1)
+  self.conv3_3 = _nn.Conv2D(in_channels=256, channels=256, kernel_size=3, padding=1)
+
+  self.conv4_1 = _nn.Conv2D(in_channels=256, channels=512, kernel_size=3, padding=1)
+  self.conv4_2 = _nn.Conv2D(in_channels=512, channels=512, kernel_size=3, padding=1)
+  self.conv4_3 = _nn.Conv2D(in_channels=512, channels=512, kernel_size=3, padding=1)
+*/
```
nickjong: What's this comment for?

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
```diff
@@ -0,0 +1,657 @@
...
+
++ (TCMPSVgg16Descriptor *) defineVGG16Descriptor:(NSUInteger)numStyles {
+  TCMPSVgg16Descriptor* vgg16Desc;
```
nickjong: Another case where you need to actually create an instance. ObjC pointers are like C and C++ pointers: by default they are just undefined values (invalid pointers)

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
```diff
@@ -0,0 +1,657 @@
...
+  vgg16Desc.block4.conv3.kernelWidth = 3;
+  vgg16Desc.block4.conv3.kernelHeight = 3;
+  vgg16Desc.block4.conv3.inputFeatureChannels = 256;
+  vgg16Desc.block4.conv3.outputFeatureChannels = 512;
+  vgg16Desc.block4.conv3.strideWidth = 1;
+  vgg16Desc.block4.conv3.strideHeight = 1;
+  vgg16Desc.block4.conv3.paddingWidth = 1;
+  vgg16Desc.block4.conv3.paddingHeight = 1;
+  vgg16Desc.block4.conv3.label = @"vgg16_block_4_conv_3";
+  vgg16Desc.block4.conv3.updateWeights = NO;
```
nickjong: I see what you mean with your offline comment about this being verbose. FWIW, in hindsight I think we could have done something more programmatic, looping over block index and conv index, if these were stored as arrays somehow instead of named components. See, for example, how the Python OD toolkits defines the tiny darknet model: https://github.com/apple/turicreate/blob/master/src/python/turicreate/toolkits/object_detector/_model.py

Not anything we need to change now, just something to consider for future reference

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
```diff
@@ -0,0 +1,657 @@
...
+  transformerWeightConv[@"weights"] = weights[@"transformer_conv5_weight"];
+  transformerWeightConv[@"biases"] = NULL;
+
+  transformerWeightInst[@"weights"] = weights[@"transformer_instancenorm5_gamma"];
+  transformerWeightInst[@"biases"] = weights[@"transformer_instancenorm5_beta"];
+
+  transformerWeight[@"transformer_conv"] = [transformerWeightConv copy];
+  transformerWeight[@"transformer_inst"] = [transformerWeightInst copy];
+
+  return [transformerWeight copy];
```
nickjong: I do think it might be worth reconsidering how we handle these dictionaries. Maybe I just need to understand the hierarchical structure, but I wonder if we can't handle everything via keys that have some hierarchical nomenclature (in the key names) instead

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
```diff
@@ -0,0 +1,657 @@
...
+  NSUInteger dataSize = (data.length)/sizeof(float);
+  NSAssert((dataSize) % 3 == 0, @"Data must follow a 3 channel format");
+  float meanWeights[3] = { 123.68, 116.779, 103.939 };
+  for(NSUInteger x = 0; x < (dataSize/3); x++){
+    for(NSUInteger c = 0;c < 3;c++){
+      NSUInteger offset = x * 3;
+      NSUInteger index = offset + c;
+      ((float *)data.mutableBytes)[index] = meanWeights[c];
+    }
+  }
```
nickjong: FWIW, consider this alternative implementation of this loop:
```
float* ptr = (float*) data.mutableBytes;
float* end = ptr + dataSize;
while (ptr < end) {
  ptr[0] = meanWeights[0];
  ptr[1] = meanWeights[1];
  ptr[2] = meanWeights[2];
  ptr += 3;
}
```

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+}
+
+- (NSDictionary<NSString *, NSData *> *)predict:(NSDictionary<NSString *, NSData *> *)inputs {
+  NSUInteger contentWidth;
+  NSUInteger contentHeight;
+  NSUInteger contentChannels;
+
+  [inputs[@"imageWidth"] getBytes:&contentWidth length:sizeof(contentWidth)];
+  [inputs[@"imageHeight"] getBytes:&contentHeight length:sizeof(contentHeight)];
+  [inputs[@"imageChannels"] getBytes:&contentChannels length:sizeof(contentChannels)];
```
nickjong: This seems pretty brittle. Does this mean that some of the `NSData` values store arrays of floats, and others are just `NSUInteger`? This is especially dangerous because on modern Macs, `float` is 32-bit and `NSUInteger` is 64-bit.  It seems safer either to pass this kind of thing around outside of these dictionaries of float arrays, or at least to encode them as floating point values.

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+
+  MPSImageDescriptor *imgDesc = [MPSImageDescriptor
+    imageDescriptorWithChannelFormat:MPSImageFeatureChannelFormatFloat32
+                               width:contentWidth
+                              height:contentHeight
+                     featureChannels:contentChannels
+                      numberOfImages:1
+                               usage:MTLTextureUsageShaderWrite | MTLTextureUsageShaderRead];
+
+  NSMutableArray<MPSImage *> *contentImageArray;
```
nickjong: Another case where you should instantiate an object.

In Obj-C code you should basically never have code that looks like
```
NSObject *foo;
```
unless it's part of something like
```
NSObject *foo;
if (...) {
  foo = ...;
} else {
  foo = ...;
}
```

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+    MPSImage *contentImage = [[MPSImage alloc] initWithDevice:_dev imageDescriptor:imgDesc];
+    [contentImage writeBytes:inputs[key].bytes dataLayout:(MPSDataLayoutHeightxWidthxFeatureChannels)imageIndex:0];
+    [contentImageArray addObject:contentImage];
+  }
+
+  MPSImageBatch *contentImageBatch = [contentImageArray copy];  
+
+  id<MTLCommandBuffer> cb = [_commandQueue commandBuffer];
+
+  NSMutableArray *intermediateImages = [[NSMutableArray alloc] initWithArray:@[]];
```
nickjong: The RHS here can just be `[[NSMutableArray alloc] init]`

nickjong: (Some people prefer `[NSMutableArray new]` which in theory is equivalent and simpler, but I'm personally just not used to it)

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+                                                       sourceStates:nil
+                                                 intermediateImages:intermediateImages
+                                                  destinationStates:destinationStates];
+
+  for (MPSImage *image in stylizedImages) {
+    [image synchronizeOnCommandBuffer:cb];  
+  }
+
+  [cb commit];
+  [cb waitUntilCompleted];
```
nickjong: This is okay for now, but in the long term we'll want to allow this to be asynchronous, so that we can start preparing the next mini-batch on CPU while this one executes on GPU. See for example the code in `mps_graph_cnnmodule::predict`

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+    NSMutableData* styleData = [NSMutableData dataWithLength:(NSUInteger)sizeof(float) * contentWidth * contentHeight * contentChannels];
+    [image readBytes:styleData.mutableBytes dataLayout:(MPSDataLayoutHeightxWidthxFeatureChannels)imageIndex:0];
+    NSString* key = [NSString stringWithFormat:@"%@%lu", @"stylizedImage", [stylizedImages indexOfObject:image]];
+    imagesOut[key] = styleData;
+  }
+
+  return [imagesOut copy];
+}
+
+- (void) setLearningRate:(NSNumber *)lr {
```
nickjong: Why `NSNumber` here?

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+    [image synchronizeOnCommandBuffer:cb];  
+  }
+
+  [cb commit];
+  [cb waitUntilCompleted];
+
+  // AAAK: What about overflow?
+  float lossValue = 0.0f;
+  MPSImageBatch *outBatch = [intermediateImages objectAtIndex:0];
+  for (NSUInteger index = 0; index < _batchSize; index++) {
```
nickjong: Can just use
```
for (MPSImage *lossImage in outBatch) {
```
I think

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+    float loss;
+    MPSImage* lossImage = outBatch[index];
+    [lossImage readBytes:&loss dataLayout:(MPSDataLayoutHeightxWidthxFeatureChannels)imageIndex:0];
+    lossValue += loss;
+  }
+  lossValue /= _batchSize;
+
+  NSMutableData * lossData = [NSMutableData dataWithCapacity:0];
```
nickjong: Here too use `dataWithLength:` instead of `dataWithCapacity`.

The length should be `sizeof(float)`, right?

Might be worth factoring out these NSData operations into a helper function or category method.

src/ml/neural_net/mps_layer_helper.h
abhishekpratapa: The helper category was to help create an `MPSCNNFullyConnectedNode` from the helper method. The `TCMPSLayerProperty` adds a property to the interface `MPSCNNFullyConnectedNode`. Functionally I though of them as two different things.

src/ml/neural_net/mps_layer_helper.h
abhishekpratapa: It's currently not used anywhere. It'll be used in the future implementations of `Drawing Classifier` and `Sound Classifier`. Since we'd already completed the helper methods for Convolution and Instance Normalization, it only felt apt to add the property to the Fully Connected helper method.

src/ml/neural_net/style_transfer/mps_style_transfer.h
abhishekpratapa: `finetuneAllParams` is `updateAllParams `. It's the name of the property we use  in Python.

Screen Shot 2019-07-25 at 12 54 28 PM

src/ml/neural_net/style_transfer/mps_style_transfer.h
abhishekpratapa: Not particularly. Seemed more Objective-C like.

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa:
```
NSMutableData * lossData = [NSMutableData dataWithLength:0];
[lossData appendBytes:&lossValue length:sizeof(float)];
```

In this instance I'm appending the bytes so the length being 0 is fine right?

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: No good, reason! Just thought it was more Objective-C like. I'll change it to float.

src/ml/neural_net/mps_layer_helper.h
abhishekpratapa: Fixed to use strong instead of retain

src/ml/neural_net/style_transfer/mps_style_transfer.h
abhishekpratapa: Done for all my files!

src/ml/neural_net/style_transfer/mps_style_transfer_decoding_node.mm
```diff
@@ -7,6 +7,9 @@
 #import <ml/neural_net/style_transfer/mps_style_transfer_decoding_node.h>
 #import <ml/neural_net/mps_layer_helper.h>
 
+#include <ml/neural_net/mps_weight.h>
```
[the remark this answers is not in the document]
abhishekpratapa: Done!

src/ml/neural_net/style_transfer/mps_style_transfer_decoding_node.mm
abhishekpratapa: Proper. Changed the method.

src/ml/neural_net/style_transfer/mps_style_transfer_encoding_node.mm
abhishekpratapa: Done! Allocated NSMutableDictionary before accessing it!

src/ml/neural_net/style_transfer/mps_style_transfer_pre_processing.h
abhishekpratapa: Perhaps, I refrain from using `clang-format` on Objective-c code. It just messes things up sometimes. I aligned the parentheses in all of the files I found mis-aligned!

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
abhishekpratapa: It's a left-over remnant. I removed it!

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
abhishekpratapa: Got it! Made sure to allocate instances to pointer I had defined!

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: Consistently allocated and initiated NS* stuff.

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: Good point! Changed!

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: Fixed this!

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: Refactored this out to properties.

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
abhishekpratapa: Fixed!

src/ml/neural_net/mps_layer_helper.h
nickjong: IIRC, properties in categories cannot be auto-synthesized, since the category cannot modify the in-memory layout of the object. So defining the property here without providing implementations (that don’t rely on introducing new state) is just creating an interface that will yield an error if anybody tries to use it. We should either add an implementation or remove the interface until we need it.

(BTW, I should have mentioned that with ARC, `strong` is the default anyway...)

src/ml/neural_net/style_transfer/mps_style_transfer.h
nickjong: Eh, I prefer `updateAllParams` in a vacuum, but I see the value in preserving consistency with the Python code, while it still exists...

src/ml/neural_net/style_transfer/mps_style_transfer.h
nickjong: Most ObjC developers will avoid `NSNumber` when possible. It only exists when object semantics are required, e.g. when creating an `NSArray` of numbers.

src/ml/neural_net/style_transfer/mps_style_transfer.m
nickjong: Ah, I believe you’re right. Probably simplest to use either `[NSMutableData data]` or `[NSMutableData dataWithCapacity:sizeof(float)]`, though

src/ml/neural_net/style_transfer/mps_style_transfer.h
abhishekpratapa: Sounds good, switching to float.

src/ml/neural_net/style_transfer/mps_style_transfer.h
abhishekpratapa: Switched to update all params. The python side can adapt later.

src/ml/neural_net/mps_layer_helper.h
abhishekpratapa: Hopefully, I kinda fixed it with `@dynamic`?

src/ml/neural_net/mps_layer_helper.h
abhishekpratapa: Removed the separate category, refactored into one category.

src/ml/neural_net/style_transfer/mps_style_transfer_utils.h
abhishekpratapa: Added helper descriptions.

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -0,0 +1,419 @@
...
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#import <ml/neural_net/mps_device_manager.h>
+#import <ml/neural_net/mps_layer_helper.h>
+#import <ml/neural_net/mps_node_handle.h>
+
+#import <ml/neural_net/style_transfer/mps_style_transfer.h>
```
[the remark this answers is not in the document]
abhishekpratapa: Corrected.

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: Added a todo for a future PR.

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: fixed

src/ml/neural_net/style_transfer/mps_style_transfer_transformer_network.mm
abhishekpratapa: Perhaps, I wanted the output keys to match the output keys in the MxNet model. That's why the keys were changing.

src/ml/neural_net/style_transfer/mps_style_transfer_decoding_node.mm
```diff
@@ -71,4 +74,34 @@ - (MPSNNImageNode *) backwardPass:(MPSNNImageNode *) inputNode {
...
 
+- (void)setLearningRate:(float)lr {
+  [[_conv weights] setLearningRate:lr];
+  [[_instNorm weights] setLearningRate:lr];
+}
+
+- (NSDictionary<NSString *, NSData *> *)exportWeights {
+  NSMutableDictionary<NSString *, NSData *> *weights;
+
+  NSUInteger convWeightSize = (NSUInteger)([[_conv weights] weight_size] * sizeof(float));
```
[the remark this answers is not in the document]
abhishekpratapa: Fixed!

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
abhishekpratapa: Alright, I'll keep that in mind when doing future implementations in MPS!

src/ml/neural_net/style_transfer/mps_style_transfer_utils.m
abhishekpratapa: Converted to flat dictionary

src/ml/neural_net/style_transfer/mps_style_transfer_utils.h
abhishekpratapa: Converted to flat dictionary

src/ml/neural_net/style_transfer/mps_style_transfer_transformer_network.mm
abhishekpratapa: Converted to using the MPS model as the source of truth.

src/ml/neural_net/mps_layer_helper.mm
```diff
@@ -44,12 +45,15 @@ + (MPSCNNFullyConnectedNode *) createFullyConnected:(MPSNNImageNode *)inputNode
...
                                      weights:fullyConnectedDataLoad];
+
+  fullyConnectedNode.weights = fullyConnectedDataLoad;
   
   return fullyConnectedNode;
 }
 @end
 
 @implementation MPSCNNConvolutionNode (TCMPSLayerHelper)
+@dynamic weights;
```
nickjong: `@dynamic` doesn't create the getter and setter, it just promises that one will be available at runtime. Where are these defined?

If not defined or used, please just remove the property for now, until you actually implement it.

abhishekpratapa: These are set in the helper methods `createConvolutional `, and accessed in the exportWeights method.

nickjong: But you haven't provided an implementation, so this will in fact be a run-time error

PR #2143, opened 2019-07-18
Move warning flag to hit both clang and g++.

hoytak:


TobyRoseman:
Based on internal tests this doesn't seem to solve the problem. I think we should back out #2130 until we have a solution.

hoytak:
@TobyRoseman -- I can't compile on my end without https://github.com/apple/turicreate/pull/2130, so I don't think that's the right solution.  

Did you do a `configure --cleanup`?

TobyRoseman:
@hoytak - yes.

PR #2145, opened 2019-07-18
adding src location info in exception messages to accelerate debugging

Jarvi-Izana:
In an ideal world, the exception message should be adequate for us to locate where the exception is thrown. But many times the error reports from those exceptions are too opaque, such as "file not found".

In the current logger, `filename`, `function name` and `line number` are actually included in the logging body. Initially, I thought we don't have those in our logger. Now my work simplifies by adding those into exception message for the debug mode.

I find it's inconvenient to check the logging file all the time to find where the exception is thrown. Now if we have `filename`, `function name` and `line number` in the exception message, it's straightforward for us to look into the issue and search solutions.

Also, it's easy for users to report issues since most of them may not know where to find the log files to copy & paste error logs. Often the time, the exception message body tells you very little useful info.

### DIFF
--- a/src/core/logging/logger.hpp
+++ b/src/core/logging/logger.hpp
@@ -335,7 +335,7 @@
   do {                                                \
     auto throw_error = [&]() GL_COLD_NOINLINE_ERROR { \
       logstream(LOG_ERROR) << (message) << std::endl; \
-      throw(std::runtime_error(message);              \
+      throw(std::runtime_error(message));             \
     };                                                \
     throw_error();                                    \
   } while (0)
--- a/src/core/logging/logger.hpp
+++ b/src/core/logging/logger.hpp
@@ -265,56 +265,61 @@
 
 #else
 
-#define logger(lvl,fmt,...)                 \
-    (log_dispatch<(lvl >= OUTPUTLEVEL)>::exec(lvl,__FILE__, __func__ ,__LINE__,fmt,##__VA_ARGS__))
+#define logger(lvl, fmt, ...)                                                  \
+  (log_dispatch<(lvl >= OUTPUTLEVEL)>::exec(lvl, __FILE__, __func__, __LINE__, \
+                                            fmt, ##__VA_ARGS__))
 
+#define logbuf(lvl, buf, len)                                                  \
+  (log_dispatch<(lvl >= OUTPUTLEVEL)>::exec(lvl, __FILE__, __func__, __LINE__, \
+                                            buf, len))
 
-#define logbuf(lvl,buf,len)                 \
-    (log_dispatch<(lvl >= OUTPUTLEVEL)>::exec(lvl,__FILE__,     \
-                        __func__ ,__LINE__,buf,len))
+#define logstream(lvl)                                                      \
+  if (lvl >= global_logger().get_log_level())                               \
+  (log_stream_dispatch<(lvl >= OUTPUTLEVEL)>::exec(lvl, __FILE__, __func__, \
+                                                   __LINE__))
 
-#define logstream(lvl)                      \
-    if(lvl >= global_logger().get_log_level()) (log_stream_dispatch<(lvl >= OUTPUTLEVEL)>::exec(lvl,__FILE__, __func__ ,__LINE__) )
+#define logger_once(lvl, fmt, ...)                                 \
+  {                                                                \
+    static bool __printed__ = false;                               \
+    if (!__printed__) {                                            \
+      __printed__ = true;                                          \
+      (log_dispatch<(lvl >= OUTPUTLEVEL)>::exec(                   \
+          lvl, __FILE__, __func__, __LINE__, fmt, ##__VA_ARGS__)); \
+    }                                                              \
+  }
 
-#define logger_once(lvl,fmt,...)                 \
-{    \
-  static bool __printed__ = false;    \
-  if (!__printed__) {                 \
-    __printed__ = true;               \
-    (log_dispatch<(lvl >= OUTPUTLEVEL)>::exec(lvl,__FILE__, __func__ ,__LINE__,fmt,##__VA_ARGS__)); \
-  }  \
-}
+#define logstream_once(lvl)                                     \
+  (*({                                                          \
+    static bool __printed__ = false;                            \
+    bool __prev_printed__ = __printed__;                        \
+    if (!__printed__) __printed__ = true;                       \
+    &(log_stream_dispatch<(lvl >= OUTPUTLEVEL)>::exec(          \
+        lvl, __FILE__, __func__, __LINE__, !__prev_printed__)); \
+  }))
 
-#define logstream_once(lvl)                      \
-(*({    \
-  static bool __printed__ = false;    \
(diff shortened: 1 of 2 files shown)


src/core/logging/logger.hpp
```diff
@@ -324,14 +329,16 @@
...
-    };                                                              \
-    throw_error();                                                  \
-  } while(0)
+#ifdef NDEBUG
+
+#define log_and_throw(message)                        \
+  do {                                                \
+    auto throw_error = [&]() GL_COLD_NOINLINE_ERROR { \
+      logstream(LOG_ERROR) << (message) << std::endl; \
+      throw(std::runtime_error(message);              \
```
znation: Will this change how errors are handled in Python or through the C API `tc_error *` mechanism?

Jarvi-Izana: I forget to mention one more important point: 

Change the default `log_and_throw` behavior. Previous behavior is to `throw(std::string(message))`, which is really hard to catch since it's not an `std::exception` type and you need to specify error handling logic like `catch(const string& ex)`. That's why there're lots of `catch(...)` around the codebase, which is a bad practice since you miss the chance to interpret the exception message or propagate the original exception message. After we set `runtime_error` as default exception, we can use `catch (const std::exception& ex)` to avoid `catch(...)`.

Meanwhile, we also need to restrict the exception categories.  Unlimited different types of expcetions will make you lazy and tempt you to use `catch (const std::exception& ex)` as a brutal solution. Even though it can help, but we should have provided error handling for certain recoverable exceptions. 

Therefore, I think we need to use only few common exception types like `runtime`, `io` and `engine_optimizer` specific exception. Those types of exceptions should be the only types of exceptions we expect. In this way, we can expect a fairly hierarchical  error-handling or logging to react to specific errors.
```cpp
try {
} catch (const std::runtime_error & ex) {
} catch (const std::turi::fileio::io_error& ex) {
} catch (const std::exception& ex) {
} catch (...) {
// unknown which should not happen
}
```

Jarvi-Izana: Let me check how error is handled in C API.

hoytak: This should be fine.

hoytak: @ylow -- do you think this is okay to do in general?

Jarvi-Izana: or with a more general io_failiure exception handling.

Jarvi-Izana:
@znation When I run tests, I found this error which I can do a little to address:
```
test/vega_renderer/base_fixture.mm:129: error: in "vega_lite_example_tests/testSimpleBarChart": check false has failed
```
Do you know if this is affected by my change of switch throwing `std::string` exception to `std::runtime_error`?

Jarvi-Izana:
@hoytak @znation @ylow I just checked that the work is non-trivial because there're around 20 places where string type exception is handled. If we can make a call that we want to adopt runtime_error to replace the general string typed exception, then I can start working on it.

src/core/logging/logger.hpp
```diff
@@ -343,42 +350,96 @@
...
 
-#define log_func_entry()                                       \
-  do {                                                         \
-    logstream(LOG_INFO) << "Function entry" << std::endl;      \
-  } while(0)
+#define log_and_throw(message)                                          \
+  do {                                                                  \
+    auto throw_error = [&]() GL_COLD_NOINLINE_ERROR {                   \
+      std::stringstream _turi_ss;                                       \
+      _turi_ss << (message) << ". " << __func__ << " from " << __FILE__ \
```
znation: Just out of curiosity, why do we only want to include filename and line numbers in debug builds, and not in release too? This message is only ever going to a debug log anyway.

src/core/logging/logger.hpp
```diff
@@ -324,14 +329,16 @@
...
-    };                                                              \
-    throw_error();                                                  \
-  } while(0)
+#ifdef NDEBUG
+
+#define log_and_throw(message)                        \
+  do {                                                \
+    auto throw_error = [&]() GL_COLD_NOINLINE_ERROR { \
+      logstream(LOG_ERROR) << (message) << std::endl; \
+      throw(std::runtime_error(message));             \
```
znation: I would leave this as `std::string` for now, as per your comment re: ~20 places that would need to be updated. We could take that change as a separate PR in the future (changing it to runtime_error here, and updating all 20 places in the same commit), but for now I would just open a GitHub issue to track it (if there is not one already), and get this PR in first.

src/core/logging/logger.hpp
Jarvi-Izana: I think we mainly work on debug mode and it's reasonable to include those detailed bug info. But for the end-users, we don't want too much irrelevant info to distract them. I personally prefer to include in release mode so that the end-user can simply copy and paste everything to report.

src/core/logging/logger.hpp
Jarvi-Izana: we have a bunch of places catching by `const char*` and `const std::string&`, which is kind of confusing to me at first. That's probably also why there are lots of `catch (...)` which doesn't really help and may get things worse because we are tend to be lazy without trying to exhaust handling all possible exception types. 

20 places are not that many. I think they are side effects of this PR and we should include those in this PR so that I can run all unit tests and make sure I don't break anything.

src/core/logging/logger.hpp
znation: Regular users don't see the debug log at all, do they? I thought it just goes to disk.

src/core/logging/logger.hpp
znation: It would be great to do that as a separate PR, since the two work items are essentially unrelated except that they happen to touch some of the same parts of the code:

1. Get filename and line number logging on log_and_throw
2. Have log_and_throw use std::runtime_error as its exception type

These should be two separate PRs, since it's easier to review and comment on them separately, and there are various situations where this makes sense in general:

* If one PR should go in, and the other shouldn't, it's harder to separate them after the fact than to keep them separate to begin with.
* If one PR needs feedback and changes, and the other is ready to go in, they aren't both delayed.
* Reviewers who are best suited to review and discuss one of them may not be the best reviewers for the other.

src/core/logging/logger.hpp
znation: I think it's ok to do in general, but I would prefer to see a separate PR for it.

src/core/logging/logger.hpp
Jarvi-Izana: That's valuable advice. I can separate them now.

Jarvi-Izana:
@znation this commit only requires including src info, I think we are good to go.

src/core/logging/logger.hpp
Jarvi-Izana: The log and exception are different. I doubt they would like to read the log since it's very troublesome to do so (you have to find log file location and then copy-paste the corresponding logs). Most cases I've seen so far are people just copy-paste what's reported in the interactive shell.

if we include src info in exception, we can debug fairly faster within the shell instead of tailing the log.

PR #2149, opened 2019-07-19
Fix for dylib compilation issue.

hoytak:
It seems the dylib building process does not actually pick up flags set by add_compile_definitions; they should set the variables we track for these things instead.


hoytak:
This isn't the issue.  Closing til it's resolved in a reasonable way.

PR #2151, opened 2019-07-19
Change VegaRenderer to expose canvas through DOM API

Saba9:
Expose canvas to Vega through the DOM API by partially partially implementing the `document` object.

### DIFF
--- a/src/visualization/vega_renderer/JSCanvas.m
+++ b/src/visualization/vega_renderer/JSCanvas.m
@@ -623,14 +623,6 @@
 }
 
 - (void)fill {
-    NSArray *args = [JSContext currentArguments];
-    (void)args;
-    assert(args != nil);
-    if(args.count > 0) {
-        // Vega occasionally makes calls to fill with four arguments when zero are expected.
-        // Might have intended to call fillRect instead.
-        NSLog(@"warning: 'fill' called with %lu arguments when none were expected", args.count);
-    }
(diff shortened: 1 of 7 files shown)


src/visualization/vega_renderer/JSDocument.m
```diff
@@ -0,0 +1,23 @@
...
+    return self;
+}
+
+- (NSObject<VegaHTMLElement>*)createElementWithString:(NSString*)element {
+    if([element isEqualToString:@"canvas"]){
+        return _canvas;
+    } else {
+        NSLog(@"warning: creating elements of type '%@' is not supported", element);
+    }
+    return nil;
```
Saba9: It might be better to throw rather than returning `nil`. I'm open to either option.

znation: I think we could `assert(false)` here.

src/visualization/vega_renderer/JSCanvas.m
```diff
@@ -627,7 +626,11 @@ - (void)fill {
     NSArray *args = [JSContext currentArguments];
     (void)args;
     assert(args != nil);
-    assert(args.count == 0);
+    if(args.count > 0) {
```
znation: I think we can just remove this entirely at some point (as discussed offline, doesn't need to be addressed in this PR).

PR #2152, opened 2019-07-19
Upgrade to Vega 5.4.0

### LINKED ISSUE #2119: Reconcile Vega.js versions and upgrade to Vega 5
The current [latest release](https://github.com/vega/vega/releases) of Vega is 5.4.0.

Currently in Turi Create, we are using the following versions:

* 3.2.1 in Turi Create Python (web based rendering).
* 4.4.0 in Vega Renderer (macOS native/C API rendering).

They give slightly different behavior, and 5.x has features we'd like to adopt. We should upgrade.

znation:
* Upgrades Vega to 5.4.0 and Vega-Lite to 3.3.0 consistently across vega_renderer (JSC), tcviz (WebKit/CEF), and node runtime environments.
* Allows for non-minified JS dependencies in debug builds and minified in release, automatically.

Fixes #2119


src/visualization/vega_renderer/JSCanvas.m
```diff
@@ -278,8 +278,48 @@ - (void)setFont:(NSString *)fontStr {
                 if ([fontProperties.fontWeight isEqualToString:@"bold"]) {
                     newFont = [fontManager convertFont:newFont toHaveTrait:NSBoldFontMask];
                     assert(newFont != nil);
+                } else if (fontProperties.fontWeight.length == 3 &&
+                           [[fontProperties.fontWeight substringFromIndex:1] isEqualToString:@"00"]) {
+                    NSString *weightString = [fontProperties.fontWeight substringToIndex:1];
+                    NSInteger weightInt = weightString.intValue;
+                    NSFontWeight weight = NSFontWeightRegular;
+                    switch (weightInt) {
+                        case 1:
```
kanitw: Why does this PR need to contain code for Font weight?

In any case, I'm curious where this `weightInt` is coming from.  Note that Vega/Vega-Lite use CSS font-weight, which would be `100`, `200`, ..., `900`.

znation: With the existing specs in the unit tests, Vega 5.4.0 now sets font weight to some numeric weights (I saw "600" in the debugger) where it used to use "bold", I think. Either one should work, so I took the opportunity to implement here.

As for "1" vs "100", the code a few lines above already pattern matched the string for `x00` where `x` is an integer, so we only need to look at the `x` part. We are expecting font weights in the form like "100" or "600".

PR #2156, opened 2019-07-22
Add compute_context factory and registration mechanism to clean up platform-specific dependencies

nickjong:


src/ml/neural_net/compute_context.cpp
```diff
@@ -6,21 +6,51 @@
...
+                      needle);
+  if (it != get_registry().end()) {
+    get_registry().erase(it);
+  }
+}
+
 // static
 std::unique_ptr<compute_context> compute_context::create() {
-
-#if defined(HAS_MPS) && !defined(TC_BUILD_IOS)
```
nickjong: Note that this file no longer requires any conditional compilation at all: it returns `nullptr` by default, but first checks to see if any of the injected callbacks returns a non-NULL value

src/ml/neural_net/mps_compute_context.mm
```diff
@@ -16,6 +16,19 @@
...
 
+namespace {
+
+std::unique_ptr<compute_context> create_mps_compute_context() {
+  return std::unique_ptr<compute_context>(new mps_compute_context);
+}
+
+// At static-init time, register create_mps_compute_context().
+// TODO: Codify priority levels?
+static auto* mps_registration = new compute_context::registration(
```
nickjong: Note that with this chance, `unity_neural_net` no longer even links against `tcmps`... The only reason there's still a dependency in the CMakeLists is as a convenient way to force `tcmps` to be built and linked in the final binary

For a Python-based library, we can just have that library sit on top (not be a dependency of unity_shared), and just inject itself at run time in the same way as this code does

PR #2159, opened 2019-07-23
Implemented save and load for OD C-API

fareeha:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -539,6 +554,132 @@ variant_map_type object_detector::perform_evaluation(gl_sframe data,
...
+  // Append conv8.
+  static constexpr float CONV8_MAGNITUDE = 0.00005f;
+  const size_t num_classes = read_state<size_t>("num_classes");
+  const size_t num_predictions = 5 + num_classes;  // Per anchor box
+  const size_t conv8_c_out = anchor_boxes().size() * num_predictions;
+  auto conv8_weight_init_fn = [&random_engine](float* w, float* w_end) {
+    std::uniform_real_distribution<float> dist(-CONV8_MAGNITUDE,
+                                               CONV8_MAGNITUDE);
+    while (w != w_end) {
+      *w++ = dist(random_engine);
```
TobyRoseman: I'm a big believer that if you're ever relying on the difference between pre and post incrementing, you're trying to do too much on one line.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -539,6 +554,132 @@ variant_map_type object_detector::perform_evaluation(gl_sframe data,
...
+                         /* epsilon */               0.00001f);
+
+  // Append leakyrelu7.
+  nn_spec->add_leakyrelu(/* name */                  "leakyrelu0_fwd",
+                         /* input */                 "batchnorm0_fwd",
+                         /* alpha */                 0.1f);
+
+  num_features = 16;
+  // Initilize Layer 1 to Layer 6
+  const std::map<int, int> idx_channels = {{1, 32}, {2,64}, {3, 128}, {4, 256}, {5,512}, {6, 1024}};
```
TobyRoseman: Would an array make more sense here? 

Maybe something like:
`const int layer_idx_to_channels[] = {NULL, 32, 64, 128, 256, 512, 1024};`

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -539,6 +554,132 @@ variant_map_type object_detector::perform_evaluation(gl_sframe data,
...
+
+  std::unique_ptr<model_spec> nn_spec(new model_spec);
+
+  // Initialize a random number generator for weight initialization.
+  std::seed_seq seed_seq = { read_state<int>("random_seed") };
+  std::mt19937 random_engine(seed_seq);
+
+  // Initilize Layer 0
+  int num_features = 3;
+  xavier_weight_initializer conv0_init_fn(16*3*3, 16*3*3, &random_engine);
```
TobyRoseman: Should some of the `3`s on the next several lines be replaced with `num_features`?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -539,6 +554,132 @@ variant_map_type object_detector::perform_evaluation(gl_sframe data,
...
+
+  // Append leakyrelu7.
+  nn_spec->add_leakyrelu(/* name */                  "leakyrelu0_fwd",
+                         /* input */                 "batchnorm0_fwd",
+                         /* alpha */                 0.1f);
+
+  num_features = 16;
+  // Initilize Layer 1 to Layer 6
+  const std::map<int, int> idx_channels = {{1, 32}, {2,64}, {3, 128}, {4, 256}, {5,512}, {6, 1024}};
+  for (auto const& x : idx_channels)
```
TobyRoseman: Maybe `auto const&` -> `const int&`

src/toolkits/object_detection/object_detector.cpp
fareeha: 'num_features' refers to kernel channels here, not image channels so I didn't want to mix them up by replacing the '3's in num of image channels.

src/toolkits/object_detection/object_detector.cpp
fareeha: I'm mapping layer number to num of channels. I feel map makes it more readable. Let me know if you feel this isn't the right way to go about it.

fareeha:
TODO: Add test

src/toolkits/object_detection/od_serialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+void init_darknet_yolo(neural_net::model_spec& nn_spec, const size_t num_classes,
+	const std::vector<std::pair<float, float>>& anchor_boxes) {
+
+  // Initialize a random number generator for weight initialization.
+  std::seed_seq seed_seq = { rand() };
+  std::mt19937 random_engine(seed_seq); //TODO: make zero init
+
+  int num_features = 3;
```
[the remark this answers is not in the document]
fareeha: Because it changes in each iteration based on the output size of the previous layer.

PR #2160, opened 2019-07-24
Parallelized data synthesis and also potentially fixed increasing RAM

### LINKED ISSUE #2023: One-Shot Object Detector data synthesis should use lambda apply
We’re currently iterating through all the images in a single-threaded loop. This should be parallelized.

### LINKED ISSUE #2037: one shot object detector crashes colab session during annotation
the 12 GB ram in the colab session is not enough as it seems
Can you fix this?

### LINKED ISSUE #2038: one shot object detector seems unusable on an iMac with 24GB!
training starts after 30min without the time measured for downloading background data.
Annotation uses all the ram and afterwards, my iMac is swapping for about 30 minutes before training starts.
Training time seems to be like 8 hours for my 12 test classes - which is in the range of expectations.
-- hmm maybe not - cpu and gnu usage is nearly non existing during the training - while hd io is high.

update: training speed is nearly non existing due to heavy hd io.....

Maybe you should not test turicreate only on iMac Pros with 64 GB of ram.

### LINKED ISSUE #2045: One shot detector dose not work on Ubuntu
I am trying to use `tc.one_shot_object_detector.create()` function of Turi Create on Ubuntu on AWS EC2. But I always get some error while Aurgumenting input images phase like this:

```
ubuntu@ip-172-31-7-98:~$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="18.04.2 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.2 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic

>> model = tc.one_shot_object_detector.create(imgs, 'label')
...
Downloading https://docs-assets.developer.apple.com/turicreate/data/one_shot_backgrounds.sarray.tar
Download completed: /var/tmp/data_cache/one_shot_backgrounds.sarray.tar
Augmenting input images using 951 background images.
+------------------+--------------+------------------+
| Images Augmented | Elapsed Time | Percent Complete |
+------------------+--------------+------------------+
| 0                | 5.89s        | 0%               |
| 1                | 5.89s        | 0%               |
| 2                | 5.90s        | 0%               |
| 3                | 5.91s        | 0%               |
| 4                | 5.93s        | 0%               |
| 5                | 5.95s        | 0%               |
| 10               | 6.06s        | 0%               |
| 50               | 7.19s        | 0.25%            |
| 100              | 8.17s        | 0.75%            |
| 500              | 17.99s       | 4%               |
| 1000             | 34.73s       | 8%               |
| 1500             | 46.22s       | 12%              |
| 2000             | 1m 2s        | 16%              |
| 2500             | 1m 13s       | 20%              |
| 3000             | 1m 29s       | 24.25%           |
| 3500             | 1m 42s       | 28.25%           |
| 4000             | 1m 58s       | 32.25%           |
| 4500             | 2m 10s       | 36.25%           |
| 5000             | 2m 25s       | 40.25%           |
| 5500             | 2m 37s       | 44.25%           |
| 6000             | 2m 53s       | 48.5%            |
| 6500             | 3m 4s        | 52.5%            |
| 7000             | 3m 21s       | 56.5%            |
| 7500             | 3m 33s       | 60.5%            |
python2.7: malloc.c:4023: _int_malloc: Assertion `(unsigned long) (size) >= (unsigned long) (nb)' failed.
```
or 
```
ubuntu@ip-172-31-53-120:~$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial

Downloading https://docs-assets.developer.apple.com/turicreate/data/one_shot_backgrounds.sarray.tar
Download completed: /var/tmp/data_cache/one_shot_backgrounds.sarray.tar
Augmenting input images using 951 background images.
+------------------+--------------+------------------+
| Images Augmented | Elapsed Time | Percent Complete |
+------------------+--------------+------------------+
| 0                | 5.89s        | 0%               |
| 1                | 5.89s        | 0%               |
| 2                | 5.90s        | 0%               |
| 3                | 5.91s        | 0%               |
| 4                | 5.93s        | 0%               |
| 5                | 5.95s        | 0%               |
| 10               | 6.06s        | 0%               |
| 50               | 7.19s        | 0.25%            |
| 100              | 8.17s        | 0.75%            |
| 500              | 17.99s       | 4%               |
| 1000             | 34.73s       | 8%               |
| 1500             | 46.22s       | 12%              |
| 2000             | 1m 2s        | 16%              |
| 2500             | 1m 13s       | 20%              |
| 3000             | 1m 29s       | 24.25%           |
| 3500             | 1m 42s       | 28.25%           |
| 4000             | 1m 58s       | 32.25%           |
| 4500             | 2m 10s       | 36.25%           |
| 5000             | 2m 25s       | 40.25%           |
| 5500             | 2m 37s       | 44.25%           |
| 6000             | 2m 53s       | 48.5%            |
| 6500             | 3m 4s        | 52.5%            |
| 7000             | 3m 21s       | 56.5%            |
| 7500             | 3m 33s       | 60.5%            |
*** Error in `python': double free or corruption (!prev): 0x000055d5c08b87b0 ***
======= Backtrace: =========
....
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7f286c42d7e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7f286c43637a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7f286c43a53c]
....
6000-7f286c576000 r-xp 00000000 ca:01 21482                      /lib/x86_64-linux-gnu/libc-2.23.so
7f286c576000-7f286c776000 ---p 001c0000 ca:01 21482                      /lib/x86_64-linux-gnu/libc-2.23.so
7f286c776000-7f286c77a000 r--p 001c0000 ca:01 21482                      /lib/x86_64-linux-gnu/libc-2.23.so
```
I used Ubuntu OS with various entivironments but I always get Error when Image Arugumantation progress reaches to around 60%. 

Is anyone facing  the same problem ?

### LINKED ISSUE #2191: one_shot_object_detector gets killed because high memory usage
I'm running to this issue, with a simple data set of 180 images, pretty small, on the peak before getting killed python process uses 62 GB of memory 16 is physical and the rest is a swap.

Here's the project with all the images and the script: [https://github.com/MaximBazarov/experiment_one_shot_object_detector](https://github.com/MaximBazarov/experiment_one_shot_object_detector)

**python3 model_create.py**          
**Python 3.6.5**
**pip freeze**
```sh
certifi==2019.6.16
chardet==3.0.4
coremltools==2.1.0
decorator==4.4.0
graphviz==0.8.4
idna==2.8
llvmlite==0.29.0
mxnet==1.1.0.post0
numba==0.45.0
numpy==1.17.0
pandas==0.25.0
Pillow==6.1.0
prettytable==0.7.2
protobuf==3.9.0
python-dateutil==2.8.0
pytz==2019.2
requests==2.22.0
resampy==0.2.1
scipy==1.3.0
six==1.12.0
turicreate==5.6
urllib3==1.25.3
```

the last breath: 
```sh
| 23750            | 22m 34s      | 8.75%            |
| 24000            | 22m 51s      | 9%               |
| 24250            | 22m 59s      | 9%               |
| 24500            | 23m 7s       | 9.25%            |
[1]    5051 killed     python3 model_create.py

```

shantanuchhabra:
Fixes #2037
Fixes #2038 
Fixes #2045
Fixes #2191


src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +135,96 @@ gl_sframe augment_data(const gl_sframe &data,
...
+  size_t nsegments = output_writer.num_segments();
+  size_t segment_size = backgrounds.size()/nsegments;
+  std::mutex counter_lock;
+  size_t augmented_counter = 0; // read and write protected by counter_lock mutex
+  std::mutex segment_locks[nsegments];
+
+  parallel_for(0, nsegments, [&](size_t segment_id){
+    auto segment_start = segment_id * segment_size;
+    auto segment_end = (segment_id+1) * segment_size;
+    parallel_for(segment_start, segment_end, [&](size_t background_image_index){
```
nickjong: Generally, nested `parallel_for` invocations is almost a guaranteed deadlock. Do you not observe this in your testing? My experience has been that the outer invocation will fill the entire global thread pool with instances of its lambda, so that all n threads are inside this call:
https://github.com/apple/turicreate/blob/38c18e019be5b0c202790bedd6113d13d5d59796/src/core/parallel/lambda_omp.hpp#L115
But then that lambda will itself call `parallel_for`, submitting the inner lambda do the pool and waiting here:
https://github.com/apple/turicreate/blob/38c18e019be5b0c202790bedd6113d13d5d59796/src/core/parallel/lambda_omp.hpp#L120
But the join never finishes because the inner lambdas never run because all n threads are still waiting for the outer lambdas to finish.

nickjong: Regardless, I think we want a single `parallel_for` or `in_parallel` call here. Can you articulate (in words) what work should happen on what thread? Then the code should as simply as possibly express that design

znation:
> Maybe fixed

Let's figure out whether this PR fixes those issues before we merge it. It's good to know what the code we're changing is intended to fix.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +135,95 @@ gl_sframe augment_data(const gl_sframe &data,
...
-      decode_image_inplace(object);
-    }
-    int row_number = -1;
-    for (const auto& background_ft: backgrounds.range_iterator()) {
-      row_number++;
-      flex_image flex_background = background_ft.to<flex_image>();
-      if (!(flex_background.is_decoded())) {
-        decode_image_inplace(flex_background);
+  size_t nsegments = output_writer.num_segments();
+  size_t segment_size = backgrounds.size()/nsegments;
```
hoytak: Move this computation to line 151-152 to avoid problems with interger flooring.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +135,95 @@ gl_sframe augment_data(const gl_sframe &data,
...
+  size_t augmented_counter = 0; // read and write protected by counter_lock mutex
+  
+  parallel_for(0, nsegments, [&](size_t segment_id){
+    auto segment_start = segment_id * segment_size;
+    auto segment_end = (segment_id+1) * segment_size;
+    for (size_t background_image_index = segment_start;
+                background_image_index < segment_end;
+                background_image_index++){
+      for (const auto& row: data.range_iterator()) {
+        flex_image object = row[column_index_map[image_column_name]].to<flex_image>();
```
hoytak: Use const flex_image& and .get<flex_image>();

hoytak: Check outside of loop that that column is indeed an Image type.  Then get<> will work fine. 

Also, save the column index outside the loop as an integer -- avoid doing this lookup all the time.

hoytak: Actually, with modifying in place, this isn't the best.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +135,95 @@ gl_sframe augment_data(const gl_sframe &data,
...
+          background_height,
+          (boost::gil::rgb8_pixel_t*)(flex_background.get_image_data()),
+          background_channels * background_width // row length in bytes
+          );
+        flex_image synthetic_image = create_synthetic_image(starter_image_view, background_view, parameter_sampler, object);
+        output_writer.write({
+          synthetic_image,
+          annotation
+        }, segment_id);
+        counter_lock.lock();
```
hoytak: Use atomics instead of locks.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +135,95 @@ gl_sframe augment_data(const gl_sframe &data,
...
+    for (size_t background_image_index = segment_start;
+                background_image_index < segment_end;
+                background_image_index++){
+      for (const auto& row: data.range_iterator()) {
+        flex_image object = row[column_index_map[image_column_name]].to<flex_image>();
+        std::string label = row[column_index_map[target_column_name]].to<flex_string>();
+        if (!(object.is_decoded())) {
+          decode_image_inplace(object);
+        }
+        auto background_ft = backgrounds[background_image_index];
```
nickjong: Retrieving each background by index one at a time is very inefficient. Use `gl_sarray::range_iterator` for this purpose. Note that you can specify the range of indices

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +135,95 @@ gl_sframe augment_data(const gl_sframe &data,
...
-      if (!(flex_background.is_decoded())) {
-        decode_image_inplace(flex_background);
+  size_t nsegments = output_writer.num_segments();
+  size_t segment_size = backgrounds.size()/nsegments;
+  std::mutex counter_lock;
+  size_t augmented_counter = 0; // read and write protected by counter_lock mutex
+  
+  parallel_for(0, nsegments, [&](size_t segment_id){
+    auto segment_start = segment_id * segment_size;
+    auto segment_end = (segment_id+1) * segment_size;
```
nickjong: This code assumes that the number of background images is an even multiple of the number of segments.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed the code to have only one parallel_for. Every thread superimposes all the starter images on a segment of the backgrounds.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Done.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Done.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Done.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Done.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +134,100 @@ gl_sframe augment_data(const gl_sframe &data,
...
+      for (const auto& row: data.range_iterator()) {
+        // go through all the starter images and create augmented images for
+        // all starter images and the respective chunk of background images
+        flex_image object = row[image_column_index].to<flex_image>();
+        std::string label = row[target_column_index].to<flex_string>();
+        if (!(object.is_decoded())) {
+          decode_image_inplace(object);
+        }
+        flex_image flex_background = background_ft.to<flex_image>();
+        if (!(flex_background.is_decoded())) {
```
nickjong: Why decode the background image each time we do a new starter image?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +134,100 @@ gl_sframe augment_data(const gl_sframe &data,
...
       row_number++;
-      flex_image flex_background = background_ft.to<flex_image>();
-      if (!(flex_background.is_decoded())) {
-        decode_image_inplace(flex_background);
+      for (const auto& row: data.range_iterator()) {
+        // go through all the starter images and create augmented images for
+        // all starter images and the respective chunk of background images
+        flex_image object = row[image_column_index].to<flex_image>();
+        std::string label = row[target_column_index].to<flex_string>();
+        if (!(object.is_decoded())) {
```
nickjong: Any reason not to decode the start images upfront, before the `parallel_for`? Instead of requiring each thread to create its own local decoded version?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +134,100 @@ gl_sframe augment_data(const gl_sframe &data,
...
+        boost::gil::rgba8_image_t::view_t starter_image_view = create_starter_image_view(object);
+
+        boost::gil::rgb8_image_t::view_t background_view = interleaved_view(
+          background_width,
+          background_height,
+          (boost::gil::rgb8_pixel_t*)(flex_background.get_image_data()),
+          background_channels * background_width // row length in bytes
+          );
+        flex_image synthetic_image = create_synthetic_image(starter_image_view, background_view, parameter_sampler, object);
+        encode_image_inplace(synthetic_image);
```
nickjong: Could we factor out a helper function that produces the synthetic `flex_image` from an input background `flex_image` and starter `flex_image`?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -133,85 +134,100 @@ gl_sframe augment_data(const gl_sframe &data,
...
+          annotation
+        }, segment_id);
+        augmented_counter++;
+        if (verbose) {
+          std::ostringstream d;
+          // For pretty printing, floor percent done
+          // resolution to the nearest .25% interval.  Do this by multiplying by
+          // 400, then do integer division by the total size, then float divide
+          // by 4.0
+          int augmented_rows_completed = augmented_counter;
```
nickjong: You should not read the atomic again after you increment above; another thread may already have changed it.

Touch the atomic once. I also strongly recommend using the more verbose API in std::atomic to make it clear that this is an atomic, not a normal integer.

https://en.cppreference.com/w/cpp/atomic/atomic/fetch_add

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -124,6 +124,44 @@ boost::gil::rgba8_image_t::view_t create_starter_image_view(flex_image &object_i
   return starter_image_view;
 }
 
+std::pair<flex_image,flex_dict> create_synthetic_image_from_background_and_starter(
+  flex_image starter,
+  const flex_image &background,
+  std::string label,
+  size_t seed,
+  size_t row_number) {
```
nickjong: The indentation of these arguments, at the same level as the first statement in the body, hides the beginning of the body.

Could we just use `git-clang-format` to normalize the formatting here?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -124,6 +124,44 @@ boost::gil::rgba8_image_t::view_t create_starter_image_view(flex_image &object_i
...
+  if (!(starter.is_decoded())) {
+    decode_image_inplace(starter);
```
nickjong: This implies that we're still decoding each starter image n times, where n is the number of background images.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -157,47 +195,19 @@ gl_sframe augment_data(const gl_sframe &data,
     size_t row_number=segment_start;
     for (const auto& background_ft: backgrounds.range_iterator(segment_start, segment_end)) {
       row_number++;
+      flex_image flex_background = background_ft.to<flex_image>();
```
nickjong: Could we just use `turi::image_util::decode_image` to make this one, more readable, line:
```
flexible_type flex_background = decode_image(background_ft.to<flex_image());
```
?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Done, thanks for the reference!

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Done, thanks for the tip!

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Done.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Should be fixed now.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,208 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
   return starter_image_view;
 }
 
+std::pair<flex_image, flex_dict>
+create_synthetic_image_from_background_and_starter(const flex_image &starter,
+                                                   const flex_image &background,
+                                                   std::string label,
+                                                   size_t seed,
+                                                   size_t row_number) {
+  // if (!(starter.is_decoded())) {
```
nickjong: Do we still need this?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,208 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+    size_t segment_start = (segment_id * backgrounds.size()) / nsegments;
+    size_t segment_end = ((segment_id + 1) * backgrounds.size()) / nsegments;
+    size_t row_number = segment_start;
+    for (const auto &background_ft :
+         backgrounds.range_iterator(segment_start, segment_end)) {
       row_number++;
-      flex_image flex_background = background_ft.to<flex_image>();
+      flex_image flex_background =
+          image_util::decode_image(background_ft.to<flex_image>());
       if (!(flex_background.is_decoded())) {
```
nickjong: Do we still need this?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,208 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+        // annotation to output SFrame.
+        output_writer.write({synthetic_image, annotation}, segment_id);
+        augmented_counter.fetch_add(1, std::memory_order_relaxed);
+        if (verbose) {
+          std::ostringstream d;
+          // For pretty printing, floor percent done
+          // resolution to the nearest .25% interval.  Do this by multiplying by
+          // 400, then do integer division by the total size, then float divide
+          // by 4.0
+          int augmented_rows_completed = augmented_counter;
```
nickjong: You're still reading from the atomic in a separate statement from when you increment it. The primary benefit of `std::atomic` is that you can both increment and read the value at once, atomically

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,208 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+              << '%';
+            table.print_progress_row(augmented_rows_completed,
+                                     augmented_rows_completed, progress_time(),
+                                     d.str());
+          }
+        }
       }
-    }
-  }
+    };
```
nickjong: Nit: what is this semi-colon for?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Nope, removed, thanks for pointing it out!

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Should be fixed now.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed now, thanks!

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -225,17 +218,14 @@ gl_sframe augment_data(const gl_sframe &data,
...
           // 400, then do integer division by the total size, then float divide
           // by 4.0
-          int augmented_rows_completed = augmented_counter;
-          if (augmented_rows_completed % 100 == 0) {
-            d << augmented_rows_completed * 400 / total_augmented_rows / 4.0
-              << '%';
-            table.print_progress_row(augmented_rows_completed,
-                                     augmented_rows_completed, progress_time(),
-                                     d.str());
+          if (augmented_counter % 100 == 0) {
```
nickjong: `augmented_counter` is an atomic variable. You should only touch it once per iteration. Here, you are touching it up to five times. There's no guarantee it will have the same value each time.

You need to touch the value exactly once, e.g.
```
int augmented_rows_completed = augmented_counter.fetch_add(1);
```
Then use only the local copy, `augmented_rows_completed` in this iteration.

If you're unsure how to use `std::atomic` or any other language construct, please ask or read the relevant docs. Don't just guess!

shantanuchhabra: Thanks for the correction, should be fixed now.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -211,17 +211,19 @@ gl_sframe augment_data(const gl_sframe &data,
         // write the synthetically generated image and the constructed
         // annotation to output SFrame.
         output_writer.write({synthetic_image, annotation}, segment_id);
-        augmented_counter.fetch_add(1, std::memory_order_relaxed);
+        size_t augmented_rows_completed = augmented_counter.fetch_add(1);
```
nickjong: Hmm, what does this print in your testing?

Upon closer inspection, `fetch_add` returns the value *before* the addition, so you might need to add one here

shantanuchhabra: It was printing a 0 that I should have thought of as anomalous earlier. Sorry about that. Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,200 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+      normalize(mat * bottom_left_corner),
+      normalize(mat * bottom_right_corner)};
   parameter_sampler.set_warped_corners(warped_corners);
 
   float min_x = std::numeric_limits<float>::max();
   float max_x = std::numeric_limits<float>::min();
   float min_y = std::numeric_limits<float>::max();
   float max_y = std::numeric_limits<float>::min();
-  for (auto corner: warped_corners) {
+  for (auto corner : warped_corners) {
```
hoytak: Change this here to const auto&

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,200 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+      (boost::gil::rgba8_pixel_t *)(object.get_image_data()),
+      object.m_channels * object.m_width  // row length in bytes
   );
   return starter_image_view;
 }
 
+std::pair<flex_image, flex_dict>
+create_synthetic_image_from_background_and_starter(const flex_image &starter,
+                                                   const flex_image &background,
+                                                   std::string label,
```
hoytak: If possible, change this to const std::string& as well -- this avoids a copy / reallocation.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,200 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
-    flex_image object = row[column_index_map[image_column_name]].to<flex_image>();
-    std::string label = row[column_index_map[target_column_name]].to<flex_string>();
-    if (!(object.is_decoded())) {
-      decode_image_inplace(object);
-    }
-    int row_number = -1;
-    for (const auto& background_ft: backgrounds.range_iterator()) {
+  size_t image_column_index = column_index_map[image_column_name];
+  size_t target_column_index = column_index_map[target_column_name];
+  DASSERT_EQ(data[image_column_name].dtype(), flex_type_enum::IMAGE);
```
hoytak: If there's a chance these are user facing, they should be if(!(...) ){ log_and_throw(...); }. DASSERT_XXX is only for internal programmer consistency checks, not user-accessible code paths.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,200 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+  DASSERT_EQ(data[target_column_name].dtype(), flex_type_enum::STRING);
```
hoytak: Same here.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,200 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
-        d << augmented_rows_completed * 400 / total_augmented_rows / 4.0 << '%';
-        table.print_progress_row(augmented_rows_completed,
-                                  augmented_rows_completed,
-                                  progress_time(), d.str());
+      flex_image flex_background =
+          image_util::decode_image(background_ft.to<flex_image>());
+      for (const auto &row : decompressed_data.range_iterator()) {
+        // go through all the starter images and create augmented images for
+        // all starter images and the respective chunk of background images
+        flex_image object = row[image_column_index].to<flex_image>();
```
hoytak: If we are confident that this is an image column, then we should use `const flex_image& object = row[image_column_index].get<flex_image>();`.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,200 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+        std::string label = row[target_column_index].to<flex_string>();
+        std::pair<flex_image, flex_dict> synthetic_row =
+            create_synthetic_image_from_background_and_starter(
+                object, flex_background, label, seed, row_number);
+        flex_image synthetic_image = synthetic_row.first;
+        flex_dict annotation = synthetic_row.second;
+        // write the synthetically generated image and the constructed
+        // annotation to output SFrame.
+        output_writer.write({synthetic_image, annotation}, segment_id);
```
hoytak: Add std::move() around both synthetic_image and annotation to avoid a copy of the data.  flex_dict copy is expensive.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -63,181 +62,200 @@ flex_dict build_annotation( ParameterSampler &parameter_sampler,
...
+          // resolution to the nearest .25% interval.  Do this by multiplying by
+          // 400, then do integer division by the total size, then float divide
+          // by 4.0
+          if (augmented_rows_completed % 100 == 0) {
+            d << augmented_rows_completed * 400 / total_augmented_rows / 4.0
+              << '%';
+            table.print_progress_row(augmented_rows_completed,
+                                     augmented_rows_completed, progress_time(),
+                                     d.str());
+          }
```
hoytak: Nice.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

hoytak:
Looks good.  We should be careful about missing values, but looks great.

PR #2164, opened 2019-07-25
Support enough of the DOM so that Vega attempts to register event listeners

Saba9:
In this PR we implement enough of the DOM API so that Vega will now attempt to add event listeners for interactivity. Our current Obj-C implementation of `addEventListener` is a no-op and will need to be expanded in a later PR; it is however being called by Vega which is promising.

The relevant changes include:
1. Expanding the `VegaHTMLElementInterface` so that it now supports calls to methods such as `addChild`, `removeChild`, `addEventListener`, etc.
2. Implementing versions of these methods in `VegaHTMLElement`.
3. Instantiating a container element which Vega will place our canvas in `VegaRenderer` and referencing this element during Vega initialization.
4. Adding the `body` property to `document`.

### DIFF
--- a/src/visualization/vega_renderer/JSCanvas.m
+++ b/src/visualization/vega_renderer/JSCanvas.m
@@ -695,7 +695,9 @@
 
 - (instancetype)initWithContext:(CGContextRef)parentContext {
     self = [super initWithTagName:@"canvas"];
-    self.context = [[VegaCGContext alloc] initWithContext:parentContext];
+    if(self) {
+        self.context = [[VegaCGContext alloc] initWithContext:parentContext];
+    }
     return self;
 }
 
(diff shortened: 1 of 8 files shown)


src/visualization/vega_renderer/JSCanvas.m
```diff
@@ -694,13 +694,12 @@ - (VegaCGLinearGradient *)createLinearGradientWithX0:(double)x0 y0:(double)y0 x1
 @implementation VegaCGCanvas
 
 - (instancetype)initWithContext:(CGContextRef)parentContext {
-    self = [super init];
+    self = [super initWithTagName:@"canvas"];
```
syoutsey: As we're setting the `context` property on `self` it might be a good idea to check if the call to `super` returned non-nil, as in [here](https://github.com/apple/turicreate/blob/bca0270cc1090c5c9c8b0a3526d2d4c756a23cdf/src/ml/neural_net/TCMPSImageAugmenting.m#L33).

Saba9: Good catch. There are a few other spots where I need to add this as well. I'll push an update.

PR #2166, opened 2019-07-26
Passes in --no-python to our cpp test runners.

hoytak:
--no-python works fine now.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,8 +12,8 @@ build_wheel_debug_linux_python27:
   tags:
     - docker
   services:
-    - docker:dind
-  image: docker:stable
+    - docker:18.09-dind
+  image: docker:18.09
   stage: build
   variables:
     DOCKER_HOST: tcp://docker:2375/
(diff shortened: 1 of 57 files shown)


TobyRoseman:
@hoytak - I think something is wrong with your branch. You seem to have a lot of unrelated commits.

.gitlab-ci.yml
```diff
@@ -192,8 +192,8 @@ test_cpp_mac_10_13:
     - build_wheel_mac_10_14_python27
   script:
     - export VIRTUALENV="/Library/Frameworks/Python.framework/Versions/2.7/bin/virtualenv"
-    - (test -d debug && test -d release) || ./configure
-    - cd release/test
+    - (test -d debug && test -d release) || ./configure --no-python
+    - cd debug/test
```
TobyRoseman: Why do we want to use debug rather release?

hoytak: Because debug actually runs all the checks; release does not.  When the NDEBUG flag is passed in, there are numerous internal consistency checks that are turned off, and many of the unit tests in some of the deeper parts of the engine rely on hitting these to get full test coverage.

nickjong: What is the effect on the runtime of the unit tests?

hoytak: It's probably faster to compile + run in debug mode vs. release mode.  They still run in < 15-20 min or so on my desktop -- the CPP tests are much cheaper than the python tests.   I haven't timed the tests in release mode, but the extra compile time required for building in release (2x or so) is comparable to that.

Issue #2170, opened 2019-07-29
`dropna` followed by `unpack` is not lazy.


shantanuchhabra:
We observed that right before object detector training, [this line](https://github.com/apple/turicreate/blob/master/src/python/turicreate/toolkits/object_detector/object_detector.py#L315) forces materialization of the SFrame. The expectation is that the code should execute lazily.

The `dropna` implementation creates a lambda function over all the rows and does the selection in the lambda function itself.  This means that the `unpack` call ends up cycling through all the images in the dataset instead of just the annotation columns.

PR #2171, opened 2019-07-29
Updates and optimizations to logic for dropna.

hoytak:
This PR updates the logic in unity_sframe and unity_sarray for dealing
with dropping missing values.

- Fixes https://github.com/apple/turicreate/issues/2170.  The root issue
  there was that the dropna function was defined as a function over all
  the rows instead of just the rows needed.  Thus the optimizer would
  not treat it correctly.

- Eliminated duplicated code between unity_sarray and unity_sframe.
  Dropna on a single column now uses unity_sarray implementation to
  provide the mask column.

### DIFF
--- a/src/core/data/flexible_type/flexible_type.cpp
+++ b/src/core/data/flexible_type/flexible_type.cpp
@@ -487,34 +487,47 @@ void flexible_type::erase(const flexible_type& index) {
   }
 }
 
-bool flexible_type::is_zero() const {
-  switch(get_type()) {
-    case flex_type_enum::INTEGER:
-      return get<flex_int>() == 0;
-    case flex_type_enum::FLOAT:
-      return get<flex_float>() == 0.0;
-    case flex_type_enum::STRING:
-      return get<flex_string>().empty();
-    case flex_type_enum::VECTOR:
-      return get<flex_vec>().empty();
-    case flex_type_enum::LIST:
-      return get<flex_list>().empty();
-    case flex_type_enum::DICT:
-      return get<flex_dict>().empty();
-    case flex_type_enum::IMAGE:
-      return get<flex_image>().m_format == Format::UNDEFINED;
-    case flex_type_enum::UNDEFINED:
-      return true;
-    default:
-      log_and_throw("Unexpected type!");
-  };
-  __builtin_unreachable();
-}
+bool flexible_type::contains_na() const {
 
-bool flexible_type::is_na() const {
-  auto the_type = get_type();
-  return (the_type == flex_type_enum::UNDEFINED) ||
-          (the_type == flex_type_enum::FLOAT && std::isnan(get<flex_float>()));
+  switch (get_type()) {
+    case flex_type_enum::VECTOR:
+      for (const auto& cc : get<flex_vec>()) {
+        if (std::isnan(cc)) {
+          return true;
+        }
+      }
+      return false;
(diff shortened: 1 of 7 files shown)


src/core/data/flexible_type/flexible_type.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright © 2017 Apple Inc. All rights reserved.
+    /* Copyright © 2017 Apple Inc. All rights reserved.
```
TobyRoseman: probably don't want this space.

hoytak: > Looks like you have changes here from #2166. That's probably not what you want.

That is what I want; gitlab doesn't seem to pass without #2166, so this branch is based off of that so I can get it to pass.  Once #2166 gets merged into master, it will drop from this PR.  Yes, dependent PRs are annoying, but it's the best I can do here.

src/core/data/flexible_type/flexible_type.cpp
```diff
@@ -487,34 +487,47 @@ void flexible_type::erase(const flexible_type& index) {
...
-bool flexible_type::is_zero() const {
-  switch(get_type()) {
-    case flex_type_enum::INTEGER:
-      return get<flex_int>() == 0;
-    case flex_type_enum::FLOAT:
-      return get<flex_float>() == 0.0;
-    case flex_type_enum::STRING:
-      return get<flex_string>().empty();
+bool flexible_type::contains_na() const {
+
```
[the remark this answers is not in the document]
hoytak: Let's hold off on that example until we have a better idea of what that would look like and if it's needed.

src/core/data/flexible_type/flexible_type.hpp
```diff
@@ -2630,6 +2636,39 @@ inline FLEX_ALWAYS_INLINE_FLATTEN EXT_ENABLE_IF(bool) operator<=(const T& other,
  */
 template <typename T, typename S>
 inline FLEX_ALWAYS_INLINE_FLATTEN EXT_ENABLE_IF(bool) operator>=(const T& other, const S& f) { return (f <= other); }
+
+
+bool FLEX_ALWAYS_INLINE_FLATTEN flexible_type::is_na() const {
```
znation: Why does this need to be in the header file? Isn't it sufficient to put it in the .cpp and let the linker do the inlining across translation units?

I'm also curious about the performance difference with and without inlining this function. What does the difference look like for various scenarios?

hoytak: We don't have link time optimization turned on, so the linker doesn't do anything. 

I haven't benchmarked it, but since it's called in an inner loop, is extremely simple, and can easily benefit from type-aware optimizations, so I felt inlining it would be a good optimization to do.

znation: Oh, right. We should probably turn on link time optimization (or at least try it out) and benchmark the results.

I'm skeptical of any so-called optimization that doesn't come with a before & after performance profile, taking into account our real user scenarios. I think we can potentially add a lot of complexity, increased build time, etc. in the name of optimization without actually making any measurable improvements if we aren't being data-driven here.

hoytak: Fair, agreed.  We should chat with @Jarvi-Izana on Monday about a good benchmark suite setup.

PR #2172, opened 2019-07-29
WIP : Adding Pybind11 as a dependency and integrated it in our build system

shreyajain17:
The goal is ultimately to plumb Object Detector/Activity Classifier C-API to Python. The C-API at this time can only be called with MPS. To be able to plumb it through we need it to be able to call MXNet (which is in python) for training on Linux (CPU and GPU) and Mac (CPU) through the C-API.

Added tcmxnet as a shared library which can depend on unity_shared and use pybind11 as well.


nickjong:
To simplify reasoning about our git history, I would recommend breaking this into at least two separate PRs (and therefore two separate commits in the final repo): one that is a pure source drop of the external code (with NO modifications), and a second one that incorporates it.

Also makes it much easier to reason about reviewing. Especially when source changes are required to the deps.

shreyajain17:
> To simplify reasoning about our git history, I would recommend breaking this into at least two separate PRs (and therefore two separate commits in the final repo): one that is a pure source drop of the external code (with NO modifications), and a second one that incorporates it.
> 
> Also makes it much easier to reason about reviewing. Especially when source changes are required to the deps.

That makes sense. To address this, I will be put up two different PRs and will be closing this PR for now.

PR #2173, opened 2019-07-29
Updated backgrounds URL to use smaller backgrounds

### LINKED ISSUE #2169: One Shot Object Detector has high disk footprint.
Currently, disk usage in the data cache during augmentation alone is as large as 100 GB. We can reduce this footprint by decreasing the background dimensions, and/or by adding image compression.

shantanuchhabra:
Fixes #2169

### DIFF
--- a/src/python/turicreate/toolkits/_data_zoo.py
+++ b/src/python/turicreate/toolkits/_data_zoo.py
@@ -19,8 +19,8 @@ DATA_URL_ROOT = 'https://docs-assets.developer.apple.com/turicreate/data/'
 
 class OneShotObjectDetectorBackgroundData(object):
     def __init__(self):
-        self.source_tar_filename = "one_shot_backgrounds.sarray.tar"
-        self.destination_tar_filename = "one_shot_backgrounds.sarray.tar"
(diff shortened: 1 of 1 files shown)


nickjong:
How certain are we that we've finalized the sizes we want here? Would it be safe for now just to go with a software change?

In general, what's the naming/versioning scheme we want to use?

shantanuchhabra:
> How certain are we that we've finalized the sizes we want here? Would it be safe for now just to go with a software change?

I wouldn't say that these sizes are final, but the results from these sizes should be the same as the results from the earlier backgrounds, if not better. Definitely cuts augmentation time and reduces the disk footprint by 75%. 
Which software change are you referring to here? Do you mean that we should have the software change from #2160 without using smaller backgrounds?

> 
> In general, what's the naming/versioning scheme we want to use?

How about a versioning scheme like "one_shot_backgrounds_v{1,2,3...}"? I can update the data if this sounds good. I'd like to understand better about how to decide a good versioning scheme in production.

nickjong:
I think we should punt the design of a versioning scheme until later and perform the resizing of the background images in the toolkit for now

shantanuchhabra:
Sounds good! Will put up another PR to make the software change!

PR #2176, opened 2019-07-30
Handles missing values in feature columns in AC, OD, One-shot, Style Transfer, Image Classifier and Drawing Classifier consistently

### LINKED ISSUE #1994: One Shot - Create doesn't handle missing values
If you try to train a One Shot Model using an SFrame with `None`s in the image column, this is what you get:
```
In [23]: tc.one_shot_object_detector.create(data, 'label', max_iterations=20)                                                                                                                                                                                                                                                                                             

RuntimeError                              Traceback (most recent call last)
~/anaconda3/envs/mygreatenvironment/lib/python3.6/site-packages/turicreate/extensions.py in __run_class_function(self, fnname, args, kwargs)
    273         try:
--> 274             ret = self._tkclass.call_function(fnname, argument_dict)
    275         except RuntimeError as exc:

turicreate/_cython/cy_model.pyx in turicreate._cython.cy_model.UnityModel.call_function()

turicreate/_cython/cy_model.pyx in turicreate._cython.cy_model.UnityModel.call_function()

RuntimeError: Cannot decode image. Unknown format.

During handling of the above exception, another exception occurred:

<ipython-input-23-b9a4a4b470b7> in <module>
----> 1 tc.one_shot_object_detector.create(data, 'label', max_iterations=20)

~/anaconda3/envs/mygreatenvironment/lib/python3.6/site-packages/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py in create(data, target, backgrounds, batch_size, max_iterations, verbose)
     55         If True, print progress updates and model details.
     56     """
---> 57     augmented_data = preview_augmented_images(data, target, backgrounds)
     58     model = _tc.object_detector.create( augmented_data,
     59                                         batch_size=batch_size,

~/anaconda3/envs/mygreatenvironment/lib/python3.6/site-packages/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py in preview_augmented_images(data, target, backgrounds, verbose, **kwargs)
     66                                             target_column_name,
     67                                             backgrounds,
---> 68                                             options_for_augmentation)
     69     return augmented_data

~/anaconda3/envs/mygreatenvironment/lib/python3.6/site-packages/turicreate/extensions.py in <lambda>(*args, **kwargs)
    288         elif name in self._functions:
    289             # is it a function?
--> 290             ret = lambda *args, **kwargs: self.__run_class_function(name, args, kwargs)
    291             ret.__doc__ = "Name: " + name + "\nParameters: " + str(self._functions[name]) + "\n"
    292             try:

~/anaconda3/envs/mygreatenvironment/lib/python3.6/site-packages/turicreate/extensions.py in __run_class_function(self, fnname, args, kwargs)
    275         except RuntimeError as exc:
    276             # Expose C++ exceptions using ToolkitError.
--> 277             raise _ToolkitError(exc)
    278         ret = _wrap_function_return(ret)
    279         return ret

ToolkitError: Cannot decode image. Unknown format.
```

shreyajain17:
fixes #1994


src/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
```diff
@@ -39,7 +39,12 @@ def preview_synthetic_training_data(data,
     out : SFrame
         An SFrame of sythetically generated annotated training data.
     """
+
     dataset_to_augment, image_column_name, target_column_name = check_one_shot_input(data, target)
+    original_dataset_size = len(dataset_to_augment)
+    dataset_to_augment = dataset_to_augment.dropna(columns=image_column_name)
```
srikris: If this is a common operation. Might be worth turning into a toolkit utility function.

src/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
```diff
@@ -39,7 +39,12 @@ def preview_synthetic_training_data(data,
...
+    if original_dataset_size != len(dataset_to_augment):
+        print("Dropping " + str(original_dataset_size - len(dataset_to_augment)) + " rows for missing features in the training dataset.")
```
TobyRoseman: At this point it's already been dropped, so I would say "Dropped" rather than "Dropping". Also I think being specific is good: I would say "for images with values of None" rather than "for missing features".

nickjong: What do other toolkits print out? Where is this warning defined for those toolkits?

src/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
nickjong: FYI, just from code inspection, I see that supervised learning toolkits either error out to use NAN for missing values (silently). They have user-controllable options for e.g. `predict` and `evaluate` but none of these drop any rows.

shreyajain17: We don't have consistency for this.

For example, the logistic classifier errors out with the error message "Missing value (None) encountered in column 'column_name'. Use the SFrame's dropna function to drop rows with 'None' values in them."
Random forest silently uses NAN values for missing values.
Image Classifier fails with the error message "ToolkitError: Cannot resize non-image type" because the image extractor can probably not handle None values. 
Drawing classifier uses exactly this since I put this in before WWDC.

nickjong: To close the loop from offline discussion: let's be consistent with the oldest, most testing toolkits and error out.

Given the shared part of this code is really just the error message, I wouldn't bother factoring it out, especially in Python code that is going away eventually

src/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
shreyajain17: @TobyRoseman @nickjong 
After discussions we have decided to make the toolkits consistent with what logistic classifier does in case missing values.

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -177,6 +177,10 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
         else:
             dataset, validation_set = _random_split_by_session(dataset, session_id)
 
+    for feature in features:
+        if any(feat is None for feat in dataset[feature]):
+            raise _ToolkitError("Missing value (None) encountered in column " + feature + " in the training dataset. Use the SFrame's dropna function to drop rows with 'None' values in them.")
```
nickjong: Ah, this makes sense, but in my head I had naively assumed we could just check for missing values at the point when we actually process the data. This approach is much cleaner. But: can you sanity check how much overhead this applies at the beginning of training? It's nice to minimize the number of passes we take through the data.

Also, given this uniform approach, I now more agree with Krishna's point about factoring this out. You can create a function somewhere that just passes in as an argument e.g. the column name

shreyajain17: Factored out the code into a toolkit utility function. It is done before training. When I compare time for preprocessing time, I don't see a overhead in checking for missing values.

PR #2178, opened 2019-07-30
Inline explore jupyter

abhishekpratapa:
Added Inline Jupyter Explore.

### DIFF
--- a/src/python/turicreate/data_structures/sframe.py
+++ b/src/python/turicreate/data_structures/sframe.py
@@ -2373,7 +2373,7 @@ class SFrame(object):
         """
         return SFrame(_proxy=self.__proxy__.head(n))
 
-    def to_dataframe(self):
+    def to_dataframe(self, explore=False):
         """
         Convert this SFrame to pandas.DataFrame.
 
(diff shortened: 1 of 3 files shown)


src/python/turicreate/data_structures/sframe.py
```diff
@@ -2373,7 +2373,7 @@ def head(self, n=10):
         """
         return SFrame(_proxy=self.__proxy__.head(n))
 
-    def to_dataframe(self):
+    def to_dataframe(self, explore=False):
```
znation: Let's not make explore be a parameter on another method. It should stay as its own method.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -2385,11 +2385,18 @@ def to_dataframe(self):
...
         assert HAS_PANDAS, 'pandas is not installed.'
         df = pandas.DataFrame()
         for i in range(self.num_columns()):
             column_name = self.column_names()[i]
-            df[column_name] = list(self[column_name])
+            if(self.column_types()[i] == _Image and explore):
+                pandas.set_option('display.max_colwidth', -1)
+                df[column_name] = [_image_resize(x[column_name])._to_pil_image() for x in self.select_columns([column_name])]
+            else:
+                df[column_name] = list(self[column_name])
```
znation: Is there not a proper way to put a PIL image inside a Pandas dataframe? That could allow us to reconcile these two codepaths and just have the dataframe actually contain the image.

znation: Check out https://www.kaggle.com/stassl/displaying-inline-images-in-pandas-dataframe for an example of storing the image directly in the dataframe.

src/python/turicreate/data_structures/sframe.py
znation: Actually, if we can get the contents of the image into the dataframe, we can just do that all the time (and not add a new parameter here).

abhishekpratapa:
Good?

PR #2179, opened 2019-07-30
API avaliability

abhishekpratapa:
- added api Availability for 10.15

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,7 +194,6 @@ if(APPLE)
     set(HAS_MLCUSTOMMODEL TRUE)
   endif()
 
-   # MPSForward Loss is only present on macOS 10.15 or higher.
   if(NOT TC_BASE_SDK_VERSION VERSION_LESS 10.15)
     add_definitions(-DHAS_MACOS_10_15)
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,6 +193,11 @@ if(APPLE)
     # CoreML MLCustomModel requires macOS 10.14 or higher
     set(HAS_MLCUSTOMMODEL TRUE)
   endif()
+
+   # MPSForward Loss is only present on macOS 10.15 or higher.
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -193,6 +193,11 @@ if(APPLE)
     # CoreML MLCustomModel requires macOS 10.14 or higher
     set(HAS_MLCUSTOMMODEL TRUE)
   endif()
+
+   # MPSForward Loss is only present on macOS 10.15 or higher.
```
nickjong: I don't think we need this comment, as the new preprocessor symbol is not specific to `MPSForwardLoss`

src/ml/neural_net/style_transfer/mps_style_transfer.h
```diff
@@ -13,6 +13,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+#ifdef DHAS_MACOS_10_15
```
nickjong: Should be `HAS_MACOS_10_15`, no `D` at the front

src/ml/neural_net/style_transfer/mps_style_transfer.h
```diff
@@ -33,4 +35,6 @@ API_AVAILABLE(macos(10.15))
 
 @end
 
+#endif
```
nickjong: Nit: let's comment with ```// #ifdef HAS_MACOS_10_15```

abhishekpratapa: Solid

src/ml/neural_net/style_transfer/mps_style_transfer.h
abhishekpratapa: Done!

CMakeLists.txt
abhishekpratapa: Removed comment!

PR #2182, opened 2019-07-31
Adds pybind11 as a dependency

shreyajain17:


znation:
Pybind11 is intended to be used as a header-only library, as I understand. I think based on @hoytak's comment above, this implies it should go into `src/external`, and only include the headers and not the full contents of the repo.

shreyajain17:
@TobyRoseman @znation @hoytak 
I think I will address the feedback and put the pybind11 headers in ```src/external```. This will also eliminate all the doc files which are not needed.

PR #2183, opened 2019-07-31
Fix for logging in build system

hoytak:
Some of the cpp test runners were failing due to significant spewing of linker warnings about mismatched symbol visibility when linking the tests.  These warnings don't go away easily it seems, so this PR filters these out of the logging process so the tests pass.


.gitlab-ci.yml
```diff
@@ -194,7 +194,7 @@ test_cpp_mac_10_13:
     - export VIRTUALENV="/Library/Frameworks/Python.framework/Versions/2.7/bin/virtualenv"
     - (test -d debug && test -d release) || ./configure --no-python
     - cd debug/test
-    - make -j8
+    - bash -c "make -j8 2>&1 | grep -v -E 'ld[^a-z]*warning[^a-z]*direct access'"
```
TobyRoseman: Could we either start the expression with a `^` or end it with a `$`?

Same question for line 216.

PR #2189, opened 2019-08-01
Ran clang-format on all of one-shot OD

shantanuchhabra:


src/toolkits/object_detection/one_shot_object_detection/CMakeLists.txt
```diff
@@ -1,14 +1,7 @@
...
-    util/parameter_sampler.cpp
-    util/superposition.cpp
-    one_shot_object_detector.cpp
-  REQUIRES
-    image_io
-    random
-    table_printer
-    unity_core
-)
+    make_library(unity_one_shot_object_detection OBJECT SOURCES
```
znation: Let's revert the changes in this file - I'm not sure clang-format got it right.

nickjong: Yes, in fact, this no longer builds...

shantanuchhabra: Fixed.

PR #2190, opened 2019-08-01
Software change to resize backgrounds as part of OSOD toolkit

### LINKED ISSUE #2169: One Shot Object Detector has high disk footprint.

shantanuchhabra:
Fixed #2169


src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -220,20 +220,23 @@ one_shot_object_detector::one_shot_object_detector() {
...
-                                                             options["verbose"]);
+gl_sframe one_shot_object_detector::augment(
+    const gl_sframe& data, const std::string& image_column_name,
+    const std::string& target_column_name, gl_sarray& backgrounds,
+    std::map<std::string, flexible_type>& options) {
+  gl_sarray small_backgrounds = backgrounds.apply(
+      [](const flexible_type& bg) {
+        flex_image background = bg.get<flex_image>();
+        return image_util::resize_image(bg, background.m_width / 2,
+                                        background.m_height / 2,
```
nickjong: This code path also applies to user-provided backgrounds too, right? How do we know that halving these is the correct thing to do?

shantanuchhabra: Oh good point, will fix.

shantanuchhabra: Fixed!

src/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
```diff
@@ -52,6 +54,7 @@ def preview_synthetic_training_data(data,
     # {'mlmodel_path':'darknet.mlmodel', 'max_iterations' : 25}
     options_for_augmentation = {
         "seed": seed,
+        "user_provided": user_provided,
```
nickjong: Instead of adding a user option just for this purpose, wouldn’t it be simpler just to resize the images on the Python side?

shantanuchhabra: Fixed.

src/python/turicreate/toolkits/one_shot_object_detector/util/_augmentation.py
```diff
@@ -48,6 +48,12 @@ def preview_synthetic_training_data(data,
         backgrounds_tar = _tarfile.open(backgrounds_tar_path)
         backgrounds_tar.extractall()
         backgrounds = _tc.SArray("one_shot_backgrounds.sarray")
+        backgrounds = backgrounds.apply(lambda im: _tc.image_analysis.resize(
```
nickjong: Could you add a comment explaining the reason for the resizing?

nickjong:
If we've only added comments since a commit with a successful Travis run, I think it's safe to merge without waiting for the with-comments version to pass Travis

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -149,8 +143,12 @@ create_synthetic_image_from_background_and_starter(const flex_image &starter,
     log_and_throw("Background image is not decoded into raw format.");
   }
 
-  boost::gil::rgba8_image_t::view_t starter_image_view =
-      create_starter_image_view(starter);
+  flex_image rgba_flex_image = create_rgba_flex_image(starter);
+  boost::gil::rgba8_image_t::view_t starter_image_view = interleaved_view(
+      rgba_flex_image.m_width, rgba_flex_image.m_height,
+      (boost::gil::rgba8_pixel_t *)(rgba_flex_image.get_image_data()),
```
nickjong: Ah, yes, you cannot have a function that safely returns a view to a local variable inside that function.

While we're at it, could you replace this C-style cast with `reinterpret_cast<boost::gil::rgba8_pixel_t *>`?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -263,9 +261,9 @@ one_shot_object_detector::one_shot_object_detector() {
 }
 
 gl_sframe one_shot_object_detector::augment(
-    const gl_sframe &data, const std::string &image_column_name,
-    const std::string &target_column_name, const gl_sarray &backgrounds,
-    std::map<std::string, flexible_type> &options) {
+    const gl_sframe& data, const std::string& image_column_name,
+    const std::string& target_column_name, gl_sarray& backgrounds,
```
nickjong: Why do the backgrounds need to be mutable?

shantanuchhabra: Fixed.

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
shantanuchhabra: Fixed.

shantanuchhabra:
Internal tests passed, merging.

Issue #1899, opened 2019-05-19
Bad result with 100.000 iterations


banoslo:
Thanks for your great work on the framework. I am trying to apply style transfer on an image. But why is the result so bad looking? (I tested 100.000 and 10.000 iterations the results are similar)
 
This is my setting:
I use Ubuntu 18 and the recent Turicreate version for Python 3.6 with mxnet CUDA 9.1 on a NVIDIA GTX 950. I use the code lines from the Docs with a 100.000 training iterations.

Style image: 
![gogh](https://user-images.githubusercontent.com/2702147/57988954-8c5a7e00-7a94-11e9-96ac-9b161aacd925.jpg)

Test image:
![architecture-buildings-chicago-416942](https://user-images.githubusercontent.com/2702147/57988957-97151300-7a94-11e9-9dac-0bf5c6d58ea2.jpg)

Result image:
![result8](https://user-images.githubusercontent.com/2702147/57988959-9d0af400-7a94-11e9-81d1-2ee2853b897f.png)

srikris:
@banoslo What version of Turi Create as you using? We recently made some improvements that improved the quality of style transfer drastically.

banoslo:
It is version 5.5. Should this show better results?

nickjong:
The improvements that @srikris referenced should be in 5.5, so we should take a look at what's going on here

abhishekpratapa:
# Parameters to help train Style Transfer Models

Hopefully these tips will help you achieve a better result @banoslo. It'll go over the things you could experiment with and modify to iteratively achieve the model you desire.

#### style_loss_mult

The larger the values of the `style_loss_mult`, the more stylized the image 
is. The diagram below illustrates this effect (Note: the model was trained for 
`64,000` iterations.)

![style_loss_computation](https://user-images.githubusercontent.com/4616962/62800465-52e44e00-ba98-11e9-8874-d8a104fa42db.png)

An example of setting the `style_loss_mult` flag is show below:

```
model = tc.style_transfer.create(styles, content, _advanced_parameters={"style_loss_mult":[1e-4, 1e-4, 1e-4, 1e-4]})
```

#### finetune_all_params

There are two different approaches to updating parameters in the style transfer 
network.

- Only update the instance normalization layers (`finetune_all_params: False`)
- Update both the convolutional and instance normalization layers. (`finetune_all_params: True`)

An example of setting the `finetune_all_params` flag is show below:

```
[...]

PR #2192, opened 2019-08-01
Added checkpointing functionality to the style transfer model

abhishekpratapa:
Helps diagnose the problems present in #1899

### DIFF
--- a/src/python/turicreate/toolkits/style_transfer/style_transfer.py
+++ b/src/python/turicreate/toolkits/style_transfer/style_transfer.py
@@ -34,8 +34,7 @@ def _vgg16_data_prep(batch):
 
 def create(style_dataset, content_dataset, style_feature=None,
         content_feature=None, max_iterations=None, model='resnet-16',
-        verbose=True, batch_size = 6, checkpoint=False,
-        checkpoint_prefix="style_transfer", checkpoint_increment=1000, **kwargs):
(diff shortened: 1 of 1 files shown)


nickjong:
I don’t think we want to introduce this as officially supported API at this time. We’d have to give some more thought to an API design that will provide a good user experience across toolkits, instead of doing something idiosyncratic for each toolkit. I think I can get behind doing something like this as a hidden/advanced option, to address an immediate need.

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -346,6 +353,31 @@ def create(style_dataset, content_dataset, style_feature=None,
             else:
                 smoothed_loss = 0.9 * smoothed_loss + 0.1 * cur_loss
             iterations += 1
+
+            if params['checkpoint'] and iterations % params['checkpoint_increment'] == 0:
+                checkpoint_filename = params['checkpoint_prefixs'] + "-" + str(iterations) + ".model"
```
nickjong: Shouldn't the dict key be `'checkpoint_prefix'`, no trailing `'s'`?

abhishekpratapa: Yes, thank for the catch!

PR #2196, opened 2019-08-02
Added OD C-API predict()

fareeha:
Replaced use of `perform_evaluate()` with `perform_predict()`and used callback functions to determine code path for `predict()` and `evaluate()`.


src/toolkits/object_detection/object_detector.hpp
```diff
@@ -73,11 +73,12 @@ class EXPORT object_detector: public ml_model_base {
...
-  REGISTER_CLASS_MEMBER_FUNCTION(object_detector::evaluate, "data", "metric",
-                                 "options");
+  REGISTER_CLASS_MEMBER_FUNCTION(object_detector::evaluate, "data", "metric");
   register_defaults("evaluate",
-      {{"metric", std::string("auto")},
-       {"options", to_variant(std::map<std::string, flexible_type>())}});
+      {{"metric", std::string("auto")}});
+
+  REGISTER_CLASS_MEMBER_FUNCTION(object_detector::predict, "data");
+  register_defaults("predict", {{"data", to_variant(gl_sarray())}});
```
nickjong: Let's not use a default for the SFrame, which should be a required argument

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -379,35 +379,86 @@ void object_detector::train(gl_sframe data,
...
+}
+
+gl_sarray object_detector::predict(gl_sframe data) {
+  gl_sarray_writer result(flex_type_enum::LIST, 1);
+
+  auto consumer = [&](const std::vector<image_annotation>& predicted_row,
+                      const std::vector<image_annotation>& groundtruth_row) {
+    // Convert predicted_row to flex_type list to call gl_sarray_writer
+    flex_list predicted_row_ft;
+    for (auto const& each_row : predicted_row) {
```
nickjong: Let's use `const auto&` instead of `auto const&` for consistency with other areas of our code, and since it seems much more idiomatic

nickjong: Actually, could we do `const image_annotation&` for more clarity, since it doesn't cost much in complexity?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -379,35 +379,86 @@ void object_detector::train(gl_sframe data,
...
+      flex_list fl = {each_row.bounding_box.x, each_row.bounding_box.y,
```
nickjong: The bounding box should be a dictionary, not a list. Reference the parsing code from https://github.com/apple/turicreate/blob/becc15cb1bd32cbcae50d0d2fbb65810508b51c9/src/toolkits/object_detection/od_data_iterator.cpp#L53

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -379,35 +379,86 @@ void object_detector::train(gl_sframe data,
...
+                      each_row.bounding_box.width,
+                      each_row.bounding_box.height};
+      auto sa = flex_list({flex_dict{{"confidence", each_row.confidence},
+                                     {"bounding_box", fl},
+                                     {"identifier", each_row.identifier}}});
+      predicted_row_ft.push_back(sa);
```
nickjong: If `sa` is a list than you're making each row of the final SArray a list of lists. I believe we want a list of dictionaries here.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -522,27 +565,14 @@ variant_map_type object_detector::perform_evaluation(gl_sframe data,
...
-  }
-  if (std::find(metrics.begin(), metrics.end(), MAP) == metrics.end()) {
-    result_map.erase(MAP);
-  }
-
-  return result_map;
+// TODO: Should accept model_backend as an optional argument to avoid
+// instantiating a new backend during training. Or just check to see if an
+// existing backend is available?
+variant_map_type object_detector::perform_evaluation(gl_sframe data,
```
nickjong: I don't think we need `perform_evaluation` at all now...

src/toolkits/object_detection/object_detector.cpp
fareeha: What would be the key for each of the annotations? For a given image.

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -743,7 +768,333 @@ BOOST_AUTO_TEST_CASE(test_object_detector_auto_split) {
...
+    shared_float_array arr = shared_float_array::wrap(
+        buffer,
+        {OUTPUT_GRID_SIZE, OUTPUT_GRID_SIZE, anchor_boxes.size() * NUM_PREDS});
+    result["output"] = arr;
+    return result;
+  };
+  mock_nn_model->predict_calls_.push_back(predict_impl);
+
+  auto convert_yolo_impl =
+      [](const neural_net::float_array& yolo_map,
```
nickjong: Given that you went through all the trouble to specify the `yolo_map` returned from the backend (in `predict_impl`), I'm surprised that we don't assert that the `yolo_map` argument here matches. I don't think we need to check everything, but enough to sanity check that the input to `convert_yolo` was the output of `predict`...

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -20,6 +20,12 @@ namespace turi {
 namespace object_detection {
 namespace {
 
+const std::vector<std::pair<float, float>> anchor_boxes = {
```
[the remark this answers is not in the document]
fareeha: Good catch, ty!

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -439,14 +439,14 @@ gl_sarray object_detector::predict(gl_sframe data) {
...
-      auto sa = flex_list({flex_dict{{"confidence", each_row.confidence},
-                                     {"bounding_box", fl},
-                                     {"identifier", each_row.identifier}}});
-      predicted_row_ft.push_back(sa);
+    for (const image_annotation& each_row : predicted_row) {
+      flex_dict bb_dict = {{"x", each_row.bounding_box.x}, {"y", each_row.bounding_box.y},
+                      {"width", each_row.bounding_box.width},
+                      {"height", each_row.bounding_box.height}};
+      flex_dict each_annotation = {{"confidence", each_row.confidence},
+                                     {"bounding_box", bb_dict},
```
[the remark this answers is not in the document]
fareeha: Good idea! thanks

PR #2200, opened 2019-08-05
Sanity check max_iterations in Style Transfer

### LINKED ISSUE #2013: Type check for value of max_iterations
In Style Transfer, the model `create()` method accepts strings as a valid input for `max_iterations`

```
In [40]: model = tc.style_transfer.create(styles, content, max_iterations='dummy_string')
Using 'image' in style_dataset as feature column and using 'image' in content_dataset as feature column
Using CPU to create model
Analyzing visual features of the style images
| Iteration    | Loss         | Elapsed Time |
| 1            | 10.074       | 30.2         |
| 2            | 12.195       | 51.9         |
| 3            | 20.362       | 73.7         |
| 4            | 27.661       | 95.8         |
| 5            | 47.275       | 117.6        |
| 6            | 43.512       | 139.5        |
| 7            | 48.951       | 161.2        |
| 8            | 47.235       | 183.0        |
| 9            | 61.822       | 204.7        |
| 10           | 65.091       | 226.7        |
| 11           | 67.914       | 249.0        |
```

fareeha:
Closes #2013 

Added check for `max_iterations` in Style Transfer `create()` to ensure in integer >=0. Checking for incorrect formats: strings, floats and negative integers. Added tests.

For consistency in toolkits, input of `max_iterations=0` matches use in OD and defaults to same behavior as `max_iterations=None` i.e. automatically determine based on amount of data. (Tests where max_iterations=0 was previously set have been modified to max_iterations=1)

### DIFF
--- a/src/python/turicreate/toolkits/style_transfer/style_transfer.py
+++ b/src/python/turicreate/toolkits/style_transfer/style_transfer.py
@@ -110,6 +110,8 @@ def create(style_dataset, content_dataset, style_feature=None,
         raise _ToolkitError("content_dataset SFrame cannot be empty")
     if(batch_size < 1):
         raise _ToolkitError("'batch_size' must be greater than or equal to 1")
+    if max_iterations is not None and (type(max_iterations)==str or max_iterations<0):
+        raise _ToolkitError("'max_iterations' must be an integer greater than or equal to 0")
 
     from ._sframe_loader import SFrameSTIter as _SFrameSTIter
     import mxnet as _mx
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -110,6 +110,8 @@ def create(style_dataset, content_dataset, style_feature=None,
         raise _ToolkitError("content_dataset SFrame cannot be empty")
     if(batch_size < 1):
         raise _ToolkitError("'batch_size' must be greater than or equal to 1")
+    if max_iterations is not None and (type(max_iterations)==str or type(max_iterations)==float or max_iterations<0):
```
TobyRoseman: It's probably cleaner to write this line as:
```python
if max_iterations is not None and (isinstance(max_iterations, (str, float)) or max_iterations < 0):
```
But I think the logic should actually be:
```python
if max_iterations is not None and (not isinstance(max_iterations, int) or max_iterations < 0):
```

fareeha: Thanks! This reads a lot cleaner

PR #2208, opened 2019-08-06
Helper function wrapping std::function in ObjC block must capture by value, not by reference

### LINKED ISSUE #2202: Test failures in test_mps_image_augmentation.cxx
While running the C++ unit tests on a 10.14 macOS machine from https://github.com/apple/turicreate/commit/2e40f1207662ca0847752c18c1ab03d6f84eaacc, the test run hung in test_mps_image_augmentation.cxxtest. Attempt to reproduce locally yielded a memory error.

nickjong:
Fixes #2202


nickjong:
The semantics of ObjC block capture are different from C++11 lambda capture (with = or &). It captures references by reference and values by value.

Issue #1956, opened 2019-05-25
stroke based model can have significant differences between Linux and macOS


TobyRoseman:
**Setup**
 - On Linux - train a drawing classifier model using stroke based data.
 - On Linux - observe probability values for the test set.
 - On Linux - save the model and the test set.
 - On macOS - open model and test set.
 -  On macOS - look at the test set probabilities.

**Result**
The probabilities do not match. In some cases the difference are quite significant (ex: ~0.92 for the right class vs ~0.30 for the right class).

**Investigation**
For one example which was particularly off, I took a look at its output of `tc.drawing_classifier.util.draw_strokes(...)` from both Linux and macOs. The images look quite similar but are clearly not identical. 

I saved the resulting image on Linux as a png file (jpg is lossy). The probability for that image on macOs match those from Linux. So it seems clear the mismatch is cased not by the model itself but by the conversion of stroke based data to bitmap.

Looking at the code, it seems we have two entirely different implementations of the stroke conversion for macOs and Linux. I suggest we use the same implementation on both Linux and macOS. However it's probably not easy to also have the same implementation for deployment to Core ML.

PR #2209, opened 2019-08-06
Adding information about known issue in DC in user guide

shreyajain17:
Related to #1956

### DIFF
--- a/userguide/drawing_classifier/how-it-works.md
+++ b/userguide/drawing_classifier/how-it-works.md
@@ -26,7 +26,7 @@ example_drawing = [
 ]
 ```
 
-When stroke-based drawing data is given as input to `turicreate.drawing_classifier.create`, a series of pre-processing steps are applied to convert the stroke-based drawings to bitmaps which is then consumed by the neural network. This data transformation can be visualized through the `turicreate.drawing_classifier.util.draw_strokes(...)` Python API as well. It is a known issue with models trained on Linux stroke data not generalizing well to drawings rendered on Macs and vice-versa.
+When stroke-based drawing data is given as input to `turicreate.drawing_classifier.create`, a series of pre-processing steps are applied to convert the stroke-based drawings to bitmaps which is then consumed by the neural network. This data transformation can be visualized through the `turicreate.drawing_classifier.util.draw_strokes(...)` Python API as well. It is a known issue that the model may be sensitive to platform-specific rendering from strokes. For the best generalization performance, train on the same platform where the model will ultimately be deployed.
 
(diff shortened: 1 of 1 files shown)


userguide/drawing_classifier/how-it-works.md
```diff
@@ -26,7 +26,7 @@ example_drawing = [
 ]
 ```
 
-When stroke-based drawing data is given as input to `turicreate.drawing_classifier.create`, a series of pre-processing steps are applied to convert the stroke-based drawings to bitmaps which is then consumed by the neural network. This data transformation can be visualized through the `turicreate.drawing_classifier.util.draw_strokes(...)` Python API as well. 
+When stroke-based drawing data is given as input to `turicreate.drawing_classifier.create`, a series of pre-processing steps are applied to convert the stroke-based drawings to bitmaps which is then consumed by the neural network. This data transformation can be visualized through the `turicreate.drawing_classifier.util.draw_strokes(...)` Python API as well. It is a known issue with models trained on Linux stroke data not generalizing well to drawings rendered on Macs and vice-versa.
```
srikris: Why are we adding this to the user guide as opposed to opening an issue about it? It's going to be hard to remember to come back and update this when we fix it.

srikris: Also, it's not clear what trained on "Linux stroke data" means. Models trained on Linux using stroke data collected from Macs?

nickjong: We do have an issue about it. We will not be able to resolve it for our present release. It's not obvious from usage of the toolkit that models trained from stroke data on Linux seem not to generalize well to Mac.

shreyajain17: It is an open issue. (refer to #1956)
I referred the PR here so we remember to update the guide. The point according to me was to not let users have a bad experience since it a known issue that cannot be fixed yet. 
It doesn't matter where data was collected. It is the stroke data converted to bitmap done while training in Linux which will not generalize well on stroke data converted to bitmap in Mac.

nickjong: Hmm, maybe it would be clearer to flip this and suggest how to maximize the likelihood of good generalization. Something along the lines of "It is a known issue that the model may be sensitive to platform-specific rendering from strokes. For the best generalization performance, train on the same platform where the model will be deployed."
?

shreyajain17: @nickjong That does sound better.

nickjong:
What wording would you prefer? FWIW, if you could train on iOS, it does seem more likely to me that the model will generalize correctly to deployment on iOS. I think most people will intuit that macOS is closer to iOS than Linux

PR #2212, opened 2019-08-06
Fixes API docs 

### LINKED ISSUE #2112: No draw_strokes link in API docs
The "util.draw_strokes" entry in the "utilities" table at the bottom of [the Drawing Classifier API docs](https://apple.github.io/turicreate/docs/api/turicreate.toolkits.drawing_classifier.html) looks like a link and should be a link, but isn't actually a link.

Also when building the documentation this warning appears **hundreds** of time:
`<partial node>:: WARNING: toctree contains reference to nonexisting document 'generated/util.draw_strokes'`

shreyajain17:
fixes #2112 

This has some fixes for drawing classifier user guide that I found yesterday. 

I have locally built the API docs for one-shot object detector and drawing classifier. There is still one warning which I cannot get rid of. We should verify this works during 5.7 release.

### DIFF
--- a/src/python/turicreate/toolkits/drawing_classifier/__init__.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/__init__.py
@@ -10,4 +10,4 @@ from .drawing_classifier import create, DrawingClassifier
 from ..image_classifier._annotate import annotate, recover_annotation
 from . import util
 
-__all__ = ['create', 'DrawingClassifier', 'util']
+__all__ = ['create', 'DrawingClassifier', 'util', 'annotate', 'recover_annotation']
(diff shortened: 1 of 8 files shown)


src/python/turicreate/toolkits/drawing_classifier/__init__.py
```diff
@@ -10,4 +10,4 @@
 from ..image_classifier._annotate import annotate, recover_annotation
 from . import util
 
-__all__ = ['create', 'DrawingClassifier', 'util', 'annotate', 'recover_annotation']
```
znation: Why don't we want `annotate` and `recover_annotation` here?

shreyajain17: Ah. I took it out because I was trying to see what was working or not. This has nothing to do with the fix. Will put it back. Thanks

PR #2213, opened 2019-08-07
Better JavaScript error handling in GUI app

### LINKED ISSUE #2137: Plot & explore JavaScript errors aren't debuggable without stack traces
Theoretically, we can get stack traces from WebKit and JavaScriptCore using this API: https://stackoverflow.com/questions/34273540/ios-javascriptcore-exception-detailed-stacktrace-info

We probably need to implement this in several places:

* tcviz client app (macOS)
* vega_renderer

znation:
* Improves JavaScript error handling and debugging; error messages in
  the VegaContainer in the GUI app are now largely on par with the
  VegaRenderer (backend) code, with a nicely formatted message
  containing the source, line number, column number, and exception message.

* Still missing: call stacks. This is seemingly not possible in
  `WKWebView` with a `file://` URL -- the exception object is undefined
  and doesn't provide a stack trace (for that matter, all other JS-side
  debug info is missing). In the future, if we are loading over http(s)
  instead, we could potentially get call stacks on error.

### DIFF
--- a/src/visualization/client/Turi Create Visualization/src/VegaContainer.swift
+++ b/src/visualization/client/Turi Create Visualization/src/VegaContainer.swift
@@ -5,20 +5,49 @@
 
 import WebKit
 
-func log(_ message: String) {
-    let withNewline = String(format: "%@\n", message)
-    guard let data = withNewline.data(using: .utf8) else {
-        assert(false)
-        return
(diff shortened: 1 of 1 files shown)


znation:
Note: this doesn't really fix #2137, so that needs to be deferred to another PR.

PR #2219, opened 2019-08-08
Changed default max_iterations for DC

### LINKED ISSUE #2092: Model output sensitive to Xavier initialization
In some cases Xavier init of the softmax layer for the same model, same classes and same data produced accuracy disparity > 10% across two different trials. It's not clear how much a random seed effects a model's accuracy. 

TC Training - classes axis

TC train repeat

fareeha:
Closes #2092 

- Changed `max_iterations` from 100 to 500.
- Added caveat in user-guide for using more epochs when training model for a larger number of classes.

### DIFF
--- a/userguide/drawing_classifier/advanced-usage.md
+++ b/userguide/drawing_classifier/advanced-usage.md
@@ -18,7 +18,7 @@ If you specify both `max_iterations` and `num_epochs`, `max_iterations` would
 take precedence and the model would train for that number of iterations and the 
 provided value of `num_epochs` would be ignored.
 
-When training a model for a large number of classes, it may be helpful to increase `max_iterations` to allow the model more revisions of the data to learn better.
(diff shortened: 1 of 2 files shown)


userguide/drawing_classifier/advanced-usage.md
```diff
@@ -18,6 +18,8 @@ If you specify both `max_iterations` and `num_epochs`, `max_iterations` would
 take precedence and the model would train for that number of iterations and the 
 provided value of `num_epochs` would be ignored.
 
+When training a model for a large number of classes, it may be helpful to increase `max_iterations` to allow the model more revisions of the data to learn better.
```
nickjong: Let's phrase this advice in a way that makes it easier for the user to tell whether `max_iterations` needs to be higher. Can they look at the metrics output during training?

PR #2227, opened 2019-08-12
unity_sframe::append reorders sframe if needed

Jarvi-Izana:
As it is documented in #2222, I move the logic from the frontend to the backend.

Issue #1977, opened 2019-05-29
Annotation should skip image with missing values 


srikris:
```
<ipython-input-40-956fe9007890> in <module>()
----> 1 data_a = tc.image_classifier.annotate(test_a, 'image')

/Users/srikris/.virtualenvs/turicreate-5.6rc/lib/python2.7/site-packages/turicreate/toolkits/image_classifier/_annotate.pyc in annotate(data, image_column, annotation_column)
    180
    181     with QuietProgress(False):
--> 182         annotation_window.annotate(_get_client_app_path())
    183         return annotation_window.returnAnnotations()
    184

/Users/srikris/.virtualenvs/turicreate-5.6rc/lib/python2.7/site-packages/turicreate/extensions.pyc in <lambda>(*args, **kwargs)
    288         elif name in self._functions:
    289             # is it a function?
--> 290             ret = lambda *args, **kwargs: self.__run_class_function(name, args, kwargs)
    291             ret.__doc__ = "Name: " + name + "\nParameters: " + str(self._functions[name]) + "\n"
    292             try:

/Users/srikris/.virtualenvs/turicreate-5.6rc/lib/python2.7/site-packages/turicreate/extensions.pyc in __run_class_function(self, fnname, args, kwargs)
    275         except RuntimeError as exc:
    276             # Expose C++ exceptions using ToolkitError.
--> 277             raise _ToolkitError(exc)
    278         ret = _wrap_function_return(ret)
    279         return ret

ToolkitError: Cannot resize non-image type
```

znation:
@Jarvi-Izana After discussing with @hoytak, this one makes sense for you to look at while you are learning the codebase -- this will give you some exposure to the visualization/UI backend code.

Jarvi-Izana:
sounds good to me! Happy to take a look at.

Jarvi-Izana:
hi @znation, I'm not familiar with image annotation. Could you introduce me to the typical image annotation process? I'm not sure how I could reproduce this.

PR #2228, opened 2019-08-12
1977 annotation skip invalid images

Jarvi-Izana:
It depends on #2227, which will make `unity_sframe::append` work as the same as what frontend part does.

The new behavior will:
1. prohibit undefined images from being shown in the annotation GUI window.
2. will throw if no valid images there to be ready to render in GUI.
3. will return result-set including all rows having missing values in their `image` column, with the corresponding annotation column value being set as None (undefined).
```
In [6]: images = tc.SFrame("my_image/")
In [6]: images = tc.SFrame("my_image/")

In [7]: images
Out[7]:
Columns:
        path    str
        image   Image

Rows: 2

Data:
+-------------------------------+------------------------+
|              path             |         image          |
+-------------------------------+------------------------+
| /Users/guihaoliang/Work/te... | Height: 412 Width: 430 |
|              None             |          None          |
+-------------------------------+------------------------+
[2 rows x 2 columns]

In [8]: tc.image_classifier.annotate(images)
Out[8]:
Columns:
        path    str
        image   Image
        annotations     str

Rows: 2

Data:
+-------------------------------+------------------------+-------------+
|              path             |         image          | annotations |
+-------------------------------+------------------------+-------------+
| /Users/guihaoliang/Work/te... | Height: 412 Width: 430 |     Cat     |
|              None             |          None          |     None    |
+-------------------------------+------------------------+-------------+
[2 rows x 3 columns]

In [9]: tc.image_classifier.annotate(images[1:2])
<ipython-input-11-05069e9b96e8> in <module>
----> 1 tc.image_classifier.annotate(images[1:2])

~/Work/guicreate-1/debug/src/python/turicreate/toolkits/image_classifier/_annotate.py in annotate(data, image_column, annotation_column)
    111                             data,
    112                             [image_column],
--> 113                             annotation_column
    114                         )
    115     if __platform == "linux" or __platform == "linux2":

~/Work/guicreate-1/debug/src/python/turicreate/extensions.py in <lambda>(*args, **kwargs)
    168
    169 def _make_injected_function(fn, arguments):
--> 170     return lambda *args, **kwargs: _run_toolkit_function(fn, arguments, args, kwargs)
    171
    172 def _class_instance_from_name(class_name, *arg, **kwarg):

~/Work/guicreate-1/debug/src/python/turicreate/extensions.py in _run_toolkit_function(fnname, arguments, args, kwargs)
    157     if not ret[0]:
    158         if len(ret[1]) > 0:
--> 159             raise _ToolkitError(ret[1])
    160         else:
    161             raise _ToolkitError("Toolkit failed with unknown error")

ToolkitError: Expect positive number of images. Found 0 valid images and 1 missing images.
```

### DIFF
--- a/src/visualization/annotation/annotation_base.cpp
+++ b/src/visualization/annotation/annotation_base.cpp
@@ -63,7 +63,7 @@ AnnotationBase::returnAnnotations(bool drop_null) {
   std::shared_ptr<unity_sframe> copy_data;
   {
     // append undefined values
-    if (m_data_na || !m_data_na->size()) {
+    if (m_data_na && m_data_na->size()) {
       auto cast_type = m_data->select_column(m_annotation_column)->dtype();
       auto na_type = m_data_na->select_column(m_annotation_column)->dtype();
       if (cast_type != na_type) {
--- a/src/visualization/annotation/annotation_base.cpp
+++ b/src/visualization/annotation/annotation_base.cpp
@@ -22,7 +22,7 @@ AnnotationBase::AnnotationBase(const std::shared_ptr<unity_sframe> &data,
   /* Copy as so not to mutate the sframe passed into the function */
   m_data = std::static_pointer_cast<unity_sframe>(
       m_data->copy_range(0, 1, m_data->size()));
-  this->_addIndexColumn();
+  _addIndexColumn();
 }
 
 void AnnotationBase::annotate(const std::string &path_to_client) {
@@ -58,11 +58,32 @@ void AnnotationBase::_sendProgress(double value) {
 
 std::shared_ptr<unity_sframe>
 AnnotationBase::returnAnnotations(bool drop_null) {
-  this->cast_annotations();
+  cast_annotations();
 
-  std::shared_ptr<unity_sframe> copy_data =
-      std::static_pointer_cast<unity_sframe>(
+  std::shared_ptr<unity_sframe> copy_data;
+  {
+    // append undefined values
+    if (m_data_na || !m_data_na->size()) {
+      auto cast_type = m_data->select_column(m_annotation_column)->dtype();
+      auto na_type = m_data_na->select_column(m_annotation_column)->dtype();
+      if (cast_type != na_type) {
+        auto data_sarray = std::static_pointer_cast<unity_sarray>(
+            m_data_na->select_column(m_annotation_column));
+        auto integer_annotations =
+            data_sarray->astype(cast_type, true);
+        m_data_na->remove_column(m_data_na->column_index(m_annotation_column));
+        m_data_na->add_column(integer_annotations, m_annotation_column);
+      }
+      copy_data =
+          std::static_pointer_cast<unity_sframe>(m_data->append(m_data_na));
+    } else {
+      copy_data = std::static_pointer_cast<unity_sframe>(
           m_data->copy_range(0, 1, m_data->size()));
+    }
+  }
+  // temporary solution
+  copy_data = std::static_pointer_cast<unity_sframe>(
+      copy_data->sort({"__idx"}, {true}));
 
   size_t id_column = copy_data->column_index("__idx");
   copy_data->remove_column(id_column);
@@ -73,18 +94,19 @@ AnnotationBase::returnAnnotations(bool drop_null) {
   if (!drop_null) {
     annotation_global->annotation_sframe = copy_data;
     return copy_data;
-  }
+    }
 
-  std::vector<std::string> annotation_column_name = {m_annotation_column};
-  std::list<std::shared_ptr<unity_sframe_base>> dropped_missing =
-      copy_data->drop_missing_values(annotation_column_name, false, false, false);
+    std::vector<std::string> annotation_column_name = {m_annotation_column};
+    std::list<std::shared_ptr<unity_sframe_base>> dropped_missing =
+        copy_data->drop_missing_values(annotation_column_name, false, false,
+                                       false);
 
-  std::shared_ptr<unity_sframe> final_sf =
-      std::static_pointer_cast<unity_sframe>(dropped_missing.front());
+    std::shared_ptr<unity_sframe> final_sf =
+        std::static_pointer_cast<unity_sframe>(dropped_missing.front());
 
-  annotation_global->annotation_sframe = final_sf;
+    annotation_global->annotation_sframe = final_sf;
 
-  return final_sf;
+    return final_sf;
 }
 
 std::shared_ptr<annotation_global> AnnotationBase::get_annotation_registry() {
@@ -95,20 +117,18 @@ std::shared_ptr<annotation_global> AnnotationBase::get_annotation_registry() {
 
 size_t AnnotationBase::size() { return m_data->size(); }
 
+/* must be called after _addIndexColumn */
+void AnnotationBase::splitUndefined(const std::string& column_names, bool how, bool recursive) {
+  DASSERT_TRUE(m_data->contains_column("__idx"));
+  const auto &image_column_name = m_data_columns.at(0);
+  auto split_data_set =
+      m_data->drop_missing_values({image_column_name}, how, true, recursive);
+  m_data = std::static_pointer_cast<unity_sframe>(split_data_set.front());
+  m_data_na = std::static_pointer_cast<unity_sframe>(split_data_set.back());
+}
+
 void AnnotationBase::_addIndexColumn() {
-  std::vector<flexible_type> indicies;
-
-  for (size_t x = 0; x < m_data->size(); x++) {
-    indicies.push_back(x);
-  }
-
-  std::shared_ptr<unity_sarray> empty_annotation_sarray =
-      std::make_shared<unity_sarray>();
-
-  empty_annotation_sarray->construct_from_vector(indicies,
-                                                 flex_type_enum::INTEGER);
-
-  m_data->add_column(empty_annotation_sarray, "__idx");
+  m_data->add_column(unity_sarray::create_sequential_sarray(m_data->size(), 0, false), "__idx");
 }
 
 void AnnotationBase::_reshapeIndicies(size_t &start, size_t &end) {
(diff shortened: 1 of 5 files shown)


Jarvi-Izana:
Big thanks to @abhishekpratapa for explaining the annotation code design.

src/visualization/annotation/annotation_base.cpp
```diff
@@ -95,20 +117,18 @@ std::shared_ptr<annotation_global> AnnotationBase::get_annotation_registry() {
...
+  DASSERT_TRUE(m_data->contains_column("__idx"));
+  const auto &image_column_name = m_data_columns.at(0);
+  auto split_data_set =
+      m_data->drop_missing_values({image_column_name}, how, true, recursive);
+  m_data = std::static_pointer_cast<unity_sframe>(split_data_set.front());
+  m_data_na = std::static_pointer_cast<unity_sframe>(split_data_set.back());
+}
 
-  m_data->add_column(empty_annotation_sarray, "__idx");
+void AnnotationBase::_addIndexColumn() {
```
Jarvi-Izana: more efficient impl of constructing seq.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -3778,48 +3778,8 @@ def append(self, other):
         if type(other) is not SFrame:
             raise RuntimeError("SFrame append can only work with SFrame")
```
Jarvi-Izana: code change for #2227; irrelevant to this PR; Feel free to ignore.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -426,9 +426,9 @@ std::shared_ptr<unity_sarray_base> unity_sframe::select_column(const std::string
   size_t column_index = _column_index_iter - _column_names.begin();
   auto ret = select_column(column_index);
```
Jarvi-Izana: code change for #2227; irrelevant to this PR; Feel free to ignore.

src/visualization/annotation/annotation_base.cpp
```diff
@@ -58,11 +58,32 @@ void AnnotationBase::_sendProgress(double value) {
...
+      if (cast_type != na_type) {
+        auto data_sarray = std::static_pointer_cast<unity_sarray>(
+            m_data_na->select_column(m_annotation_column));
+        auto integer_annotations =
+            data_sarray->astype(cast_type, true);
+        m_data_na->remove_column(m_data_na->column_index(m_annotation_column));
+        m_data_na->add_column(integer_annotations, m_annotation_column);
+      }
+      copy_data =
+          std::static_pointer_cast<unity_sframe>(m_data->append(m_data_na));
```
Jarvi-Izana: `append` requires the change #2227 to behave correctly.

src/core/storage/sframe_interface/unity_sframe.cpp
znation: Can you rebase your changes against master? It's better (for reviewers) if you put up independent PRs. Thanks!

znation: Nevermind, I see you actually have a dependency here.

test/annotation/image_classifier_tests.cxx
```diff
@@ -346,7 +346,7 @@ struct image_classification_test {
 
     TS_ASSERT(annotation_testing::check_equality(annotation_sf, returned_sf));
 
-    
+
```
znation: Since this is only a whitespace change, please revert the changes in this file.

userguide/image_classifier/advanced-usage.md
```diff
@@ -20,7 +20,12 @@ annotated_data = tc.image_classifier.annotate(data)
 
 ```
 
-If you forget to assign the output of your annotation to a variable, 
+This utility will **only collect valid images**. All undefined values
```
znation: I'm not sure what "only collect valid images" means.

How about "only present rows from the SFrame that have non-missing images"? And I don't think I would make it bold... as a user, I would not think it's too surprising or needs to be taken note of, just good to know.

src/visualization/annotation/annotation_base.cpp
```diff
@@ -58,11 +58,32 @@ void AnnotationBase::_sendProgress(double value) {
...
+    } else {
+      copy_data = std::static_pointer_cast<unity_sframe>(
           m_data->copy_range(0, 1, m_data->size()));
+    }
+  }
+  // temporary solution
```
znation: What is temporary about this solution? What problem is it solving? Why is the solution temporary?

Or if this comment is outdated, please remove it.

src/visualization/annotation/annotation_base.cpp
```diff
@@ -58,11 +58,32 @@ void AnnotationBase::_sendProgress(double value) {
...
+  std::shared_ptr<unity_sframe> copy_data;
+  {
+    // append undefined values
+    if (m_data_na && m_data_na->size()) {
+      auto cast_type = m_data->select_column(m_annotation_column)->dtype();
+      auto na_type = m_data_na->select_column(m_annotation_column)->dtype();
+      if (cast_type != na_type) {
+        auto data_sarray = std::static_pointer_cast<unity_sarray>(
+            m_data_na->select_column(m_annotation_column));
+        auto integer_annotations =
```
znation: These aren't necessarily integer annotations, right? Couldn't they be string annotations (depending on cast_type)?

src/visualization/annotation/annotation_base.cpp
```diff
@@ -73,18 +94,19 @@ AnnotationBase::returnAnnotations(bool drop_null) {
...
     annotation_global->annotation_sframe = copy_data;
     return copy_data;
-  }
+    }
 
-  std::vector<std::string> annotation_column_name = {m_annotation_column};
-  std::list<std::shared_ptr<unity_sframe_base>> dropped_missing =
-      copy_data->drop_missing_values(annotation_column_name, false, false, false);
+    std::vector<std::string> annotation_column_name = {m_annotation_column};
+    std::list<std::shared_ptr<unity_sframe_base>> dropped_missing =
```
znation: I'm not sure I understand the overall approach here. Why do we drop missing values at all, only to add them back in upon return? Wouldn't it be simpler to send all rows (including those with missing data) and receive all rows? We don't necessarily need to display those rows on the client, in order to send and receive them.

If this approach was considered and rejected, please add a comment here as to why; if not, I think we should consider it.

src/visualization/annotation/annotation_base.hpp
```diff
@@ -90,6 +90,9 @@ class EXPORT AnnotationBase : public ml_model_base {
 
   virtual void checkDataSet() = 0;
 
+  /* must be called after construction */
```
znation: If it's intended to be called once, and only from the constructor, I wouldn't put it in the public API. It should ideally be private in that case and non-virtual (or protected and virtual, if subclass constructors need to call it directly instead of through their superclass constructor).

test/annotation/image_classifier_tests.cxx
Jarvi-Izana: Yeah, I could. IMO, it can keep the source pkg smaller XD. Besides, you can set ignore whitespaces change when you do code review. 
https://github.blog/2018-05-01-ignore-white-space-in-code-review/
@znation

src/visualization/annotation/annotation_base.cpp
Jarvi-Izana: Yeah, but the first line `cast` will try to probe whether the annotation can be convertible to integer or not. This is just to align with it.

src/visualization/annotation/annotation_base.cpp
Jarvi-Izana: Haha, I forget to change its position.

Well, the temp solution is to sort by `__idx` column to restore the original order, which is not that efficient. But in the future, if we hit any perf issue, we should come up with a solution that we don't have to split the image set into defined part and undefined part. We can do all operations inline. Since the design by @abhishekpratapa requires all images feeding into neural network to be valid, we come up with this **temp** solution.

userguide/image_classifier/advanced-usage.md
Jarvi-Izana: Oh, sure.

src/visualization/annotation/annotation_base.hpp
Jarvi-Izana: That's a good point! I should provide it as a private virtual.

src/visualization/annotation/annotation_base.cpp
Jarvi-Izana: Sure. It's explained just now in previous threads.

In short, we can do everything inline but it's tricky and requires non-trivial work. Maybe @abhishekpratapa can explain it more. So we designed this workaround with the consideration that annotation is done by humans and each batch job won't be that large (less than 1000 pics per batch) for them. So split the data set won't be the bottle neck.

src/visualization/annotation/annotation_base.hpp
Jarvi-Izana: Ah, I just realized it's a mistake to call virtual function during construction. I will make it non-virtual protected.

Jarvi-Izana: it's tricky to have virtual functions involved in construction. I will make it non-virtual for this case.

Jarvi-Izana: @znation as is recorded in #2234. Will be a feature request for @abhishekpratapa. For now, it's okay since base class is pure virtual and derived class directly inherits base. If the inheritance chain grows, then we should consider changing the code structure first. For now, it's irrelevant to this pr.

Jarvi-Izana:
referenced by #1977

znation:
Please rebase against latest master; this PR now has changes in it that (seem to be) already in master.

Jarvi-Izana:
@znation rebased.

Jarvi-Izana:
Once it passes the internal CI, I will merge it.

PR #2239, opened 2019-08-14
Drawing assertions

shantanuchhabra:


src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
     for (int dx = -pad; dx < pad; dx++) {
         for (int dy = -pad; dy < pad; dy++) {
-            if (in_bounds(x+dx, y+dy, dimension)) {
-                bitmap[(y+dy) * dimension + (x+dx)] = 1.0;
+            if (in_bounds(x + dx, y + dy, x_dimension, y_dimension)) {
+                DASSERT_GE(x + dx, 0);
+                DASSERT_LT(x + dx, x_dimension);
+                DASSERT_GE(y + dy, 0);
+                DASSERT_LT(y + dy, y_dimension);
+#ifndef NDEBUG
```
shantanuchhabra: @hoytak Do we need an #ifndef NDEBUG ... #endif here because `indices` might be an unused variable in `release`?

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -172,7 +173,7 @@ flex_list ramer_douglas_peucker(
     return compressed_stroke;
 }
 
-flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
+flex_list simplify_drawing(flex_list &raw_drawing, int row_number) {
```
shantanuchhabra: Can't use a `const flex_list &` here since we make calls to `mutable_get` on elements of `raw_drawing`.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
-    } else {
-        slope = (end.get_y() - start.get_y())/(end.get_x() - start.get_x());
-    }
-    int pad = (int)(stroke_width/2);
-    along_x = (fabs(slope) < 1);
-    if ((along_x && (start.get_x() > end.get_x()))
-        || (!along_x && (start.get_y() > end.get_y()))) {
-        std::swap(start, end);
+flex_nd_vec paint_stroke(flex_nd_vec &bitmap, Point &start, Point &end,
+                         float stroke_width) {
```
shantanuchhabra: All parameters here are being modified in this function, so we can't use a const type& here.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
+  bool along_x;
+  float slope;
+  if (floorf(end.get_x()) == floorf(start.get_x())) {
+    slope = std::numeric_limits<float>::max();
+  } else {
+    // Make sure denominator of slope is non-zero, not necessarily positive.
+    DASSERT_NE(end.get_x() - start.get_x(), 0);
```
shantanuchhabra: Since either `end.get_x()` or `start.get_x()` could be greater than the other but the difference must be non-zero (for the division in the next line to not be undefined), we check for non-zeroness only.

src/toolkits/drawing_classifier/data_preparation.cpp
hoytak: Can you add a comment as to how exactly raw_drawing is changed?  If it's not changed, this should be const flex_list& and the rest should be modified as needed. 

Hmm.  don't actually see any reason to use mutable_get below.  mutable_get should only be used when the contents are explicitly being modified, but it seems the access is purely read-only.

src/toolkits/drawing_classifier/data_preparation.cpp
hoytak: Yes, likely.

src/toolkits/drawing_classifier/data_preparation.cpp
hoytak: How are start and end being modified?

src/toolkits/drawing_classifier/data_preparation.cpp
hoytak: DASSERT_NE(end.get_x(), start.get_x()) would be clearer.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
-        slope = (end.get_y() - start.get_y())/(end.get_x() - start.get_x());
-    }
-    int pad = (int)(stroke_width/2);
-    along_x = (fabs(slope) < 1);
-    if ((along_x && (start.get_x() > end.get_x()))
-        || (!along_x && (start.get_y() > end.get_y()))) {
-        std::swap(start, end);
+flex_nd_vec paint_stroke(flex_nd_vec &bitmap, Point &start, Point &end,
+                         float stroke_width) {
+  bool along_x;
```
hoytak: Good habit -- set all variables on initialization to a default.  No harm in it and guards against initialization errors.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
+  }
+  int pad = (int)(stroke_width / 2);
+  along_x = (fabs(slope) < 1);
+  if ((along_x && (start.get_x() > end.get_x())) ||
+      (!along_x && (start.get_y() > end.get_y()))) {
+    std::swap(start, end);
+  }
+  DASSERT_LE(start.get_x(), end.get_x());
+  DASSERT_LE(start.get_y(), end.get_y());
+  int x1 = (int)(start.get_x());
```
hoytak: Use std::floor, std::ceil, or std::round to make it clear what's going on with this casting.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
+  int y1 = (int)(start.get_y());
+  int x2 = (int)(end.get_x());
+  int y2 = (int)(end.get_y());
+  if (along_x) {
+    for (int x = x1; x <= x2; x++) {
+      int y = (int)(slope * (x - x1) + y1);
```
hoytak: Generally, with this kind of geometry it's better to do all math in doubles until the very end, where you convert it to the pixel values.  This version has a bunch of numerical dragons here that could easily be problematic -- rounding errors on the slope vs x - x1 could cause some weird out-of-bounds things in the bitmap.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
-            paint_point(bitmap, x, y, pad);
-        }
-    } else {
-        for (int y = y1; y <= y2; y++) {
-            int x = (int)(x1 + ((y - y1) / slope));
-            paint_point(bitmap, x, y, pad);
-        }
+  } else {
+    for (int y = y1; y <= y2; y++) {
+      int x = (int)(x1 + ((y - y1) / slope));
```
hoytak: This is quite prone to numerical instability with the slope.  Again, it'd be less error prone to do all math in doubles and then do one rounding at the end.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -230,84 +231,132 @@ flex_list simplify_drawing(flex_list raw_drawing, int row_number) {
...
+  if ((along_x && (start.get_x() > end.get_x())) ||
+      (!along_x && (start.get_y() > end.get_y()))) {
+    std::swap(start, end);
+  }
+  DASSERT_LE(start.get_x(), end.get_x());
+  DASSERT_LE(start.get_y(), end.get_y());
+  int x1 = (int)(start.get_x());
+  int y1 = (int)(start.get_y());
+  int x2 = (int)(end.get_x());
+  int y2 = (int)(end.get_y());
```
hoytak: Generally speaking, it's good practice to use static_cast<int> and avoid the C-style casts.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -422,7 +471,8 @@ flex_image rasterize(flex_list simplified_drawing) {
 }
 
 flex_image convert_stroke_based_drawing_to_bitmap(
-    flex_list stroke_based_drawing, int row_number) {
+    flex_list &stroke_based_drawing, int row_number) {
```
hoytak: Again, not sure why this can't be const flex_list&

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -390,7 +439,7 @@ flex_image rasterize_on_mac(flex_list simplified_drawing) {
 }
 #endif // __APPLE__
 
-flex_image rasterize(flex_list simplified_drawing) {
+flex_image rasterize(flex_list &simplified_drawing) {
```
hoytak: Needs to either be const flex_list& or have documentation explaining how it's changed.  It's not clear to me why it is changed.

Issue #2153, opened 2019-07-21
Front-end UI testing infrastructure


znation:
We have no way to author and run automated UI tests in a consistent way. We should adopt a framework and integrate it in our codebase so we can author and run:

* Web front-end UI tests (in a browser or browser-like enviroment).
* macOS native UI tests? (Probably unnecessary.)
* Linux native UI tests? (Probably unnecessary.)

PR #2243, opened 2019-08-15
Browser target for SFrame.explore

### LINKED ISSUE #2240: SFrame.explore should support browser target

znation:
* Adds support for the `browser` target for SFrame explore (table view).
* Adds proper data handling on the server, using POST requests and
  `boost::property_tree` for JSON decoding.
* Supports scrolling, jump to row, and accordion expansion.
* Refactors the table spec & data functions out of unity_sframe into
  visualization/server/table.
* Refactors empty_tz construction out of unity_sframe and into static
  local initialization within the escaping functions.
* Refactors image hover to load over HTTP from localhost, for both GUI
  and browser target explore. This appears to be much faster than the
  server queue/push-based approach, since now the client handles its own
  request queue and the server can respond on multiple threads.

Fixes #2240


src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1832,323 +1831,29 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
...
-    };
+    // Get a reader just once.
+    auto reader = underlying_sframe->get_reader();
 
-    // pass the first 1k rows
-    getRows(0, 100);
+    ew << visualization::table_data(self, reader.get(), 0, 100);
 
-    const size_t resized_height = 200;
+    //const size_t resized_height = 200;
```
fareeha: Should this line be removed?

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1832,323 +1831,29 @@ void unity_sframe::explore(const std::string& path_to_client, const std::string&
...
-        DASSERT_LT(index, self->size());
-        DASSERT_GE(index, 0);
+    std::stringstream table_spec;
+    table_spec << "{\"table_spec\":" << visualization::table_spec(self, title) << "}" << std::endl;
+    ew << table_spec.str();
 
-        auto accordion_sa = self->select_column(column_name);
-        auto gl_sa = gl_sarray(accordion_sa);
+    // This materializes if not already
+    auto underlying_sframe = self->get_underlying_sframe();
```
TobyRoseman: `auto` -> `sframe`

src/core/storage/sframe_interface/unity_sframe.cpp
znation: Yes, good catch.

src/core/storage/sframe_interface/unity_sframe.cpp
znation: Thanks, good catch. I think I was lazily copy/pasting.

znation:
> Is there any way we can add unit tests for any of this?

Good question. I'll look into it. We don't have any infrastructure yet (tracked by https://github.com/apple/turicreate/issues/2153) but it might not be too hard to use something built into Python or already in the Python ecosystem, now that our web UI is exposed in a regular browser (not just custom GUI), and the protocol is regular HTTP (not just custom IPC).

znation:
After some initial investigation, it looks like it'll be non-trivial to set up a browser test environment and some package that allows us to automate the browser. For now I'd like to merge this PR without unit test coverage, and I have issues tracking:

* #2153 - web front-end test infrastructure
* #2247 - SFrame.explore should have UI tests
* #2248 - SFrame.show should have UI tests

I'll also see if there is anything on the C++ side we can unit test independently.

PR #2245, opened 2019-08-15
Style Transfer Backend Code

abhishekpratapa:
- Corrected un-initalized code in the style transfer descriptors
- Added a `model_backend` implementation for the Style Transfer network.
- Tested weight initialization for the training and inference graph

### DIFF
--- a/src/ml/neural_net/mps_descriptor_utils.m
+++ b/src/ml/neural_net/mps_descriptor_utils.m
@@ -10,7 +10,7 @@
 - (instancetype) init {
   self = [super init];
   if (self) {
-    _label = [NSString alloc];
+    _label = [[NSString alloc] init];
   }
   return self;
 }
@@ -20,7 +20,7 @@
 - (instancetype) init {
   self = [super init];
   if (self) {
-    _label = [NSString alloc];
+    _label = [[NSString alloc] init];
   }
   return self;
 }
--- a/src/ml/neural_net/mps_descriptor_utils.m
+++ b/src/ml/neural_net/mps_descriptor_utils.m
@@ -7,9 +7,23 @@
 #import <ml/neural_net/mps_descriptor_utils.h>
 
 @implementation TCMPSConvolutionDescriptor
+- (instancetype) init {
+  self = [super init];
+  if (self) {
+    _label = [NSString alloc];
+  }
+  return self;
+}
 @end
 
 @implementation TCMPSInstanceNormalizationDescriptor
+- (instancetype) init {
+  self = [super init];
+  if (self) {
+    _label = [NSString alloc];
+  }
+  return self;
+}
 @end
 
 @implementation TCMPSUpsamplingDescriptor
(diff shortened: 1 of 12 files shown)


src/ml/neural_net/mps_descriptor_utils.m
```diff
@@ -7,9 +7,23 @@
 #import <ml/neural_net/mps_descriptor_utils.h>
 
 @implementation TCMPSConvolutionDescriptor
+- (instancetype) init {
+  self = [super init];
+  if (self) {
+    _label = [NSString alloc];
```
nickjong: Need to init here. You almost never want to do just `alloc` by itself

src/ml/neural_net/mps_layer_helper.mm
```diff
@@ -46,7 +46,7 @@ + (MPSCNNFullyConnectedNode *) createFullyConnected:(MPSNNImageNode *)inputNode
     [MPSCNNFullyConnectedNode nodeWithSource:inputNode
                                      weights:fullyConnectedDataLoad];
 
-  fullyConnectedNode.weights = fullyConnectedDataLoad;
+  // fullyConnectedNode.weights = fullyConnectedDataLoad;
```
nickjong: Why are these commented out?

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -44,30 +44,25 @@ @interface TCMPSStyleTransfer()
...
-                          numStyles:(NSUInteger)numStyles {
+
+- (instancetype) initWithDev:(id<MTLDevice>) dev
+                commandQueue:(id<MTLCommandQueue>) commandQueue
+                     weights:(NSDictionary<NSString *, NSData *> *) weights
+                   numStyles:(NSUInteger) numStyles {
   self = [super init];
   if (self) {
-    // Set Default parameters values
+    // Allocate a _dev and _commandQueue
```
nickjong: Nit: You're not actually allocating anything here. Allocation typically implies new memory is being appropriated from the system

src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
```diff
@@ -0,0 +1,44 @@
...
+#define TURI_STYLE_TRANSFER_STYLE_TRANSFER_H_
+
+#include <functional>
+#include <map>
+#include <memory>
+
+#include <ml/neural_net/float_array.hpp>
+#include <ml/neural_net/model_backend.hpp>
+#include <ml/neural_net/model_spec.hpp>
+#include <ml/neural_net/mps_device_manager.h>
```
nickjong: This is an Obj C import, which you probably don't want in this C++ header?

src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+using namespace turi::neural_net;
```
nickjong: Never use `using` declarations at the top level in headers. This will affect anyone who recursively includes this header, which may surprise and confuse others

src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace turi {
+namespace style_transfer {
+
+class EXPORT style_transfer : public model_backend {
+public:
+  style_transfer(const float_array_map &config,
+                 const float_array_map &weights);
+  
+  ~style_transfer() {}
```
nickjong: Nit: in C++11, prefer
```
~style_transfer() = default;
```

src/ml/neural_net/style_transfer/mps_style_transfer_backend.mm
```diff
@@ -0,0 +1,145 @@
...
++(float_array_map) fromNSDictionary: (NSDictionary<NSString *, NSData *> *) dictionary {
+  float_array_map map;
+  for (NSString* key in dictionary) {
+    NSData* value = [dictionary objectForKey:key];
+
+    size_t dataLength = (size_t) (value.length / sizeof(float));
+    std::vector<size_t> dataShape = { dataLength }; 
+    float *dataBytes = (float *) (value.bytes);
+    
+    shared_float_array array = shared_float_array::copy(dataBytes, dataShape);
```
nickjong: Could you leave a TODO here to optimize away this copy? Also worth filing a GH issue for the convergence sprint. This can instead be a wrapper around the NSData: a custom subclass of `float_array` that preserves a strong reference to the NSData instance

src/ml/neural_net/style_transfer/mps_style_transfer_backend.mm
```diff
@@ -0,0 +1,145 @@
...
+
+    map.insert(std::pair<std::string, shared_float_array>([key UTF8String], array));
```
nickjong: Nit: in C++11, prefer
```
map.emplace(key.UTF8String, array);
```

src/ml/neural_net/style_transfer/mps_style_transfer_backend.mm
```diff
@@ -0,0 +1,145 @@
...
++(NSDictionary<NSString *, NSData *> *) toNSDictionary: (float_array_map) map {
+  NSMutableDictionary<NSString *, NSData *> * dictionary 
+      = [[NSMutableDictionary alloc] init];
+
+  for (const auto &element : map) {
+    const float* elementData = element.second.data();
+    const size_t elementSize = element.second.size() * sizeof(float);
+
+    NSMutableData * mutableElementData = [NSMutableData dataWithLength:elementSize];
+    memcpy(mutableElementData.mutableBytes, elementData, elementSize);
```
nickjong: For now, these two lines should just be
```
NSData *data = [NSData dataWithBytes:elementData length:elementSize];
```

Let's also add a TODO here. The issue you file for wrapping `NSData` with a `float_array` subclass can go the other way, too. We can construct an `NSData` that wraps a `shared_float_array`, using this API: https://developer.apple.com/documentation/foundation/nsdata/1417337-initwithbytesnocopy?language=objc

Let's not do this now, since it's not on the critical path to testing the MPS implementation. But we should avoid these copies in general before we release.

src/ml/neural_net/style_transfer/mps_style_transfer_decoding_node.mm
```diff
@@ -28,11 +28,12 @@ - (instancetype) initWithParameters:(NSString *)name
   self = [super init];
   
   if (self) {
-  	_upsample = [MPSCNNUpsamplingNearestNode nodeWithSource:inputNode
+    _upsample = [MPSCNNUpsamplingNearestNode nodeWithSource:inputNode
                                        integerScaleFactorX:descriptor.upsample.scale
                                        integerScaleFactorY:descriptor.upsample.scale];
 
-  	_conv = [MPSCNNConvolutionNode createConvolutional:[_upsample resultImage]
+    NSMutableData* convBiases = [NSMutableData dataWithLength:descriptor.conv.outputFeatureChannels*sizeof(float)];
```
nickjong: Why are we just creating a zeroed out `NSData` here?

src/ml/neural_net/style_transfer/mps_style_transfer_encoding_node.mm
```diff
@@ -27,6 +27,7 @@ - (instancetype) initWithParameters:(NSString *)name
   self = [super init];
   
   if (self) {
+    NSMutableData* convBiases = [NSMutableData dataWithLength:descriptor.conv.outputFeatureChannels*sizeof(float)];
```
nickjong: And here?

src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
abhishekpratapa: Thanks! That actually solved another problem! Genius.

src/ml/neural_net/style_transfer/mps_style_transfer_backend.mm
abhishekpratapa: Simple. Thanks!

src/ml/neural_net/mps_layer_helper.mm
```diff
@@ -9,8 +9,18 @@
 #import <ml/neural_net/mps_layer_instance_norm_data_loader.h>
 #import <ml/neural_net/mps_weight.h>
 
+#import <objc/runtime.h>
+
 @implementation MPSCNNFullyConnectedNode (TCMPSLayerHelper)
-@dynamic weights;
+
+- (void)setWeights:(TCMPSConvolutionWeights *)weights {
+    objc_setAssociatedObject(self, @selector(tag), weights, OBJC_ASSOCIATION_RETAIN);
```
nickjong: The key used for associated objects must be unique. What is `tag` in this case? How can you be sure no one else will use it?

I would recommend the officially endorsed pattern:
```
static char kAssociatedObjectKey;
```
And use `&kAssociatedObjectKey` as the key for these objc functions

abhishekpratapa: Ah yes, that `tag` was supposed to be `weights`! I did look into the `kAssociatedObjectKey` pattern according to https://nshipster.com/associated-objects/. It was suggested that the selector pattern was better.

nickjong: AFAICT, `@selector` only gives a unique value if no one else uses the same selector name. So if you use `tag` or `weights`, anyone else who uses the same name could in theory collide. It's safer to use a name that at least includes a TC prefix, but at that point I figure why not go the guaranteed route and use a static variable.

Although maybe it's reasonable to apply the same standard to the category property itself. Maybe it should be `tc_weights` to avoid conflicts with any other category property that somebody else might add. Then I think it would be reasonable to use `@selector(tc_weights)`

abhishekpratapa: I'm not sure I understand. Is this a per @instance thing or a per @implementation thing. Would the static variable collide in the multiple implementations?

nickjong: The main thing to understand is that `objc_setAssociatedObject` is just a global function allowing you to specify values for any pair of object instance and "key", where the "key" is just a `void*` value. It's global, so any other code in the same process can call the same function to access the values associated with a given object instance. The safest way to guarantee that some other code doesn't use the same key in a different way is to use a key that no one else knows about. Using `@selector` only satisfies this requirement if the name of the selector is something no one else knows about. A static variable by definition is unique, even if another .m or .cpp file uses the exact same name.

abhishekpratapa:
WAHOO!!!!!!

PR #2251, opened 2019-08-19
Fixed namespace conflicts Obj-C

abhishekpratapa:
- weights key conflicts with property already existent on `MPSCNNConvolutionNode`

### DIFF
--- a/src/ml/neural_net/mps_layer_helper.h
+++ b/src/ml/neural_net/mps_layer_helper.h
@@ -15,7 +15,7 @@ NS_ASSUME_NONNULL_BEGIN
 @class TCMPSInstanceNormDataLoader;
 
 @interface MPSCNNFullyConnectedNode (TCMPSLayerHelper)
-@property (nonatomic, strong) TCMPSConvolutionWeights *tcWeightsData API_AVAILABLE(macosx(10.14));
+@property (nonatomic, strong) TCMPSConvolutionWeights *tc_weightsData API_AVAILABLE(macosx(10.14));
 + (TCMPSConvolutionWeights *) createFullyConnected:(MPSNNImageNode *)inputNode
                               inputFeatureChannels:(NSUInteger)inputFeatureChannels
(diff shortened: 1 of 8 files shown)


src/ml/neural_net/mps_layer_helper.h
```diff
@@ -15,7 +15,7 @@ NS_ASSUME_NONNULL_BEGIN
 @class TCMPSInstanceNormDataLoader;
 
 @interface MPSCNNFullyConnectedNode (TCMPSLayerHelper)
-@property (nonatomic, strong) TCMPSConvolutionWeights *weights API_AVAILABLE(macosx(10.14));
+@property (nonatomic, strong) TCMPSConvolutionWeights *tcWeightsData API_AVAILABLE(macosx(10.14));
```
nickjong: Let's just go ahead and follow the officially documented naming guidelines, which would yield something like `tc_weights` or `tc_weightsData`. See "Avoiding Category Method Name Clashes" here: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html

src/ml/neural_net/style_transfer/mps_style_transfer_decoding_node.mm
```diff
@@ -77,27 +77,27 @@ - (MPSNNImageNode *) backwardPass:(MPSNNImageNode *) inputNode {
 }
 
 - (void) setLearningRate:(float)lr {
-  [[_conv weights] setLearningRate:lr];
-  [[_instNorm weights] setLearningRate:lr];
+  [[_conv tcWeightsData] setLearningRate:lr];
```
nickjong: Nit: This statement, and others like it, would be more readable as e.g.
```
[_conv.tc_weights setLearningRate:lr];
```
May as well actually use the property syntax since we've defined our category methods as a property

PR #2259, opened 2019-08-21
Mac OS 10.14 Compile Time Check

abhishekpratapa:
- changed Mac OS version checking from runtime to compile time check


nickjong:
First, I wouldn't think of this as replacing the runtime check with the compile-time check. We need both. This toolkit will only compile if the base SDK supports 10.15 or later. For the built Turi Create, that Python package can only support the toolkit when run on a Mac running 10.15 or later.

Second, why confine the compile-time check to the specific Obj-C interfaces here? Do we expect the `style_transfer` class to be useful on 10.14? Could we just use one #ifdef here for the entire .mm file, and perhaps one more wherever we actually instantiate it? Seems more robust

PR #2210, opened 2019-08-06
groupby optimization

Jarvi-Izana:
provide the benchmark regards to #2175 

Running against 1M rows with binary category data. The first half is all 1, the rest is 0. Each test will repeat ~100~ >= 3 times. The aggregation is only using **count** for proof of concept purpose. If this works really well, then further change will apply to other built-in group-by operators.

---

A brief summary of improvements in v2:
1 thread, 16% faster.
2 threads, 9% faster.
4 threads, 32% faster.
8 threads,  53 % faster.
12 threads (max concurrency on my machine), 72% faster.

A brief summary of improvements in v1:
1 thread, 6% faster.
2 threads, 2% faster.
4 threads, 19% faster.
8 threads,  40% faster.
12 threads (max concurrency on my machine), 64% faster.

---

**without** any optimization enabled:
```
nrows: 1000000
reps: 100
=============== bench_test_aggreate_count ================
nthreads: 1
Elapsed time: 68031.1 ms
Average single pass: 680.311 ms
========================== END ===========================
nthreads: 2
Elapsed time: 35193.7 ms
Average single pass: 351.937 ms
========================== END ===========================
nthreads: 4
Elapsed time: 25837 ms
Average single pass: 258.37 ms
========================== END ===========================
nthreads: 8
Elapsed time: 30760.8 ms
Average single pass: 307.608 ms
========================== END ===========================
nthreads: 12
Elapsed time: 43284.2 ms
Average single pass: 432.842 ms
========================== END ===========================

```

---

v2 optimization:
a finite set of cyclic local buffers (resolve the file handler hog) + uplifting first local buffer as the global buffer (saves effort to open a new buffer as global one) + skipping empty segments (avoid unnecessary context switches)
```
nrows: 1000000
reps: 3
=========== bench_test_aggreate_count summary ============
nthreads: 1
Elapsed time: 1723.51 ms
Average single pass: 574.507 ms
========================== END ===========================
nthreads: 2
Elapsed time: 956.427 ms
Average single pass: 318.811 ms
========================== END ===========================
nthreads: 4
Elapsed time: 524.067 ms
Average single pass: 174.691 ms
========================== END ===========================
nthreads: 8
Elapsed time: 435.011 ms
Average single pass: 145.006 ms
========================== END ===========================
nthreads: 12
Elapsed time: 365.199 ms
Average single pass: 121.735 ms
========================== END ===========================
```

---

v1 optimization:
[...]

PR #2260, opened 2019-08-22
benchmark for groupby operation

Jarvi-Izana:
#2210 pr is too large and it contains too many topics and therefore it's hard to be reviewed.

I separate benchmark test code from #2210 in order to polish it first and then share it among other branches. 

Let's make it into master first, and then I can start bench testing other ideas of optimization by branching from master to share the same benchmark test code.


Jarvi-Izana:
If we have more benchmark code, then we need to move the `benchmark` dir under top-level dir. Currently, I don't think we need.

test/benchmark/sframe_bench_aggregate.cpp
```diff
@@ -0,0 +1,287 @@
...
+#include <thread>
+#include <random>
+
+namespace turi {
+
+using row_gen_func_t = std::function<std::vector<flexible_type>(size_t)>;
+
+/* generate sframe */
+/* I need to make this a fixture */
+static sframe bench_test_sframe_generator(
```
hoytak: Use random sframe generator code in <core/storage/sframe_data/testing_utils.hpp> -- that would be easier to use. We don't want to duplicate code if we have good reason not to.

hoytak:
I think this is ready -- want to merge it?

Jarvi-Izana:
I still have some changes. Let me clean them up and then we are good to go!

Jarvi-Izana:
Refactored my code and it should be very easy to add more benchmark test cases.

sample output. A force update is due to I mistakenly amended all my changes to a previous commit.
```
=========== count on binary categorical data =============

nrows: 100000
reps: 3
avg time to run w/  2 threads: 42.1337 ms.
avg time to run w/  4 threads: 30.247 ms.
avg time to run w/  8 threads: 28.779 ms.
avg time to run w/ 12 threads: 26.7163 ms.
========================== END ===========================

=========== bench_test_aggreate_min summary ============

nrows: 100000
reps: 3
users: 1000
avg time to run w/  2 threads: 187.639 ms.
avg time to run w/  4 threads: 163.525 ms.
avg time to run w/  8 threads: 154.321 ms.
avg time to run w/ 12 threads: 145.579 ms.
========================== END ===========================

=========== bench_test_aggreate_avg summary ============

nrows: 100000
reps: 3
users: 1000
avg time to run w/  2 threads: 180.494 ms.
avg time to run w/  4 threads: 158.512 ms.
avg time to run w/  8 threads: 156.701 ms.
avg time to run w/ 12 threads: 146.346 ms.
========================== END ===========================
```

Jarvi-Izana:
ready to go! @hoytak

Jarvi-Izana:
added data generator for skewed data distribution. 

It will be later used for validating the `rwlock` implementation perf since if a key appears a lot more frequent, then it will cause more contentions when sframe performs aggregations under multithreading context.

test/benchmark/sframe_bench_aggregate.cpp
```diff
@@ -0,0 +1,482 @@
...
+ * 3. number of unique keys except count operator
+ *
+ * the generated sframe will be reused for different
+ * sized thread pool settings.
+ * */
+void bench_test_aggreate_summary_count_bin(size_t nrows, size_t reps) {
+  ASSERT_MSG(reps > 0, "reps shouldn't be 0");
+  ASSERT_MSG(nrows > 0, "nrows shouldn't be 0");
+
+  sframe sf = make_random_sframe(nrows, "b", false);
```
Jarvi-Izana: adopted Hoyt's advice.

test/benchmark/sframe_bench_aggregate.cpp
```diff
@@ -0,0 +1,482 @@
...
+        ret.emplace_back(values[itr - samples.begin()]);
+      } else {
+        ret.emplace_back(dist_dat(gen));
+      }
+    }
+
+    return ret;
+  };
+
+  sframe sf = make_testing_sframe(keys, {flexible_type(T()).get_type()}, nrows,
```
Jarvi-Izana: added a new utility routine into `testing_utils`.

hoytak:
Looks good!  Go ahead and merge it!

PR #2262, opened 2019-08-22
[WIP] make atomic.value private and provide a getter for it

Jarvi-Izana:
```
  public:
    //! The current value of the atomic number
    volatile T value
```
the client can mistakenly override the value without using the correct setters.

I plan to make it private; If you think it's something we need to do, I will make it private and change all other locations where value is used without protection.

### DIFF
--- a/src/core/parallel/atomic.hpp
+++ b/src/core/parallel/atomic.hpp
@@ -36,6 +36,8 @@ namespace turi_impl {
     //! Performs an atomic decrement by 1, returning the new value
     T dec() { return __sync_sub_and_fetch(&value, 1);  }
 
+    T load() const {return value; }
+
     //! Lvalue implicit cast
     operator T() const { return value; }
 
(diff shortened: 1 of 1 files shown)


Jarvi-Izana:
to avoid inadvertent overriding member value directly.
```
atomic.value = 10;
```
Besides, you have setters to guard the member and why not have a corresponding getter?

znation:
Oh. Perhaps I am missing something -- I don't see where in the diff it actually makes it private (thus doesn't prevent anything as of yet).

znation:
Assuming you plan to make `value` a private member, I think this change is good 👍

nickjong:
As an aside, any reason not to use `std::atomic` given that we require C++11?

I could even see creating a wrapper around that, if we want to remove the ability to treat atomics like ordinary variables

Jarvi-Izana:
std::atomic doesn't support float type. So that's why the turi::atomic exists.

nickjong:
I just tested this simple program on my Mac:
```
    std::atomic<float> x;
    x.store(5.0f);
    std::cout << x.is_lock_free() << std::endl;
```
and it compiles and outputs `1`...

Does the same not hold in the general case, on other platforms?

Jarvi-Izana:
I believe is architecture-dependent.

Jarvi-Izana:
too many places have direct access to member `value`. It's a non-trivial work-load. will drop it here.

PR #2263, opened 2019-08-22
[WIP] proof of concept: aggregate with rwlock

Jarvi-Izana:
compare with perf with the implementation in #2210. 

This implementation requires:
* atomic aggregate operations
* reader and writer lock to protect the bucket containing k-v pairs.

For this version, we only prove the feasibility in COUNT operation.


Jarvi-Izana:
This branch is branched from #2260. Once #2260 is merged into master, the diff should only include the relevant changes. Right now, just keep this PR open for recording progress.

Issue #2184, opened 2019-07-31
Import of turicreate from python takes > 2 seconds


hoytak:
Currently, importing turicreate from python takes more than 2 seconds.  This is compared with ~200ms for scipy or numpy.  This should be reduced as much as possible -- ideally, we should be in the same range.

Jarvi-Izana:
In ipython3 client,
```
In [1]: %time import turicreate
CPU times: user 602 ms, sys: 347 ms, total: 948 ms
Wall time: 1.58 s
```

Jarvi-Izana:
```
CPU times: user 441 ms, sys: 108 ms, total: 548 ms
Wall time: 288 ms
```
optimization v1. @hoytak

Jarvi-Izana:
```
CPU times: user 504 ms, sys: 116 ms, total: 621 ms
Wall time: 243 ms
```
phrase 1 is done for now.

PR #2272, opened 2019-08-29
reduce the import time of turicreate

Jarvi-Izana:
#2184 

implemented a naive lazy import.


Jarvi-Izana:
v1:
```
(env) ➜  debug git:(2184-import-reduction) ✗ ipython3 -c "%time import pandas"
CPU times: user 565 ms, sys: 123 ms, total: 688 ms
Wall time: 238 ms
(env) ➜  debug git:(2184-import-reduction) ✗ ipython3 -c "%time import numpy"
CPU times: user 248 ms, sys: 68.8 ms, total: 317 ms
Wall time: 59.5 ms
(env) ➜  debug git:(2184-import-reduction) ✗ ipython3 -c "%time import turicreate"
<class 'turicreate.data_structures.sgraph.Vertex'>
CPU times: user 441 ms, sys: 109 ms, total: 550 ms
Wall time: 289 ms
(env) ➜  debug git:(2184-import-reduction) ✗ ipython3 -c "%time import sklearn"
CPU times: user 586 ms, sys: 167 ms, total: 753 ms
Wall time: 168 ms
```

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,59 +9,122 @@
...
+                        'Numpy support will be disabled.')
+                        % (numpy.__version__, NUMPY_MIN_VERSION) )
+    if not HAS_NUMPY:
+        from . import numpy_mock as _ret
+    return _ret
 
+numpy = LazyModuleLoader('numpy', _dynamic_load_numpy)
 
-HAS_SKLEARN = True
+HAS_SKLEARN = __has_module('sklearn')
```
TobyRoseman: So it doesn't look like anything is actually using `HAS_SKLEARN`. So let's just rip this code out completely.

src/python/turicreate/__init__.py
```diff
@@ -16,24 +16,27 @@
 __version__ = '{{VERSION_STRING}}'
 from turicreate.version_info import __version__
 
+from turicreate._deps import LazyModuleLoader
```
TobyRoseman: You should hide this import, i.e. `from turicreate._deps import _LazyModuleLoader`.

Jarvi-Izana: will do

src/python/turicreate/visualization/_plot.py
```diff
@@ -369,13 +369,13 @@ def image_formatter(im):
         im.save(image_buffer, format='PNG')
         return "<img src=\"data:image/png;base64," + base64.b64encode(image_buffer.getvalue()).decode() + "\"/>"
 
-    import pandas as pd
+    from turicreate._deps import pandas as pd
```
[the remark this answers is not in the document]
znation: From my understanding, TMTOWTDI ("there's more than one way to do it") is more the Perl philosophy than Python. Supposedly, "there should be one obvious way to do it" in Python (see https://www.python.org/dev/peps/pep-0020/). In practice, as we observe (re: imports), this is unfortunately not always true.

src/python/turicreate/__init__.py
znation: `_deps` itself is already hidden, so it would also be fine to do `from turicreate._deps import LazyModuleLoader as _LazyModuleLoader`.

Jarvi-Izana:
```
CPU times: user 414 ms, sys: 98.4 ms, total: 512 ms
Wall time: 259 ms
```
slightly improved.

src/python/turicreate/visualization/_plot.py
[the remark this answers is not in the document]
Jarvi-Izana: Oh well, in this case, I should refrain from blaming python.

Jarvi-Izana:
```
CPU times: user 408 ms, sys: 102 ms, total: 510 ms
Wall time: 247 ms
```
If we get rid of numpy from `cy_flexible_type`; then we can possibly reach to ~ 170 ms; that will be a huge gain.

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,6 +9,224 @@
...
+            if name in LazyModuleLoader.registry_:
+                raise RuntimeError("pkg {} is already taken by other LazyModuleLoader instance.".format(name))
+            LazyModuleLoader.registry_.add(name)
+            self.module_ = sys.modules.get(name, None)
+            self.loaded_ = self.module_ is not None
+
+            if self.loaded_:
+                if not isinstance(self.module_, LazyModuleLoader):
+                    _logging.debug("turicreate: {} is loaded ahead and it cannot be"
+                                   " be deffered by LazyModuleLoader".format(name))
```
hoytak: nit: spelled "Deferred"

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,59 +9,346 @@
...
+        # moulde and pkg naming should follow pep8
+        if not all([x.isalnum() or x in '._' for x in name]):
+            raise ValueError('invalid module name')
+        exec("import %s as _py2_ret" % name)
+        return _py2_ret
+    else:
+        return importlib.import_module(name)
+
+class LazyModuleLoader(ModuleType, object):
+    """ defer to load a module. If it's loaded, no exception but warning will be given. """
```
hoytak: Change this comment?  I believe this is now just a no-op?

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,59 +9,346 @@
...
+    # read-only member list -> used by __setattr__; see below
+    _my_attrs_ = ['is_loaded', 'reload', 'get_module',
+                  '_name', '_init_func', '_module', '_loaded']
+    # ensure singleton; must be guarded by _ImportLockContext lock
```
hoytak: Why do we need a lock around this?  Python is single-threaded.

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,59 +9,346 @@
...
+        by __getattr__, __setattr__, or __delattr__ on `attr` other than
+        `dir(LazyModuleLoader())` The side effect is that sys.modules[name]
+        will be set to real module object.
+
+        >>> np = LazyModuleLoader('numpy')
+        >>> import numpy # won't actually load
+        >>> np.ndarray # load the numpy module
+
+        If the module is loaded before LazyModuleLoader loads the underlying module,
+        then LzyModuleLoader won't do anything but just a thin wrapper to forward
```
hoytak: Typo

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,59 +9,346 @@
...
+            raise ValueError("name must be str type")
+        if len(name) == 0:
+            raise ValueError("module name should not be empty")
+        if name.startswith('.'):
+            raise ValueError("only support absolute path import style")
+        # must set first to avoid recursion
+        self._name = name
+        self._init_func = init_func
+
+        with _ImportLockContext():
```
hoytak: I still don't understand why an import lock is needed.

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,59 +9,346 @@
...
+        LazyModuleLoader hijacks sys.modules[name] to defer the
+        use of the module if the module is not already loaded yet.
+
+        The load of module is deferred until LazyModuleLoader is called
+        by __getattr__, __setattr__, or __delattr__ on `attr` other than
+        `dir(LazyModuleLoader())` The side effect is that sys.modules[name]
+        will be set to real module object.
+
+        >>> np = LazyModuleLoader('numpy')
+        >>> import numpy # won't actually load
```
hoytak: I don't understand why this is the case...  To do this, it would actually have to do injection into sys.modules, which I don't see happening below.

Jarvi-Izana:
passed internal test, we are good to go!

src/python/turicreate/_deps/__init__.py
Jarvi-Izana: Yeah, CPython is single-threaded. But iron-python is not. Well, I just follow what `importlib` does by providing an import lock.

src/python/turicreate/_deps/__init__.py
Jarvi-Izana: cleared

src/python/turicreate/_deps/__init__.py
Jarvi-Izana: oh yeah, I forget to clean it up. Will do.

src/python/turicreate/_deps/__init__.py
Jarvi-Izana: the importlib has it. I guess probably the import is done in C thread.

src/python/turicreate/_deps/__init__.py
Jarvi-Izana: Good catch! That's the previous behavior. I will hijack sys.modules['name'].

src/python/turicreate/_deps/__init__.py
Jarvi-Izana: nice catch!

src/python/turicreate/_deps/__init__.py
Jarvi-Izana: removed.

PR #2279, opened 2019-09-05
Consolidated VegaRenderer fixes

znation:
* Ensure a `TC` prefix is on all exported Objective C symbols.
* Simplify the `#include` process for generated headers.
* Add preliminary CSS cursor support.
* Support a wider variety of Vega specifications by implementing a tiny
  bit of fake DOM (canvas and div HTML elements; style attribute).
* Use explicit `strong`/`atomic`/etc. on properties.
* Use explicit synthesis of properties.
* Add support for dictionary-based linear gradient specifications.
* Use a backing CGBitmapContext instead of a CGLayer; turns out it's
  actually much faster.
* Fix font weight handling; the enum values are floats and don't
  correspond to the expected integers.
* Add preliminary support for UIKit-based platforms.
* Add isPointInPath checking in the Canvas implementation, along with
  unit tests for it.
* Refactor out console logging and other types of logging.


znation:
All visualization C++/ObjC unit tests pass locally for me.

PR #2289, opened 2019-09-06
build support

Jarvi-Izana:
check commit

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -84,9 +84,17 @@ Optionally, set an environment variable named `VIRTUALENV` if you want to use a 
     # /usr/local/bin/virtualenv is using Python 2.7
     export VIRTUALENV=/Library/Frameworks/Python.framework/Versions/3.6/bin/virtualenv
 
+If you use `pyenv`, you can just set your local pyenv python version under the same directory as `.config` file lives in to set right python version to compile with:
+    # using python2
+    # when built from source, turicreate will link over python2 C headers and dynamic libraries
+    pyenv local 2.7.16
+    # if virtualenv is not installed; pip3 for python3
+    pip2 install virtualenv
+
 Then, run `./configure` (optionally with command line arguments to control what is built):
 
     ./configure
+    # for more help, you can run ./configure --help
 
 Running `./configure` will create two sub-directories, `release/` and
 `debug/` . cd into `src/` under either of these directories and running make will build the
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -84,9 +84,17 @@ Optionally, set an environment variable named `VIRTUALENV` if you want to use a
     # /usr/local/bin/virtualenv is using Python 2.7
     export VIRTUALENV=/Library/Frameworks/Python.framework/Versions/3.6/bin/virtualenv
 
+If you use `pyenv`, you can just set your local pyenv python version under the same directory as `.config` file lives in to set right python version to compile with:
```
znation: I think this is going to be a bit confusing for users who don't use pyenv (which is almost everyone). Is this something users of `pyenv` (who I assume are going to be a bit more advanced) could figure out for themselves? If so, I'd avoid putting it in the `BUILD.md`.

Otherwise, I think we should consider restructuring this document to accommodate a choice of different tools/environments -- but right now the document is structured around the assumption that there is a single, correct set of tools/environments (just virtualenv, really) and if you want to use others you're on your own.

Jarvi-Izana: I think `pyenv` is used by tons of people. At least people around me, the adoption rate is pretty high (except our team).

znation: I haven't used or seen it before (or seen anyone else use it). It does have 16k stars on GitHub though...

znation: But my point is more that the current structure of the document doesn't really allow for the insertion of random tools at random stages, without adding even more confusion. It's already confusing enough to figure out how to build Turi Create from source and what the requirements are. Adding `pyenv` here is going to lead to people reading this document thinking that `pyenv` is required, and I don't think it's required. Frankly I'm not even sure what value it adds here. Is the purpose of using pyenv for building Turi Create, to avoid having to install virtualenv globally?

Jarvi-Izana: If I use pyenv and I will be much appreciated if Turi guys have this considerate thought for me. Otherwise I would waste some time search for solution or just give it up. Since build from source is already very troublesome, if I cannot take advantage of my tool, then there will be some learning curve for me to get thins right. Then I would stop and chill. I think most mediocre programmers like me will have similar minds.

As a tool, we should give support as much as we can.

znation: What problem is `pyenv` solving as part of this workflow? I'm struggling to understand how involving `pyenv` is simpler or easier than following this document as written in master.

Jarvi-Izana: manage different python versions and make it easy to switch python versions for development.

Jarvi-Izana: the problem it tries to solve is to specify the python version the virtualenv should use when you run `./config`.

TobyRoseman: I agree with Zach. I think it would be confusing to include this information. I just don't think pyenv is a common enough tool to warrant its inclusion here.

Jarvi-Izana: Then where should we put `pyenv` support? Maybe I'm biased, I was taught with pyenv in my grad school and all the people around me use pyenv, except our team.

PR #2297, opened 2019-09-10
Changed seeding strategy and have one generator per row of output data

shantanuchhabra:


src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -32,9 +32,11 @@ namespace one_shot_object_detection {
 namespace data_augmentation {
 
 flex_dict build_annotation(ParameterSampler &parameter_sampler,
-                           std::string label, size_t object_width,
-                           size_t object_height, long seed) {
-  parameter_sampler.sample(seed);
+                           std::string label,
```
TobyRoseman: can `label` be `const`?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -209,23 +210,28 @@ gl_sframe augment_data(const gl_sframe &data,
...
-         backgrounds.range_iterator(segment_start, segment_end)) {
-      row_number++;
-      flex_image flex_background =
-          image_util::decode_image(background_ft.to<flex_image>());
-      for (const auto &row : decompressed_data.range_iterator()) {
-        // go through all the starter images and create augmented images for
-        // all starter images and the respective chunk of background images
-        const flex_image &object = row[image_column_index].get<flex_image>();
-        std::string label = row[target_column_index].to<flex_string>();
+  for (const auto &row : decompressed_data.range_iterator()) {
```
TobyRoseman: Can you avoid using `auto` here and on line 225?

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
nickjong: Typically this should either be just `std::string` if we want to pass by value and `const std::string&` if we want to pass by reference. See https://google.github.io/styleguide/cppguide.html#Use_of_const

> For a function parameter passed by value, const has no effect on the caller, thus is not recommended in function declarations.

Here I would recommend a const reference, since we don't modify the value

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
```diff
@@ -149,14 +150,14 @@ create_synthetic_image_from_background_and_starter(const flex_image &starter,
                        reinterpret_cast<const boost::gil::rgba8_pixel_t *>(
                            rgba_flex_image.get_image_data()),
                        rgba_flex_image.m_channels *
-                           rgba_flex_image.m_width  // row length in bytes
+                           rgba_flex_image.m_width // row length in bytes
```
nickjong: Any reason to remove the space here? I recommend we just use `clang-format` to resolve formatting decisions and be done with it.

In this case, I believe clang-format should leave two spaces: https://google.github.io/styleguide/cppguide.html#Implementation_Comments

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
nickjong: FYI: https://google.github.io/styleguide/cppguide.html#Type_deduction

> The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type.

src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
```diff
@@ -64,33 +65,31 @@ void ParameterSampler::set_warped_corners(
...
-  phi_mean = phi_means_[std::rand() % phi_means_.size()];
-  std::srand(seed + 2);
-  gamma_mean = gamma_means_[std::rand() % gamma_means_.size()];
+  std::seed_seq seed_seq = {static_cast<int>(seed), static_cast<int>(row_number)};
+  std::mt19937 engine(seed_seq);
+  
+  std::uniform_int_distribution<int> index_distribution(0, INT_MAX);
+  theta_mean = theta_means_[index_distribution(engine) % theta_means_.size()];
+  phi_mean = phi_means_[index_distribution(engine) % phi_means_.size()];
+  gamma_mean = gamma_means_[index_distribution(engine) % gamma_means_.size()];
```
nickjong: It seems strange to take the result modulo the number of means. Why not just generate a random integer in the desired range right off the bat? Could be a useful helper function here: given a non-inclusive upper bound on the range of the random variable, and the random engine, return a random index

src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
```diff
@@ -64,33 +65,31 @@ void ParameterSampler::set_warped_corners(
...
+  
   std::normal_distribution<double> theta_distribution(theta_mean, angle_stdev_);
   std::normal_distribution<double> phi_distribution(phi_mean, angle_stdev_);
   std::normal_distribution<double> gamma_distribution(gamma_mean, angle_stdev_);
-  std::normal_distribution<double> focal_distribution((double)width_,
+  std::normal_distribution<double> focal_distribution((double)background_width,
```
nickjong: Let's use `static_cast` here.

https://google.github.io/styleguide/cppguide.html#Casting

src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
TobyRoseman: Agreed. It should be a const reference.

PR #2300, opened 2019-09-10
Pybind11 code to call TensorFlow (Python) from C++ implementation

shreyajain17:


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -28,7 +33,9 @@ static auto* tf_registration = new compute_context::registration(
 }  // namespace
 
 tf_compute_context::tf_compute_context()
-{}
+{
```
nickjong: Nit: why change this? If we change this, may as well use the modern syntax:
```
tf_compute_context::tf_compute_context() = default;
```

src/ml/neural_net/tf_activity_classifier_backend.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <ml/neural_net/model_backend.hpp>
+
+#include <pybind11/pybind11.h>
+#include <pybind11/numpy.h>
+
+namespace turi {
+namespace neural_net {
+
+namespace py = pybind11;
```
nickjong: Let's avoid using namespace aliases in header files.

https://google.github.io/styleguide/cppguide.html#Aliases

src/ml/neural_net/tf_activity_classifier_backend.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+namespace py = pybind11;
+
+using turi::neural_net::float_array;
+using turi::neural_net::float_array_map;
+using turi::neural_net::shared_float_array;
+
+
+template <typename CallFunc> 
+auto call_pybind_function(CallFunc&& func) -> decltype(func()) {
```
nickjong: Why is the return type `auto`? Can't it just be `void` here? We don't return anything.

I think we should also prefer `const CallFunc&` as the argument type, since all we do with `func` is call it? (https://google.github.io/styleguide/cppguide.html#Rvalue_references)

src/ml/neural_net/tf_activity_classifier_backend.cpp
```diff
@@ -0,0 +1,195 @@
...
+    PyGILState_STATE gstate;
+  	gstate = PyGILState_Ensure();
```
nickjong: Inconsistent indentation. Please run `git-clang-format`

src/ml/neural_net/tf_activity_classifier_backend.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    turi::scoped_finally gstate_restore([&](){PyGILState_Release(gstate);});
+
+	try {
+		func();
+	} catch(...){
+	    	log_and_throw("An error occurred!");
```
nickjong: Could you add a TODO for improved error logging?

src/ml/neural_net/tf_activity_classifier_backend.cpp
```diff
@@ -0,0 +1,195 @@
...
+	    }
+}
+
+
+static std::vector<size_t> get_dimensions( const float_array& num) {
+  std::vector<size_t> result(num.dim());
+  std::memcpy(result.data(), num.shape(), num.dim()* sizeof(size_t));
```
nickjong: This function could be more clearly implemented as 
```
return std::vector<size_t>(num.shape(), num.shape() + num.dim());
```

Could we just call it `get_shape` for consistency?

src/ml/neural_net/tf_activity_classifier_backend.cpp
```diff
@@ -0,0 +1,195 @@
...
+  return result;
+}
+
+static std::vector<size_t> get_strides( const float_array& num) {
+  std::vector<size_t> result(num.dim());
+  std::fill(result.begin(), result.end(), sizeof(float));
+  return result;
```
nickjong: I don't think this computation is correct, if the `float_array` has more than one dimension. For a `float_array` with shape `[3,3]`, for example, the strides are
`[12,4]`, since it takes 12 bytes to step from one row to the next (and 4 bytes to step from one column within one row to the next)

src/ml/neural_net/tf_activity_classifier_backend.cpp
```diff
@@ -0,0 +1,195 @@
...
+	          const_cast<float*>(m.data()),                   /* Pointer to buffer */
+	          sizeof(float),                                  /* Size of one scalar */
+	          py::format_descriptor<float>::format(),         /* Python struct-style format descriptor */
+	          m.dim(),                                        /* Number of dimensions */
+	          get_dimensions(m),                              /* Buffer dimensions */
+	          get_strides(m)                                  /* Strides (in bytes) for each index */
+	           
+	      );
+	});
+	py::class_<shared_float_array>(m, "SharedFloatArray", py::buffer_protocol())
```
nickjong: Is this necessary? `shared_float_array` inherits from `float_array`, at least on the C++ side

src/ml/neural_net/tf_activity_classifier_backend.cpp
```diff
@@ -0,0 +1,195 @@
...
+	// Make an instance of python object 
+	activity_classifier_ = tf_ac_backend.attr("ActivityTensorFlowModel")(weights, 
+		batch_size, num_features, num_classes, pw, predictions_in_chunk);
+
+    });
+
+}
+
+
+float_array_map tf_activity_classifier_backend::train(const float_array_map& inputs) {
```
nickjong: I don't see anything specific to AC in these method implementations. Could we factor it out into a C++ base class for all TF models?

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: Will change it to the modern syntax.

src/ml/neural_net/tf_activity_classifier_backend.cpp
shreyajain17: Got it

src/ml/neural_net/tf_activity_classifier_backend.cpp
shreyajain17: Yes, it was not working with just `float_array`.

src/ml/neural_net/tf_activity_classifier_backend.cpp
shreyajain17: All the methods except the model object right?

nickjong: Yeah, I think the things to factor out all the implementations of the four methods in the `model_backend` interface

src/ml/neural_net/tf_model_backend.cpp
```diff
@@ -48,10 +34,9 @@ static std::vector<size_t> get_shape(const float_array& num) {
 static std::vector<size_t> get_strides(const float_array& num) {
   std::vector<size_t> result(num.dim());
   const size_t* shape = num.shape();
-  for (size_t i = num.dim() - 1; i < 0; i--) {
+  for (size_t i = num.dim() - 1; i > 0; i--) {
```
nickjong: With this change, you're never setting `result[0]`. I recommend something like:
```
result[num.dim() - 1] = sizeof(float);
for (size_t i = num.dim() - 1; i > 0; --i) {
  result[i - 1] = result[i] * shape[i];
}
```

src/ml/neural_net/tf_model_backend.hpp
```diff
@@ -15,6 +16,21 @@
 namespace turi {
 namespace neural_net {
 
+template <typename CallFunc>
+auto call_pybind_function(const CallFunc&& func) -> decltype(func()) {
```
nickjong: Does this need to return `auto` or can it return `void`?

PR #2305, opened 2019-09-13
Python Frontend AC will call C++ implementation of AC

shreyajain17:
When we set `use_tensorflow=True` in `_advanced_parameters`, we basically use the C++ implementation of AC now. The methods create, predict and evaluate work for now. I still have to get the other stuff working. 

Please only look at `_activity_classifier.py` only. Working on the other files in a different PR (will rebase before merging).


src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -14,6 +14,7 @@
 import time as _time
 import six as _six
 
+import turicreate as _turicreate
```
nickjong: Why do we need this at the top level?

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -165,7 +167,36 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
...
+
+        name = 'activity_classifier'
+
+        # Imports tensorflow
+        import turicreate.toolkits.libtctensorflow
+        from ._tf_model_architecture import ActivityTensorFlowModel
+
+
+        model = _turicreate.extensions.activity_classifier()
+        options = {}
```
nickjong: What about `features`?

src/ml/neural_net/tf_compute_context.hpp
```diff
@@ -47,6 +67,22 @@ class tf_compute_context: public compute_context {
   
 };
 
+class tf_model_backend : public model_backend {
```
nickjong: I don't think this even needs to be in the header file. If we move this to the top of the `.cpp` file, then all the pybind headers can go there too

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
shreyajain17: passing it in the dataset, right?

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
shreyajain17: Won't I need it to call extensions?

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
shreyajain17: The options dictionary I made to sent the arguments needed for init_options. There aren't any features. I am sending the features when I call train right?

src/ml/neural_net/tf_compute_context.hpp
shreyajain17: Doing it in the other PR.

PR #2306, opened 2019-09-13
Resolves #2236: Fixed jupyter notebook inline title

### LINKED ISSUE #2236: Jupyter notebook SFrame.explore ignores title parameter
The Jupyter Notebook integrated SFrame.explore does not honor the `title=` parameter to `explore`. See screenshots for comparison with GUI explore:

Screen Shot 2019-08-13 at 11 29 35 PM
Screen Shot 2019-08-13 at 11 29 06 PM

dhivyaaxim:
This PR resolves #2236 
Fixed Jupyter notebook inline title

### DIFF
--- a/src/python/turicreate/visualization/_plot.py
+++ b/src/python/turicreate/visualization/_plot.py
@@ -360,6 +360,8 @@ def display_table_in_notebook(sf, title=None):
     from PIL import Image
 
     import base64
+    import cgi
+    import html
     from io import BytesIO
     from IPython.display import HTML
     from ..data_structures.image import Image as _Image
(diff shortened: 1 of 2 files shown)


src/python/turicreate/visualization/_plot.py
```diff
@@ -450,6 +450,7 @@ def image_formatter(im):
                             </style>                               \
                           </head>                                  \
                           <body>                                   \
+                            <h1> '+ title +' </h1>                 \
```
znation: `title` needs to be HTML encoded; otherwise some values of `title` could cause malformed HTML. You can use the [HTML.escape](https://docs.python.org/3/library/html.html#html.escape) in Python 3 and [cgi.escape](https://docs.python.org/3.6/library/cgi.html#cgi.escape) in Python 2 for this.

dhivyaaxim: Thanks @znation !  Will check and raise PR

src/python/turicreate/visualization/_plot.py
```diff
@@ -360,6 +360,8 @@ def display_table_in_notebook(sf, title=None):
     from PIL import Image
 
     import base64
+    import cgi
```
znation: Please put these imports within the if/else conditions where they are used. Otherwise one of the imports may fail on Python 2 or Python 3 (I think the html module is new in Python 3).

dhivyaaxim: @znation Thanks for your suggestion! Changes are implemented and request for PR

PR #2307, opened 2019-09-13
Added a fix for building on a system lower than MacOS 10.15

abhishekpratapa:
- Added an `#ifdef` for a MacOS version lower than 10.15

### DIFF
--- a/src/ml/neural_net/mps_graph_trainer.mm
+++ b/src/ml/neural_net/mps_graph_trainer.mm
@@ -19,7 +19,10 @@ using turi::neural_net::float_scalar;
 using turi::neural_net::make_array_map;
 using turi::neural_net::mps_graph_cnn_module;
 using turi::neural_net::shared_float_array;
+
+#ifdef HAS_MACOS_10_15
 using turi::style_transfer::style_transfer;
+#endif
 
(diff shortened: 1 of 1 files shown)


src/ml/neural_net/mps_graph_trainer.mm
```diff
@@ -63,23 +66,27 @@ int TCMPSDeleteGraphModule(MPSHandle handle) {
...
-  shared_float_array index_array(
-      std::make_shared<float_scalar>(index));
-
-  auto outputs = obj->train({ { "input",  inputs_array },
-                              { "labels", labels_array },
-                              { "index",  index_array  } });
-  shared_float_array* loss = new shared_float_array(outputs.at("loss"));
-  *loss_out = reinterpret_cast<TCMPSFloatArrayRef>(loss);
+  #ifdef HAS_MACOS_10_15
+    style_transfer *obj = reinterpret_cast<style_transfer *>(handle);
```
nickjong: Don't change the indent level of the code.

The preprocessor direction should be at zero indent

https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives

Consider just using clang-format

abhishekpratapa: Used clang-format on the file!

PR #2310, opened 2019-09-14
Sound Classifier predict for clips less than 975 milliseconds

rplom:
https://github.com/apple/turicreate/issues/1712

Update the audio input example to chunk the data in with zero padding such that it can work with any legnth material.

### DIFF
--- a/userguide/sound_classifier/export-coreml.md
+++ b/userguide/sound_classifier/export-coreml.md
@@ -85,63 +85,40 @@ guard let audioData = try? MLMultiArray(shape:[windowSize as NSNumber],
 }
 
 
-let frameLength = Int64(buffer.frameLength)
-var windowNumber = 0
-var offset = 0
-var startIndex = 0
+
 var results = [Dictionary<String, Double>]()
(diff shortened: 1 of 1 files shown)


userguide/sound_classifier/export-coreml.md
```diff
@@ -83,20 +84,63 @@ guard let audioData = try? MLMultiArray(shape:[windowSize as NSNumber],
...
+    
+    guard let modelOutput = try? model.prediction(input: modelInput) else {
+        fatalError("Error calling predict")
+    }
+    results.append(modelOutput.categoryProbability)
+    windowNumber += 1
+}
+
+if windowNumber > 0 {
+    offset = offset + windowSize
```
TobyRoseman: I'm not really understanding why this is necessary.

Would it be cleaner to just delete this code paragraph (i.e. lines 111-113) and change line 116 to be:
`if windowNumber == 0 {`?

rplom: glad i revisited this, much simpler now.

PR #2315, opened 2019-09-16
One Shot C++ unit tests: a start

shantanuchhabra:
Wrote one test to sanity check the distributions of the parameters sampled. This test should catch any software bugs with regard to generating angles not in tune with the mean and standard deviation specified.

### DIFF
--- a/src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
+++ b/src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
@@ -20,11 +20,15 @@ namespace one_shot_object_detection {
 
 ParameterSampler::ParameterSampler(size_t starter_width, size_t starter_height,
                                    size_t dx, size_t dy)
-    : starter_width_(starter_width), starter_height_(starter_height), dx_(dx), dy_(dy) {}
+    : starter_width_(starter_width),
+      starter_height_(starter_height),
+      dx_(dx),
+      dy_(dy) {}
 
 double deg_to_rad(double angle) { return angle * M_PI / 180.0; }
 
-std::vector<double> vector_deg_to_rad(const std::vector<double> &means_in_degrees) {
+std::vector<double> vector_deg_to_rad(
+    const std::vector<double> &means_in_degrees) {
   std::vector<double> means_in_radians;
   means_in_radians.reserve(means_in_degrees.size());
   for (double mean : means_in_degrees) {
@@ -48,11 +52,17 @@ double ParameterSampler::get_phi() { return phi_; }
 
 double ParameterSampler::get_gamma() { return gamma_; }
 
-std::vector<double> ParameterSampler::get_theta_means() { return vector_deg_to_rad(theta_means_); }
+std::vector<double> ParameterSampler::get_theta_means() {
+  return vector_deg_to_rad(theta_means_);
+}
 
-std::vector<double> ParameterSampler::get_phi_means() { return vector_deg_to_rad(phi_means_); }
+std::vector<double> ParameterSampler::get_phi_means() {
+  return vector_deg_to_rad(phi_means_);
+}
 
-std::vector<double> ParameterSampler::get_gamma_means() { return vector_deg_to_rad(gamma_means_); }
+std::vector<double> ParameterSampler::get_gamma_means() {
+  return vector_deg_to_rad(gamma_means_);
+}
 
 double ParameterSampler::get_theta_stdev() { return deg_to_rad(angle_stdev_); }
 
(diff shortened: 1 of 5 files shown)


src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
```diff
@@ -22,6 +22,14 @@ ParameterSampler::ParameterSampler(size_t width, size_t height, size_t dx,
 
 double deg_to_rad(double angle) { return angle * M_PI / 180.0; }
 
+std::vector<double> vector_deg_to_rad(std::vector<double> means_in_degrees) {
```
hoytak: Where possible, get used to passing in parameters using `const type&` instead of type.  Passing in `std::vector<double>` triggers a copy and memory allocation.  It's not noticeably expensive in each operation, but if it happens across a code base it can make it a lot slower.

src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
```diff
@@ -22,6 +22,14 @@ ParameterSampler::ParameterSampler(size_t width, size_t height, size_t dx,
...
+  std::vector<double> means_in_radians;
```
hoytak: It's also good practice to add a reserve() call when using the push_back pattern and you know how many elements will be added.  This avoids a lot of unnecessary reallocations and copies.  Here, add 
`means_in_radians.reserve(means_in_degrees.size());`

test/unity/toolkits/object_detection/one_shot_object_detection/test_parameter_sampler.cxx
```diff
@@ -0,0 +1,105 @@
...
+  int count = 0;
+  for (double sample : samples) {
+    if (sample >= lower && sample <= upper) {
+      count += 1;
+    }
+  }
+  return count;
+}
+
+bool angles_match_distribution(std::vector<double> angles,
```
hoytak: It's testing code, so less important, but still a good idea to get used to the `const type&` pattern.

test/unity/toolkits/object_detection/one_shot_object_detection/test_parameter_sampler.cxx
```diff
@@ -0,0 +1,105 @@
...
+#define BOOST_TEST_MODULE test_parameter_sampler
+
+#include <toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.hpp>
+
+#include <boost/test/unit_test.hpp>
+#include <core/util/test_macros.hpp>
+
+#include <unordered_map>
+
+static constexpr int max_dimension = 6000;
```
nickjong: Let's give constants a distinctive syntax, either `MAX_DIMENSION` or `k_max_dimension`

(FWIW, https://google.github.io/styleguide/cppguide.html#Constant_Names, although it'd be a pretty big change for us to adopt Google's CamelCase for C++ names in the general case)

test/unity/toolkits/object_detection/one_shot_object_detection/test_parameter_sampler.cxx
```diff
@@ -0,0 +1,105 @@
...
+  
+  size_t starter_width = dist(engine);
+  size_t starter_height = dist(engine);
+  size_t dx = 0;
+  size_t dy = 0;
+  ParameterSampler sampler(starter_width, starter_height, dx, dy);
+  return sampler;
+}
+
+std::unordered_map<std::string, std::vector<double> > get_all_angles(
```
nickjong: `get_all_angles` seems like a strange name for this function. `get` normally connotes a simple accessor. Here you're actually sampling from the `sampler`

shantanuchhabra: Changed it to `construct_angle_samples_map`.

test/unity/toolkits/object_detection/one_shot_object_detection/test_parameter_sampler.cxx
shantanuchhabra: Done.

src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
shantanuchhabra: Done.

src/toolkits/object_detection/one_shot_object_detection/util/parameter_sampler.cpp
shantanuchhabra: Done.

test/unity/toolkits/object_detection/one_shot_object_detection/test_parameter_sampler.cxx
shantanuchhabra: Done. Followed the Google style guide.

PR #2319, opened 2019-09-17
Upgraded version of libpng for bugfix.

hoytak:
Noted crashes in some libpng files; upgrading libpng to version 1.6.37 to incorporate recent bugfixes.


hoytak:
No harm in it -- there are a bunch of bugfixes in that code, so it's probably good to pull it in regardless.

PR #2326, opened 2019-09-18
Better error and validation for #2301

### LINKED ISSUE #2301: 'tuple' object has no attribute 'column_names'
Im new to Python while inserting images in DB, why am I getting this error, how can we solve this issue.

File "/usr/local/lib/python3.6/site-packages/turicreate/toolkits/image_classifier/image_classifier.py", line 242, in create
AttributeError: 'tuple' object has no attribute 'column_names'

sirahd:
- fix #2301

### DIFF
--- a/src/python/turicreate/toolkits/image_classifier/image_classifier.py
+++ b/src/python/turicreate/toolkits/image_classifier/image_classifier.py
@@ -235,6 +235,8 @@ def create(dataset, target, feature=None, model = 'resnet-50',
     _tkutl._check_categorical_option_type('model', model, allowed_models)
 
     # Check dataset parameter
+    if not isinstance(dataset, _tc.SFrame):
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -235,6 +235,8 @@ def create(dataset, target, feature=None, model = 'resnet-50',
     _tkutl._check_categorical_option_type('model', model, allowed_models)
 
     # Check dataset parameter
+    if not isinstance(dataset, _tc.SFrame):
+        raise _ToolkitError("Unrecognized value for 'dataset'. An SFrame is expected.")
```
TobyRoseman: I think you should say "type" rather than "value" in the error message.

Also I think it makes more sense to raise a `TypeError` rather than a `_ToolkitError`.

PR #2327, opened 2019-09-18
Python frontend calls C++ implementation of AC/ Tensorflow

shreyajain17:
To check if this works  -
Also `pip install tensorflow==1.14.0` in the virtual environment 
```model = tc.activity_classifier.create(data, session_id='exp_id', target='activity', _advanced_parameters={'use_tensorflow':True})```

### DIFF
--- a/src/ml/neural_net/compute_context.hpp
+++ b/src/ml/neural_net/compute_context.hpp
@@ -21,14 +21,9 @@ namespace neural_net {
  * hardware resources.
  */
 class compute_context {
-public:
-
+ public:
   /** Function that yields a compute context. */
   using factory = std::function<std::unique_ptr<compute_context>()>;
-
-  /** Function that yields a compute context (made to return
-   * tf_compute_context). */
-  using tf_factory = std::function<std::unique_ptr<compute_context>()>;
 
   /**
    * To solve for layering/dependency issues, we allow compute_context::factory
@@ -36,16 +31,16 @@ public:
    * static init time, adjusts the behavior of the create() function below.
    */
   class registration {
-  public:
+   public:
     // Registers `factory_fn` at the given priority.
-   registration(int priority, factory factory_fn, tf_factory tf_factory_fn_);
+    registration(int priority, factory factory_fn, factory tf_factory_fn_);
 
-   // Removes the registration. In practice, simplest just not to deallocate...
-   ~registration();
+    // Removes the registration. In practice, simplest just not to deallocate...
+    ~registration();
 
-   int priority() const { return priority_; }
-   std::unique_ptr<compute_context> create_context() const {
-     return factory_fn_();
+    int priority() const { return priority_; }
+    std::unique_ptr<compute_context> create_context() const {
+      return factory_fn_();
     }
 
     std::unique_ptr<compute_context> create_tensorflow_context() const {
@@ -53,7 +48,7 @@ public:
       ;
     }
 
-  private:
+   private:
     int priority_;
     factory factory_fn_;
     tf_factory tf_factory_fn_;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py [hunk 1]
```diff
@@ -14,6 +14,7 @@
 import time as _time
 import six as _six
 
+import turicreate as _turicreate
```
nickjong: Why is this needed?

shreyajain17: I am calling _turicreate.extensions in create function.

nickjong: Could we just import it in the `create` function then?

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -637,9 +642,14 @@ std::unique_ptr<data_iterator> activity_classifier::create_iterator(
 }
 
 std::unique_ptr<compute_context>
-activity_classifier::create_compute_context() const
+activity_classifier::create_compute_context(bool use_tensorflow) const
 {
-  return compute_context::create();
+  if (use_tensorflow) {
```
nickjong: This could use `read_state("use_tensorflow")` instead of adding an argument to the function, right?

shreyajain17: Yes

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
shreyajain17: I will change it

src/ml/neural_net/compute_context.hpp
```diff
@@ -26,6 +26,10 @@ class compute_context {
   /** Function that yields a compute context. */
   using factory = std::function<std::unique_ptr<compute_context>()>;
 
+  /** Function that yields a compute context (made to return
+   * tf_compute_context). */
+  using tf_factory = std::function<std::unique_ptr<compute_context>()>;
```
nickjong: You don't need a new name here. You can just use `factory` everywhere you were using `tf_factory` since they are both identical aliases for `std::function<std::unique_ptr<compute_context>()>`

src/ml/neural_net/compute_context.hpp
```diff
@@ -34,19 +38,25 @@ class compute_context {
...
     // Registers `factory_fn` at the given priority.
-    registration(int priority, factory factory_fn);
+   registration(int priority, factory factory_fn, tf_factory tf_factory_fn_);
+
+   // Removes the registration. In practice, simplest just not to deallocate...
+   ~registration();
 
-    // Removes the registration. In practice, simplest just not to deallocate...
-    ~registration();
+   int priority() const { return priority_; }
```
nickjong: Did `clang-format` do this, give this an overall indent level of 3 spaces? If so, try giving the `public:` above an indent level of 1 space. I think `clang-format` thinks that stuff should be indented by 1 space relative to `public:`

In short, the number of leading spaces for code should just about always be a multiple of 2, except for `public:`

(See https://google.github.io/styleguide/cppguide.html#Class_Format)

src/toolkits/activity_classification/activity_classifier.hpp
```diff
@@ -14,6 +14,7 @@
 #include <ml/neural_net/compute_context.hpp>
 #include <ml/neural_net/model_backend.hpp>
 #include <ml/neural_net/model_spec.hpp>
+#include <ml/neural_net/tf_compute_context.hpp>
```
nickjong: I don't think you need this....

PR #2329, opened 2019-09-19
Random x-y translation to object in the superimposed image

shantanuchhabra:
Currently, objects in the superimposed image always have the same center coordinate. This hasn't adversely affected mAP performance so far because the object ends up with different relative positions in the augmented image (because of varying sizes of the backgrounds). This PR seeks to add a random amount of x-y translation to the object in the superimposed image in order to prevent the object being lodged at the same place while also making an effort to keep the object inside the background image as much as possible. Performance testing in progress.

### DIFF
--- a/src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
+++ b/src/toolkits/object_detection/one_shot_object_detection/one_shot_object_detector.cpp
@@ -36,58 +36,9 @@ flex_dict build_annotation(ParameterSampler &parameter_sampler,
                            size_t object_width, size_t object_height,
                            size_t background_width, size_t background_height,
                            size_t seed, size_t row_number) {
+  
   parameter_sampler.sample(background_width, background_height, seed, row_number);
-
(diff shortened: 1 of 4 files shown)


hoytak:
Code looks good to me.  Would really like to see C++ unit test on this — even just running 100 random examples through to make sure everything is within bounds.  But this code looks good.

shantanuchhabra:
> Code looks good to me. Would really like to see C++ unit test on this — even just running 100 random examples through to make sure everything is within bounds. But this code looks good.

@hoytak Cool, I have an open PR (#2315 ) for the architecture of OSOD C++ unit tests (along with one unit test). Should I:
A) put up another PR with a unit test for this functionality (which would involve copying over the architecture from #2315 )?
B) add a unit test for this functionality and push to the branch for #2315 ?
C) copy over the architecture of OSOD C++ unit tests from #2315 and push a commit with a unit test to this PR?

Please let me know which of the above I should do!

shantanuchhabra:
Closing the PR since we are scoping this out for later.

PR #2331, opened 2019-09-19
New API for SArray's read_json

### LINKED ISSUE #89: SArray.read_json() should be able to read an array
The following code fails:
```
import turicreate as tc
import json

with open('/tmp/test.json', 'w') as f:
    json.dump([1,2,3], f)

tc.SArray.read_json('/tmp/test.json')
```
with this error message:
```
Parsing JSON records from /tmp/test.json
/tmp/test.json has non-dictionary elements which are ignored.
```

sirahd:
- fix #89 
- update the documented behavior of read_json to allow reading JSON list with homogeneous type.
- since SFrame's read_json also depends on the modified SArray's read_json, I add assertion to make sure the JSON file read into SFrame is a list of dictionary

### DIFF
--- a/src/core/storage/sframe_interface/unity_sarray.cpp
+++ b/src/core/storage/sframe_interface/unity_sarray.cpp
@@ -223,6 +223,7 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
   // create sarray and output iterator
   auto sarray_ptr = std::make_shared<sarray<flexible_type>>();
   sarray_ptr->open_for_write(1);
+  sarray_ptr->set_type(flex_type_enum::DICT);
   auto output = sarray_ptr->get_output_iterator(0);
 
   flexible_type_parser parser(",", true, '\\',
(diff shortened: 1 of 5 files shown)


src/core/storage/sframe_interface/unity_sarray.cpp
```diff
@@ -276,28 +275,50 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
...
-            bool has_non_dict_elements = false;
-            for (const auto& element : parse_result.first.get<flex_list>()) {
-              if (element.get_type() == flex_type_enum::DICT ||
-                  element.get_type() == flex_type_enum::UNDEFINED) {
-                (*output) = element;
-                ++output;
-                ++num_elems_parsed;
-              } else {
-                has_non_dict_elements = true;
+            auto elements = parse_result.first.get<flex_list>();
```
hoytak: Let's use const auto& here instead of auto to avoid a copy.

src/core/storage/sframe_interface/unity_sarray.cpp
```diff
@@ -276,28 +275,50 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
...
+            if (!elements.empty()) {
+              auto sarray_type = elements.front().get_type();
+              sarray_ptr->set_type(sarray_type);
+              for (auto it = elements.begin(); it != elements.end(); it++) {
+                auto element = *it;
```
hoytak: Same here -- use `const auto&`

src/core/storage/sframe_interface/unity_sarray.cpp
```diff
@@ -276,28 +275,50 @@ void unity_sarray::construct_from_json_record_files(std::string url) {
...
+                  element_type = sarray_type;
+                }
+                // throw an error if the type doesn't match
+                if (element_type != sarray_type) {
+                  std::stringstream error_msg;
+                  error_msg << "sequence item " << it - elements.begin() 
+                            << ": expected " << flex_type_enum_to_name(sarray_type)
+                            << ", but get " << flex_type_enum_to_name(element_type)
+                            << std::endl;
+                  sarray_ptr->close();
```
hoytak: Good catch.

PR #2333, opened 2019-09-20
issue fix for recommend_from_interactions --exclude doesn't work properly

### LINKED ISSUE #2044: recommend_from_interactions --exclude doesn't work properly
图片
图片
recommend can gets rid of 14 correctly,but recommend_from_interactions  doesn't work properly.

dhivyaaxim:
This PR resolves #2044 
recommend_from_interactions --exclude doesn't work properly

### DIFF
--- a/src/python/turicreate/test/test_recommender.py
+++ b/src/python/turicreate/test/test_recommender.py
@@ -1353,6 +1353,14 @@ U1103,135104,0'''
             self._test_coreml_export(m4, ['135085','135038'])
         '''
 
+    def test_recommend_from_interactions(self):
+        data=tc.SFrame({'userId':[1,1,1,2,2,2,3,3,3],'movieId':[10,11,12,10,13,14,10,11,14]})
+        exclude_pairs=tc.SFrame({'movieId':[14]})
+        observed_data=tc.SFrame({'movieId':[10]})
+        data_set=tc.item_similarity_recommender.create(data,user_id='userId',item_id='movieId')
+        observed_data=data_set.recommend_from_interactions(observed_items=observed_data,exclude=exclude_pairs)
+        assert data['movieId'][8] not in observed_data['movieId']
+
     def test_compare_models(self):
         from turicreate.toolkits.recommender.util import compare_models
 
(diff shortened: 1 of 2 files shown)


dhivyaaxim:
> @TobyRoseman Thanks for your suggestion! Will update unit test and let you know

src/python/turicreate/test/test_recommender.py
```diff
@@ -1353,6 +1353,14 @@ def test_implicit(self):
             self._test_coreml_export(m4, ['135085','135038'])
         '''
 
+    def test_recommend_from_interactions(self):
+        data=tc.SFrame({'userId':[1,1,1,2,2,2,3,3,3],'movieId':[10,11,12,10,13,14,10,11,14]})
+        exclude_pairs=tc.SFrame({'movieId':[14]})
+        observed_data=tc.SFrame({'movieId':[10]})
+        data_set=tc.item_similarity_recommender.create(data,user_id='userId',item_id='movieId')
```
TobyRoseman: This is confusing. What you're creating on this line is a model not a data set. I would just call the variable `model` rather than `data_set`.

src/python/turicreate/test/test_recommender.py
```diff
@@ -1353,6 +1353,14 @@ def test_implicit(self):
...
+        observed_data=data_set.recommend_from_interactions(observed_items=observed_data,exclude=exclude_pairs)
```
TobyRoseman: These are recommendations not observed data. I think `recommendations` makes a lot more sense than `observed_data`.

src/python/turicreate/test/test_recommender.py
```diff
@@ -1353,6 +1353,14 @@ def test_implicit(self):
...
+        assert data['movieId'][8] not in observed_data['movieId']
```
TobyRoseman: I think this line would be a lot easier to reason about if it was just `assert 14 not in observed_data['movieId']`.

dhivyaaxim:
Thanks @TobyRoseman  All changes are addressed !

TobyRoseman:
Thanks @dhivyaaxim

PR #2334, opened 2019-09-20
Fixed offset for the initialization of the instance norm data loader

abhishekpratapa:
Missed an offset in the Instance norm data loader on initialization of weights.

### DIFF
--- a/src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
+++ b/src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
@@ -95,15 +95,19 @@ API_AVAILABLE(macos(10.14))
                                                    dataType:(MPSDataTypeFloat32)];
 
     _style_props = [[NSMutableArray alloc] init];
-    
+
     for (NSUInteger index = 0; index < styles; index ++) {
       TCMPSInstanceNormDataLoaderProps *style_property = [[TCMPSInstanceNormDataLoaderProps alloc] init];
 
-      id<MTLBuffer> gammaBuffer = [dev newBufferWithBytes:_gamma_weights.bytes + (index * _numberOfFeatureChannels * sizeof(float))
+      size_t gammaBufferOffset = _gamma_weights.bytes + (index * _numberOfFeatureChannels * sizeof(float));
+
+      size_t betaBufferOffset = _beta_weights.bytes + (index * _numberOfFeatureChannels * sizeof(float))
+
+      id<MTLBuffer> gammaBuffer = [dev newBufferWithBytes:gammaBufferOffset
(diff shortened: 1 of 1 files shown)


src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -95,15 +95,15 @@ - (instancetype) initWithParams:(NSString *)name
                                                    dataType:(MPSDataTypeFloat32)];
 
     _style_props = [[NSMutableArray alloc] init];
-
+
```
nickjong: Please don't introduce any trailing whitespace

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -95,15 +95,15 @@ - (instancetype) initWithParams:(NSString *)name
                                                    dataType:(MPSDataTypeFloat32)];
 
     _style_props = [[NSMutableArray alloc] init];
-
+    
     for (NSUInteger index = 0; index < styles; index ++) {
       TCMPSInstanceNormDataLoaderProps *style_property = [[TCMPSInstanceNormDataLoaderProps alloc] init];
 
-      id<MTLBuffer> gammaBuffer = [dev newBufferWithBytes:_gamma_weights.mutableBytes
+      id<MTLBuffer> gammaBuffer = [dev newBufferWithBytes:_gamma_weights.bytes + (index * _numberOfFeatureChannels * sizeof(float))
```
nickjong: For readability, could we just factor out the computation of the offset into something like
```
size_t offset = ...;
```

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -95,15 +95,19 @@ - (instancetype) initWithParams:(NSString *)name
                                                    dataType:(MPSDataTypeFloat32)];
 
     _style_props = [[NSMutableArray alloc] init];
-    
+
     for (NSUInteger index = 0; index < styles; index ++) {
       TCMPSInstanceNormDataLoaderProps *style_property = [[TCMPSInstanceNormDataLoaderProps alloc] init];
 
-      id<MTLBuffer> gammaBuffer = [dev newBufferWithBytes:_gamma_weights.bytes + (index * _numberOfFeatureChannels * sizeof(float))
+      size_t gammaBufferOffset = _gamma_weights.bytes + (index * _numberOfFeatureChannels * sizeof(float));
```
nickjong: These aren't offsets, these are addresses. Just factor out the offset, which is shared: `(index * _numberOfFeatureChannels * sizeof(float)`

PR #2335, opened 2019-09-20
Added check for MacOS Platform

abhishekpratapa:
Checks if the platform is MacOS 10.15

### DIFF
--- a/src/python/turicreate/toolkits/style_transfer/style_transfer.py
+++ b/src/python/turicreate/toolkits/style_transfer/style_transfer.py
@@ -20,7 +20,6 @@ from turicreate.toolkits._model import PythonProxy as _PythonProxy
 import turicreate as _tc
 import numpy as _np
 import math as _math
-import platform as _platform
 import six as _six
 from .._mps_utils import (use_mps as _use_mps,
                           mps_device_memory_limit as _mps_device_memory_limit,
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -209,7 +210,7 @@ def create(style_dataset, content_dataset, style_feature=None,
     _style_loss_mult = params['style_loss_mult']
 
     num_gpus = _mxnet_utils.get_num_gpus_in_use(max_devices=params['batch_size'])
-    use_mps = _use_mps() and num_gpus == 0
+    use_mps = _use_mps() and num_gpus == 0 and (float(_platform.mac_ver()[0]) >= 10.15)
```
nickjong: We should try to keep things consistent in our code base to make it easier to understand and to update. Look for how other places in our code have solved the same problem you're trying to solve!

https://github.com/apple/turicreate/blob/037e112032c40c330885bf9cb7ba136d7a61e020/src/python/turicreate/toolkits/_internal_utils.py#L706

abhishekpratapa: Thanks for pointing out that function!

PR #2336, opened 2019-09-20
Object Detector TF beta release

fareeha:
Object Detector TF integration in C++

### DIFF
--- a/src/ml/neural_net/mps_compute_context.mm
+++ b/src/ml/neural_net/mps_compute_context.mm
@@ -25,7 +25,8 @@ std::unique_ptr<compute_context> create_mps_compute_context() {
 // At static-init time, register create_mps_compute_context().
 // TODO: Codify priority levels?
 static auto* mps_registration = new compute_context::registration(
-    /* priority */ 0, &create_mps_compute_context);
+    /* priority */ 1, &create_mps_compute_context);
+///* priority */ 0, &create_mps_compute_context); //restore
 
 }  // namespace
 
--- a/src/ml/neural_net/mps_compute_context.mm
+++ b/src/ml/neural_net/mps_compute_context.mm
@@ -25,8 +25,7 @@ std::unique_ptr<compute_context> create_mps_compute_context() {
 // At static-init time, register create_mps_compute_context().
 // TODO: Codify priority levels?
 static auto* mps_registration = new compute_context::registration(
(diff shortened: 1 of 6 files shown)


src/ml/neural_net/mps_compute_context.mm
```diff
@@ -25,7 +25,8 @@
 // At static-init time, register create_mps_compute_context().
 // TODO: Codify priority levels?
 static auto* mps_registration = new compute_context::registration(
-    /* priority */ 0, &create_mps_compute_context);
+    /* priority */ 1, &create_mps_compute_context);
```
nickjong: Shreya is working on solving this problem as we speak... Her changes should make it so you don't have to do this, FYI

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -493,6 +493,22 @@ void object_detector::perform_predict(gl_sframe data,
   // For each anchor box, we have 4 bbox coords + 1 conf + one-hot class labels
   int num_outputs_per_anchor = 5 + static_cast<int>(class_labels.size());
   int num_output_channels = num_outputs_per_anchor * anchor_boxes().size();
+
+  float_array_map pred_config = get_prediction_config();
+  pred_config["num_classes"] =
+      shared_float_array::wrap(data_iter->class_labels().size());
```
nickjong: Let's move these into the `get_prediction_config` and `get_training_config` functions. These functions may need to take arguments for things like the number of classes. Per offline discussion, looks like `get_base_config` could be useful for the shared portion

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -493,6 +493,22 @@ void object_detector::perform_predict(gl_sframe data,
...
+  pred_config["lmb_noobj"] = shared_float_array::wrap(5.0);
+  pred_config["lmb_obj"] = shared_float_array::wrap(100.0);
+  pred_config["lmb_coord_xy"] = shared_float_array::wrap(10.0);
+  pred_config["lmb_coord_wh"] = shared_float_array::wrap(10.0);
+  pred_config["lmb_class"] = shared_float_array::wrap(2.0);
```
nickjong: Oh, let's also add a TODO wherever this ends up to make the MPS path actually use these parameters, instead of the versions hardcoded in our MPS code

Issue #1600, opened 2019-03-14
Inefficient implementation of SFrame.sample when fraction=1


dbickson:
There are a couple issues in TC sample implementation of sample in the case of fraction=1

1) When fraction=1 there is no sampling so the full SFrame is returned. However it seems that the internal implementation tries to sample 100% of the SFrame in a slow manner.

2) In pandas, when sampling with frac=1 the DataFrame is shuffled. In TC it is not shuffled.

3) The documentation does not state what happens when fraction=1. Strangely this is not documented in pandas either.

It is rather confusing when porting code from pandas to TC. Regarding the justification of using frac=1 imagine a dynamic code which samples in a function based on the SFrame size, thus small SFrame does not need to sampled while big dataset need to be sampled with a higher ratio.

```
>  a = pd.DataFrame({'a':np.random.rand((1000000))}).astype(str)
> b = tc.SFrame(a)

 a.sample(frac=1).head(5)
Out[80]:
                          a
416813   0.2570231164299698
612139   0.7297101984023956
456607   0.1511116535755409
959180  0.06478008039001715
132543   0.6147032089691518

> a.sample(frac=1).head(5)

                          a
597857  0.32254294052590005
283246  0.17014579668655405
382578  0.04829261879567148
770500   0.6205931460140915
211697   0.8465029355843529

> b.sample(fraction=1).head(5)

Columns:
	a	str

Rows: 5

Data:
+---------------------+
|          a          |
+---------------------+
| 0.38618082793554964 |
|  0.478491292822116  |
|  0.1816862818356344 |
| 0.21641780723877713 |
| 0.29920091097653223 |
+---------------------+
[5 rows x 1 columns]

> b.sample(fraction=1).head(5)
Columns:
	a	str

Rows: 5

Data:
+---------------------+
|          a          |
+---------------------+
| 0.38618082793554964 |
|  0.478491292822116  |
|  0.1816862818356344 |
| 0.21641780723877713 |
| 0.29920091097653223 |
+---------------------+
[5 rows x 1 columns]


```

PR #2343, opened 2019-09-23
Make SFrame's Sample function more efficient when percent = 1.

jakesabathia2:
Better Sample function for Frame when percent = 1. (#1600)
Also add a copy function for unity_sframe class.


src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -494,6 +494,10 @@ std::shared_ptr<unity_sframe_base> unity_sframe::select_columns(
   return ret;
 }
 
+std::shared_ptr<unity_sframe_base> unity_sframe::copy(){
+  return select_columns(column_names());
```
hoytak: Instead of this -- which requires computation to make sure the sets of names match up, let's use the copying code at lines 923-925 of this PR: 

```
    auto ret = std::make_shared<unity_sframe>();
    auto new_planner_node = std::make_shared<planner_node>(*(this->get_planner_node()));
    ret->construct_from_planner_node(new_planner_node, this->column_names());
    return ret;
```

This doesn't add any overhead and just creates a new copy by duplicating the planner node.

PR #2344, opened 2019-09-23
Python frontend calls C++ OD TF impl

fareeha:
Usage from Python:

`model = tc.object_detector.create(data, 'annotation', max_iterations=25, _advanced_parameters= {'mlmodel_path' : 'darknet.mlmodel', 'use_tensorflow' : True})`

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/python/turicreate/toolkits/object_detector/object_detector.py
@@ -260,6 +260,8 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
         # This large buffer size (8 batches) is an attempt to mitigate against
         # the SFrame shuffle operation that can occur after each epoch.
         'io_thread_buffer_size': 8,
+        # TODO: Have the Python side download the Core ML version of the
+        # pre-trained model from somewhere.
         'mlmodel_path': None,
         'use_tensorflow': False
     }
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -259,6 +260,7 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
         # This large buffer size (8 batches) is an attempt to mitigate against
         # the SFrame shuffle operation that can occur after each epoch.
         'io_thread_buffer_size': 8,
+        'mlmodel_path': None,
```
nickjong: Huh, how come we don’t hit this exception: https://github.com/apple/turicreate/blob/386efa4eb5033573ee9120704a8c88a9a6151133/src/ml/neural_net/model_spec.cpp#L486

Either way, worth a TODO to have the Python side download the Core ML version of the pre-trained model from somewhere

fareeha: Because I pass in the mlmodel in create : `model = tc.object_detector.create(data, 'annotation', max_iterations=25, _advanced_parameters= {'mlmodel_path' : 'darknet.mlmodel', 'use_tensorflow' : True})` It was hitting that exception when I didn't pass the model path

PR #2345, opened 2019-09-23
Updated Gitlab Builds to 10.15

abhishekpratapa:
Didn't update the cpp tests or the `dylib` runners because they don't depend on 10.15 api.

Plus the tests run only on 10.14 because currently we don't have tests to cover the code that is specific to 10.15.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -80,9 +80,9 @@ build_wheel_linux_python36:
     paths:
       - target/
 
-build_wheel_mac_10_14_python27:
+build_wheel_mac_10_15_python27:
   tags:
-    - macos10.14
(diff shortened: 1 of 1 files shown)


nickjong:
I’d like @znation to review, but offhand, why run the C++ tests on 10.13 and 10.14 but not 10.15? Why build the macOS dylib (for e.g. recommenders) on 10.14?

abhishekpratapa:
They can all be 10.15 but I wanted to build and test those targets on the mimimim required machines. We don't have any unit-tests currently that are specifically 10.15 so it seemed unnecessary to add them in right now. When we have 10.15 specific unit-tests and requirements for recommenders we can bump the OS version number was my reasoning.

PR #2347, opened 2019-09-24
Implement equality operator for image_type

### LINKED ISSUE #2042: SArray == doesn't work for tc.Image types
Using turicreate 5.6 the following code:
```python
import turicreate as tc

i = tc.load_images('./test/images/sample.png')
i = i['image']

print("Images are equal: " + str(
    i[0] == i[0]
))
print("SArray indexes equals: " + str(
    i == i
))
```
Produces: 
```
Images are equal: True
SArray indexes equals: [0]
```

So `==` works when comparing instances of `tc.Image` directly but doesn't work when they are inside an `SArray`.

sirahd:
- fix #2042 to allow comparing SArray with Image type
- Currently, implementation of the equality operator mirrors the python implementation where two images are equal *only* when every property (`m_height`, `m_width`, ..., `m_image_data`) are identical, but we might want to allow comparing two identical images with different lossless types (png vs. RAW) by decoding to RAW type

### DIFF
--- a/src/core/data/image/image_type.cpp
+++ b/src/core/data/image/image_type.cpp
@@ -5,6 +5,7 @@
  */
 #include <core/data/image/image_type.hpp>
 #include <boost/gil/gil_all.hpp>
+#include <algorithm>
 
 namespace turi{
 
@@ -86,7 +87,9 @@ bool image_type::operator==(const image_type& other) const {
       m_image_data_size == other.m_image_data_size &&
       m_version == other.m_version &&
       m_format == other.m_format &&
-      strcmp(m_image_data.get(), other.m_image_data.get()) == 0;
+      std::equal(m_image_data.get(),
+                 m_image_data.get() + m_image_data_size,
+                 other.m_image_data.get());
 }
 
 }
--- a/src/core/data/image/image_type.cpp
+++ b/src/core/data/image/image_type.cpp
@@ -75,4 +75,18 @@ const unsigned char* image_type::get_image_data() const {
   }
 }
 
+// The equality behavior is currently mirrored from python implementation 
+// where every property is compared directly, but we might want to later
+// support comparing two images with different lossless types.
+bool image_type::operator==(const image_type& other) const {
+  return
+      m_height == other.m_height &&
+      m_width == other.m_width &&
+      m_channels == other.m_channels && 
+      m_image_data_size == other.m_image_data_size &&
+      m_version == other.m_version &&
+      m_format == other.m_format &&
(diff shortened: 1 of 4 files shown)


Jarvi-Izana:
Great job! How about adding a unittest for your change?

src/core/data/image/image_type.cpp
```diff
@@ -75,4 +75,18 @@ const unsigned char* image_type::get_image_data() const {
...
+// support comparing two images with different lossless types.
+bool image_type::operator==(const image_type& other) const {
+  return
+      m_height == other.m_height &&
+      m_width == other.m_width &&
+      m_channels == other.m_channels && 
+      m_image_data_size == other.m_image_data_size &&
+      m_version == other.m_version &&
+      m_format == other.m_format &&
+      strcmp(m_image_data.get(), other.m_image_data.get()) == 0;
```
hoytak: This works only with C style strings, which means it will stop at the first zero byte.  Use std::equal instead:
`std::equal(m_image_data.get(), m_image_data.git() + m_image_data_size, other.m_image_data.get())`

src/python/turicreate/test/test_sarray.py
```diff
@@ -11,6 +11,7 @@
 from ..data_structures.sarray import SArray
 from ..data_structures.sframe import SFrame
 from ..data_structures.sarray import load_sarray
+from ..toolkits.image_analysis import image_analysis
```
Jarvi-Izana: let's don't introduce toolkit code in this stage since sarray is the foundation of other high-level application code. Let's make it just self-contained by just using bare flexible type and sarray itself. Use the image type defined in data_structure.Image to explicit add image to sarray.

Jarvi-Izana: Ok, I just check the implementation. The actual image load happens inside our cpp backend.  You are good to go. Better use `from ... import load_images` to only bind name load_image to the current namespace. So far so good! Great work!

src/core/data/image/image_type.cpp
sirahd: Thanks for pointing out! I was not aware of the EOF character that could potentially be in the char array.

src/core/data/image/image_type.hpp
```diff
@@ -7,6 +7,7 @@
 #define TURI_IMAGE_IMAGE_TYPE_HPP
 
 #include <string>
+#include <string.h>
```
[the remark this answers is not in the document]
sirahd: I just removed the header because we no longer need it for strcmp per suggestion above

PR #2350, opened 2019-09-25
Added the correct way to update weights from MPS to Python

### LINKED ISSUE #2349: Style Transfer Mac GPU results not looking good.
Python 3.6
macOS 10.15

**Repro script**

```python
import turicreate as tc

style_images = tc.load_images("style")[:1]
content_images = tc.load_images("content")[:50]

model = tc.style_transfer.create(style_dataset=style_images, content_dataset=content_images, max_iterations=5000)

test_image = content_images[50]['image']
image_output = model.stylize(test_image, style=0)
image_output.show()
```

**Style image**

style

**Content image**

content

**Output image**

image_output

abhishekpratapa:
Fixes #2349

### DIFF
--- a/src/python/turicreate/toolkits/style_transfer/style_transfer.py
+++ b/src/python/turicreate/toolkits/style_transfer/style_transfer.py
@@ -409,13 +409,12 @@ def create(style_dataset, content_dataset, style_feature=None,
 
         mps_mxnet_key_map = _MpsStyleGraphAPI.mps_mxnet_weight_dict()
 
+
         for key in mps_weights:
-            if "transformer" in key and "conv" in key:
+            if "transformer" in key and ("inst" in key or "conv" in key):
                 weight = transformer.collect_params()[mps_mxnet_key_map[key]].data()
-                transformer.collect_params()[mps_mxnet_key_map[key]].set_data(_mx.nd.array(_mps_to_mxnet(mps_weights[key]).reshape(weight.shape)))
-            if "transformer" in key and "inst" in key:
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -412,10 +412,10 @@ def create(style_dataset, content_dataset, style_feature=None,
         for key in mps_weights:
             if "transformer" in key and "conv" in key:
```
TobyRoseman: Should the two if-statement be combined? The bodies of each if-statement now (I think) exactly the same.
The combined test could be `if "transformer" in key and ("inst" in key or "conv" in key):`

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -412,10 +412,10 @@ def create(style_dataset, content_dataset, style_feature=None,
...
                 weight = transformer.collect_params()[mps_mxnet_key_map[key]].data()
-                weight = _mx.nd.array(_mps_to_mxnet(mps_weights[key]))
+                transformer.collect_params()[mps_mxnet_key_map[key]].set_data(_mx.nd.array(_mps_to_mxnet(mps_weights[key]).reshape(weight.shape)))
```
TobyRoseman: These lines are really. Introducing temporary variables would also help readability.

src/python/turicreate/toolkits/style_transfer/style_transfer.py
abhishekpratapa: Good catch. Thanks!

PR #2351, opened 2019-09-25
Added `.so` to setup.py

abhishekpratapa:
Added the tensorflow so file to the `setup.py`.


src/python/setup.py
```diff
@@ -173,7 +173,7 @@ def run(self):
         package_data={
             'turicreate': [
                 '_cython/*.so', '_cython/*.pyd',
-                '*.so', '*.dylib',
+                '*.so', '*.dylib', 'toolkits/*.so',
```
TobyRoseman: Why is this change needed? Are we putting `.so` files in the `toolkits` folder now?

PR #2356, opened 2019-09-25
Eliminated CMake source and moved some functionality into recyclable scripts.

hoytak:
Now that we know cmake ships with the xcode compiler toolchain, there isn't really a need to build it from source.  This PR drops that source code and adds additional checks for the compiler. 

Also, in the process of updating our build system, it makes sense to isolate some components to make the code more readable and to be able to recycle functionality across other scripts.  This PR does this with the CMake and compiler finding code.


znation:
Please change the title - `Eliminated CMake source` seems to be in another PR.

PR #2359, opened 2019-09-25
Fixed bug concerning numpy import in sframe.

hoytak:
Error in to_numpy() on fresh install.  This fixes it.


Jarvi-Izana:
Hmmm, that worries me. HAS_NUMPY should be True, isn't it?

Jarvi-Izana:
```
In [2]: import turicreate as tc                                                                                                                                                                                                                        

In [3]: tc._deps.HAS_NUMPY                                                                                                                                                                                                                             
Out[3]: True
```

hoytak:
FIxed by the other PR.

PR #2360, opened 2019-09-25
address concern in HAS_NUMPY

Jarvi-Izana:
fix always true for HAS_NUMPY.
better use class_name.static_variable.

In related to #2359

### DIFF
--- a/src/python/turicreate/_deps/__init__.py
+++ b/src/python/turicreate/_deps/__init__.py
@@ -276,7 +276,7 @@ def __get_version(version):
 def __has_module(name):
     if six.PY2:
         try:
-            spec = imp.find_module(name)[0]
+            spec = imp.find_module(name)
             # file or module path
             return not (spec[0] is None and spec[1] is None)
         except ImportError:
--- a/src/python/turicreate/_deps/__init__.py
+++ b/src/python/turicreate/_deps/__init__.py
@@ -164,7 +164,7 @@ class LazyModuleLoader(ModuleType):
         # for interactive autocompletion
         if self._module:
             return dir(self._module)
-        return self._my_attrs_ + super(LazyModuleLoader, self).__dir__()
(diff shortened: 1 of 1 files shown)


src/python/turicreate/_deps/__init__.py
```diff
@@ -275,7 +275,12 @@ def __get_version(version):
 
 def __has_module(name):
     if six.PY2:
-        return imp.find_module(name) is not None
+        try:
+            spec = imp.find_module(name)
+            # file or module path
```
TobyRoseman: I don't understand this comment or the conditional below it.

Jarvi-Izana: from [imp.find_module](https://docs.python.org/2/library/imp.html) doc,
```
If search is successful, the return value is a 3-element tuple (file, pathname, description):

file is an open file object positioned at the beginning, pathname is the pathname of the file found, and description is a 3-element tuple as contained in the list returned by get_suffixes() describing the kind of module found.

If the module does not live in a file, the returned file is None, pathname is the empty string, and the description tuple contains empty strings for its suffix and mode; the module type is indicated as given in parentheses above. If the search is unsuccessful, ImportError is raised. Other exceptions indicate problems with the arguments or environment.

If the module is a package, file is None, pathname is the package path and the last item in the description tuple is PKG_DIRECTORY.
```

Jarvi-Izana: spec[0] is file and spec[1] is package (or module) path.

TobyRoseman: Perhaps the following would be clearer:
```python
(file, pathname, _) = imp.find_module(name)
return not (file is None and pathname is None)
```

Jarvi-Izana: Perfect! I will use this form. Thanks!

PR #2364, opened 2019-09-26
Fix the bug that OD confidence score is always near 0.999

### LINKED ISSUE #1314: Object Detector deployed to Vision framework unexpectedly yields confidences at 0.999
I've observed that training and exporting an object detector, and integrating into the Vision sample code [here](https://developer.apple.com/documentation/vision/recognizing_objects_in_live_capture) yields confidences very close to 1. Training and exporting a model with the same data but without NMS yields confidences that span the range from 0 to 1 (after tweaking the sample code just to output the single highest-confidence box).

Our docs describe NMS as a filtering mechanism; we don't expect any "merging" of confidence scores. So there appears to be an error in either the docs, the CoreML implementation, or the Vision integration.

jakesabathia2:
close #1314


userguide/object_detection/export-coreml.md
```diff
@@ -25,17 +25,18 @@ let objectRecognition = VNCoreMLRequest(model: visionModel,
     for case let foundObject as VNRecognizedObjectObservation in results {
         let bestLabel = foundObject.labels.first! // Label with highest confidence
         let objectBounds = foundObject.boundingBox
+        let confidence = foundObject.confidence // Confidence for the predicted class
+        // bestLabel.confidence is the normalized scroe across all classes, which might be misleading.
```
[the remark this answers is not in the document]
jakesabathia2: done

userguide/object_detection/export-coreml.md
```diff
@@ -25,17 +25,18 @@ let objectRecognition = VNCoreMLRequest(model: visionModel,
     for case let foundObject as VNRecognizedObjectObservation in results {
         let bestLabel = foundObject.labels.first! // Label with highest confidence
         let objectBounds = foundObject.boundingBox
+        let confidence = foundObject.confidence // Confidence for the predicted class
+        // bestLabel.confidence is the normalized score across all classes, which might be misleading.
```
TobyRoseman: Given the bug you're fixing, I understand why you mention this here, but perhaps it would be cleaner for our users just to omit this comment.

jakesabathia2: Just deleted the comment .

PR #2370, opened 2019-09-26
Added support for single image input to OSOD, along with a unit test.

### LINKED ISSUE #2363: Passing a single image to `preview_synthetic_training_data` fails
Repro script:
```python
import turicreate as tc                                                                                                
starter = tc.Image('image.png')               
data = tc.one_shot_object_detector.util.preview_synthetic_images(starter, 'label')
```

We get the following error.
```python
AttributeError: 'Image' object has no attribute 'column_names'
```

The fix is to move [this line](https://github.com/apple/turicreate/blob/master/src/python/turicreate/toolkits/one_shot_object_detector/util/_error_handling.py#L18) inside the `if` statement for when `dataset` is an `SFrame`.

shantanuchhabra:
Fixes #2363

### DIFF
--- a/src/python/turicreate/test/test_one_shot_object_detector.py
+++ b/src/python/turicreate/test/test_one_shot_object_detector.py
@@ -110,6 +110,16 @@ class OneObjectDetectorSmokeTest(unittest.TestCase):
         }
         self.fields_ans = self.get_ans.keys()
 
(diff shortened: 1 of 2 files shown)


shantanuchhabra:
> Are there any docstrings which need to be updated?

No, the docstring already reflects the expected behavior.

Issue #2378, opened 2019-09-30
AC import from custom model expects "random_seed" state not present in custom models


nickjong:
The initial implementation for importing an AC model saved with the old Python implementation attempts to read the "random_seed", but "random_seed" is only populated from C++.

https://github.com/apple/turicreate/blob/432af44763b18753e4524e9ba025e8915ac884eb/src/toolkits/activity_classification/activity_classifier.cpp#L523

PR #2379, opened 2019-10-01
Import pretained python AC model to C++

jakesabathia2:
#2378


jakesabathia2:
So I just find out the evaluation need random seed anyway,
so I just generate a random seed when importing the old Python model.

src/ml/neural_net/weight_init.cpp
```diff
@@ -61,7 +61,29 @@ xavier_weight_initializer::xavier_weight_initializer(
   result.output_gate_bias_fn = zero_weight_initializer();
 
   return result;
-}
+  }
```
nickjong: This curly brace is now mis-aligned. Please consider running git-clang-format (just on your changes)

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -33,6 +33,7 @@ using neural_net::lstm_weight_initializers;
 using neural_net::shared_float_array;
 using neural_net::xavier_weight_initializer;
 using neural_net::zero_weight_initializer;
+using neural_net::weight_initializer;
```
nickjong: Nit: let's keep these in alphabetical order. (clang-format may or may not do this automatically for you)

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -524,6 +525,11 @@ void activity_classifier::import_from_custom_model(
   flex_dict pred_model = variant_get_value<flex_dict>(it->second);
   model_data.erase(it);
 
+  //generate a random seed for the model
```
nickjong: Why is a random seed required for evaluation?

src/toolkits/activity_classification/activity_classifier.hpp
```diff
@@ -201,7 +201,7 @@ class EXPORT activity_classifier: public ml_model_base {
       const;
 
   // Returns the initial neural network to train
-  virtual std::unique_ptr<neural_net::model_spec> init_model() const;
+  virtual std::unique_ptr<neural_net::model_spec> init_model(bool use_zero_init = false) const;
```
nickjong: We shouldn't use default arguments for virtual functions, due to nonintuitive semantics. See also https://google.github.io/styleguide/cppguide.html#Default_Arguments

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -652,15 +659,23 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
   size_t prediction_window = read_state<flex_int>("prediction_window");
   const flex_list &features_list = read_state<flex_list>("features");
 
-  // Initialize a random number generator for weight initialization.
-  std::seed_seq seed_seq = { read_state<int>("random_seed") };
-  std::mt19937 random_engine(seed_seq);
-
+  //Only to create a random engine for weight initialization if use_zero_init = false
+  std::mt19937 random_engine;
+  if (use_zero_init == false){
```
nickjong: Our code tends to the style `if (!use_zero_init) {`

Might be more clear to use the inverse: `use_random_init`

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -652,15 +659,23 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
...
+    std::seed_seq seed_seq { read_state<int>("random_seed") };
```
nickjong: Let's resolve whether we really need random seed or not for evaluation. If we do, there's no need to make this conditional: we can always initialize the random_engine from the random seed

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -652,15 +659,23 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
...
       "features",
       std::vector<std::string>(features_list.begin(), features_list.end()));
   result->add_reshape("reshape", "features",
                       {{1, num_features, 1, prediction_window}});
+  weight_initializer initializer;
+  lstm_weight_initializers lstm_initializer;
+  initializer = (use_zero_init) ? (weight_initializer) zero_weight_initializer() :
+  (weight_initializer) xavier_weight_initializer(num_features * prediction_window,
+                                NUM_CONV_FILTERS * prediction_window,
+                                &random_engine);
```
nickjong: I think you can just initialize `initializer` to a `zero_weight_initializer`, then only overwrite it when use random initialization.

Then we can avoid the unnecessary use of the ternary ?: operator

src/toolkits/activity_classification/activity_classifier.cpp
jakesabathia2: it somehow need to sample the time series data with a random seed.

src/toolkits/activity_classification/activity_classifier.cpp
jakesabathia2: So the evaluation always need a random seed.
However, I guess in here we want to use the previous produced random seed 
from the saved model to reproduce the result.
Hmmm so maybe I should change the name of the bool.

src/ml/neural_net/weight_init.cpp
```diff
@@ -40,49 +40,49 @@ xavier_weight_initializer::xavier_weight_initializer(
...
-  result.forget_gate_bias_fn = zero_weight_initializer();
-  result.block_input_bias_fn = zero_weight_initializer();
-  result.output_gate_bias_fn = zero_weight_initializer();
+    result.input_gate_bias_fn = zero_weight_initializer();
+    result.forget_gate_bias_fn = zero_weight_initializer();
+    result.block_input_bias_fn = zero_weight_initializer();
+    result.output_gate_bias_fn = zero_weight_initializer();
 
-  return result;
+    return result;
```
nickjong: These lines should not be indented further. Is this from clang-format? It looks like it's being thrown off by the incorrect indentation in line 31.

Don't just blindly accept what clang-format outputs!

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -639,7 +635,12 @@ std::unique_ptr<data_iterator> activity_classifier::create_iterator(
   data_params.prediction_window = read_state<flex_int>("prediction_window");
   data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
   data_params.use_data_augmentation = use_data_augmentation;
-  data_params.random_seed = read_state<int>("random_seed");
+  if (!use_data_augmentation){
+    data_params.random_seed = 0;
+  }
+  else{
+    data_params.random_seed = read_state<int>("random_seed");
+  }
```
nickjong: Did you run clang-format here? In most of our code, the formatting should be
```
if (condition) {
  ...
} else {
  ...
}
```

jakesabathia2: hmmm it seems like the clang-format didn't run when I commit it.
Will take a look

PR #2385, opened 2019-10-02
TCMPS Style Transfer Fixes

abhishekpratapa:
Added necessary changes to make TCMPS Style Transfer converge.
- Changed weight export from Instance Norm Data Loader
- Added `resultImages` to training graph definition
- Change the the weight reshape in Python

### DIFF
--- a/src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
+++ b/src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
@@ -177,11 +177,11 @@ API_AVAILABLE(macos(10.14))
 
 - (float *) beta {
   NSUInteger previousStyle = _styleIndex;
-  _betaPlaceHolder = [NSMutableData data];
+  _betaPlaceHolder.length = 0;
   for (NSUInteger index = 0; index < _styles; index++) {
     _styleIndex = index;
     [self checkpointWithCommandQueue:_cq];
-    float* betaWeights = (float *) [[[_style_props objectAtIndex: _styleIndex] betaBuffer] contents];
+    float* betaWeights = (float *) [_style_props[_styleIndex].betaBuffer contents];
     [_betaPlaceHolder appendBytes:betaWeights length:sizeof(float)*_numberOfFeatureChannels];
   }
   _styleIndex = previousStyle;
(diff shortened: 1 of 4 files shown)


src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -175,10 +177,11 @@ - (void) loadBeta:(float *)beta {
 
 - (float *) beta {
   NSUInteger previousStyle = _styleIndex;
+  _betaPlaceHolder = [NSMutableData data];
   for (NSUInteger index = 0; index < _styles; index++) {
     _styleIndex = index;
     [self checkpointWithCommandQueue:_cq];
-    float* betaWeights = (float *) [[[[_style_props objectAtIndex: _styleIndex] betaVector] data] contents];
+    float* betaWeights = (float *) [[[_style_props objectAtIndex: _styleIndex] betaBuffer] contents];
```
nickjong: Nit: I think code like this would be more readable as e.g.
```
float* betaWeights = (float *) [_style_props[_styleIndex].betaBuffer contents];
```

src/ml/neural_net/mps_layer_instance_norm_data_loader.mm
```diff
@@ -175,10 +177,11 @@ - (void) loadBeta:(float *)beta {
 
 - (float *) beta {
   NSUInteger previousStyle = _styleIndex;
+  _betaPlaceHolder = [NSMutableData data];
```
nickjong: Why do we need to allocate a new instance here? Don't we allocate in the data loader's initializer?

Note that you could do `_betaPlaceHolder.length = 0;` to clear out any current contents (and reuse the existing buffer)

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -409,11 +409,13 @@ def create(style_dataset, content_dataset, style_feature=None,
...
-
         for key in mps_weights:
-            if "transformer" in key and ("inst" in key or "conv" in key):
+            if "transformer" in key:
                 weight = transformer.collect_params()[mps_mxnet_key_map[key]].data()
-                mxnet_weight = _mx.nd.array(_mps_to_mxnet(mps_weights[key]).reshape(weight.shape))
+                if "inst" in key:
+                    mxnet_weight = _mx.nd.array(_mps_to_mxnet(mps_weights[key].reshape((weight.shape))))
+                elif "conv" in key:
+                    mxnet_weight = _mx.nd.array(_mps_to_mxnet(mps_weights[key].reshape((weight.shape[0], weight.shape[2], weight.shape[3], weight.shape[1]))))
```
nickjong: Could you document the logic here?

PR #2387, opened 2019-10-02
Cleaned up the use of REMOTEFS flags.

hoytak:
Currently, the handling of the building of the optional remote file system / networking libraries wasn't implemented correctly.  This resulted in some of the remote fs libraries being built even when they didn't need to be.  This PR consolidates the handling of these into a single build variable, TC_BUILD_REMOTEFS, and a single compile time define, TC_DISABLE_REMOTEFS.  Incompatible configurations with other build variables, e.g. TC_BUILD_IOS, now simply error out.


hoytak:
This PR is finally passing all internal tests.

src/core/storage/fileio/file_download_cache.cpp
```diff
@@ -60,12 +60,8 @@ std::string file_download_cache::get_file(const std::string& url) {
   int status; bool is_temp;
   std::tie(status, is_temp, localfile) = download_url(url);
   if (status) {
-#ifdef TC_ENABLE_REMOTEFS
```
Jarvi-Izana: should we keep this prompt? I think it's kind of informational,

src/core/storage/fileio/union_fstream.cpp
```diff
@@ -76,7 +73,11 @@ union_fstream::union_fstream(std::string url,
     } catch(...) {
       log_and_throw_io_failure("Unable to open " + url);
     }
+#else
```
Jarvi-Izana: corresponding to comment above; I think we should explicitly tell users (well, us) the download failure is due to no remote fs compilation, instead of bug in our code base.

test/annotation/CMakeLists.txt
```diff
@@ -1,29 +1,23 @@
...
-  REQUIRES
-    annotation_test_utils
-  COMPILE_FLAGS_EXTRA_GCC
-    -Wno-unknown-pragmas  # NOTE: used for auto-generated protobuf source files
-    -Wno-unused-function  # NOTE: used for auto-generated protobuf source files
-)
+if(${TC_BUILD_REMOTEFS})
 
-make_boost_test(object_detector_tests.cxx
+make_boost_test(image_classifier_tests.cxx
```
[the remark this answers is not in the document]
hoytak: It downloads the model, so yes.

src/core/storage/fileio/file_download_cache.cpp
hoytak: The logic has changed -- it's now one macro, TC_DISABLE_REMOTEFS.  and the original logic was wrong.

src/core/storage/fileio/union_fstream.cpp
hoytak: I think we do.  If it doesn't we should resolve that.

Jarvi-Izana: good to know!

PR #2391, opened 2019-10-03
TCMPS Style Transfer fixes

abhishekpratapa:
## Completed
- Added checkpointing functionality to shuttle gamma and beta weights back and forth from the `CPU` to the `GPU`

## Work Items Left
- #2392 There are still issues with the style and content loss multipliers. These values are different for the MPS GPU and need to be fine-tuned to be correct


abhishekpratapa:
@shantanuchhabra may be a bit dangerous.

PR #2395, opened 2019-10-03
Import OD python model to c++

jakesabathia2:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -772,6 +772,62 @@ std::shared_ptr<MLModelWrapper> object_detector::export_to_coreml(
...
+  model_data.erase(it);
+  auto it2 = model_data.find("_grid_shape");
+  std::vector<size_t> shape = variant_get_value<std::vector<size_t>>(it2->second);
+  model_data.insert(std::pair<std::string, size_t>("grid_height", shape[0]));
+  model_data.insert(std::pair<std::string,size_t>("grid_width",shape[1]));
+  model_data.insert(std::pair<std::string,std::string>("annotation_scale","pixel"));
+  model_data.insert(std::pair<std::string,std::string>("annotation_origin","top_left"));
+  model_data.insert(std::pair<std::string,std::string>("annotation_position","center"));
+  
+  state.clear();
```
fareeha: Just curious, why do we need to clear the state?

jakesabathia2: I think it is a unity check say if you somehow import model twice :)

PR #2397, opened 2019-10-03
Removed unused code paths from configure and CMakeLists.txt

hoytak:
Removes functionality in the build scripts that is no longer present in the code itself.


hoytak:
Waiting until all of this works end to end.

Issue #2402, opened 2019-10-04
segfault when dropping words in text_classifier on table containing undefined values.


Jarvi-Izana:
```
import turicreate as tc
data = tc.SFrame.read_json('corpus.json', orient='records')
model = tc.text_classifier.create(data, 'intent', features=['text'], word_count_threshold=1)
```
The `corpus.json` has a trailing comma, which is invalid under json standard but we can parse it at our loose internal standard. It will introduce an extra row with all fields filled with undefined.
``` 
[
  {
    "intent": "getWeather",
    "text": "Is it sunny outside"
  },
]
```

PR #2403, opened 2019-10-04
skip missing values when dropping words

Jarvi-Izana:
otherwise it will cause segfault. #2402 

Current, behavior will be during the model training session, it will throw ToolkitError to remind user using dropna to clean up data.

### DIFF
--- a/src/python/turicreate/test/test_text_classifier.py
+++ b/src/python/turicreate/test/test_text_classifier.py
@@ -168,7 +168,7 @@ class TextClassifierCreateTests(unittest.TestCase):
         sf = tc.SFrame({'a': a, 'b': b})
         with self.assertRaises(ToolkitError):
             tc.text_classifier.create(sf, target='a', features=['b'], word_count_threshold=1)
-        # feature contains none, #2402
+        # feature contains none, Github #2402
         sf = tc.SFrame({'b': a, 'a': b})
         with self.assertRaises(ToolkitError):
             tc.text_classifier.create(sf, target='b', features=['a'], word_count_threshold=1)
--- a/src/python/turicreate/test/test_text_classifier.py
+++ b/src/python/turicreate/test/test_text_classifier.py
@@ -36,9 +36,9 @@ class TextClassifierTest(unittest.TestCase):
         self.target = 'score'
         self.method = 'bow-logistic'
         self.model = tc.text_classifier.create(self.docs,
-                                                   target=self.target,
-                                                   features=self.features,
-                                                   method='auto')
+                                               target=self.target,
+                                               features=self.features,
+                                               method='auto')
 
         self.num_examples = 4
 
@@ -159,6 +159,20 @@ class TextClassifierCreateTests(unittest.TestCase):
         data_str['rating'] = data_str['rating'].astype(str)
         model = tc.text_classifier.create(data_str, target='rating')
         self.assertTrue(isinstance(model, tc.text_classifier.TextClassifier))
+
+    def test_invalid_data_set(self):
+        # infer dtype str
+        a = tc.SArray(['str', None])
+        b = tc.SArray(['str', 'str'])
+        # target contains none
+        sf = tc.SFrame({'a': a, 'b': b})
+        with self.assertRaises(ToolkitError):
+            tc.text_classifier.create(sf, target='a', features=['b'], word_count_threshold=1)
+        # feature contains none, #2402
+        sf = tc.SFrame({'b': a, 'a': b})
+        with self.assertRaises(ToolkitError):
+            tc.text_classifier.create(sf, target='b', features=['a'], word_count_threshold=1)
+
 
     def test_validation_set(self):
         train = self.data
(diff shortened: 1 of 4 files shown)


src/toolkits/feature_engineering/word_trimmer.cpp
```diff
@@ -148,7 +144,7 @@ void word_trimmer_topk_index_mapping(const gl_sarray& src,
     size_t end_idx = src_size * (thread_idx + 1) / num_threads;
 
     for (const auto& v: src.range_iterator(start_idx, end_idx)) {
-      switch(run_mode) {
+      switch(v.get_type()) {
```
sirahd: Now that we check the type of each element individually, do we have to check that every element in `src` has the same type?

src/toolkits/feature_engineering/word_trimmer.cpp
```diff
@@ -204,6 +200,12 @@ void word_trimmer_topk_index_mapping(const gl_sarray& src,
           break;
         }
 
+        case flex_type_enum::UNDEFINED:
+          logstream(LOG_WARNING)
+              << "Skip undefined value in 'word_trimmer_topk_index_mapping'; "
```
TobyRoseman: I don't think the user is going to have any idea what 'word_trimmer_topk_index_mapping' is. I would just say `Skipping undefined values. Consider calling dropna on the feature columns.`

src/toolkits/feature_engineering/word_trimmer.cpp
```diff
@@ -241,6 +243,9 @@ flexible_type word_trimmer_apply(const flexible_type& input,
                || run_mode == flex_type_enum::UNDEFINED
                || run_mode == flex_type_enum::DICT);
 
+  // do nothing; gigo
```
TobyRoseman: I don't know what "gigo" mean.

src/python/turicreate/test/test_text_classifier.py
```diff
@@ -160,6 +160,20 @@ def test_sentiment_create_string_target(self):
...
 
+    def test_invalid_data_set(self):
+        # infer dtype str
+        a = tc.SArray(['str', None])
+        b = tc.SArray(['str', 'str'])
+        # target contains none
+        sf = tc.SFrame({'a': a, 'b': b})
+        with self.assertRaises(ToolkitError):
+            tc.text_classifier.create(sf, target='a', features=['b'], word_count_threshold=1)
+        # feature contains none, #2402
```
TobyRoseman: It might not be clear that `#2402` is referring to a GitHub issue. Please clarify.

src/toolkits/feature_engineering/word_trimmer.cpp
```diff
@@ -241,6 +243,9 @@ flexible_type word_trimmer_apply(const flexible_type& input,
...
+  if (run_mode == flex_type_enum::UNDEFINED) return input;
```
sirahd: we should document the new behavior of the function when the input has `undefined` type.

src/toolkits/feature_engineering/word_trimmer.cpp
Jarvi-Izana: garbage in garbage out.

src/toolkits/feature_engineering/word_trimmer.cpp
Jarvi-Izana: I believe this warning will actually go into the server log. So the user won't actually see this.

src/toolkits/feature_engineering/word_trimmer.cpp
Jarvi-Izana: you can think undefined here means the value is missing, not saying the **type** is undefined. You can rephrase it as this element should be type `blah` but the value is missing.

src/toolkits/feature_engineering/word_trimmer.cpp
srikris: It's still worth fixing the log message for better clarity.

Jarvi-Izana: anyway, log stream will append src context. Added more user friendly warning.

src/toolkits/feature_engineering/word_trimmer.cpp
```diff
@@ -204,6 +200,12 @@ void word_trimmer_topk_index_mapping(const gl_sarray& src,
           break;
         }
 
+        case flex_type_enum::UNDEFINED:
+          logstream(LOG_WARNING)
```
znation: I think we might want to make this a user-facing warning rather than just sending it to a log stream. Not sure what the best way to do that is. For visualization uses, I think we typically just `printf`, but that may not be good from code that could be called programmatically vs. interactively.

Jarvi-Izana: `cerr` might be enough. Let me check if there's any user facing log stream.

Jarvi-Izana: @znation now I remove all warnings. Skipping missing value should be an expected behavior. The training phase will decide skip or raise error. For this task, it will raise
```
Please use dropna() to drop rows with missing target values.. operator() from ../src/toolkits/supervised_learning/supervised_learning.cpp at 122'
```
if target or feature columns contain missing values.

Issue #2373, opened 2019-09-27
AC with TF does not use GPU and breaks with ToolkitError


fareeha:
On Linux, Python 3.5

AC with TF defaults to using CPU (even when 4 GPUs available) and breaks with the following error:

```
Pre-processing 525156 samples...
Using sequences of size 1000 for model creation.
Processed a total of 43 sessions.
Using CPU to create model
Traceback (most recent call last):
  File "/task_runtime/venv/lib/python3.5/site-packages/turicreate/extensions.py", line 274, in __run_class_function
    ret = self._tkclass.call_function(fnname, argument_dict)
  File "cy_model.pyx", line 35, in turicreate._cython.cy_model.UnityModel.call_function
  File "cy_model.pyx", line 40, in turicreate._cython.cy_model.UnityModel.call_function
RuntimeError: An error occurred!

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "run_ac_linux.py", line 89, in <module>
    prediction_window=50, _advanced_parameters={'use_tensorflow':True})
  File "/task_runtime/venv/lib/python3.5/site-packages/turicreate/toolkits/activity_classifier/_activity_classifier.py", line 196, in create
    model.train(dataset, target, session_id, validation_set, options)
  File "/task_runtime/venv/lib/python3.5/site-packages/turicreate/extensions.py", line 290, in <lambda>
    ret = lambda *args, **kwargs: self.__run_class_function(name, args, kwargs)
  File "/task_runtime/venv/lib/python3.5/site-packages/turicreate/extensions.py", line 277, in __run_class_function
    raise _ToolkitError(exc)
turicreate.toolkits._main.ToolkitError: An error occurred!
```

Disabling TF flag runs AC with the underlying GPUs.

fareeha:
The issue happens when TF installation is missing.

Fix:
To enable running TF with CPU: `pip install tensorflow`
To run using GPU(s): `pip install tensorflow-gpu`
Checked with version 1.14.0

PR #2404, opened 2019-10-04
Adds error logging for pybind11

shreyajain17:
Would print out the errors from Python side correctly. Won't show vague errors like #2373 anymore.


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -52,10 +52,12 @@ void call_pybind_function(const CallFunc& func) {
 
   try {
     func();
+  } catch (const pybind11::error_already_set& e) {
+    log_and_throw(std::string("An error occurred: ") + e.what());
   } catch (...) {
-    // TODO: Do better error logging
-    log_and_throw("An error occurred!");
+    log_and_throw("Unknown Error occurred");
```
nickjong: Nit: let's not capitalize "Error" here

Issue #2384, opened 2019-10-02
HTTPS downloads fail with Turi Create 5.7


syoutsey:
HTTPS downloads fail with Turi Create 5.7. This issue does not reproduce on Turi Create 5.6.

System: macOS 10.15 beta

Error: `OSError: Cannot open https://static.turi.com/datasets/audioscrobbler for read. Not implemented: compiled without support for http(s):// URLs.`

Reproduction steps - working version:

```
(venv) ➜  turi-create-3.6-5.6 pip list | grep turicreate
turicreate      5.6        
(venv) ➜  turi-create-3.6-5.6 python --version
Python 3.6.8
(venv) ➜  turi-create-3.6-5.6 python
Python 3.6.8 (v3.6.8:3c6b436a57, Dec 24 2018, 02:04:31) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import turicreate as tc
>>> sf = tc.SFrame('https://static.turi.com/datasets/audioscrobbler')
Downloading https://static.turi.com/datasets/audioscrobbler/dir_archive.ini to /var/folders/bt/d7_f7y897cz8f7256sj30j1c0000gn/T/turicreate-sam/41175/2c04d8ef-585a-4937-89f5-3c8ba78d52b9.ini
Downloading https://static.turi.com/datasets/audioscrobbler/objects.bin to /var/folders/bt/d7_f7y897cz8f7256sj30j1c0000gn/T/turicreate-sam/41175/1f709ccd-edfe-4cc0-9c4c-175ef6f38ef4.bin
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.frame_idx to /var/folders/bt/d7_f7y897cz8f7256sj30j1c0000gn/T/turicreate-sam/41175/aa6f3d62-9b01-4071-8d8c-db5291cf3276.frame_idx
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.sidx to /var/folders/bt/d7_f7y897cz8f7256sj30j1c0000gn/T/turicreate-sam/41175/561e4baa-7d91-4a81-a5a4-24d5785c5a9d.sidx
```

Reproduction steps - nonfunctional version:

```
(venv) ➜  turi-create-3.6-5.7 pip list | grep turicreate                                                    
turicreate      5.7        
[...]

PR #2405, opened 2019-10-04
bug fix for https load failure

Jarvi-Izana:
#2384

### DIFF
--- a/src/core/storage/sframe_interface/unity_sframe.cpp
+++ b/src/core/storage/sframe_interface/unity_sframe.cpp
@@ -101,12 +101,12 @@ void unity_sframe::construct_from_sframe_index(std::string location) {
 
   auto status = fileio::get_file_status(location);
 
-  // web protocol is a bad name, it should mean read-only protocol
-  if (status.first == fileio::file_status::REGULAR_FILE &&
+  if ((status.first == fileio::file_status::REGULAR_FILE ||
+       status.first == fileio::file_status::FS_UNAVAILABLE ||
+       status.first == fileio::file_status::MISSING) &&
       fileio::is_web_protocol(fileio::get_protocol(location))) {
     // if it is a web protocol, we cannot be certain what type of file it is.
     // HEURISTIC:
-    //   assume it is a "directory" and try to load dir_archive.ini
     //   if we can open it, it is a regular file. Otherwise not.
     if (fileio::try_to_open_file(location + "/dir_archive.ini")) {
       status.first = fileio::file_status::DIRECTORY;
--- a/src/core/storage/sframe_interface/unity_sframe.cpp
+++ b/src/core/storage/sframe_interface/unity_sframe.cpp
@@ -101,8 +101,8 @@ void unity_sframe::construct_from_sframe_index(std::string location) {
 
   auto status = fileio::get_file_status(location);
 
-  if ((status.first == fileio::file_status::FS_UNAVAILABLE ||
-       status.first == fileio::file_status::MISSING) &&
+  // web protocol is a bad name, it should mean read-only protocol
+  if (status.first == fileio::file_status::REGULAR_FILE &&
       fileio::is_web_protocol(fileio::get_protocol(location))) {
     // if it is a web protocol, we cannot be certain what type of file it is.
     // HEURISTIC:
@@ -110,10 +110,6 @@ void unity_sframe::construct_from_sframe_index(std::string location) {
     //   if we can open it, it is a regular file. Otherwise not.
     if (fileio::try_to_open_file(location + "/dir_archive.ini")) {
       status.first = fileio::file_status::DIRECTORY;
-      status.second.clear();
-    }
-    else {
-      status.first = fileio::file_status::REGULAR_FILE;
       status.second.clear();
     }
   }
@@ -1165,7 +1161,7 @@ std::shared_ptr<unity_sframe_base> unity_sframe::sample(float percent,
   }
   auto logical_filter_array = std::static_pointer_cast<unity_sarray>(
     unity_sarray::make_uniform_boolean_array(size(), percent, random_seed, exact));
-  
+
   return logical_filter(logical_filter_array);
 }
 


Jarvi-Izana:
It takes time for my fat fingers. ;-)

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -101,8 +101,8 @@ void unity_sframe::construct_from_sframe_index(std::string location) {
 
   auto status = fileio::get_file_status(location);
 
-  if ((status.first == fileio::file_status::FS_UNAVAILABLE ||
-       status.first == fileio::file_status::MISSING) &&
+  // web protocol is a bad name, it should mean read-only protocol
```
znation: This doesn't seem like the right fix to me; status shouldn't give `REGULAR_FILE` for a directory path over http(s). Is that the behavior you're seeing for URLs to SFrames?

Jarvi-Izana: But it actually works. Under the hood, it will try to figure out the instance behind the protocol is a directory contains sframe.idx or it's a pure file. If you read the `get_file_status`, it will blindly tell you the URI points to a file.

Jarvi-Izana:
```
tc.SFrame('https://static.turi.com/datasets/audioscrobbler')
Downloading https://static.turi.com/datasets/audioscrobbler/dir_archive.ini to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/02d09699-2bdf-4164-8ac2-bb578c36cf1a.ini
Downloading https://static.turi.com/datasets/audioscrobbler/objects.bin to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/fd45f74e-7604-41a2-9fcd-3993d0a9c734.bin
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.frame_idx to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/07f16828-af8d-45ea-b2c0-37d672b77823.frame_idx
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.sidx to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/e9e00719-1c4c-4cc1-92f5-5efc39fcb005.sidx
Out[3]: Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0000 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/e23d74e9-c13b-413e-a131-88daec109dce.0000
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0001 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/db7664e2-2c6e-41b3-955b-cd6975223b1b.0001
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0002 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/47fc0d15-ae41-4c1b-a742-6097155fb7f7.0002
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0003 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/d199c616-56a6-463f-a55a-c131a51bc3c1.0003
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0004 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/caf93341-386d-4563-a63b-8c03c6ef02db.0004
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0005 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/4a6e7edd-abeb-4e7e-9dd7-e63c655c96a5.0005
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0006 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/36221993-abd7-4b8c-ae2f-08f944a03f2b.0006
Downloading https://static.turi.com/datasets/audioscrobbler/m_74b0dc51.0007 to /var/folders/n9/wscz46nj4_10nz3f2y2r0pnr0000gn/T/turicreate-guihaoliang/6732/30c0d6aa-0b05-4b60-8647-f0a9e9f68e0a.0007

Columns:
        user_id str
        item_id str
        target  int

Rows: 418036

Data:
+---------+----------------+--------+
| user_id |    item_id     | target |
+---------+----------------+--------+
| 1001606 | Raver's Nature |   2    |
| 1002598 | Raver's Nature |   1    |
| 1009768 | Raver's Nature |   29   |
| 1020855 | Raver's Nature |   9    |
| 1024677 | Raver's Nature |   1    |
| 1025593 | Raver's Nature |   3    |
| 1036036 | Raver's Nature |   51   |
| 1037250 | Raver's Nature |   3    |
| 1037372 | Raver's Nature |   1    |
| 1037666 | Raver's Nature |   1    |
+---------+----------------+--------+
[418036 rows x 3 columns]
```

znation: I get that it works, but to me that is very strange and something doesn't seem right. The directory path doesn't actually exist at that URL -- it should give a 404 not found, which I would think would translate into a `MISSING` rather than `REGULAR_FILE`. Is it possible the bug is deeper in the fileio code?

Jarvi-Izana: oh, right. I forget to push the other code. Just a minute.

Jarvi-Izana: nope, the GitHub folds that section:
```
if (fileio::try_to_open_file(location + "/dir_archive.ini")) {
      status.first = fileio::file_status::DIRECTORY;
      status.second.clear();
    }
```
If it's non writeable remote file, it doesn't do further ado to make sure it exist in remote or not. So here, we have extra logic to try to figure out it means a directory or a file.

Jarvi-Izana: You might need extra credentials to see the dir. For me, I cannot curl the URL neither.

Jarvi-Izana:
since the data in `https://static.turi.com/datasets/audioscrobbler` is small; only ~3MB, I will add it into our unittest.

src/core/storage/sframe_interface/unity_sframe.cpp
hoytak: @znation -- this was triggered by the https being a thin wrapper to an s3 service, so it ends up not being a directory because s3 doesn't have directories, just buckets and keys.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -101,8 +101,8 @@ void unity_sframe::construct_from_sframe_index(std::string location) {
...
+  if (status.first == fileio::file_status::REGULAR_FILE &&
```
hoytak: Shouldn't this test for both regular file and missing?  Why do we ignore the missing path?

znation: I actually don't get why it works in the repro case when only testing for regular. The repro involves a 404 at the directory path -- I would think this would come in as MISSING.

Jarvi-Izana: for remote fs, unlike hdfs or s3, http/https protocol is not handled eagerl, and the URI will be marked as regular file instead of marking it as MISSING or UNAVAILABLE (at least this is what original `get_file_status` does). 

Now we can understand, regular file can either be eagerly verified (s3, local file), or can be file behind protocol like http/https, whose real file status verification is deferred until reading.

znation: That's really strange. Was this always the behavior? How did the regression happen on this issue, and what code change caused it?

Jarvi-Izana: Yeah. The regression is that I modified this file by moving http probing logic into this deleted section. 

The original behavior is `try_open_file` blindly regardless the file status returned from `get_file_status`, 
```
auto status = fileio::get_file_status(location);
  if (fileio::is_web_protocol(fileio::get_protocol(location))) {
    // if it is a web protocol, we cannot be certain what type of file it is.
    // HEURISTIC:
    //   assume it is a "directory" and try to load dir_archive.ini
    //   if we can open it, it is a regular file. Otherwise not.
    if (fileio::try_to_open_file(location + "/dir_archive.ini")) {
      status = fileio::file_status::DIRECTORY;
    } else {
      status = fileio::file_status::REGULAR_FILE;
    }
  }
```
it discards the info of MISSING and FS_UNAVAILABLE, which is problematic.

In this PR, 
1. for MISSING and FS_UNAVAILABLE, we can skip probing which saves network. 
2. For DIRECTORY, no need to probe. 
3. For REGULAR file, if it's so called `web_protocol`, we need to probe its real file status eagerly through network. For s3 or hdfs, the really file status is already probed through network.

Essentially, we should move the probing into `get_file_status`. But here, let's get it in.

src/core/storage/sframe_interface/unity_sframe.cpp
Jarvi-Izana: @hoytak Thanks for the explanation!

src/core/storage/sframe_interface/unity_sframe.cpp
Jarvi-Izana: @hoytak No need to handle MISSING or UNAVAILABLE since these file statuses are verified eagerly through network already.

Jarvi-Izana:
I talked with @hoytak and come up with a more readable version. We will handle the problematic `is_web_protocol` later when we have time.

Issue #1367, opened 2019-01-29
SFrame.read_csv()'s API documentation is missing an argument


igiloh:
In the [API doc](https://apple.github.io/turicreate/docs/api/generated/turicreate.SFrame.read_csv.html?highlight=nrows_to_infer) of `SFrame.read_csv()`, there's no description of the `nrows_to_infer` param.

PR #2408, opened 2019-10-05
Doc: Add documentation for SFrame.read_csv() nrows_to_infer

tapaswenipathak:
Fixes https://github.com/apple/turicreate/issues/1367.


src/python/turicreate/data_structures/sframe.py
```diff
@@ -1270,6 +1270,9 @@ def read_csv(cls,
             continue parsing skipping lines which fail to parse correctly.
             A sample of the first 10 encountered bad lines will be printed.
 
+        nrows_to_infer: integer
```
TobyRoseman: Let's have the ordering of parameters in the docstring match the order of parameters in the function definition.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -1270,6 +1270,9 @@ def read_csv(cls,
...
+            Get the number of nrows_to_infer rows enabling the arguments as specified.
```
TobyRoseman: I don't really understand this description. I would just say something like: `The number of rows used to infer column types.`

tapaswenipathak:
@TobyRoseman: review ping.

TobyRoseman:
@tapaswenipathak - thanks for the ping. Also thanks for the helpful pull request and addressing my feedback.

PR #2411, opened 2019-10-06
Added a skeleton for Drawing Classifier C++

shantanuchhabra:


src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -0,0 +1,28 @@
+/* Copyright © 2019 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#include <toolkits/drawing_classifier/dc_data_iterator.hpp>
```
TobyRoseman: Do we need this empty cpp file?

If we do need it, we probably don't need all these `#include` statements

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+  END_CLASS_MEMBER_REGISTRATION
+
+ protected:
+  // Override points allowing subclasses to inject dependencies
+
+  // Factory for data_iterator
+  virtual std::unique_ptr<data_iterator> create_iterator(
+      gl_sframe data, bool requires_labels, bool is_train,
+      bool use_data_augmentation) const;
```
shreyajain17: Will DC have data augmentation?

src/toolkits/drawing_classifier/dc_data_iterator.hpp
```diff
@@ -0,0 +1,164 @@
...
+  };
+
+  static preprocessed_data preprocess_data(const parameters& params);
+
+  const preprocessed_data data_;
+  
+  gl_sframe_range range_iterator_;
+  gl_sframe_range::iterator next_row_;
+  gl_sframe_range::iterator end_of_rows_;
+  size_t sample_in_row_ = 0;
```
shreyajain17: So this was added for AC to have data augmentation. You should maybe look at OD for iterator stuff!

src/toolkits/drawing_classifier/dc_data_iterator.cpp
nickjong: Let's not add empty `.cpp` files. Files should generally be added "lazily"

src/toolkits/drawing_classifier/dc_data_iterator.hpp
nickjong: For the iterator specifically, OD (without data augmentation or annotations) is a closer fit for what I expect DC needs. Please don't copy and paste blindly; only copy (or factor out!) what you need!

Note that we don't necessarily need to define this class in the initial commit. The entire data iterator class (even its header) is an implementation detail from the point of view of the `drawing_classifier` class

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -0,0 +1,40 @@
...
+using coreml::MLModelWrapper;
+using neural_net::compute_context;
+using neural_net::float_array_map;
+using neural_net::model_backend;
+using neural_net::model_spec;
+using neural_net::shared_float_array;
+using neural_net::xavier_weight_initializer;
+using neural_net::zero_weight_initializer;
+
+using padding_type = model_spec::padding_type;
```
nickjong: Let's only add imports and using declarations as we need them. I think we should also go ahead and provide stub implementations for any public methods (exposed in C++ or via C-API)

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -0,0 +1,250 @@
...
+             variant_type validation_data,
+             std::map<std::string, flexible_type> opts);
+  gl_sarray predict(gl_sframe data, std::string output_type);
+  gl_sframe predict_topk(gl_sframe data, std::string output_type,
+    size_t k, size_t batch_size);
+  variant_map_type evaluate(gl_sframe data, std::string metric,
+    size_t batch_size);
+  std::shared_ptr<coreml::MLModelWrapper> export_to_coreml(
+      std::string filename);
+  void load_pretrained_model(variant_map_type model_data, size_t version);
```
nickjong: What is this for? Why is loading a pre-trained model part of the public API, outside of the options for `train`?

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -0,0 +1,250 @@
...
+      "    prevent the model from overfitting to the training data.\n"
+      "\n"
+      "    For each row of the progress table, accuracy is measured over the\n"
+      "    provided training dataset and the `validation_data`. The format of\n"
+      "    this SFrame must be the same as the training set.\n"
+      "\n"
+      "    When set to 'auto', a validation set is automatically sampled from "
+      "the\n"
+      "    training data (if the training data has > 100 sessions).\n"
+      "warm_start : string\n"
```
nickjong: Look at how the OD class works in this case. Chances are for now we just want to pass a path to the pre-trained model on disk. This we can always add later anyway, we can focus for now on getting cold-start working

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -0,0 +1,250 @@
...
+
+  // Interface exposed via Unity server
+
+  void train(gl_sframe data,
+             std::string target_column_name, std::string feature_column_name,
+             variant_type validation_data,
+             std::map<std::string, flexible_type> opts);
+  gl_sarray predict(gl_sframe data, std::string output_type);
+  gl_sframe predict_topk(gl_sframe data, std::string output_type,
+    size_t k, size_t batch_size);
```
nickjong: Why does `batch_size` need to be a top-level argument? Could we omit this for now?

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -0,0 +1,250 @@
...
+  gl_sframe validation_data_;
+  std::unique_ptr<data_iterator> training_data_iterator_;
+  std::unique_ptr<data_iterator> validation_data_iterator_;
+  std::unique_ptr<neural_net::compute_context> training_compute_context_;
+  std::unique_ptr<neural_net::model_backend> training_model_;
+
+  // Nonnull while training is in progress, if progress printing is enabled.
+  std::unique_ptr<table_printer> training_table_printer_;
+};
+
```
nickjong: We don't necessarily have to define anything private or protected in this PR. Let's focus on the interface, and having it callable. In particular, after your PR, we should be able to do the following:
```
import turicreate as tc
model = tc.extensions.drawing_classifier()
model.train(data, 'target', 'drawing') # Shouldn't crash. Might do nothing now, but shouldn't crash.
preds = model.predict(data) # Should return an SArray
```

nickjong: @abhishekpratapa I think this would be a good first step for Style Transfer, too. First get to the point where we can instantiate a (mostly unimplemented) C++ object from Python. Then we can begin developing and testing from Python using the low-level extensions wrapper around C++

src/ml/neural_net/tf_compute_context.hpp
```diff
@@ -41,6 +41,13 @@ EXPORT class tf_compute_context: public compute_context {
       int n, int c_in, int h_in, int w_in, int c_out, int h_out, int w_out,
       const float_array_map& config, const float_array_map& weights) override;
 
+  std::unique_ptr<model_backend> create_drawing_classifier(
+      const variant_type &validation_set,
```
shreyajain17: I also don't think validation set should be passed. I don't think variant type must be passed from C++ to python while making the model. It must be passed using the iterator while training and that's it.

nickjong:
Why are we adding compute context changes to this PR? Let's just focus on getting this in with empty stubs first

src/toolkits/drawing_classifier/drawing_classifier.hpp
shantanuchhabra: Gonna take it out, since we should think about this properly at a later point.

src/toolkits/drawing_classifier/class_registrations.cpp
```diff
@@ -5,6 +5,7 @@
  */
 #include "class_registrations.hpp"
 #include "data_preparation.hpp"
+#include "drawing_classifier.hpp"
```
nickjong: Include statements in our codebase should generally be full paths

src/toolkits/drawing_classifier/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ project(unity_toolkits)
 
 make_library(unity_drawing_classifier OBJECT
 SOURCES
+	drawing_classifier.cpp
```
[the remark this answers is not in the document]
shantanuchhabra: Did not intend to tab, I hate tab. My bad!

src/toolkits/drawing_classifier/class_registrations.cpp
shantanuchhabra: Done.

src/toolkits/drawing_classifier/CMakeLists.txt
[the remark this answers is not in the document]
shantanuchhabra: Fixed.

PR #2413, opened 2019-10-07
MPS style Transfer Base Test Setup

abhishekpratapa:
## Overview
The definition for the MPS Base Test setup makes writing all of the style transfer tests easier. This base class can handle a majority of the situations for checking the network correctness. The child class only needs to define the network to be tested and provide expected input and output files.


test/neural_net/CMakeLists.txt
```diff
@@ -6,6 +6,22 @@ if(APPLE AND
...
+    REQUIRES
+      unity_shared_for_testing
+  )
+
+  target_compile_options(style_transfer_test_utils 
+    PUBLIC 
+      "-fobjc-arc -Wno-nullability-completeness -Wno-shadow-ivar"
+  )
+  
+  make_boost_test(style_transfer.cxx
```
nickjong: Nit: for clarity, could we name this file `style_transfer_tests.cxx` or `test_style_transfer.cxx`?

test/neural_net/style_transfer.cxx
```diff
@@ -89,22 +89,138 @@ struct style_transfer_test {
   }
 
   /**
-    Loss
+    Pre Processing
 
-    This test checks whether the Loss out of the total graph defined in TCMPS is
+    This correctness test checks whether the Pre Processing Node present in
+    TCMPS is performing as intended by comparing it to the golden set.
```
nickjong: What will be the general strategy for producing/maintaining the golden set? I'm wary of committing significant binary data to the repo if we can avoid it

test/neural_net/style_transfer.cxx
```diff
@@ -129,9 +245,57 @@ BOOST_AUTO_TEST_CASE(test_block_2) { style_transfer_test::test_block_2(); }
 
 BOOST_AUTO_TEST_CASE(test_vgg16) { style_transfer_test::test_vgg16(); }
 
-BOOST_AUTO_TEST_CASE(test_loss) { style_transfer_test::test_loss(); }
+BOOST_AUTO_TEST_CASE(test_pre_processing) {
+  style_transfer_test::test_pre_processing();
```
nickjong: Ignore some precedents you see in our code; there's no reason for this to call a separate `style_transfer_test::test_pre_processing()` function from here. Just define that function here. There's no reason to have every test named in two places

test/neural_net/style_transfer_testing.hpp
```diff
@@ -0,0 +1,63 @@
...
+#ifndef __TEST_NEURAL_NET_STYLE_TRANSFER_TESTING_HPP
+#define __TEST_NEURAL_NET_STYLE_TRANSFER_TESTING_HPP
+
+#include <functional>
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace neural_net_test {
```
nickjong: It will almost certainly be most convenient to use whatever namespace the code under test is in. In this case, `turi::neural_net`.

test/neural_net/style_transfer_testing.hpp
```diff
@@ -0,0 +1,63 @@
...
+              inference and the expected inference exceeds an epsilon of 5e-3.
+              If this epsilon is exceeded, the method returns `false`. If the
+              element diff is within this epsilon the method returns `true`.
+  */
+
+  virtual bool check_predict(std::string input, std::string output);
+
+ protected:
+  struct common_impl;
+  std::unique_ptr<common_impl> internal_impl;
```
nickjong: Thanks for working out how to get ObjC code into our C++ tests!

This data member, `internal_impl` needs a name like `m_internal_impl` or `internal_impl_`

test/neural_net/style_transfer_testing.hpp
```diff
@@ -0,0 +1,63 @@
...
+class BaseNetworkTest {
+ public:
+  BaseNetworkTest() = delete;
+  explicit BaseNetworkTest(float epsilon);
+
+  ~BaseNetworkTest();
+
+  /**
+    Checks the encoding prediction by using an input configuration and an input
+    bin file. The configuration file has 3 major keys:
```
nickjong: What is a "bin file"? What is a "configuration file?" Might need some higher-level context here

test/neural_net/style_transfer_testing_utils.hpp
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#import <Accelerate/Accelerate.h>
+#import <Foundation/Foundation.h>
+#import <Metal/Metal.h>
+#import <MetalPerformanceShaders/MetalPerformanceShaders.h>
+
+#include <boost/property_tree/ptree.hpp>
+
+@interface NeuralNetStyleTransferUtils : NSObject
```
nickjong: Not sure what's going on here. I'm wondering if we need some design review before we proceed further....

test/neural_net/CMakeLists.txt
abhishekpratapa: Sounds Good!

test/neural_net/style_transfer.cxx
abhishekpratapa: I'll have a separate files in Python folder that'll generate the golden set.

In the build `YAML` files I'll have a script to get to generate the golden set before the tests are built. That way the golden set doesn't have to directly be committed to the repo. And during the build process it gets generated.

test/neural_net/style_transfer.cxx
abhishekpratapa: Sounds good, it simplifies the code for sure.

test/neural_net/style_transfer_testing.hpp
abhishekpratapa: Alright!

test/neural_net/style_transfer_testing.hpp
abhishekpratapa: The bin file contains the data, the json files contain the format in which the data is stored. For example let's take an array as an input.

The array has a shape, and data. The shape information would go in the config file, whereas the actual bytes would be stored in the bin file. 

These arrays could be an inputs or outputs. 

There could be multiple array's (multiple inputs and outputs) as well so bin and configuration files seemed to be the easiest way to go for testing. Probably not ideal for production.

Earlier I had a JSON file containing both the format data and the actual data. That significantly slowed my testing process since I was directly compiling this json file into the test executable and parsing the json for the weights. Both compile time and run-time were very slow using this method.

A potential solution to this problem:
- Use Protobuf to serialize and deserialize the data.

If there are any other I'd love to hear!

test/neural_net/style_transfer_testing_utils.hpp
abhishekpratapa: I commented above on how I was loading the golden set for testing into this testing scaffolding. These functions help with that.

PR #2415, opened 2019-10-07
Darknet mlmodel hosted and called in OD

fareeha:
* `darknet.mlmodel` hosted on `devpubs`
* Added model loading in OD

To inspect the meta data in the mlmodel file: https://docs-assets.developer.apple.com/turicreate/models/darknet.mlmodel

### DIFF
--- a/src/python/turicreate/toolkits/_pre_trained_models.py
+++ b/src/python/turicreate/toolkits/_pre_trained_models.py
@@ -177,9 +177,21 @@ class DarkNetObjectDetectorBase(ObjectDetectorBasePreTrainedModel):
         ])
         return subset_params
 
+class DarkNetObjectDetectorModel(ObjectDetectorBasePreTrainedModel):
+    def __init__(self):
+        self.name = 'darknet'
+        self.source_url = _urlparse.urljoin(MODELS_URL_ROOT, 'darknet.mlmodel')
+        self.source_md5 = '2fbe8725b8041c31467dd25cf9f65aed'
+
+    def get_model_path(self):
+        model_path = _download_and_checksum_files(
+            [(self.source_url, self.source_md5)], _get_cache_dir()
+            )[0]
+        return model_path
 
 OBJECT_DETECTION_BASE_MODELS = {
     'darknet': DarkNetObjectDetectorBase,
+    'darknet_mlmodel': DarkNetObjectDetectorModel
 }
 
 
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -262,7 +265,7 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
         'io_thread_buffer_size': 8,
         # TODO: Have the Python side download the Core ML version of the
```
znation: You should remove this TODO comment now that it's done.

fareeha: Thanks for the catch!

src/python/turicreate/toolkits/_pre_trained_models.py
```diff
@@ -177,9 +177,21 @@ def available_parameters_subset(self, mx_params):
         ])
         return subset_params
 
+class DarkNetObjectDetectorModel(ObjectDetectorBasePreTrainedModel):
```
nickjong: Do we still need `DarkNetObjectDetectorBase` above?

That name seems more accurate, since this model is also just a base for an OD model, not a complete OD model itself

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -219,6 +219,9 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
     base_model = model.split('-', 1)[0]
     ref_model = _pre_trained_models.OBJECT_DETECTION_BASE_MODELS[base_model]()
 
+    pretrained_model = _pre_trained_models.OBJECT_DETECTION_BASE_MODELS['darknet_mlmodel']()
+    pretrained_model_path = pretrained_model.get_model_path()
```
nickjong: Doesn't this mean we're always downloading both the MXNet model and the Core ML model?

What gets printed to console when this code runs? It looks like we should be reaching this print statement twice: https://github.com/apple/turicreate/blob/4e191b87b028af7ae78a7d3491890a0887b9d013/src/python/turicreate/toolkits/_pre_trained_models.py#L62

fareeha: Yes we are downloading both MXNET and CoreML model and the print is executed twice (if the models are not stored in cache). I was wondering the same if we need both.

We use the darknet mxnet model in two places:
* to get `spatial_reduction` field - which we can add to the CoreML model init
* loading weights for the MXNET case when not using TF

nickjong: Okay, that makes sense as long as this is a transient state. I guess no matter what we need a later change to remove the old codepath.

PR #2416, opened 2019-10-07
Style transfer Data Iterator

abhishekpratapa:
## Overview

Setup the class for the Style Transfer Data Iterator. It takes in data iterator params. and the relevant method on this iterator is `next_batch`. `next_batch` returns a batch containing a tie of `content_image`, `style_image`, and `style_index`.

### DIFF
--- a/src/toolkits/style_transfer/style_transfer_data_iterator.cpp
+++ b/src/toolkits/style_transfer/style_transfer_data_iterator.cpp
@@ -5,19 +5,17 @@
  * https://opensource.org/licenses/BSD-3-Clause
  */
 
-#include "style_transfer_data_iterator.hpp"
+#include <toolkits/style_transfer/style_transfer_data_iterator.hpp>
 
 #include <core/data/image/io.hpp>
+#include <core/data/sframe/gl_sframe.hpp>
+#include <ml/neural_net/image_augmentation.hpp>
 #include <model_server/lib/image_util.hpp>
-
-#include <iostream>
 
 namespace turi {
 namespace style_transfer {
 
-using neural_net::image_annotation;
-using neural_net::labeled_image;
-using neural_net::shared_float_array;
+namespace {
 
 flex_image get_image(const flexible_type& image_feature) {
   if (image_feature.get_type() == flex_type_enum::STRING) {
@@ -27,19 +25,13 @@ flex_image get_image(const flexible_type& image_feature) {
   }
 }
 
-std::vector<flexible_type> get_style(const data_iterator::parameters& params) {
+gl_sarray get_style(const data_iterator::parameters& params) {
   gl_sarray style_images = params.style;
 
   if (style_images.dtype() == flex_type_enum::IMAGE)
     style_images = style_images.apply(image_util::encode_image, flex_type_enum::IMAGE);
 
-  std::vector<flexible_type> style_batch;
-  style_batch.reserve(style_images.size());
-
-  for (const auto& i : style_images.range_iterator())
-    style_batch.emplace_back(i);
-
-  return style_batch;
+  return style_images;
 }
 
 gl_sarray get_content(const data_iterator::parameters& params) {
@@ -51,17 +43,19 @@ gl_sarray get_content(const data_iterator::parameters& params) {
   return content_images;
 }
 
+}
+
 style_transfer_data_iterator::style_transfer_data_iterator(
     const data_iterator::parameters& params)
-    : m_style_vector(get_style(params)),
-      m_content(get_content(params)),
+    : m_style_images(get_style(params)),
+      m_content_images(get_content(params)),
       m_repeat(params.repeat),
       m_shuffle(params.shuffle),
-      m_content_range_iterator(m_content.range_iterator()),
+      m_content_range_iterator(m_content_images.range_iterator()),
       m_content_next_row(m_content_range_iterator.begin()),
       m_random_engine(params.random_seed) {}
 
-std::vector<st_image> style_transfer_data_iterator::next_batch(
+std::vector<st_example> style_transfer_data_iterator::next_batch(
     size_t batch_size) {
 
   std::vector<std::tuple<flexible_type, flexible_type, flexible_type>>
@@ -71,15 +65,15 @@ std::vector<st_image> style_transfer_data_iterator::next_batch(
   while (raw_batch.size() < batch_size && m_content_next_row != m_content_range_iterator.end()) {
     const turi::flexible_type& content_image = *m_content_next_row;
 
-    std::uniform_int_distribution<size_t> dist(0, m_style_vector.size() - 1);
+    std::uniform_int_distribution<size_t> dist(0, m_style_images.size() - 1);
     size_t random_style_index = dist(m_random_engine);
-    const flexible_type& style_image = m_style_vector.at(random_style_index);
+    const flexible_type& style_image = m_style_images[random_style_index];
 
     raw_batch.emplace_back(content_image, style_image, random_style_index);
 
     if (++m_content_next_row == m_content_range_iterator.end() && m_repeat) {
       if (m_shuffle) {
-        gl_sarray indices = gl_sarray::from_sequence(0, m_content.size());
+        gl_sarray indices = gl_sarray::from_sequence(0, m_content_images.size());
         std::uniform_int_distribution<uint64_t> dist(0);
         uint64_t random_mask = dist(m_random_engine);
         auto randomize_indices = [random_mask](const flexible_type& x) {
@@ -87,26 +81,26 @@ std::vector<st_image> style_transfer_data_iterator::next_batch(
           return flexible_type(hash64(masked_index));
         };
 
-        gl_sframe temp_content({{"content", m_content}});
+        gl_sframe temp_content({{"content", m_content_images}});
         temp_content.add_column(
             indices.apply(randomize_indices, flex_type_enum::INTEGER, false),
             "_random_order");
 
         temp_content = temp_content.sort("_random_order");
-        m_content = temp_content["content"];
+        m_content_images = temp_content["content"];
       }
     }
   }
 
-  std::vector<st_image> result(raw_batch.size());
+  std::vector<st_example> result(raw_batch.size());
 
   for (size_t i = 0; i < raw_batch.size(); ++i) {
     flexible_type content_image, style_image, style_index;
     std::tie(content_image, style_image, style_index) = raw_batch[i];
 
-    result[i].style = get_image(style_image);
-    result[i].content = get_image(content_image);
-    result[i].index = style_index.get<flex_int>();
+    result[i].style_image = get_image(style_image);
+    result[i].content_image = get_image(content_image);
+    result[i].style_index = style_index.get<flex_int>();
   }
 
   return result;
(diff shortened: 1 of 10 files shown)


src/toolkits/style_transfer/style_transfer_data_iterator.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <model_server/lib/extensions/ml_model.hpp>
+#include <toolkits/coreml_export/mlmodel_wrapper.hpp>
+
+namespace turi {
+namespace style_transfer {
+
+struct st_image {
+  image_type style;
+  image_type content;
+  size_t index;
```
nickjong: I'm going to obsess over names, because I think names are important.

For the two images, could we go with a more complete name, like `style_image` and `content_image`? And I presume `index` is `style_index`?

Maybe `st_image` could be `st_example`?

This what we get more readable code like:
```
st_example example = ...;
do_stuff(example.style_image, example.content_image);
```

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * The name of the column containing the style images.
+     */
+    std::string style_column_name;
+
+    /**
+     * The name of the column containing the content images.
+     */
+    std::string content_column_name;
```
nickjong: Any reason not just to pass two `gl_sarray` values, one for the styles and one for the contents?

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
```diff
@@ -0,0 +1,96 @@
...
+  style_transfer_data_iterator(const parameters& params);
+
+  style_transfer_data_iterator(const style_transfer_data_iterator&) = delete;
+  style_transfer_data_iterator& operator=(const style_transfer_data_iterator&) =
+      delete;
+
+  std::vector<st_image> next_batch(size_t batch_size) override;
+
+ private:
+  std::vector<flexible_type> m_style_vector;
```
nickjong: Could this just be `m_style_images`?

Just curious, why prefer `vector` over `gl_sarray`?

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
```diff
@@ -0,0 +1,123 @@
+/* Copyright © 2019 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#include "style_transfer_data_iterator.hpp"
```
nickjong: We should always use full paths in our `include` directives, and angle brackets

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <map>
+#include <memory>
+
+#include <core/data/sframe/gl_sframe.hpp>
+#include <core/logging/table_printer/table_printer.hpp>
+#include <ml/neural_net/compute_context.hpp>
+#include <ml/neural_net/model_backend.hpp>
+#include <ml/neural_net/model_spec.hpp>
+#include <model_server/lib/extensions/ml_model.hpp>
+#include <toolkits/coreml_export/mlmodel_wrapper.hpp>
```
nickjong: What are these `include`s for?

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
```diff
@@ -0,0 +1,96 @@
...
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#ifndef __TOOLKITS_STYLE_TRANSFER_H_
+#define __TOOLKITS_STYLE_TRANSFER_H_
+
+#include <functional>
+#include <map>
+#include <memory>
```
nickjong: I don't see any of these being used. You do however use `string` and `random`

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <model_server/lib/image_util.hpp>
+
+namespace turi {
+namespace style_transfer {
+
+using neural_net::image_annotation;
+using neural_net::labeled_image;
+using neural_net::shared_float_array;
+
+flex_image get_image(const flexible_type& image_feature) {
```
nickjong: Helper functions local to a `.cpp` file should either be declared static or put inside an anonymous namespace

https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+std::vector<flexible_type> get_style(const data_iterator::parameters& params) {
+  gl_sarray style_images = params.style[params.style_column_name];
+
+  if (style_images.dtype() == flex_type_enum::IMAGE) {
+    style_images =
+        style_images.apply(image_util::encode_image, flex_type_enum::IMAGE);
+  }
+
+  gl_sframe result({{params.style_column_name, style_images}});
```
nickjong: Why not just use `gl_sarray::range_iterator`, instead of iterating over an SFrame?

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
abhishekpratapa: Copied from OD. I'll go through and clean up the includes.

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
abhishekpratapa: Copied from OD. I'll go through and clean up the includes.

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
abhishekpratapa: It was an artifact left over from me testing things. I'll refactor to use `gl_sarray`

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
abhishekpratapa: It's the way we do it in `Python` but that's a terrible reason. I'll refactor!

src/toolkits/style_transfer/style_transfer_data_iterator.hpp
abhishekpratapa: No reason just more familiar with vector, refactored.

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
abhishekpratapa: Irrelevant now.

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
```diff
@@ -25,95 +25,82 @@ flex_image get_image(const flexible_type& image_feature) {
...
-    style_batch.emplace_back(style_image);
-  }
-
-  return style_batch;
+  return style_images;
 }
 
-gl_sframe get_content(const data_iterator::parameters& params) {
-  gl_sarray content_images = params.style[params.content_column_name];
+gl_sarray get_content(const data_iterator::parameters& params) {
```
nickjong: At this point, it's probably cleanest to unify these into something like `encode_images_if_necessary` or `ensure_encoded` and have the argument be the `gl_sarray`

test/toolkits/style_transfer/test_style_transfer_data_iterator.cxx
```diff
@@ -0,0 +1,42 @@
...
+#define BOOST_TEST_MODULE
+
+#include <boost/test/unit_test.hpp>
+#include <core/util/test_macros.hpp>
+
+#include <toolkits/style_transfer/style_transfer_data_iterator.hpp>
+
+#include <string>
+
+#include "utils.hpp"
```
nickjong: Let's continue to avoid header includes with quotation marks

test/toolkits/style_transfer/test_style_transfer_data_iterator.cxx
```diff
@@ -0,0 +1,42 @@
...
+  data_iterator::parameters params;
+
+  params.style = style_sarray;
+  params.content = content_sarray;
+
+  style_transfer_data_iterator iter(params);
+
+  for (size_t x = 0; x < expected_batch_size; x++) {
+    std::vector<st_example> test_batch = iter.next_batch(test_batch_size);
+    TS_ASSERT(test_batch.size() == expected_batch_array[x]);
```
nickjong: Can we also assert that the images are the expected random image?

test/toolkits/style_transfer/test_style_transfer_data_iterator.cxx
```diff
@@ -0,0 +1,42 @@
...
+#include <core/util/test_macros.hpp>
+
+#include <toolkits/style_transfer/style_transfer_data_iterator.hpp>
+
+#include <string>
+
+#include "utils.hpp"
+
+using namespace style_transfer_testing;
+using namespace turi::style_transfer;
```
nickjong: Could we instead just declare/define all the test code in `turi::style_transfer` directly?

abhishekpratapa: done!

test/toolkits/style_transfer/test_style_transfer_data_iterator.cxx
abhishekpratapa: Good idea! Added those tests in!

test/toolkits/style_transfer/test_style_transfer_data_iterator.cxx
abhishekpratapa: Turned out to be a bit tricky for tests. Don't know how to address that but I added an include directories to the current CMakeLists. Maybe that's not the answer. Let me know what you think!

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
abhishekpratapa: Refactored!

test/toolkits/style_transfer/test_style_transfer_data_iterator.cxx
```diff
@@ -7,26 +7,23 @@
 
 #define BOOST_TEST_MODULE
 
+#include <string>
+#include <utils.hpp>
```
nickjong: Did you try <toolkits/style_transfer/utils.hpp>?

If that path also exists on the non-test side, maybe give it a different name…

abhishekpratapa: Path wasn't directly available to us. Fixed this however by putting the include_directories in the top CMakeLists under the test directory.

nickjong: What do you mean by "path wasn't directly available to us?"

@hoytak @znation How is test code supposed to import headers defined under `/test`? Or do we just not support headers there? Does test utility code need to be in `/src`?

znation: I think I've done this before by using relative include paths (like `"../..."`). I think `<test/...>` should also work though I haven't tried it.

hoytak: Sorry to be late to the game here -- but yes, relative imports in the test directory are a good idea.  We should not risk src/ code picking up things from the test/ directory.

PR #2419, opened 2019-10-09
Fix "random_seed" state not found" bug for AC import for python model

### LINKED ISSUE #2378: AC import from custom model expects "random_seed" state not present in custom models

jakesabathia2:
close #2378 
Fix the import function for python model.

### DIFF
--- a/src/ml/neural_net/weight_init.cpp
+++ b/src/ml/neural_net/weight_init.cpp
@@ -28,63 +28,59 @@ xavier_weight_initializer::xavier_weight_initializer(
       random_engine_(*random_engine)
 {}
 
-  void xavier_weight_initializer::operator()(float* first_weight,
-                                           float* last_weight)
-  {
-    for (float* w = first_weight; w != last_weight; ++w) {
+void xavier_weight_initializer::operator()(float* first_weight,
+                                           float* last_weight) {
+  for (float* w = first_weight; w != last_weight; ++w) {
     *w = dist_(random_engine_);
-    }
   }
+}
 
-  // static
-  lstm_weight_initializers lstm_weight_initializers::create_with_xavier_method(
-    size_t input_size, size_t state_size, std::mt19937* random_engine)
-  {
-    xavier_weight_initializer i2h_init_fn(input_size, state_size,
-                                          random_engine);
-    xavier_weight_initializer h2h_init_fn(state_size, state_size,
-                                          random_engine);
+// static
+lstm_weight_initializers lstm_weight_initializers::create_with_xavier_method(
+    size_t input_size, size_t state_size, std::mt19937* random_engine) {
+  xavier_weight_initializer i2h_init_fn(input_size, state_size, random_engine);
+  xavier_weight_initializer h2h_init_fn(state_size, state_size, random_engine);
 
-    lstm_weight_initializers result;
+  lstm_weight_initializers result;
 
-    result.input_gate_weight_fn = i2h_init_fn;
-    result.forget_gate_weight_fn = i2h_init_fn;
-    result.block_input_weight_fn = i2h_init_fn;
-    result.output_gate_weight_fn = i2h_init_fn;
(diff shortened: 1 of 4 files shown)


src/ml/neural_net/weight_init.cpp
```diff
@@ -40,28 +40,51 @@ xavier_weight_initializer::xavier_weight_initializer(
...
+    result.output_gate_recursion_fn = h2h_init_fn;
 
-  result.input_gate_bias_fn = zero_weight_initializer();
-  result.forget_gate_bias_fn = zero_weight_initializer();
-  result.block_input_bias_fn = zero_weight_initializer();
-  result.output_gate_bias_fn = zero_weight_initializer();
+    result.input_gate_bias_fn = zero_weight_initializer();
+    result.forget_gate_bias_fn = zero_weight_initializer();
+    result.block_input_bias_fn = zero_weight_initializer();
+    result.output_gate_bias_fn = zero_weight_initializer();
```
nickjong: Why are we changing the indent level here? Didn't we already address this in the last PR?

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -636,12 +639,19 @@ std::unique_ptr<data_iterator> activity_classifier::create_iterator(
   data_params.prediction_window = read_state<flex_int>("prediction_window");
   data_params.predictions_in_chunk = NUM_PREDICTIONS_PER_CHUNK;
   data_params.use_data_augmentation = use_data_augmentation;
-  data_params.random_seed = read_state<int>("random_seed");
+  if (!use_data_augmentation) {
```
nickjong: Nit: it would be more readable to switch the if and else cases and have the condition just be `if (use_data_augmentation)`

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -636,12 +639,19 @@ std::unique_ptr<data_iterator> activity_classifier::create_iterator(
...
+    data_params.random_seed = 0;
+  } else {
+    data_params.random_seed = read_state<int>("random_seed");
+  }
   return std::unique_ptr<data_iterator>(new simple_data_iterator(data_params));
 }
 
 std::unique_ptr<compute_context> activity_classifier::create_compute_context()
     const {
+  if (state.find("use_tensorflow") == state.end()){
```
nickjong: I think we want to simplify this logic, not extend it. Again, please sync up with Shreya and Fareeha here

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -680,47 +703,54 @@ std::unique_ptr<model_spec> activity_classifier::init_model() const
...
+      /* hidden_input */ "hiddenIn",
+      /* cell_input */ "cellIn",
+      /* hidden_output */ "hiddenOut",
+      /* cell_output */ "cellOut",
+      /* input_vector_size */ NUM_CONV_FILTERS,
+      /* output_vector_size */ LSTM_HIDDEN_SIZE,
       /* cell_clip_threshold */ LSTM_CELL_CLIP_THRESHOLD,
-      /* initializers */  lstm_weight_initializers::create_with_xavier_method(
-          NUM_CONV_FILTERS, LSTM_HIDDEN_SIZE, &random_engine));
+      /* initializers */ lstm_initializer);
```
nickjong: If running clang-format on this existing code, could we add spaces inside the comments so that the string constants continue to line up? The main goal here is to keep the code readable

jakesabathia2: hmmm it is not smart enough to do this.
Will add spaces by my self.

PR #2420, opened 2019-10-09
Style Transfer Compute Context

abhishekpratapa:
Added an implementation for creating the style transfer compute context. Had to refactor the `style_transfer_base` code to work with an externally defined command queue.

### DIFF
--- a/src/ml/neural_net/tf_compute_context.cpp
+++ b/src/ml/neural_net/tf_compute_context.cpp
@@ -176,6 +176,19 @@ std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
       new tf_model_backend(activity_classifier));
   
 }
+
+/**
+* TODO: Add model backend for the tensorflow implementation of style transfer
+*/
+std::unique_ptr<model_backend> create_style_transfer(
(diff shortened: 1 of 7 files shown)


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -176,6 +176,19 @@ std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
...
+/**
+* TODO: Add model backend for the tensorflow implementation of style transfer
+*/
+std::unique_ptr<model_backend> create_style_transfer(
+      const float_array_map& config, const float_array_map& weights) {
+  #ifdef __APPLE__
+    return mps_compute_context().create_style_transfer(config, weights);
+  #else
+    return nullptr;
+  #endif
```
nickjong: Nit: preprocessor directives should not be indented, since they don't obey scope. (They are handled before any parsing of curly braces, for example.)

https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives

abhishekpratapa: Thanks!

PR #2421, opened 2019-10-09
Copy_image_to_memory moved to image_utils

shreyajain17:


src/model_server/lib/image_util.hpp
```diff
@@ -8,8 +8,22 @@
...
+
+void copy_image_to_memory(const image_type& input, float* outptr,
+                          const std::vector<size_t>& outstrides,
+                          const std::vector<size_t>& outshape,
+                          bool channel_last);
+
+void image_load_to_numpy(const image_type& img, float* output,
+                         const std::vector<size_t>& outstrides);
+
 namespace image_util{
```
nickjong: PR looks good overall but could we move these functions into the `image_util` namespace?

shreyajain17: Done!

PR #2424, opened 2019-10-10
Unclear error message in apply method when most values are None #2066

### LINKED ISSUE #2066: Unclear error message in apply method when most values are None
When all rows used by the `apply` method for type inference return `None`, then the resulting `SArray` defaults to float.  I know this can be solved by specifying `dtype` but the error message should be clearer, especially when considering lazy operations.

Running:

```python
import turicreate as tc
a = tc.SArray(list(range(100000)))
b = a.apply(lambda x: None if x < 1000 else str(x))
b.materialize()
```

Results in:
`TypeError: Cannot convert 41666 to float`

I suggest something along the lines of `TypeError: Cannot convert 41666 of type XXXX to float which is the type of the SArray. If this is not the correct type try specifying it during creation`

dhivyaaxim:
This PR resolves #2066 
Unclear error message in apply method when most values are None

### DIFF
--- a/src/core/storage/query_engine/operators/lambda_transform.hpp
+++ b/src/core/storage/query_engine/operators/lambda_transform.hpp
@@ -189,7 +189,10 @@ class operator_impl<planner_node_type::LAMBDA_TRANSFORM_NODE> : public query_ope
       res.soft_assign(val);
       return res;
     } else {
-      std::string message = "Cannot convert " + std::string(val) + " of type " + flex_type_enum_to_name(val.get_type()) + " to " + flex_type_enum_to_name(type) +  " which is the type of the SArray. If this is not the correct type try specifying it during creation";
+      std::string message = "Cannot convert " + std::string(val) + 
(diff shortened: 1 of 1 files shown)


src/core/storage/query_engine/operators/lambda_transform.hpp
```diff
@@ -189,8 +189,7 @@ class operator_impl<planner_node_type::LAMBDA_TRANSFORM_NODE> : public query_ope
       res.soft_assign(val);
       return res;
     } else {
-      std::string message = "Cannot convert " + std::string(val) +
-          " to " + flex_type_enum_to_name(type);
+      std::string message = "Cannot convert " + std::string(val) + " of type " + flex_type_enum_to_name(val.get_type()) + " to " + flex_type_enum_to_name(type) +  " which is the type of the SArray. If this is not the correct type try specifying it during creation";
```
TobyRoseman: This line is really long. It would be nice to break it into a few shorter lines.

PR #2432, opened 2019-10-10
Add tctensorflow to python dependencies

TobyRoseman:


TobyRoseman:
> Shouldn't it be libtctensorflow?

We want the name of the target which is `tctensorflow`: https://github.com/apple/turicreate/blob/master/src/ml/neural_net/CMakeLists.txt#L85

shreyajain17:
> > Shouldn't it be libtctensorflow?
> 
> We want the name of the target which is `tctensorflow`: https://github.com/apple/turicreate/blob/master/src/ml/neural_net/CMakeLists.txt#L85

Target is named tctensorflow. I got confused because it adds a prefix of lib on its own so when I import it, I use `libtctensorflow`.

PR #2433, opened 2019-10-10
Activity classifier bug fix: don't pass unrecognized options to C++

TobyRoseman:


jakesabathia2:
Nice! It seems like in activity_classifier.cpp the flag is not used at all :).

TobyRoseman:
> Nice! It seems like in the activity_classifier.cpp this option is never used.

Also passing unrecognized options to C++ causes an error.

jakesabathia2:
Oh !?
Since I haven't tried to use the use_tensorflow = true before.
Maybe @shreyajain17 can also check this issue :).

PR #2434, opened 2019-10-10
Style Transfer model definition

abhishekpratapa:
Added Style Transfer Model Definition with CoreML models and a CoreML spec.


src/ml/neural_net/weight_init.cpp
```diff
@@ -35,6 +35,13 @@ void xavier_weight_initializer::operator()(float* first_weight,
   }
 }
 
+void ones_weight_initializer::operator()(float* first_weight,
```
TobyRoseman: Could these be passed as const references?

abhishekpratapa: Probably just the last one.  We use the first one as an iterator of sorts and increment the pointer. I'd have to let @nickjong speak to this. He probably has a better context to this.

TobyRoseman: Oh, right. So maybe neither of them can be references, but I think they should be const.

abhishekpratapa: So I was referring to the const. Since we use the first pointer as an iterator, not sure it can be a const unless we want to do a `const_cast` later. I remember that being inadvisable. But again I'd have to deffer that to nick. Basically with the current implementation making them const means that the program wouldn't compile.

nickjong: These types cannot be `const`, since they specify a range of mutable memory that we want this function to initialize. We write to each `float` in the range in line 41. That line would generate a compiler error if `w` had type `const float *`

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -0,0 +1,69 @@
...
+using CoreML::Specification::NeuralNetwork;
+using CoreML::Specification::NeuralNetworkLayer;
+
+using turi::neural_net::model_spec;
+using turi::neural_net::ones_weight_initializer;
+using turi::neural_net::zero_weight_initializer;
+
+namespace {
+
+std::unique_ptr<neural_net::model_spec> update_num_styles(
```
nickjong: Sigh, I see this is the first time we’re trying to modify parts of an existing model spec we’ve loaded from disk. (For OD, we just append to the pre-trained model.)

But I think we should avoid leaking raw protobuf code into this file. My proposal is this: instead programmatically construct the architecture of the model from scratch using the existing layer-by-layer `model_spec` methods. Initialize all the weights to zero, except for the instance norm weights, for which you do what you do here. Then use existing functionality to “merge” the parameters for the other layers from the pre-trained model into the model you built programmatically. 

“But Nick, why do go through all the trouble?” I think you’ll need to do this anyway to robustly support loading from a saved model. (Since our file format only saves the parameters by name, not the layer types.) So you may as well do that now, and avoid the leaky `model_spec` abstraction. Happy to discuss in more detail if you have other ideas or context, of course!

abhishekpratapa: This'll be fun! Haha!

src/ml/neural_net/weight_init.hpp
```diff
@@ -48,6 +48,24 @@ class xavier_weight_initializer {
   std::mt19937& random_engine_;
 };
 
+struct const_weight_initializer {
+  /**
+   * Creates a weight initializer that initializes all of the weights to a
+   * constanst scale value.
```
nickjong: *constant

src/ml/neural_net/weight_init.hpp
```diff
@@ -48,6 +48,24 @@ class xavier_weight_initializer {
...
+   * 
+   * \param scale The scale value to initialize the weights to.
+   */
+  const_weight_initializer(float scale);
```
nickjong: Do you mean `scalar` here?

src/ml/neural_net/weight_init.hpp
```diff
@@ -48,6 +48,24 @@ class xavier_weight_initializer {
   std::mt19937& random_engine_;
 };
 
+struct const_weight_initializer {
```
nickjong: Maybe should be `scalar_weight_initializer`

src/ml/neural_net/weight_init.hpp
```diff
@@ -48,6 +48,24 @@ class xavier_weight_initializer {
...
+   * \param scale The scale value to initialize the weights to.
+   */
+  const_weight_initializer(float scale);
+  void operator()(float* first_weight, float* last_weight);
+
+ private:
+  float scale_;
+};
+
+struct ones_weight_initializer {
```
nickjong: Do we still need this if we have the above?

src/toolkits/style_transfer/style_transfer_model_definition.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <memory>
+#include <string>
+
+#include <ml/neural_net/model_spec.hpp>
+
+namespace turi {
+namespace style_transfer {
+
+std::unique_ptr<neural_net::model_spec> init_resnet(std::string& path);
```
nickjong: `std::string` arguments should essentially always be either passed by value or by `const` reference.

If you don't need to store the argument's value persistently anywhere, use a `const` reference

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -0,0 +1,1008 @@
...
+using turi::neural_net::const_weight_initializer;
+using turi::neural_net::float_array_map;
+using turi::neural_net::model_spec;
+using turi::neural_net::ones_weight_initializer;
+using turi::neural_net::zero_weight_initializer;
+using padding_type = model_spec::padding_type;
+
+namespace {
+
+void define_resnet(std::unique_ptr<model_spec>& nn_spec, size_t num_styles) {
```
nickjong: Could the definition be factored out into component substructures or repeated sub blocks? Maybe worth a TODO, no need to do it now

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -0,0 +1,1008 @@
...
+      /* input */ "transformer_instancenorm5__fwd");
+
+  nn_spec->add_scale(
+      /* name */ "stylizedImage",
+      /* input */ "transformer_activation5",
+      /* shape_c_h_w */ {1},
+      /* weight_init_fn */ const_weight_initializer(255.0));
+}
+
+void load_weights(std::unique_ptr<model_spec>& nn_spec, std::string& path) {
```
nickjong: Let's just make the first parameter have type `model_spec*`. No need to involve `unique_ptr` here

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -0,0 +1,1008 @@
...
+  std::unique_ptr<model_spec> weight_spec(new model_spec(path));
```
nickjong: I wouldn't bother wrapping the spec from disk in a `unique_ptr`. This line would be more readable as just
```
model_spec weight_spec(path);
```

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -0,0 +1,1008 @@
...
+  float_array_map nn_params = weight_spec->export_params_view();
+
+  for (auto iter = nn_params.begin(); iter != nn_params.end();) {
+    if (iter->first.find("inst") != std::string::npos) {
+      nn_params.erase(iter++);
+      continue;
+    }
+    ++iter;
+  }
```
nickjong: FWIW, I would find this loop's logic more readable as e.g.
```
auto iter = nn_params.begin();
while (iter != nn_params.end()) {
  if (...) {
    nn_param.erase(iter++);
  } else {
    ++iter;
  }
}
```
Might also be good form to make the iterator variable `map_iter`, since the logic only works correctly for the iterator invalidation semantics of `std::map` (but not, e.g. `std::vector`)

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -0,0 +1,1008 @@
...
+
+  nn_spec->update_params(nn_params);
+}
+
+}  // namespace
+
+std::unique_ptr<model_spec> init_resnet(std::string& path) {
+  std::unique_ptr<model_spec> spec(new model_spec(path));
+  return spec;
+}
```
nickjong: Honest question: what do you see as the benefit to defining this function (which has the same implementation as `init_vgg_16` below)?

test/toolkits/style_transfer/test_style_transfer_model_definition.cxx
```diff
@@ -0,0 +1,186 @@
...
+using CoreML::Specification::NeuralNetworkLayer;
+
+using namespace turi::style_transfer;
+using namespace turi::neural_net;
+
+#define RESNET_MODEL                                           \
+  "https://docs-assets.developer.apple.com/turicreate/models/" \
+  "transformer.mlmodel"
+#define VGG_16_MODEL \
+  "https://docs-assets.developer.apple.com/turicreate/models/vgg16.mlmodel"
```
nickjong: Why do we need to download the actual models to test this code? I would prefer it if our unit tests have few dependencies, and we can reserve this kind of thing for end-to-end integration tests

abhishekpratapa: Fair, I didn't know how to test the functionality of the code without the actual MLModels. For VGG_16 can't test the network without the MLModel. For Resnet 16 can only test the Network Architecture. What do you suggest I do here?

src/toolkits/style_transfer/style_transfer_model_definition.cpp
abhishekpratapa: It's for api, readability. The init_resnet has many overloaded functions and to keep it consistent I added both function. I could refactor to just init_model and call it a day there.

src/toolkits/style_transfer/style_transfer_model_definition.cpp
abhishekpratapa: Ahh got it!

src/toolkits/style_transfer/style_transfer_model_definition.cpp
abhishekpratapa: Alright! No memory leakage here because of it?

src/toolkits/style_transfer/style_transfer_model_definition.cpp
abhishekpratapa: It definitely could, I used a Python script to generate this using CoreMLtools so it isn't the most readable code but I just wanted a network I knew would work. I opened an issue to track it: #2453

src/ml/neural_net/model_spec.cpp
```diff
@@ -636,6 +643,50 @@ void model_spec::add_convolution(
...
+  left_right->set_startedgesize(padding_left);
+  left_right->set_endedgesize(padding_right);
+
+  switch (padding) {
+  case padding_type::VALID:
+    break;
+  case padding_type::SAME:
+    break;
+  case padding_type::REFLECTIVE:
+      params->reflection();
```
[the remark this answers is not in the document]
abhishekpratapa: Not sure, perhaps I need to. I was running into some trouble with this, but yeah mutable would make sense in this instance.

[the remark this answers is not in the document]
abhishekpratapa: Changing it to mutable

src/ml/neural_net/model_spec.cpp
```diff
@@ -586,6 +589,8 @@ void model_spec::add_pooling(const std::string& name, const std::string& input,
     case padding_type::SAME:
       params->mutable_same();
       break;
+    case padding_type::REFLECTIVE:
+      break;
```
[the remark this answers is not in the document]
abhishekpratapa: I believe it defaults to valid?

```
If you set a oneof field to the default value (such as setting an int32 oneof field to 0), the "case" of that oneof field will be set, and the value will be serialized on the wire.
```

https://developers.google.com/protocol-buffers/docs/proto#oneof-features

src/ml/neural_net/model_spec.hpp
```diff
@@ -21,12 +21,16 @@
...
+class WeightParams;
 }
 }
 
 namespace turi {
 namespace neural_net {
 
+void init_weight_params(CoreML::Specification::WeightParams* params, size_t size,
+                        const weight_initializer& weight_init_fn);
+
```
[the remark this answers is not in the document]
abhishekpratapa: Yes, now we're initing weights a different way, I'll refactor back to the original code.

nickjong:
@dfindlay2 FYI

src/ml/neural_net/model_spec.cpp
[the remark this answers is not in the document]
nickjong: But the point is that you're not setting the field to a default value, you're not setting anything at all. My point is, in what sense does this implement "reflective" padding?

src/toolkits/style_transfer/style_transfer_model_definition.cpp
nickjong: Thanks, although for this kind of thing I actually prefer a TODO to a GH issue. IME, "cleanup" issues don't get seen; but at least TODO items prompt anybody who sees this code to make it better

src/toolkits/style_transfer/style_transfer_model_definition.cpp
nickjong: Passing a pointer in and of itself is memory-allocation-neutral. Using a raw pointer makes it more clear that ownership of the memory is not changing; using a `unique_ptr` leaves open the possibility that the function can delete the memory.

src/toolkits/style_transfer/style_transfer_model_definition.cpp
nickjong: I wonder if we could improve the design here, but what we have is fine for now. We can always refactor it later if desired

src/ml/neural_net/model_spec.cpp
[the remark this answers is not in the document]
abhishekpratapa: I think I remember now the padding available to us in:

- `PoolingPadding`
- `ConvolutionalLayerParams`

have different padding values than:

- `PaddingLayerParams`

 I was trying to use the same enumerated types for different layers, but using the same enum wouldn't work for all types of padding.

Right now since we only handle reflective padding in our MLModels specifically in the `PaddingLayer`. I'll remove the enumerated type and add a `TODO` there for future reference in any of the other models we have specifically have other types of padding in their `PaddingLayer`.

src/ml/neural_net/weight_init.hpp
abhishekpratapa: We don't! Got rid of it!

src/toolkits/style_transfer/style_transfer_model_definition.hpp
abhishekpratapa: Used const!

test/toolkits/style_transfer/test_style_transfer_model_definition.cxx
abhishekpratapa: Defined the spec and exported to coreML in the utils file.

test/toolkits/style_transfer/utils.hpp
```diff
@@ -24,8 +24,8 @@ std::string generate_data(size_t data_size);
 turi::flex_image random_image();
 turi::gl_sarray random_image_sarray(size_t length);
 turi::gl_sframe random_sframe(size_t length, std::string image_column_name);
-std::string download_ml_model(std::string download_url,
-                              std::string download_directory);
+std::string get_vgg16_model();
+std::string get_resnet_model();
```
TobyRoseman: Is this resnet-16? resnet-50? Some other resent?

test/toolkits/style_transfer/utils.mm
```diff
@@ -57,29 +71,1189 @@
...
+  input_index_feature_desc->set_shortdescription("Stylized image");
+
+  ImageFeatureType* output_image_feature = output_feature_desc->mutable_type()->mutable_imagetype();
+  output_image_feature->set_width(256);
+  output_image_feature->set_height(256);
+  output_image_feature->set_colorspace(ImageFeatureType::RGB);
+
+  model.mutable_neuralnetwork()->MergeFrom(nn_spec->get_coreml_spec());
+
+  auto model_wrapper = std::make_shared<MLModelWrapper>(std::make_shared<CoreML::Model>(model));
```
TobyRoseman: Probably shouldn't use `auto` here.

nickjong: Honestly, I think this is one of the cases where `auto` makes sense: `std::make_shared<T>` _always_ returns a value of type `std::shared_ptr<T>`, so the reader can tell from local information (that line alone) what `auto` resolves too. (See some discussion here: https://google.github.io/styleguide/cppguide.html#Type_deduction)

My rule of thumb is that `auto` is okay if the type can be exactly deduced from immediate context, or if it saves at least one level of nested STL type nonsense. e.g. for
```
  std::map<std::string,std::vector<int>> m;
  for (const auto& kv : m) { // prefer `auto` over `std::map<std::string,std::vector<int>>::value_type` or `std::pair<std::string, std::vector<int>>`
```
but YMMV

TobyRoseman: Thanks for the explanation @nickjong.

test/toolkits/style_transfer/utils.hpp
abhishekpratapa: resent_16. I can name it more specifically to reduce confusion.

PR #2435, opened 2019-10-11
Drawing Classifier Data Iterator

shantanuchhabra:
This PR should not be merged until #2421 gets merged, since this PR uses a definition of `copy_image_to_memory` that #2421 introduces.


src/toolkits/drawing_classifier/dc_data_iterator.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include <random>
+#include <string>
+#include <vector>
+
+#include <core/data/sframe/gl_sframe.hpp>
+#include <ml/neural_net/float_array.hpp>
+
+const int kDrawingHeight = 28;
+const int kDrawingWidth = 28;
+const int kDrawingChannels = 1;
```
nickjong: These should
* Be moved inside the `drawing_classifier` namespace
* Use `constexpr` instead of `const`, since they are compile-time constants, not run-time computed (https://google.github.io/styleguide/cppguide.html#Use_of_constexpr)

shantanuchhabra: Done, thanks!

src/toolkits/drawing_classifier/dc_data_iterator.hpp
```diff
@@ -0,0 +1,178 @@
...
+#include <core/data/sframe/gl_sframe.hpp>
+#include <ml/neural_net/float_array.hpp>
+
+
+namespace turi {
+namespace drawing_classifier {
+
+constexpr int kDrawingHeight = 28;
+constexpr int kDrawingWidth = 28;
+constexpr int kDrawingChannels = 1;
```
nickjong: Sorry, now that I see that these are only used in `dc_data_iterator.cpp`, could we just move these constants there?

src/toolkits/drawing_classifier/dc_data_iterator.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Pure virtual interface for classes that produce batches of data
+ * (pre-augmentation) from a raw SFrame.
+ */
+class data_iterator {
```
nickjong: Could we add a TODO to factor out the shared structure with the other iterators?

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -0,0 +1,177 @@
+/* Copyright © 2019 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#include <model_server/extensions/additional_sframe_utilities.hpp>
+#include <toolkits/drawing_classifier/dc_data_iterator.hpp>
```
nickjong: Please include `dc_data_iterator.hpp` first, in its own grouping

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -0,0 +1,177 @@
...
+    // Start an iteration through the entire SFrame.
+    range_iterator_(data_.range_iterator()),
+    next_row_(range_iterator_.begin()),
+
+    // Initialize random number generator.
+    random_engine_(params.random_seed)
+
+{}
+
+void add_drawing_pixel_data_to_batch(
```
nickjong: Let's move internal/local helper functions (not declared in the header) to the top. These should be in an anonymous namespace or declared static
https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -0,0 +1,177 @@
...
+    float *next_drawing_pointer, const flex_image &bitmap) {
+  image_load_to_numpy(bitmap, (size_t)next_drawing_pointer,
```
nickjong: Please use C++11 `static_cast` here

https://google.github.io/styleguide/cppguide.html#Casting

nickjong: Also, consider using the `copy_image_to_memory` function that @shreyajain17  is adding. https://github.com/apple/turicreate/pull/2421

Less confusing, since numpy is irrelevant/misleading here

src/toolkits/drawing_classifier/dc_data_iterator.hpp
shantanuchhabra: Fixed,

src/toolkits/drawing_classifier/dc_data_iterator.cpp
shantanuchhabra: Done.

A quick follow-up: Should `dc_data_iterator.hpp` be included as "dc_data_iterator.hpp" or as <toolkits/drawing_classifier/dc_data_iterator.hpp> ? I got conflicting information on this.

src/toolkits/drawing_classifier/dc_data_iterator.cpp
shantanuchhabra: Done, thanks!

src/toolkits/drawing_classifier/dc_data_iterator.cpp
shantanuchhabra: Added a call to `copy_image_to_memory`, which makes this PR dependent on #2421 .

src/toolkits/drawing_classifier/dc_data_iterator.hpp
shantanuchhabra: Done!

src/toolkits/drawing_classifier/dc_data_iterator.cpp
nickjong: The latter, since we're under `src`. Things are a bit weirder under `test`. We should use fully qualified paths when possible

PR #2438, opened 2019-10-11
Unclear error message when calculating jaccard distance

### LINKED ISSUE #1233: Unclear error message when calculating jaccard and weighted_jaccard distance
For both jaccard and weighted_jaccard distances the following produces an unclear error:
 ```python
a = {'k': 'df'}
b = {'j': 1}
tc.toolkits.distances.jaccard(a, b)
```
```
ToolkitError: stod: no conversion
```

I would expect something similar to:
```
ToolkitError: at least one of the dictionary values could not be converted to a number.
```
This is especially confusing when using jaccard distance which does not consider the values of the dictionary.

dhivyaaxim:
This PR resolves #1233 
Unclear error message when calculating jaccard distance

### DIFF
--- a/src/python/turicreate/extensions.py
+++ b/src/python/turicreate/extensions.py
@@ -156,7 +156,7 @@ def _run_toolkit_function(fnname, arguments, args, kwargs):
     # handle errors
     if not ret[0]:
         if len(ret[1]) > 0:
-            raise _ToolkitError("at least one of the dictionary values could not be converted to a number.")
+            raise _ToolkitError(ret[1])
         else:
             raise _ToolkitError("Toolkit failed with unknown error")
 
--- a/src/python/turicreate/extensions.py
+++ b/src/python/turicreate/extensions.py
@@ -156,7 +156,7 @@ def _run_toolkit_function(fnname, arguments, args, kwargs):
     # handle errors
     if not ret[0]:
         if len(ret[1]) > 0:
-            raise _ToolkitError(ret[1])
+            raise _ToolkitError("at least one of the dictionary values could not be converted to a number.")
         else:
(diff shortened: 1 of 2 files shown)


dhivyaaxim:
> I don't think this is the right fix. This fix will cause the error message to get display in many cases, other than #1233, were that message doesn't make sense.
> You'll want to change this error message in the C++ toolkit code that calculates the Jaccard distance.

dhivyaaxim:
@TobyRoseman  Thanks for your suggestion . Will check and let you know

dhivyaaxim:
@TobyRoseman Exception for Jaccard in cpp has been captured dynamically using standard exception. If I change the standard exception, it will affect other exception. There is a way to check in python for Jaccard input validation. Please give suggestion

TobyRoseman:
@dhivyaaxim - I don't know where this error is being generated either. It could be in [our Cython Layer](https://github.com/apple/turicreate/tree/master/src/python/turicreate/_cython) or it could be in [the C++ code for calculating the Jaccard distance](https://github.com/apple/turicreate/blob/9281a0f8fdd0a7c8a9d77396b2c94557a5cbd147/src/toolkits/nearest_neighbors/distances.hpp#L190). I would guess the error is being generated from [this helper function](https://github.com/apple/turicreate/blob/9281a0f8fdd0a7c8a9d77396b2c94557a5cbd147/src/toolkits/nearest_neighbors/distances.hpp#L25) called by the C++ Jaccard distance function.

I suggest adding print statements or stepping through the code in a debugger.

dhivyaaxim:
Thanks @TobyRoseman for your suggestion , changes are addressed

PR #2443, opened 2019-10-11
Changes in Image Augmenter for OD in TF

shreyajain17:


src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+
+  const size_t n = opts_.batch_size;
+  constexpr size_t c = 3;
+
+  std::vector<turi::neural_net::shared_float_array> images_to_aug;
+  std::vector<turi::neural_net::shared_float_array> ann_batch;
+  std::vector<turi::neural_net::shared_float_array> pred_batch;
+
+  // Decode a batch of images to raw format
+  for (size_t i = 0; i < n; i++) {
```
nickjong: What happens if `source_batch` does not contain exactly `n` elements?

Consider looking at what was done for `mps_image_augmenter` https://github.com/apple/turicreate/blob/4658fc57dba09f01806d5d5ad1a9c6bee670e916/src/ml/neural_net/mps_image_augmentation.mm#L332

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    size_t input_height = source_batch[i].image.m_height;
```
nickjong: Please factor out all the instances of `source_batch[i]`. There's no need to perform a new array lookup each time

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    size_t input_width = source_batch[i].image.m_width;
+    std::vector<float> img(input_height * input_width * c, 0.f);
+    // unsigned char *outptr = reinterpret_cast<unsigned char *>(img.data());
```
nickjong: What is this for?

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    image_util::copy_image_to_memory(source_batch[i].image, img.data(),
+                                     {input_width * c, c, 1},
+                                     {input_height, input_width, c}, true);
```
nickjong: Could you document what these arguments are in comments? E.g., use `/* channel_last */` for the last argument

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+
+    std::transform(img.begin(), img.end(), img.begin(),
+                   [](float pixel) -> float { return pixel / 255; });
```
nickjong: Nit: prefer to use numeric literals of the target type, to make it more visibly obvious that this is floating-point arithmetic. So `255.0f` instead of `255`

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    shared_float_array image_to_aug =
+        shared_float_array::wrap(img, {input_height, input_width, c});
```
nickjong: Let's use `std::move(img)` here instead of `img` to avoid a copy. In fact, I would recommend factoring out a helper function that converts `image_type` to `shared_float_array`, so you can reason more robustly about when `img` becomes moved-from (invalid)

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    images_to_aug.push_back(image_to_aug);
+    std::vector<float> annotation(source_batch[i].annotations.size() * 6);
```
nickjong: Please factor out `source_batch[i].annotations.size()` for readability

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    size_t x = 0;
+    for (size_t j = 0; j < source_batch[i].annotations.size(); j++) {
+      annotation[x] = source_batch[i].annotations[j].identifier;
+      annotation[x + 1] = source_batch[i].annotations[j].bounding_box.x;
+      annotation[x + 2] = source_batch[i].annotations[j].bounding_box.y;
+      annotation[x + 3] = source_batch[i].annotations[j].bounding_box.height;
+      annotation[x + 4] = source_batch[i].annotations[j].bounding_box.width;
+      annotation[x + 5] = source_batch[i].annotations[j].confidence;
+      x = (j + 1) * 6;
```
nickjong: Instead of defining `x` outside the loop and updating it at the end of each iteration, could you just define it at the beginning of each iteration, as `size_t x = j * 6`? Let's also use a more informative name, something like `annotation_offset`.

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    std::transform(img.begin(), img.end(), img.begin(),
+                   [](float pixel) -> float { return pixel / 255; });
+    shared_float_array image_to_aug =
+        shared_float_array::wrap(img, {input_height, input_width, c});
+    images_to_aug.push_back(image_to_aug);
+    std::vector<float> annotation(source_batch[i].annotations.size() * 6);
+    size_t x = 0;
+    for (size_t j = 0; j < source_batch[i].annotations.size(); j++) {
+      annotation[x] = source_batch[i].annotations[j].identifier;
+      annotation[x + 1] = source_batch[i].annotations[j].bounding_box.x;
```
nickjong: Please factor out repeated sub-expressions, like `source_batch[i].annotations[j]`

(and give it a name that helps future readers/maintainers)

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+    std::vector<float> predictions(source_batch[i].predictions.size() * 6);
+    size_t y = 0;
+    for (size_t k = 0; k < source_batch[i].predictions.size(); k++) {
+      predictions[y] = source_batch[i].predictions[k].identifier;
+      predictions[y + 1] = source_batch[i].predictions[k].bounding_box.x;
+      predictions[y + 2] = source_batch[i].predictions[k].bounding_box.y;
+      predictions[y + 3] = source_batch[i].predictions[k].bounding_box.height;
+      predictions[y + 4] = source_batch[i].predictions[k].bounding_box.width;
+      predictions[y + 5] = source_batch[i].predictions[k].confidence;
+      y = (k + 1) * 6;
```
nickjong: Please don't just duplicate code like this. Define a helper function and call it twice. This is something you are expected to do in general!

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+
+    turi::neural_net::shared_float_array pred_to_aug = shared_float_array::wrap(
+        predictions, {source_batch[i].predictions.size() * 6});
+
+    pred_batch.push_back(pred_to_aug);
+  }
+  intermediate_labeled_image input_to_tf_aug;
+  input_to_tf_aug.images = images_to_aug;
+  input_to_tf_aug.annotations = ann_batch;
+  input_to_tf_aug.predictions = pred_batch;
```
nickjong: Instead of copying these into the struct here, could we just allocate the struct upfront and then build these values directly in the struct?

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+  res.image_batch = augmented_data.images;
+  std::vector<std::vector<image_annotation>> annotations_per_batch;
+  for (size_t a = 0; a < n; a++) {
+    std::vector<image_annotation> annotations_per_image;
+    const size_t* sh = augmented_data.annotations[a].shape();
+
+    for (size_t b = 0; b < sh[0]; b++) {
+      image_annotation annotation;
+      image_box bbox;
+      std::cout << a << b;
```
nickjong: Please remove debugging code from your branch before pushing for review. Don't require reviewers to find this for you

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+      const float* ptr = augmented_data.annotations[a][b].data();
+      annotation.identifier = static_cast<int>(ptr[0]);
+      bbox.x = static_cast<float>(ptr[1]);
```
nickjong: Why are these `static_cast`s needed?

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -117,5 +117,98 @@ image_augmenter::result resize_only_image_augmenter::prepare_images(
...
+      bbox.y = static_cast<float>(*augmented_data.annotations[a][b][2].data());
+      bbox.height =
+          static_cast<float>(*augmented_data.annotations[a][b][3].data());
+      bbox.width =
+          static_cast<float>(*augmented_data.annotations[a][b][4].data());
+      annotation.bounding_box = bbox;
+      annotation.confidence =
+          static_cast<float>(*augmented_data.annotations[a][b][5].data());
```
nickjong: Could we define a pair of helper functions just for converting back and forth from one annotation to a buffer of 6 floats? Then it would be easier to see from inspection that the conversions are consistent

src/ml/neural_net/image_augmentation.cpp
shreyajain17: They all are `shared_float_arrays` need to be casted in as float/int.

src/ml/neural_net/image_augmentation.cpp
shreyajain17: Okay will make two functions one to go from image_annotation to vector of shared_float_array and the other way around.

src/ml/neural_net/image_augmentation.cpp
shreyajain17: Yeah sorry missed that.

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -269,7 +269,57 @@ class resize_only_image_augmenter final: public image_augmenter {
   options opts_;
 };
 
+/**
+ * An abstract class that inherits from image_augmenter used to convert
+ * input images, annotations and predictions to shared_float_arrays for
+ * tf_image_augmenter.
+ * Subclass must be written for it if needed. The subclass must implement the
+ * pure virtaul method prepare_augmented_images.
```
nickjong: *virtual

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -269,7 +269,57 @@ class resize_only_image_augmenter final: public image_augmenter {
   options opts_;
 };
 
+/**
+ * An abstract class that inherits from image_augmenter used to convert
+ * input images, annotations and predictions to shared_float_arrays for
+ * tf_image_augmenter.
+ * Subclass must be written for it if needed. The subclass must implement the
```
nickjong: Not sure, but you may need to leave a blank line (with just a `*`) so that Doxygen won't interpret this as one big paragraph. Assuming you intend this to be two paragraphs.

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -269,7 +269,57 @@ class resize_only_image_augmenter final: public image_augmenter {
...
+class float_array_image_augmenter : public image_augmenter {
+ public:
+  float_array_image_augmenter(const options& opts) : opts_(opts) {}
+
+  const options& get_options() const override { return opts_; }
+
+  result prepare_images(std::vector<labeled_image> source_batch) override;
+
+ protected:
+  /** The output sent from TensorFlow after augmenting the images. */
```
nickjong: This isn't specific to TensorFlow. Let's remove this terminology. Maybe use "the subclass" or "the child class???"

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -269,7 +269,57 @@ class resize_only_image_augmenter final: public image_augmenter {
...
+  struct intermediate_result {
+    /** The images after augmenting sent from Tensorflow */
+    shared_float_array images;
+
+    /** The annotations associated with augmented images sent from Tensorflow */
+    std::vector<shared_float_array> annotations;
+  };
+
+  /** The output sent to TensorFlow to augment the images. */
+  struct intermediate_labeled_image {
```
nickjong: I just double-checked, and I think I gave you bad advice before. Our existing (MPS) implementation ignores the `predictions` entirely. So for now I think it's okay to ignore them here, too. Let's define just one struct, which we can use for both the input and output from the subclass / TensorFlow. Maybe something like `annotated_float_array_image` or `labeled_float_images` or some such?

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -14,6 +14,42 @@
 namespace turi {
 namespace neural_net {
 
+shared_float_array convert_to_shared_float_array(
+    std::vector<image_annotation> annotations_per_image) {
+  size_t offset = 0;
+  std::vector<float> ann(annotations_per_image.size() * 6);
+  for (size_t j = 0; j < annotations_per_image.size(); j++) {
+    offset = j * 6;
```
nickjong: Nit: you may as well do `size_t offset = j * 6` here and delete line 19

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -14,6 +14,42 @@
...
+    image_annotation annotation = annotations_per_image[j];
```
nickjong: You can make this variable a `const image_annotation&` to avoid a copy

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -14,6 +14,42 @@
...
+    ann[offset] = annotation.identifier;
+    ann[offset + 1] = annotation.bounding_box.x;
+    ann[offset + 2] = annotation.bounding_box.y;
+    ann[offset + 3] = annotation.bounding_box.height;
+    ann[offset + 4] = annotation.bounding_box.width;
+    ann[offset + 5] = annotation.confidence;
+  }
+  shared_float_array data_to_augment =
+      shared_float_array::wrap(std::move(ann), {annotations_per_image.size() * 6});
```
nickjong: Should the shape here be `{annotations_per_images.size(), 6}`? That is, rank 2 not 1?

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -14,6 +14,42 @@
...
+
+std::vector<image_annotation> convert_to_image_annotation(
+    shared_float_array augmented_annotation) {
+  const size_t* shape = augmented_annotation.shape();
+  std::vector<image_annotation> augmented_ann;
+  for (size_t b = 0; b < shape[0]; b++) {
+    image_annotation annotation = augmented_ann[b];
+    const float* ptr = augmented_annotation[b].data();
+    annotation.identifier = static_cast<int>(ptr[0]);
+    annotation.bounding_box.x = static_cast<float>(ptr[1]);
```
nickjong: `ptr[1]` is already a `float`. You don't need to cast it.

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -14,6 +14,42 @@
...
+  }
+  shared_float_array data_to_augment =
+      shared_float_array::wrap(std::move(ann), {annotations_per_image.size() * 6});
+  return data_to_augment;
+}
+
+std::vector<image_annotation> convert_to_image_annotation(
+    shared_float_array augmented_annotation) {
+  const size_t* shape = augmented_annotation.shape();
+  std::vector<image_annotation> augmented_ann;
```
nickjong: You haven't actually allocated the vector here. I think you want
```
std::vector<image_annotation> augmented_ann(shape[0]);
```
Although it would be nice to factor our `shape[0]` and give it a more readable name

src/ml/neural_net/image_augmentation.hpp
shreyajain17: I cannot use the same struct. It is because images member of the struct would be different in both. When we get images as input since their sizes can be different. It is being passed to tensorflow as `std::vector<shared_float_array>`. But when the images are augmented they are also resized in the shape`(416, 416)` and it is being passed back as one `shared_float_array`.

nickjong: Ah, makes sense. Thanks for the explanation!

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -29,23 +28,25 @@ shared_float_array convert_to_shared_float_array(
     ann[offset + 5] = annotation.confidence;
   }
   shared_float_array data_to_augment =
-      shared_float_array::wrap(std::move(ann), {annotations_per_image.size() * 6});
+      shared_float_array::wrap(ann, {annotations_per_image.size(), 6});
```
nickjong: This should still use `std::move`, no?

shreyajain17: `shared_float_array::wrap` does it in its implementation.

nickjong: That's not how C++11 works. The implementation will now move a copy being made here. `std::move` must be used at every step along the way, otherwise a copy will occur

PR #2444, opened 2019-10-11
Pybind11 code for image augmentation

shreyajain17:


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -269,5 +268,56 @@ tf_model_backend::~tf_model_backend() {
...
+              std::vector<size_t>(buf_ann.shape.begin(), buf_ann.shape.end()));
+      annotations_per_batch.push_back(annotate);
+    }
+    image_annotations.annotations = annotations_per_batch;
+  });
+
+  return image_annotations;
+}
+
+tf_image_augmenter::~tf_image_augmenter() {}
```
TobyRoseman: Why add a destructor if it's empty?

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -43,6 +41,23 @@ class tf_model_backend : public model_backend {
   pybind11::object model_;
 };
 
+class tf_image_augmenter : public processed_image_augmenter {
+ public:
+  tf_image_augmenter(const options& opts);
+  ~tf_image_augmenter();
```
TobyRoseman: Since you're using inheritance, should the destructor be virtual?

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -9,24 +9,22 @@
 #include <vector>
 
 #include <core/util/try_finally.hpp>
+#include <ml/neural_net/image_augmentation.hpp>
 #include <ml/neural_net/model_backend.hpp>
+#include <model_server/extensions/additional_sframe_utilities.hpp>
```
nickjong: Why do you need this header?

src/ml/neural_net/tf_compute_context.cpp
nickjong: The idiom to follow here is
```
~tf_image_augmenter() override = default;
```

This asserts that (a) the parent class has a virtual destructor, and (b) this class doesn't introduce any additional logic for when an instance is destroyed.

Then you can omit the empty definition below

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -269,5 +268,56 @@ tf_model_backend::~tf_model_backend() {
   call_pybind_function([&]() { model_ = pybind11::object(); });
 }
 
+tf_image_augmenter::tf_image_augmenter(const options& opts)
+    : processed_image_augmenter(opts) {}
```
nickjong: Could we put the implementation of `tf_image_augmenter` methods up top, just after the definition of the class itself? This way, all the `tf_image_augmenter` stuff is together.

It's also more common idiom in C/C++, to have the file-local helper stuff defined first, followed by the stuff using the helper stuff

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -269,5 +268,56 @@ tf_model_backend::~tf_model_backend() {
...
+
+image_augmenter::result tf_image_augmenter::prepare_images(
+    std::vector<labeled_image> source_batch) {
+  return processed_image_augmenter::prepare_images(source_batch);
+}
```
nickjong: You don't need to override a method just to call the parent class's version. Just omit this definition and the override declaration above.

You only need to override methods whose behavior you want to change

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -269,5 +268,56 @@ tf_model_backend::~tf_model_backend() {
...
+
+image_augmenter::intermediate_result
+tf_image_augmenter::prepare_augmented_images(
+    image_augmenter::intermediate_labeled_image data_to_augment) {
+  // const size_t n = opts_.batch_size;
```
nickjong: Please remove comments that don't document the shipping code or otherwise help readers and future maintainers.

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -269,5 +268,56 @@ tf_model_backend::~tf_model_backend() {
...
+  image_augmenter::intermediate_result image_annotations;
+
+  call_pybind_function([&]() {
+    // Pass the required data to tensorflow
```
nickjong: The formatting here makes it look like the comment "Pass the required data to TensorFlow" applies to the next line (287), which just accesses the relevant module from Python. Please try to write comments that explain the code accurately!

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -269,5 +268,56 @@ tf_model_backend::~tf_model_backend() {
...
+    pybind11::module tf_aug = pybind11::module::import(
+        "turicreate.toolkits.object_detector._tf_image_augmenter");
+
+    // Get augmented images and annotations from tensorflow
+    pybind11::object tf_augmentation = tf_aug.attr("DataAugmenter")(
+        data_to_augment.images, data_to_augment.annotations,
+        data_to_augment.predictions);
+    pybind11::object augmented_img =
+        tf_augmentation.attr("get_augmented_images")();
```
nickjong: Do we need a `DataAugmenter` class in Python? Can this just be a function? Is there any state maintained from one augmentation batch to the next?

shreyajain17: As per our offline discussion, it was because of the heterogeneous data that was being passed back from python. I will change it to a function now that return a pair of pybind11 buffers and vector of pybind11 buffers.

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: Will change it

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: It is not needed. I removed it now and added the change Nick suggested in the header.

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: Removed it

PR #2446, opened 2019-10-12
Changes to make OD work with TF 2.0

shreyajain17:


src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -7,7 +7,8 @@
 from __future__ import print_function as _
 from __future__ import division as _
 from __future__ import absolute_import as _
-import tensorflow as _tf
+import tensorflow.compat.v1 as _tf
+_tf.disable_v2_behavior()
```
nickjong: Could we move this line outside of the `import`s?

Is this function lightweight? Should it just be included in the functions below?

shreyajain17: Yes

srikris:
@shreyajain17 Please merge if this is good to go.

PR #2450, opened 2019-10-14
Column names containing commas should be wrapped in quotes

### LINKED ISSUE #860: Column names containing commas should be wrapped in quotes when saving to CSV
Observed Behavior:

SFrame.save("foo.csv") knows that string values containing commas should be wrapped in quotes when saved in csv (comma-delimited). But it does not do that for column names (headers).

```python
import turicreate as tc
print tc.__version__  # => 5.0b2

tc.SFrame({"look,comma": [  # column name with a comma
    "wow, amazing",         # string with comma will get wrapped in quotes
    "wonderful"             # ordinary string, no wrap
]}).save("foo.csv")
```

```bash
$ cat foo.csv   # file won't load using SFrame.read_csv
look,comma      # column name did not get wrapped
"wow, amazing"  # column value got wrapped
wonderful
```

Expected Behavior:

Just like column values, column named should be wrapped with quotes if necessary.
This is consistent with other software packages, for instance in pandas:

```python
import pandas as pd
print pd.__version__  # => 0.19.1

pd.DataFrame({"look,comma": [  # same thing
    "wow, amazing",
    "wonderful"
]}).to_csv("foo.csv")
```

```bash
$ cat foo.csv
,"look,comma"     # looks OK!
0,"wow, amazing"
1,wonderful
```

Steps To Reproduce:

See above.

dhivyaaxim:
This PR resolves #860 
Column names containing commas should be wrapped in quotes when saving to CSV

### DIFF
--- a/src/core/storage/sframe_data/csv_writer.cpp
+++ b/src/core/storage/sframe_data/csv_writer.cpp
@@ -11,7 +11,7 @@ namespace turi {
 void csv_writer::write_verbatim(std::ostream& out,
                                 const std::vector<std::string>& row) {
   for (size_t i = 0;i < row.size(); ++i) {
-    out << row[i];
+    csv_print(out, row[i], row.size() > 1);
     // put a delimiter after every element except for the last element.
     if (i + 1 < row.size()) out << delimiter;
   }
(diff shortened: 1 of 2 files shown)


hoytak:
Looks fantastic!  Thanks for your work, @dhivyaaxim!  Ready to merge.

PR #2452, opened 2019-10-14
Add memory budget to tf_compute_context

shreyajain17:
I am returning 4GB as the memory as C++ side makes batch_size 32 then (which is what we want as MXNet uses always and we want to make TF comparable to MXNet as of now)

### DIFF
--- a/src/ml/neural_net/tf_compute_context.cpp
+++ b/src/ml/neural_net/tf_compute_context.cpp
@@ -125,7 +125,8 @@ tf_compute_context::tf_compute_context() = default;
 tf_compute_context::~tf_compute_context() = default;
 
 size_t tf_compute_context::memory_budget() const {
-  return 0;
+  // TODO: Return 4GB for now. Do something that makes more sense later.
+  return 4294967296;
 }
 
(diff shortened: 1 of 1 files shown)


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -125,7 +125,9 @@ tf_compute_context::tf_compute_context() = default;
 tf_compute_context::~tf_compute_context() = default;
 
 size_t tf_compute_context::memory_budget() const {
-  return 0;
+  // TODO: Returns 4GB as that makes sure default batch size is used. 
+  // Do something that makes more sense like MPS later.
+  return 4294967296;
```
nickjong: Have you tested this? Just to be safe, try `4294967296lu`. The `lu` suffix tells the compiler that the integer literal has type `unsigned long`, which should be big enough to contain `2**32`.

Otherwise, C/C++ defaults to `int`, which is not actually big enough to contain `2**32`

shreyajain17: I will test it on Linux GPU/CPU by making an OD model with Tensorflow.

Issue #2441, opened 2019-10-11
Object Detection C++ Test Failure


shantanuchhabra:
On `master`, we have a C++ unit test failure in `test_od_serialization.cxx`

```
Running 2 test cases...
/build/test/unity/toolkits/object_detection/test_od_serialization.cxx(33): error: in "test_init_darknet_yolo": check (nn.layers_size()) == (25) has failed [31 != 25]
/build/test/unity/toolkits/object_detection/test_od_serialization.cxx(42): error: in "test_init_darknet_yolo": check convlayer_.has_convolution() has failed
/build/test/unity/toolkits/object_detection/test_od_serialization.cxx(44): error: in "test_init_darknet_yolo": check (convlayer_.name()) == ("conv" + std::to_string(x.first) + "_fwd") has failed [pool0_fwd != conv1_fwd]
/build/test/unity/toolkits/object_detection/test_od_serialization.cxx(45): error: in "test_init_darknet_yolo": check (convlayer_.convolution().outputchannels()) == (x.second) has failed [0 != 32]
/build/test/unity/toolkits/object_detection/test_od_serialization.cxx(46): error: in "test_init_darknet_yolo": check (convlayer_.convolution().kernelchannels()) == (num_features) has failed [0 != 16]
[libprotobuf FATAL /build/src/external/google/protobuf/repeated_field.h:1142] CHECK failed: (index) < (current_size_): 
unknown location(0): fatal error: in "test_init_darknet_yolo": __tc_google::protobuf::FatalException: CHECK failed: (index) < (current_size_): 
/build/test/unity/toolkits/object_detection/test_od_serialization.cxx(47): last checkpoint

*** 6 failures are detected in the test module "test_od_serialization"
```

fareeha:
#2418 Test `test_init_darknet_yolo` needs to be updated to match change in model initialization i.e. added pool layers.

PR #2454, opened 2019-10-14
Change OD unity test

jakesabathia2:
close https://github.com/apple/turicreate/issues/2441.


nickjong:
@Jarvi-Izana Here's an example of a nice unit test for (in this case) the OD model definition

test/unity/toolkits/object_detection/test_od_serialization.cxx
```diff
@@ -30,7 +30,7 @@ BOOST_AUTO_TEST_CASE(test_init_darknet_yolo) {
   init_darknet_yolo(nn_spec, num_classes, anchor_boxes);
 
   const CoreML::Specification::NeuralNetwork& nn = nn_spec.get_coreml_spec();
-  TS_ASSERT_EQUALS(nn.layers_size(), 25);
+  TS_ASSERT_EQUALS(nn.layers_size(), 31);
```
shantanuchhabra: Could you add a comment explaining the magic number here?

test/unity/toolkits/object_detection/test_od_serialization.cxx
```diff
@@ -62,7 +62,29 @@ BOOST_AUTO_TEST_CASE(test_init_darknet_yolo) {
                      "leakyrelu" + std::to_string(x.first) + "_fwd");
     TS_ASSERT_EQUALS(relulayer_.activation().leakyrelu().alpha(), 0.1f);
 
-    layer_num = layer_num + 3;
+    if (x.first <= 5){
+      const auto& poolinglayer_ = nn.layers(layer_num + 3);
```
shantanuchhabra: There seem to be a lot of magic numbers in this file. Could you add comments (or make variables) explaining what the magic numbers mean, so the code becomes more readable and inheritable?

jakesabathia2:
Just add the test for the final conversion layer which wasn't covered in the past.
And also make some comments :)

Jarvi-Izana:
> @Jarvi-Izana Here's an example of a nice unit test for (in this case) the OD model definition

Will have a look.

PR #2457, opened 2019-10-15
Add checkpointing support for OD

nickjong:
While everyone is wrapping their head around unit tests and mocks, I thought this would be an opportune time to merge a PR with a preliminary stab at checkpointing support for OD, taking this opportunity to clean up the OD tests a bit. With `object_detector::train` now split into smaller pieces, especially `init_training` and `iterating_training`, the previous monolithic test can be decomposed a bit too.

Also moved all the task-agnostic mocks to a header, where they can be used by tests for ST, DC, etc.

### DIFF
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -338,6 +338,10 @@ size_t object_detector::get_version() const {
 }
 
 void object_detector::save_impl(oarchive& oarc) const {
+  if (training_model_) {
+    // If checkpointing during training, first copy weights from the backend.
+    synchronize_model(nn_spec_.get());
+  }
   _save_impl(oarc, *nn_spec_, state);
 }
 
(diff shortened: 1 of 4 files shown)


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -972,14 +993,43 @@ void object_detector::init_train(
...
       {"training_epochs", 0},
       {"training_iterations", 0},
   });
   // TODO: The original Python implementation also exposed "anchors",
   // "non_maximum_suppression_threshold", and "training_time".
 
+  init_training_backend();
+}
+
+void object_detector::resume_training(gl_sframe data,
```
shantanuchhabra: Where do we call `resume_training`?

src/toolkits/object_detection/object_detector.hpp
```diff
@@ -109,6 +138,22 @@ class EXPORT object_detector: public ml_model_base {
   END_CLASS_MEMBER_REGISTRATION
 
  protected:
+  // Constructor allowing tests to set the initial state of this class and to
+  // inject dependencies.
+  object_detector(
```
shantanuchhabra: Nice! We should take motivation from this for other toolkits' C++ Tests as well.

test/unity/toolkits/neural_net/neural_net_mocks.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+  std::unique_ptr<model_backend> create_activity_classifier(
+      int n, int c_in, int h_in, int w_in, int c_out, int h_out, int w_out,
+      const float_array_map& config, const float_array_map& weights) override {
+    return nullptr;
+  }
+
+  std::unique_ptr<model_backend> create_style_transfer(
+      const float_array_map& config, const float_array_map& weights) override {
+    return nullptr;
```
shantanuchhabra: This is really helpful, thanks for doing this!

To confirm, for other toolkits, would we now add logic here to mock the respective compute contexts, instead of what we did earlier, which was: Copy over the mock_compute_context class, add logic for the toolkit in question, and return `nullptr` for all other toolkits?

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -16,6 +16,8 @@
 #include <boost/test/unit_test.hpp>
 #include <core/util/test_macros.hpp>
 
+#include "../neural_net/neural_net_mocks.hpp"
```
shantanuchhabra: Are we using double quotes with a relative path here because the header we're importing is part of the same project, i.e. `test`?

PR #2458, opened 2019-10-15
Move boost and other libraries from deps/ to source tree.

hoytak:
It turns out that it is possible to build boost, nanomsg, libpng, and libjpeg from within our source tree, which allows us to have all libraries used when the --no-remotefs flag is used to be built within our source tree.  This permits significant build speedups as well as more control over how these libraries are built for different architectures. 

This PR moves these dependencies and cleans up a bit of the base CMakeLists.txt file to accommodate it.


hoytak:
Closed in favor of later PR that pulls all this out end-to-end.

PR #2459, opened 2019-10-15
Added a simplified version of Pybind support for Drawing Classifier

shantanuchhabra:


test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -209,6 +209,15 @@ class mock_compute_context: public compute_context {
     return nullptr;
   }
 
+  std::unique_ptr<model_backend> create_drawing_classifier(
+      /* TODO: const float_array_map& weights
```
shreyajain17: Also will need config for whatever extra options DC has

shantanuchhabra:
Internal tests passed.

PR #2461, opened 2019-10-16
Added Train and Export to CoreML to Model Backend - Style Transfer

abhishekpratapa:
## Overview

- Added train to the style transfer model backend

## TODO

- test save and load for model backend


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+  constexpr size_t c = 3;
+
+  std::vector<float> result_array(h * w * c);
+  auto out_it = result_array.begin();
+
+  image_type resized_image = image_util::resize_image(image, w, h, c, true, 1);
+  ASSERT_EQ(resized_image.m_image_data_size, h * w * c);
+
+  const unsigned char* src_ptr = resized_image.get_image_data();
+  auto out_end = out_it + h * w * c;
```
shantanuchhabra: Same as above re: `auto`.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+    *out_it = *src_ptr / 255.f;
+    ++src_ptr;
+    ++out_it;
+  }
+
+  return shared_float_array::wrap(std::move(result_array), {h, w, c});
+}
+
+flex_int estimate_max_iterations(flex_int num_styles, flex_int batch_size) {
+  return (int)(num_styles * 10000.0f / batch_size);
```
shantanuchhabra: Could we make this a `static_cast`?

src/toolkits/style_transfer/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ project(unity_toolkits)
 
 make_library(unity_style_transfer OBJECT
 SOURCES
+	style_transfer.cpp
```
nickjong: Nit: please avoid tab characters

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
+/* Copyright © 2018 Apple Inc. All rights reserved.
```
nickjong: Nit: for new code, let's use 2019

test/toolkits/style_transfer/test_style_transfer.cxx
```diff
@@ -0,0 +1,39 @@
...
+  turi::gl_sarray style_sarray = random_image_sarray(NUM_STYLES);
+  turi::gl_sarray content_sarray = random_image_sarray(50);
+
+  std::map<std::string, turi::flexible_type> opts = {
+    {"resnet_mlmodel_path", RESNET_MODEL_PATH},
+    {"vgg_mlmodel_path", VGG_16_MODEL_PATH},
+    {"num_styles", NUM_STYLES},
+  };
+
+  model.train(style_sarray, content_sarray, opts);
```
shantanuchhabra: Why are we calling `train` and not `init_train` when we want to test initialization?

test/toolkits/style_transfer/test_style_transfer.cxx
```diff
@@ -0,0 +1,39 @@
...
+#include <core/util/test_macros.hpp>
+#include <toolkits/style_transfer/style_transfer.hpp>
+
+#include <map>
+#include <memory>
+
+#include "utils.hpp"
+
+using namespace turi::style_transfer;
+
```
shantanuchhabra: Do we want more tests here?

src/toolkits/style_transfer/style_transfer.hpp
```diff
@@ -0,0 +1,77 @@
...
+                                                 gl_sarray content, bool repeat,
+                                                 int random_seed) const;
+
+  virtual std::unique_ptr<neural_net::compute_context> create_compute_context()
+      const;
+
+  virtual void init_train(gl_sarray style, gl_sarray content,
+                          std::map<std::string, flexible_type> opts);
+
+  virtual void perform_training_iteration();
```
nickjong: Actually, could we make these public and use the patterns I (just) established for OD? We want all the toolkits to eventually have a consistent API to support checkpointing, and we may as well bake it in from the beginning for new toolkits

https://github.com/apple/turicreate/blob/master/src/toolkits/object_detection/object_detector.hpp#L50

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+using turi::neural_net::float_scalar;
+using turi::neural_net::shared_float_array;
+
+namespace {
+
+constexpr size_t STYLE_TRANSFER_VERSION = 1;
+
+constexpr int DEFAULT_BATCH_SIZE = 32;
+
+constexpr size_t MEMORY_REQUIRED_FOR_DEFAULT_BATCH_SIZE = 4294967296;
```
nickjong: Let's use the `lu` suffix here to ensure that we're defining a 64-bit constant. (The default is `int`, which on most of our platforms is 32-bit, which is not big enough to represent this number!)

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+shared_float_array prepare_images(turi::image_type image) {
```
nickjong: Pass `image` by const ref to avoid a copy

src/toolkits/style_transfer/style_transfer.cpp
nickjong: Nit: use `static_cast<flex_int>`

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+void style_transfer::load_version(iarchive& iarc, size_t version) {
+  variant_deep_load(state, iarc);
+
+  float_array_map nn_params;
+  iarc >> nn_params;
+
+  m_resnet_spec =
+      init_resnet(variant_get_value<size_t>(state.at("num_styles")));
+  m_resnet_spec->update_params(nn_params);
```
nickjong: How does the VGG spec get initialized?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+void style_transfer::infer_derived_options() {
+  if (read_state<flexible_type>("batch_size") == FLEX_UNDEFINED) {
+    flex_int batch_size = DEFAULT_BATCH_SIZE;
+
+    size_t memory_budget = m_training_compute_context->memory_budget();
+    if (memory_budget < MEMORY_REQUIRED_FOR_DEFAULT_BATCH_SIZE) {
+      batch_size /= 2;
+    }
```
nickjong: Is this copied from the Python ST or from the C++ OD? Do we know this is actually the right threshold to use?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+    turi::image_type content_image = batch[y].content_image;
+    turi::image_type style_image = batch[y].style_image;
+
+    shared_float_array index_array(std::make_shared<float_scalar>(style_index));
+
+    std::map<std::string, shared_float_array> results =
+        m_training_model->train({{"input", prepare_images(content_image)},
+                                 {"labels", prepare_images(style_image)},
+                                 {"index", index_array}});
```
nickjong: Doesn't this only pass one image at a time to the backend? Shouldn't we be passing an entire batch at a time?

test/toolkits/style_transfer/test_style_transfer.cxx
abhishekpratapa: Added TODOs to address the coming tests.

test/toolkits/style_transfer/test_style_transfer.cxx
abhishekpratapa: Wrong, name for the tests. `init_train` is protected. Updated naming.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+constexpr size_t MEMORY_REQUIRED_FOR_DEFAULT_BATCH_SIZE = 4294967296;
+
+shared_float_array prepare_images(turi::image_type image) {
+  constexpr size_t h = 256;
+  constexpr size_t w = 256;
+  constexpr size_t c = 3;
+
+  std::vector<float> result_array(h * w * c);
+  auto out_it = result_array.begin();
```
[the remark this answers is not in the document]
nickjong: Here we're in personal judgement territory, which of these is clearer:
```
auto out_it = result_array.begin();
```
or
```
std::vector<float>::iterator out_it = result_array.begin();
```
In this case, I guess I would choose the latter, although the fully specified type is becoming complex enough that it's close to preferring `auto` as more readable. In this particular case, the type can be inferred anyway just by looking at the previous line.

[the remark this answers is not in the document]
abhishekpratapa: removed auto.

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Copied this from C++ OD. Not sure if it's the right threshold to use.

nickjong: If the current MPS implementation we're shipping in 5.8 doesn't adjust batch size due to memory, let's not do it here, either

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,269 @@
+/* Copyright © 2018 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#include <toolkits/style_transfer/style_transfer.hpp>
```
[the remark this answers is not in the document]
abhishekpratapa: Screen Shot 2019-10-18 at 11 40 40 AM

In this instance: #include <toolkits/style_transfer/style_transfer.hpp> acts as (1. dir2/foo2.h). Rather than a relative import we're using an absolute import for this.

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Good catch

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+  image_type resized_image =
+      image_util::resize_image(image, width, height, channels, true, 1);
+
+  const unsigned char* resized_image_ptr = resized_image.get_image_data();
+
+  std::vector<float>::iterator start_iter = array.begin() + offset;
+  std::vector<float>::iterator end_iter = start_iter + image_size;
+
+  populate_image(start_iter, end_iter, resized_image_ptr);
```
nickjong: FYI, instead of defining `populate_image`, this could also just be a call to `std::transform`:
```
std::transform(resized_image_ptr, resized_image_ptr + image_size, start_iter,
    [](unsigned char val) { return val / 255.f; });
```
https://en.cppreference.com/w/cpp/algorithm/transform

In this case, instead of passing in `start_iter` and `end_iter` (per my last comment), you could just pass in `start_iter`.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,354 @@
...
+  variant_deep_save(state, oarc);
+  oarc << m_resnet_spec->export_params_view()
+       << m_vgg_spec->export_params_view();
+}
+
+void style_transfer::load_version(iarchive& iarc, size_t version) {
+  variant_deep_load(state, iarc);
+
+  float_array_map nn_params;
+  iarc >> nn_params;
```
nickjong: You wrote two `float_array_map` values in `save_impl` but only read one `float_array_map` value here.... Did you mean to still write the VGG spec?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+  add_or_update_state({
+    {"training_iterations", iteration_idx + 1},
+  });
+
+  shared_float_array loss_batch = results.at("loss");
+
+
+  float batch_loss = std::accumulate(loss_batch.data(), loss_batch.data() + loss_batch.size(), 0.f,
+        [&loss_batch](float a, float b) { return a + b / loss_batch.size(); });
```
nickjong: Nit: I think this will be more readable if we compute the loss batch size upfront and then just capture that `size_t` by value, instead of capturing the `shared_float_array` by reference. Keeps things simpler and easier to read

test/toolkits/style_transfer/test_style_transfer.cxx
```diff
@@ -0,0 +1,49 @@
...
+BOOST_AUTO_TEST_CASE(test_train) {
```
nickjong: This PR is already rather big, but we should make some effort to test the functions you're defining. In the future, please consider splitting this work into multiple PRs, where each PR defines one function and its related tests and support code!

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: I removed it, must have slipped my mind thanks!

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: WOAH, didn't know transform did that. That's cool!

test/toolkits/style_transfer/test_style_transfer.cxx
abhishekpratapa: I didn't expect the train PR to blow up so much. Next time I'll think about what components go into files a bit more. I did it on a macro level with model_spec, compute_context and data loaders. I'll figure out how to do it on a micro scale per file. I added a couple of more tests, nothing crazy just checking no see-faults. I'll add more for the predict PR.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -283,23 +270,26 @@ void style_transfer::iterate_training() {
...
 
   shared_float_array loss_batch = results.at("loss");
 
-
-  float batch_loss = std::accumulate(loss_batch.data(), loss_batch.data() + loss_batch.size(), 0.f,
-        [&loss_batch](float a, float b) { return a + b / loss_batch.size(); });
+  size_t loss_batch_size = loss_batch.size();
+  float batch_loss = std::accumulate(
+      loss_batch.data(), loss_batch.data() + loss_batch_size, 0.f,
+      [&loss_batch_size](float a, float b) { return a + b / loss_batch_size; });
```
nickjong: For a `size_t` just capture by value, not reference

test/toolkits/style_transfer/test_style_transfer.cxx
```diff
@@ -16,8 +16,85 @@
 
 #include "utils.hpp"
 
+/**
+ * Since the only backend is MPS, these tests won't pass on any system other
+ * than 10.15.
+ */
+#ifdef HAS_MACOS_10_15
```
nickjong: Why do our unit tests for the backend-agnostic C++ ST code depend on any particular platform or backend?

test/toolkits/style_transfer/test_style_transfer.cxx
```diff
@@ -16,8 +16,85 @@
...
+
 using namespace turi::style_transfer;
 
+
+BOOST_AUTO_TEST_CASE(test_init) {
```
nickjong: What does this test assert?

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -398,4 +410,54 @@ std::shared_ptr<coreml::MLModelWrapper> export_style_transfer_model(
   return model_wrapper;
 }
 
+std::shared_ptr<coreml::MLModelWrapper> export_drawing_classifier_model(
```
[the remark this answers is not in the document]
abhishekpratapa: I think I pulled from master incorrectly. Let me see. I didn't write that code.

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -120,33 +120,45 @@ void set_threshold_feature(FeatureDescription* feature_desc, std::string feature
...
 
-  if (use_flexible_shape) {
-    ImageFeatureType_ImageSizeRange* image_size_range = image_feature->mutable_imagesizerange();
-    
-    SizeRange* width_range = image_size_range->mutable_widthrange();
-    SizeRange* height_range = image_size_range->mutable_heightrange();
-
-    width_range->set_lowerbound(64);
-    width_range->set_upperbound(-1);
+  if (use_flexible_shape) set_image_feature_size_range(image_feature);
```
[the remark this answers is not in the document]
abhishekpratapa: Didn't know how to reason about it. I thought all image things should go in set_image_feature, but breaking it out makes sense

test/toolkits/style_transfer/test_style_transfer.cxx
abhishekpratapa: On linux there isn't a compute context for style transfer yet that's callable directly through C++. So the create_compute context will return a `nullptr` this will throw an exception. But since I can't figure out how to test that I took these tests out.

abhishekpratapa: Even the TensorFlow code is only callable through python because of PyBind11 dependencies. Not sure we can call that code directly through C++.

src/toolkits/coreml_export/neural_net_models_exporter.cpp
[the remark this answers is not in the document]
abhishekpratapa: It should be resolved.

test/toolkits/style_transfer/test_style_transfer.cxx
abhishekpratapa: I was presuming that no exceptions are thrown. I tried to add checks to see if the class was populated by checking the member variables but couldn't access them because they were private. Not exactly sure how to proceed.

test/toolkits/style_transfer/test_style_transfer.cxx
nickjong: Please look at the unit tests for OD, and the ones that @shantanuchhabra and @Jarvi-Izana are now adding for DC

PR #2462, opened 2019-10-16
AC predict_topk c++

jakesabathia2:
Implement c++ version of AC's predict_topk() function.


Jarvi-Izana:
@jakesabathia2 definitely, it helps a lot!

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,127 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
...
+  raw_preds_per_window.add_column(
+      raw_preds_per_window["preds"].apply(argsort_prob, flex_type_enum::LIST),
+      "rank");
+
+  // get top k class name
+  size_t rank_column_index = raw_preds_per_window.column_index("rank");
+  auto get_class_name = [=](const sframe_rows::row& row) {
+    const flex_list& rank_list = row[rank_column_index];
+    flex_list topk_class;
+    for (auto i : rank_list) {
```
shantanuchhabra: Could we avoid using `auto` here? See https://google.github.io/styleguide/cppguide.html#Type_deduction

"Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type."

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,127 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
...
+      raw_preds_per_window.apply(get_class_name, flex_type_enum::LIST),
+      "class");
+
+  // if output_type = probability then change rank to probabilty
+  if (output_type == "probability") {
+    size_t prob_column_index = raw_preds_per_window.column_index("preds");
+    auto get_probability = [=](const sframe_rows::row& row) {
+      const flex_list& rank_list = row[rank_column_index];
+      flex_list topk_prob;
+      for (auto& i : rank_list) {
```
shantanuchhabra: Same as above.

Jarvi-Izana:
if the `perform_inferrence` function has a mock, it should be easy to add one unittest.

jakesabathia2:
Yes I do have unit test for classify function for another pr which is waiting for approved.
so after it is done I'll make another pr for the unit test.
However I have test this function on my local :)

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,127 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
...
+  } else if (output_frequency != "per_row" &&
+             output_frequency != "per_window") {
+    log_and_throw(output_frequency +
+                  " is not a valid option for output_frequency.  "
+                  "Expected one of: per_row, per_window");
+  }
+
+  // data inference
+  std::unique_ptr<data_iterator> data_it =
+      create_iterator(data, false, false, false);
```
shantanuchhabra: Could we add a brief comment explaining what all the different arguments here mean?

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,127 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
...
+  gl_sframe raw_preds_per_window = perform_inference(data_it.get());
+
+  // argsort probability to get index (rank) for top k class
+  flex_list class_labels = read_state<flex_list>("classes");
+  k = std::min(k, class_labels.size());
+  auto argsort_prob = [=](const flexible_type& ft) {
```
shantanuchhabra: The code here on out is very dense, it might be helpful to add a few comments at places like this, where something not so obvious and self-explanatory is happening.

src/toolkits/activity_classification/activity_classifier.cpp
jakesabathia2: I think auto is used here is because it is a lambda function?
https://en.cppreference.com/w/cpp/language/lambda

src/toolkits/activity_classification/activity_classifier.cpp
jakesabathia2: Sure!

src/toolkits/activity_classification/activity_classifier.cpp
jakesabathia2: Sure!

src/toolkits/activity_classification/activity_classifier.cpp
nickjong: This usage of `auto` is unrelated to the lambda that encloses it.

In particular, why can't this just be
```
for (const flexible_type& i : rank_list)
```

Although usually we don't use `i` for feature values. More common in our code is `ft` for a `flexible_type` value

jakesabathia2: Okay! I thought @shantanuchhabra  is talking about the auto for the lambda function :)

jakesabathia2:
Just add unit test.

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,127 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
...
+    const flex_vec& prob_vec = ft.get<flex_vec>();
+    std::vector<size_t> index_vec(prob_vec.size());
+    std::iota(index_vec.begin(), index_vec.end(), 0);
+    auto compare = [&](const size_t& i, const size_t& j) {
+      return prob_vec[i] > prob_vec[j];
+    };
+    std::sort(index_vec.begin(), index_vec.end(), compare);
```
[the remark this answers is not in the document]
jakesabathia2: nice! thanks!

src/toolkits/activity_classification/activity_classifier.cpp
jakesabathia2: Nice!

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,133 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
   return result;
 }
 
+gl_sframe activity_classifier::predict_topk(gl_sframe data,
+                                            std::string output_type, size_t k,
+                                            std::string output_frequency) {
+  // check valid input arguments
+  if (output_type.empty()) {
```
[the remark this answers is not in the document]
jakesabathia2: Yes the default value is a empty string (I just follow the "predict" function).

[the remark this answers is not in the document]
jakesabathia2: But you are right make it "per_row" make much more sense. Thanks!

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,133 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
...
+    output_type = "probability";
+  } else if (output_type != "probability" && output_type != "rank") {
+    log_and_throw(output_type +
+                  " is not a valid option for output_type.  "
+                  "Expected one of: probability, rank");
+  }
+  if (output_frequency.empty()) {
```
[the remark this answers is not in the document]
jakesabathia2: :))

test/unity/toolkits/activity_classifier/test_activity_classifier.cxx
```diff
@@ -0,0 +1,615 @@
...
+  const std::vector<size_t> session_distribution = {1, 3, 5, 7, 9};
+  TS_ASSERT_EQUALS(session_num, session_distribution.size());
+
+  constexpr size_t test_num_examples = 25;
+  size_t check_sum = 0;
+  for (size_t num_per_session : session_distribution)
+    check_sum += num_per_session;
+  TS_ASSERT_EQUALS(check_sum, test_num_examples);
+
+  /** mock predict data
```
Jarvi-Izana: add a newline. that's convention.

jakesabathia2:
So I have several changes:
(1) Thanks @Jarvi-Izana 's suggest that improve the predict top k by first using quick select for the 4th element and then sort it.
(2) Add some comments on the unit test. Thanks again for @Jarvi-Izana comments.

test/unity/toolkits/activity_classifier/test_activity_classifier.cxx
jakesabathia2: :))))))

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -464,6 +464,130 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
...
+  flex_list class_labels = read_state<flex_list>("classes");
+  k = std::min(k, class_labels.size());
+  auto argsort_prob = [=](const flexible_type& ft) {
+    const flex_vec& prob_vec = ft.get<flex_vec>();
+    std::vector<size_t> index_vec(prob_vec.size());
+    std::iota(index_vec.begin(), index_vec.end(), 0);
+    auto compare = [&](size_t i, size_t j) {
+      return prob_vec[i] > prob_vec[j];
+    };
+    std::nth_element(index_vec.begin(), index_vec.begin() + k, index_vec.end(),
```
Jarvi-Izana: I just realized that you need to check k is not larger than prob_vec.size(), which means the k should be larger than the number of class labels in the current data iterator. Otherwise, the behavior will be undefined. Welcome to Cpp world. You should put a pr to prevent potential segfault.

The logic should be like, 
```
if (!raw_preds_per_window["preds"].empty())
k = std::min(k, raw_preds_per_window["preds"][0].size());
```
Let's have DASSERT to make sure each cell of raw_preds_per_window["preds"] has same size, if the sframe is not empty.

jakesabathia2: I think I do check k's range at 491 ?

PR #2464, opened 2019-10-17
Annotations should error out on empty Frame

### LINKED ISSUE #1915: Annotations should error out on empty Frame
Right now, I get back a new SFrame without an annotation column
```
In [16]: aa = tc.image_classifier.annotate(sf[0:0])

In [17]: aa
Out[17]:
Columns:
	path	str
	image	Image

Rows: 0

Data:
	[]
```

dhivyaaxim:
This PR resolves #1915 
Annotations should error out on empty Frame

### DIFF
--- a/src/python/turicreate/toolkits/image_classifier/_annotate.py
+++ b/src/python/turicreate/toolkits/image_classifier/_annotate.py
@@ -72,13 +72,13 @@ def annotate(data, image_column=None, annotation_column='annotations'):
         [4 rows x 3 columns]
 
     """
-    # Check if Value is Empty
-    if __tc.SFrame.num_rows(data)== 0:
-        raise Exception('input data cannot be empty')
-
     # Check Value of Column Variables
     if not isinstance(data, __tc.SFrame):
         raise TypeError('"data" must be of type SFrame.')
+    
+    # Check if Value is Empty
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/image_classifier/_annotate.py
```diff
@@ -72,6 +72,10 @@ def annotate(data, image_column=None, annotation_column='annotations'):
         [4 rows x 3 columns]
 
     """
+    # Check if Value is Empty
+    if __tc.SFrame.num_rows(data)== 0:
```
TobyRoseman: If that `data` is not even an SFrame, then when this line is evaluated the user will get an error message which is very difficult to understand. Let's move this code to line 79, after we've checked that `data` is an SFrame.

Calling `__tc.SFrame.num_rows(data)` to get the number of row will work but doing either `len(data)` or `data.num_rows()` is more standard and probably better.

One very minor note - the standard Python style guide lines (i.e. [PEP 8](https://pep8.org)) says you should have a space on both sides of a `==`.

TobyRoseman:
Thanks @dhivyaaxim for addressing my feedback. I've restarted the Traivs CI job which timed out. Feel free to merge once that's completed.

PR #2466, opened 2019-10-17
gitignore for CMake builds

Jarvi-Izana:
FYI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,16 @@ deps/build
 # to the root of the project for most tooling to work. This ignores the symlink.
 /compile_commands.json
 
+# CMake builds
+## dir match
+**/cmake-build-debug
+**/cmake_install.cmake
+**/CMakeFiles
+**/CTestTestfile.cmake
+## file match from root
+CMakeCache.txt
+Makefile
+
 # Visual Studio Code files
 /.vscode/
 


.gitignore
```diff
@@ -98,6 +98,16 @@ deps/build
...
 
+# CMake builds
+## dir match
+**/cmake-build-debug
+**/cmake_install.cmake
+**/CMakeFiles
+**/CTestTestfile.cmake
+## file match from root
+CMakeCache.txt
+Makefile
```
TobyRoseman: We have several `Makefile` currently checked into:

```
./userguide/supervised-learning/notebooks/Makefile
./pydocs/Makefile
./src/python/doc/Makefile
./src/external/xgboost/Makefile
./src/external/xgboost/subtree/rabit/guide/Makefile
./src/external/xgboost/subtree/rabit/test/Makefile
./src/external/xgboost/subtree/rabit/Makefile
./src/external/xgboost/subtree/rabit/doc/Makefile
./src/external/xgboost/doc/Makefile
./src/external/bzip2/Makefile
```

I don't think we want to ignore these. So I don't think we want to add this line.

Jarvi-Izana: Oh, they are not generated by CMake? I see. I run cmake outside of those dirs you mentioned. Probably I shouldn't avoid run cmake outside of release or debugs.

PR #2467, opened 2019-10-18
Added `USE_CPP` flag for AC and OD

fareeha:
**Usage:**

On CLI:

To enable C++ model codepath:
```
export TURI_OD_USE_CPP_PATH="1"
export TURI_AC_USE_CPP_PATH="1"
```
(or any input other than '0')

To disable C++ model codepath i.e. enable Python codepath:

```
export TURI_OD_USE_CPP_PATH="0"
export TURI_AC_USE_CPP_PATH="0"
```

By default or in the absence of the above env variables, the behavior is to use Python codepath.

_Setting these env variable will allow python unit testing to use the desired codepath._

### DIFF
--- a/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -30,10 +30,7 @@ from turicreate.toolkits._model import PythonProxy as _PythonProxy
 from .util import random_split_by_session as _random_split_by_session
 from .util import _MIN_NUM_SESSIONS_FOR_SPLIT
 
-import os as _os
-USE_CPP = True
-if not _os.environ.has_key('ac_use_cpp') or _os.environ.get('ac_use_cpp')=="0":
-    USE_CPP = False
+USE_CPP = _tkutl._read_env_var_cpp('TURI_AC_USE_CPP_PATH')
 
 def create(dataset, session_id, target, features=None, prediction_window=100,
            validation_set='auto', max_iterations=10, batch_size=32, verbose=True, **kwargs):
--- a/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -30,6 +30,11 @@ from turicreate.toolkits._model import PythonProxy as _PythonProxy
 from .util import random_split_by_session as _random_split_by_session
 from .util import _MIN_NUM_SESSIONS_FOR_SPLIT
 
+import os
+use_cpp = True
+if not os.environ.has_key('ac_use_cpp') or os.environ.get('ac_use_cpp')=="0":
+    use_cpp = False
+
 def create(dataset, session_id, target, features=None, prediction_window=100,
            validation_set='auto', max_iterations=10, batch_size=32, verbose=True, **kwargs):
     """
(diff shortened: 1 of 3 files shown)


hoytak:
Looks great, but to follow Turi convention, we should name them TURI_OD_USE_CPP_PATH and TURI_AC_USE_CPP_PATH

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -30,6 +30,11 @@
 from .util import random_split_by_session as _random_split_by_session
 from .util import _MIN_NUM_SESSIONS_FOR_SPLIT
 
+import os
```
srikris: Always import using _os

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -342,858 +344,859 @@ def _encode_target(data, target, mapping=None):
...
-
-    @classmethod
-    def _native_name(cls):
-        return "activity_classifier"
-
-    def _get_version(self):
-        return self._PYTHON_ACTIVITY_CLASSIFIER_VERSION
-
-    def export_coreml(self, filename):
+if use_cpp:
```
srikris: @TobyRoseman Is this an okay thing to do? i.e define a class inside an if

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
TobyRoseman: That good general practice and it's not going to hurt here, but I don't think it's needed. The filename for this file starts with an underscore so we're already in a hidden scope.

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
TobyRoseman: Well I'm not finding anything about this specifically in PEP8, but would recommend not doing it.

It adds a lot of additional nesting. Also it really break the diff; I can't tell what has changed and what hasn't.

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -41,6 +41,11 @@
 
 _MXNET_MODEL_FILENAME = "mxnet_model.params"
 
+import os
```
TobyRoseman: This is a case where you should `import os as _os`

TobyRoseman:
I'm a bit confused. Based on the description of the pull request (i.e. allowing a control flag to be specified via an environment variable) it seems like the diff is quite large. Is it just a lot of white space changes? Or is there other functionality included in this pull request?

fareeha:
> I'm a bit confused. Based on the description of the pull request (i.e. allowing a control flag to be specified via an environment variable) it seems like the diff is quite large. Is it just a lot of white space changes? Or is there other functionality included in this pull request?

Honestly I'm not sure why the diff is showing like that. I only moved the classes in if-else. I also tried to re-do it by checking out the files from master again but the diff kept showing in a weird broken way.

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
fareeha: @TobyRoseman What is an alternative to the if-else here? We want to keep the class names the same and use one or the other based on what codepath we want to select. 
PS This is a temporary hack to enable using both Python and C++ implementations till we remove Python CustomModel class completely.

TobyRoseman: Why do you need to keep the class names the same? Python is dynamically typed so I don't think it should matter. If you do really need to keep the classnames the same, I'm not sure there is another option.

nickjong: Per offline discussion, I think we should find a way to get the flag behavior testable without changing any names. The goal is to make both implementations testable without code changes, not to perform the actual migration (throwing the switch)...

Jarvi-Izana:
The internal global var is registered through `REGISTER_GLOBAL`, in `globals.hpp`.

The example can be found in `src/core/storage/fileio/fileio_constants.cpp`.

This is useful for DC as well.

PR #2469, opened 2019-10-18
Save and Load for C++ codepath for OD and AC

fareeha:
These changes make use of the `use_cpp` flag to define the codepath taken. 

Both C++ and Python models (i.e. from the new and old implementations) are loaded correctly with the C++ codepath. 

_PS Models from the new implementations are not loaded corrected in the old Python loaded._

### DIFF
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -27,6 +27,16 @@ import six as _six
 
 MODEL_NAME_MAP = {}
 
+# Object detector use C++ codepath
+od_use_cpp = True
+if not os.environ.has_key('od_use_cpp') or os.environ.get('od_use_cpp')=="0":
+    od_use_cpp = False
+
+# Activity Classifier use C++ codepath
+ac_use_cpp = True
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/_model.py
```diff
@@ -27,6 +27,16 @@
...
+# Object detector use C++ codepath
+od_use_cpp = True
+if not os.environ.has_key('od_use_cpp') or os.environ.get('od_use_cpp')=="0":
+    od_use_cpp = False
+
+# Activity Classifier use C++ codepath
+ac_use_cpp = True
+if not os.environ.has_key('ac_use_cpp') or os.environ.get('ac_use_cpp')=="0":
+    ac_use_cpp = False
+
```
nickjong: I suspect these should be encapsulated into helper functions, instead of code running at the top level on import. Probably this should be handled by https://github.com/apple/turicreate/pull/2467
?

fareeha: @nickjong That's a good idea. I'm thinking of adding a helper function to `_internal_utils.py`. Does that sound good?

PR #2470, opened 2019-10-21
dict_keys and dict_values types do not work with filter_by #2221

### LINKED ISSUE #2221: dict_keys and dict_values types do not work with filter_by
```python
import turicreate as tc                                                                                                                                                                                                                                                                                                                                           

sa = tc.SArray(range(10))                                                                                                                                                                                                                                                                                                                                         
d = {1:1}

sa.filter_by(d.keys())         # Does not work 
sa.filter_by(d.values())       # Also does not work

# of course this can be solve by doing: 
sa.filter_by(list(d.keys()))
sa.filter_by(list(d.values()))
```
This issue is only effecting Python 3. This issues should also be fixed for `sframe.filter_by`.

The fix here could be quite similar to #1723. However we should consider doing something more general, rather than special casing each possible type. We probably want to allow any iterable type as input. Determining if an object is iterable [is not straight forward](https://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable).

dhivyaaxim:
This PR resolves #2221 
dict_keys and dict_values types do not work with filter_by

### DIFF
--- a/src/python/turicreate/data_structures/sarray.py
+++ b/src/python/turicreate/data_structures/sarray.py
@@ -375,7 +375,7 @@ class SArray(object):
         # In Python 3, str implements '__iter__'.
         return (isinstance(obj, types.GeneratorType) or
                 (sys.version_info.major < 3 and isinstance(obj, six.moves.xrange)) or
-                sys.version_info.major >= 3 and isinstance(obj, (range, filter, map)))
+                sys.version_info.major >= 3 and isinstance(obj, (range, filter, map, collections.abc.KeysView, collections.abc.ValuesView)))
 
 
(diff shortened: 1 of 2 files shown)


hoytak:
This looks great, but it would be fantastic to add a unit test for this.   Please just add a "this works" test in test/test_sframe.py.  Thanks @dhivyaaxim for the great work!!

hoytak:
The error that caused the above CI failures has been fixed elsewhere.  Could you merge in master and push again?

dhivyaaxim:
@hoytak Thanks!  Can you merge this changes to master?

TobyRoseman:
@dhivyaaxim - Thanks for another great bug fix. For future reference: once your changes have been approved and passed Travis CI, feel free to merge.

PR #2474, opened 2019-10-22
Augmentation done using TF

shreyajain17:


nickjong:
FWIW, given the circumstances I think it's fine to merge Nivii's work as-is and then just modify it as necessary for our purposes

shreyajain17:
> FWIW, given the circumstances I think it's fine to merge Nivii's work as-is and then just modify it as necessary for our purposes

I didn't see this. I did some refactoring but it still keeps her code quite intact.

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -223,6 +223,13 @@ class image_augmenter {
      * -0.05 * pi and 0.05 * pi.
      */
     float hue_max_jitter = 0.f;
+
+    /**
+     * When set to true, the images will only be resized in the output shape
+     * Used by the subclass of float_array_image_augmenter, to only resize
+     * images in case of prediction and evaluation.
+     */
+    bool resize_only = false;
```
nickjong: Why is this needed? The default values of this struct imply resize-only anyway. See https://github.com/apple/turicreate/blob/108139b62f77a2965de4c858ec21290e30e8123e/src/toolkits/object_detection/object_detector.cpp#L576

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -308,7 +315,8 @@ class float_array_image_augmenter : public image_augmenter {
   };
 
   virtual float_array_result prepare_augmented_images(
-      labeled_float_image data_to_augment) = 0;
+      labeled_float_image data_to_augment, size_t output_height,
+      size_t output_width, bool resize_only) = 0;
```
nickjong: Why do these parameters need to be added? Isn't this captured in the `options` struct?

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -212,7 +215,8 @@ float_array_image_augmenter::float_array_result tf_image_augmenter::prepare_augm
 
     // Get augmented images and annotations from tensorflow
     pybind11::object augmented_data = tf_aug.attr("get_augmented_data")(
-        data_to_augment.images, data_to_augment.annotations);
+        data_to_augment.images, data_to_augment.annotations, output_height,
+        output_width, resize_only);
```
nickjong: Can't we just pull e.g. `output_width` from `get_options`? Why do they need to be passed explicitly to this method?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -579,6 +582,7 @@ void object_detector::perform_predict(gl_sframe data,
   augmenter_opts.batch_size = batch_size;
   augmenter_opts.output_height = grid_height * SPATIAL_REDUCTION;
   augmenter_opts.output_width = grid_width * SPATIAL_REDUCTION;
+  augmenter_opts.resize_only = true;
```
nickjong: The fact that this is resize-only is implied by the options being set. This flag is redundant

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -51,7 +53,7 @@ def __init__(self, input_h, input_w, batch_size, output_size, init_weights, conf
                                         name="global_step")
 
         self.loss = self.loss_layer(self.tf_model, self.labels)
-        self.base_lr = _utils.convert_shared_float_array_to_numpy(config['learning_rate'])
+        self.base_lr = _utils.convert_shared_float_array_to_numpy(config['learning_rate'])
```
nickjong: Please don't introduce trailing whitespace

src/ml/neural_net/image_augmentation.hpp
shreyajain17: So I can just check if the default settings are on and I should just pass resize flag to python based on the check.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
shreyajain17: Okay

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: I will refactor and only pass these flags to tensorflow.

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -201,18 +201,28 @@ class tf_image_augmenter : public float_array_image_augmenter {
 
 tf_image_augmenter::tf_image_augmenter(const options& opts) : float_array_image_augmenter(opts) {}
 
-float_array_image_augmenter::float_array_result tf_image_augmenter::prepare_augmented_images(
+float_array_image_augmenter::float_array_result
+tf_image_augmenter::prepare_augmented_images(
     float_array_image_augmenter::labeled_float_image data_to_augment) {
+  opts_ = tf_image_augmenter::get_options();
```
nickjong: Why is this needed?

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -201,18 +201,28 @@ class tf_image_augmenter : public float_array_image_augmenter {
...
 
   call_pybind_function([&]() {
     // Import the module from python that does data augmentation
     pybind11::module tf_aug = pybind11::module::import(
         "turicreate.toolkits.object_detector._tf_image_augmenter");
 
+    const size_t output_height = opts_.output_height;
+    const size_t output_width = opts_.output_width;
+    bool resize_only = false;
+    if (opts_.crop_prob == 0.f) {
```
nickjong: Worth a TODO that in theory you need to test all the augmentation probabilities, but in practice this suffices until we plumb through all the options to Python

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: I am calling this to get options from its base class.

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: Okay

shreyajain17:
> You shouldn’t need to assign the options, they were provided to the constructor. What breaks if you just delete the line in question?

I get zeros for all the options if I remove the line `opts_ = tf_image_augmenter::get_options();`

nickjong:
> > You shouldn’t need to assign the options, they were provided to the constructor. What breaks if you just delete the line in question?
> 
> I get zeros for all the options if I remove the line `opts_ = tf_image_augmenter::get_options();`
> 
> `ValueError: all dims of 'image.shape' must be > 0: (0, 0, 3)`
> It is because output height and output width sent to tensorflow is 0.

Ah, I see what happened (via visual code debugging). You created a shadow `opts_` data member here:
https://github.com/apple/turicreate/blob/944a7e3ae6f98b2eefa1628e63415c56fb16a54a/src/ml/neural_net/tf_compute_context.cpp#L199
This `tf_image_augmenter::opts_` variable shadows the `float_array_image_augmenter::opts_` from the parent class. The correct fix is just to delete the shadow `opts_` and to use `get_options()` directly. No need to write `tf_image_augmenter::get_options()`, just `get_options()` will suffice. Your class inherits this stuff from the base class. There's no need to manually duplicate stuff you inherit, in OOP.

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -194,25 +194,34 @@ class tf_image_augmenter : public float_array_image_augmenter {
...
-  options opts_;
 };
 
 tf_image_augmenter::tf_image_augmenter(const options& opts) : float_array_image_augmenter(opts) {}
 
-float_array_image_augmenter::float_array_result tf_image_augmenter::prepare_augmented_images(
+float_array_image_augmenter::float_array_result
+tf_image_augmenter::prepare_augmented_images(
     float_array_image_augmenter::labeled_float_image data_to_augment) {
+  options opts_ = get_options();
```
nickjong: Generally, local variables shouldn't get the `_` suffix

PR #2479, opened 2019-10-22
dc save and load

Jarvi-Izana:
__MY OVERVIEW__
- save and load impl
- corresponding test

__MY SUMMARY__
[Y] provide unittest? [Y/N]
[Y] pass the test pipeline? [Y/N]

### DIFF
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -50,6 +50,39 @@ struct result {
 
 }  // namespace
 
+const size_t drawing_classifier::DRAWING_CLASSIFIER_VERSION = 1;
+
+size_t drawing_classifier::get_version() const {
+  return DRAWING_CLASSIFIER_VERSION;
+}
+
+void drawing_classifier::save_impl(oarchive& oarc) const {
+  if (!nn_spec_)
+    log_and_throw(
+        "model spec is not initalized, pls call `init_train` before saving model");
+
+  // Save model attributes.
+  variant_deep_save(state, oarc);
+
+  // Save neural net weights.
+  oarc << nn_spec_->export_params_view();
+}
+
+void drawing_classifier::load_version(iarchive& iarc, size_t version) {
+  if (!nn_spec_)
+    log_and_throw(
+        "model spec is not initalized, pls call `init_train` before loading model");
+  // Load model attributes.
+  variant_deep_load(state, iarc);
+
+  // Load neural net weights.
+  float_array_map nn_params;
+  iarc >> nn_params;
+
+  nn_spec_ = init_model();
+  nn_spec_->update_params(nn_params);
+}
+
 std::unique_ptr<model_spec> drawing_classifier::init_model() const {
   std::unique_ptr<model_spec> result(new model_spec);
 
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -223,7 +223,6 @@ void drawing_classifier::init_options(
       1,
       std::numeric_limits<int>::max());
 
-
   // Validate user-provided options.
   options.set_options(opts);
 
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -72,6 +72,7 @@ void drawing_classifier::load_version(iarchive& iarc, size_t version) {
   if (!nn_spec_)
     log_and_throw(
         "model spec is not initalized, pls call `init_train` before loading model");
+
   // Load model attributes.
   variant_deep_load(state, iarc);
 
@@ -310,7 +311,7 @@ void drawing_classifier::init_training(
 
   // Initialize the neural net. Note that this depends on statistics computed by
   // the data iterator.
-  nn_spec_ = init_model();
+  init_model_spec();
 
   // TODO: Do not hardcode values
   training_model_ = training_compute_context_->create_drawing_classifier(
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -59,7 +59,7 @@ size_t drawing_classifier::get_version() const {
 void drawing_classifier::save_impl(oarchive& oarc) const {
   if (!nn_spec_)
     log_and_throw(
-        "model spec is not initalized, pls call `init_train` before saving model");
+        "model spec is not initalized, please call `init_train` before saving model");
 
   // Save model attributes.
   variant_deep_save(state, oarc);
@@ -71,7 +71,7 @@ void drawing_classifier::save_impl(oarchive& oarc) const {
 void drawing_classifier::load_version(iarchive& iarc, size_t version) {
   if (!nn_spec_)
     log_and_throw(
-        "model spec is not initalized, pls call `init_train` before loading model");
+        "model spec is not initalized, please call `init_train` before loading model");
 
   // Load model attributes.
   variant_deep_load(state, iarc);
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -311,7 +311,7 @@ void drawing_classifier::init_training(
 
   // Initialize the neural net. Note that this depends on statistics computed by
   // the data iterator.
-  init_model_spec();
+  nn_spec_ = init_model();
 
   // TODO: Do not hardcode values
   training_model_ = training_compute_context_->create_drawing_classifier(
--- a/src/toolkits/drawing_classifier/drawing_classifier.hpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.hpp
@@ -218,6 +218,14 @@ class EXPORT drawing_classifier : public ml_model_base {
 
   END_CLASS_MEMBER_REGISTRATION
 
+  inline std::unique_ptr<neural_net::model_spec> get_model_spec_copy() const {
+    if (nn_spec_) {
+      return nn_spec_->clone();
+    } else {
+      return nullptr;
+    }
+   }
+
  protected:
   // Constructor allowing tests to set the initial state of this class and to
   // inject dependencies.
@@ -262,16 +270,13 @@ class EXPORT drawing_classifier : public ml_model_base {
     return variant_get_value<T>(get_state().at(key));
   }
 
-  /* mock test only ; not intended for inheritance */
-  inline void init_model_spec() { nn_spec_ = init_model(); }
+  /* mock test only */
 
-  inline std::unique_ptr<neural_net::model_spec> get_model_spec_copy() const {
-    if (nn_spec_) {
-      return nn_spec_->clone();
-    } else {
-      return nullptr;
-    }
-   }
+  /**
+   * not intended for inheritance
+   * only initialize nn_spec_ for test convenience
+   **/
+  inline void init_model_spec() { nn_spec_ = init_model(); }
 
  private:
   // Primary representation for the trained model.
--- a/src/toolkits/drawing_classifier/drawing_classifier.hpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.hpp
@@ -280,7 +280,9 @@ class EXPORT drawing_classifier : public ml_model_base {
 };
 
 /* minimum test suite friends */
-inline void init_dc_model_spec_for_test(drawing_classifier& dc) { dc.nn_spec_ = dc.init_model(); };
+inline void init_dc_model_spec_for_test(drawing_classifier& dc) {
+  dc.nn_spec_ = dc.init_model();
+};
 
 }  // namespace drawing_classifier
 }  // namespace turi
--- a/src/toolkits/drawing_classifier/drawing_classifier.hpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.hpp
@@ -23,6 +23,8 @@ namespace drawing_classifier {
 
 class EXPORT drawing_classifier : public ml_model_base {
  public:
+  static const size_t DRAWING_CLASSIFIER_VERSION;
+
   drawing_classifier() = default;
 
   /**
@@ -215,8 +217,6 @@ class EXPORT drawing_classifier : public ml_model_base {
   REGISTER_CLASS_MEMBER_FUNCTION(drawing_classifier::iterate_training);
 
   END_CLASS_MEMBER_REGISTRATION
-
-  static const size_t DRAWING_CLASSIFIER_VERSION;
 
  protected:
   // Constructor allowing tests to set the initial state of this class and to
--- a/src/toolkits/drawing_classifier/drawing_classifier.hpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.hpp
@@ -218,14 +218,6 @@ class EXPORT drawing_classifier : public ml_model_base {
 
   END_CLASS_MEMBER_REGISTRATION
 
-  inline std::unique_ptr<neural_net::model_spec> get_model_spec_copy() const {
-    if (nn_spec_) {
-      return nn_spec_->clone();
-    } else {
-      return nullptr;
-    }
-   }
-
  protected:
   // Constructor allowing tests to set the initial state of this class and to
   // inject dependencies.
@@ -269,6 +261,17 @@ class EXPORT drawing_classifier : public ml_model_base {
   T read_state(const std::string& key) const {
     return variant_get_value<T>(get_state().at(key));
   }
+
+  std::unique_ptr<neural_net::model_spec> clone_model_spec_for_test() const {
+    if (nn_spec_) {
+      return std::unique_ptr<neural_net::model_spec>(
+          new neural_net::model_spec(nn_spec_->get_coreml_spec()));
+    }
+    else {
+      return nullptr;
+    }
+  }
+
 
  private:
   // Primary representation for the trained model.
--- a/src/toolkits/drawing_classifier/drawing_classifier.hpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.hpp
@@ -34,7 +34,7 @@ class EXPORT drawing_classifier : public ml_model_base {
   void init_options(const std::map<std::string, flexible_type>& opts) override;
   /* Commented out for the purpose of a skeleton. */
 
-  // void init_options(const std::map<std::string, flexible_type>& opts) override;
+  void init_options(const std::map<std::string, flexible_type>& opts) override;
 
   size_t get_version() const override;
 
--- a/src/toolkits/drawing_classifier/drawing_classifier.hpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.hpp
@@ -25,7 +25,9 @@ class EXPORT drawing_classifier : public ml_model_base {
  public:
   drawing_classifier() = default;
 
-  // ml_model_base interface
+  /**
+   * ml_model_base interface
+   */
 
   void init_options(const std::map<std::string, flexible_type>& opts) override;
   /* Commented out for the purpose of a skeleton. */
@@ -38,7 +40,9 @@ class EXPORT drawing_classifier : public ml_model_base {
 
   void load_version(iarchive& iarc, size_t version) override;
 
-  /* Interface exposed via Unity server */
+  /**
+   * Interface exposed via Unity server
+   */
 
   void train(gl_sframe data, std::string target_column_name,
              std::string feature_column_name, variant_type validation_data,
@@ -258,12 +262,13 @@ class EXPORT drawing_classifier : public ml_model_base {
     return variant_get_value<T>(get_state().at(key));
   }
 
-  // for test svae & load purpose
-  // well, for some reason nn_spec_ should not be exposed to user
+  /**
+   * for test svae & load purpose. well, nn_spec_ should not be exposed
+   * to user. I have to do this hack for test purpose
+   */
   friend void init_dc_model_spec_for_test(drawing_classifier& dc);
 
  private:
-
   // Primary representation for the trained model.
   std::unique_ptr<neural_net::model_spec> nn_spec_;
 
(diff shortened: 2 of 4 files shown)


Jarvi-Izana:
> Can you resolve conflicts against master?

You get it.

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -209,6 +214,8 @@ class EXPORT drawing_classifier : public ml_model_base {
 
   END_CLASS_MEMBER_REGISTRATION
 
+  static const size_t DRAWING_CLASSIFIER_VERSION;
```
nickjong: Let's put public constants at the top of the `public` section

https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -254,6 +261,12 @@ class EXPORT drawing_classifier : public ml_model_base {
     return variant_get_value<T>(get_state().at(key));
   }
 
+  /**
+   * for test svae & load purpose. well, nn_spec_ should not be exposed
+   * to user. I have to do this hack for test purpose
+   */
+  friend void init_dc_model_spec_for_test(drawing_classifier& dc);
+
```
nickjong: The pattern we're using in other toolkits is to provide a protected constructor that can directly initialize data members. Let's avoid the usage of `friend` if not necessary, since it's harder to reason about robustly

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -174,8 +179,140 @@ BOOST_AUTO_TEST_CASE(test_export_coreml) {
...
+        }
+      };
+
+  /**
+   * This is a hack because nn_spec_ is a private member. Thus, I can't
+   * compare it directly after loading from file (load_version call).
+   *
+   * The solution is to call `load_version` first and write serialized
+   * `nn_spec_` to disk. In this way, I can indirectly verify the `nn_spec_`
+   * successively loads all params from the source model spec file.
```
nickjong: Could we instead just add a protected getter for `nn_spec_`?

The other thing we can do is implement the core logic for save and load as free functions, as we did with OD.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -174,8 +179,140 @@ BOOST_AUTO_TEST_CASE(test_export_coreml) {
...
+      std::stringstream ss;
+      ss << "fail to open" << my_file << std::endl;
+      throw std::ios_base::failure(ss.str());
+    }
+    std::streamsize file_size = file.tellg();
+    file.seekg(0, std::ios::beg);
+
+    std::vector<char> model_spec_src(file_size);
+    file.read(model_spec_src.data(), file_size);
+    file.close();
```
nickjong: These are a lot of low-level filesystem operations for this kind of test. @hoytak Do we have any abstractions for testing serialization code?

Alternatively, typically I would expect that we could just test that a save/load round trip preserves the model spec and state.

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: Sure!

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: But `friend` is only used for test purpose and it's very common in test case where a private member is involved. Thus, I think here it's okay.

I don't really think we should put anything that's not intended to be inherited in `protected` section.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: These operations just tell me how large the file is. And then I load all bytes of those files and compare the contents all together to make sure we have the same results after loading.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: should we expose the `nn_spec_` to user? or I can have return a copy to user. Either way.

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: done.

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: I should rephrase in this way: this function only makes sense for testing.

What this function does is to only initialize `nn_spec_`  in order to avoid segfault when calling save and load. 

But from the user's perspective, they should expect an exception to be thrown when `save` or `load` is called on an uninitialized model, which means the `init_train` is not called beforehand.

Jarvi-Izana: rewrote the impl. Pls have a look.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -175,22 +209,17 @@ std::unique_ptr<model_spec> drawing_classifier::init_model() const {
 }
 
 void drawing_classifier::init_options(
-    const std::map<std::string, flexible_type>& opts) {
+    const std::map<std::string, flexible_type> &opts) {
```
Jarvi-Izana: lines after in this file are all style changes. Feel free to skip.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -49,6 +49,40 @@ struct result {
...
+const size_t drawing_classifier::DRAWING_CLASSIFIER_VERSION = 1;
+
+size_t drawing_classifier::get_version() const {
+  return DRAWING_CLASSIFIER_VERSION;
+}
+
+void drawing_classifier::save_impl(oarchive& oarc) const {
+  if (!nn_spec_)
+    log_and_throw(
+        "model spec is not initalized, pls call `init_train` before saving model");
```
shantanuchhabra: pls -> please

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -174,8 +180,148 @@ BOOST_AUTO_TEST_CASE(test_export_coreml) {
...
+        std::stringstream ss;
+        ss << "cannot remove file: " << fname << std::endl;
+        throw(turi::error::io_error(ss.str()));
+      }
+    }
+  };
+
+  /**
+   * The test calls `load_version` first and write serialized
+   * `nn_spec_` to disk. In this way, I can verify the `nn_spec_`
```
shantanuchhabra: I -> we

Jarvi-Izana: great

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: fantastic

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -261,11 +260,16 @@ class EXPORT drawing_classifier : public ml_model_base {
     return variant_get_value<T>(get_state().at(key));
   }
 
-  /**
-   * for test svae & load purpose. well, nn_spec_ should not be exposed
-   * to user. I have to do this hack for test purpose
-   */
-  friend void init_dc_model_spec_for_test(drawing_classifier& dc);
+  /* mock test only ; not intended for inheritance */
+  inline void init_model_spec() { nn_spec_ = init_model(); }
```
nickjong: Why can't we just use the protected `drawing_classifier` constructor provided above?

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -261,11 +260,16 @@ class EXPORT drawing_classifier : public ml_model_base {
...
+
+  inline std::unique_ptr<neural_net::model_spec> get_model_spec_copy() const {
```
nickjong: Why can't we just provide a protected accessor like
```
const neural_net::model_spec* get_model_spec() const { return nn_spec_.get(); }
```

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -24,11 +21,15 @@ namespace {
 class drawing_classifier_mock
     : public turi::drawing_classifier::drawing_classifier {
  public:
-  std::unique_ptr<neural_net::model_spec> my_init_model() const {
+  std::unique_ptr<neural_net::model_spec> my_get_model_spec() const {
     return init_model();
   }
```
nickjong: Instead of defining `my_get_model_spec` (we generally avoid first-person nomenclature), could we just expose the base class function like this:
```
public:
  using drawing_classifier::init_model;
```
(see also https://en.cppreference.com/w/cpp/language/using_declaration)

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -184,47 +185,44 @@ BOOST_AUTO_TEST_CASE(test_load_version) {
...
   /**
-   * This is a hack because nn_spec_ is a private member. Thus, I can't
-   * compare it directly after loading from file (load_version call).
-   *
-   * The solution is to call `load_version` first and write serialized
-   * `nn_spec_` to disk. In this way, I can indirectly verify the `nn_spec_`
-   * successively loads all params from the source model spec file.
+   * The test calls `load_version` first and write serialized
+   * `nn_spec_` to disk. In this way, the `nn_spec_` can be verified if
+   * it successively loads all params from the source model spec file.
```
nickjong: I still don't understand the core logic of this test. How are we verifying that all the parameters are loaded successfully?

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -184,47 +185,44 @@ BOOST_AUTO_TEST_CASE(test_load_version) {
...
-  auto load_save_compare = [&remove_if_exist](drawing_classifier_mock& dc,
-                              drawing_classifier_mock& dc_other) {
+  auto load_save_compare = [&remove_if_exist](
+                               drawing_classifier_mock& dc,
+                               drawing_classifier_mock& dc_other) {
     // random init; avoid segfault
-    dc.my_init_model_spec();
-    dc_other.my_init_model_spec();
+    dc.my_model_init_and_copy();
+    dc_other.my_model_init_and_copy();
```
nickjong: What is this doing? It's not clear to me

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: but you can still do a const_cast to that original pointer.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: init_model is protected.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: well, I shouldn't be lazy. Let me add some verbose explanations.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: init model and get a copy of nn_spec.

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: right. But I have to provide other parameters.

Besides that, `init_model` is a virtual function, call virtual function during construction is not a good practice.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: okay, I should say goodbye to `my`, which I have a lot of fun with.

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: FYI, I move` get_model_sepc_copy` to the public section.

Jarvi-Izana:
@nickjong let me clarify.

`init_training` is under development. It needs a special state set up to use that. Troublesome.

The protected constructor contains too many other parts (compute_context, model_backend) that potentially will make the test fail if we call `init_train` later. Isolation is the key part of the unittest and this approach contradicts the unittest idea. 
```
  drawing_classifier(
      const std::map<std::string, variant_type>& initial_state,
      std::unique_ptr<neural_net::model_spec> nn_spec,
      std::unique_ptr<neural_net::compute_context> training_compute_context,
      std::unique_ptr<data_iterator> training_data_iterator,
      std::unique_ptr<neural_net::model_backend> training_model)
      : nn_spec_(std::move(nn_spec)),
        training_data_iterator_(std::move(training_data_iterator)),
        training_compute_context_(std::move(training_compute_context)),
        training_model_(std::move(training_model)) {
    add_or_update_state(initial_state);
  ```
I **ONLY** need one simple line to test everything, which is `nn_spec_ = init_model()`. :-)

File: test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx (hunk 3 above)
Comment (nickjong) [reply test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx]: I'm suggesting that `drawing_classifier` expose a read-only view to subclasses, then the testing subclass can expose it publicly to tests

File: src/toolkits/drawing_classifier/drawing_classifier.hpp (hunk 8 above)
Comment (nickjong) [reply src/toolkits/drawing_classifier/drawing_classifier.hpp]: We shouldn't call `init_model` in the constructor, but the constructor can simply accept the intended values for the data members. Please see the OD tests for an example.

**Comment by Jarvi-Izana** [comment]
> > @nickjong let me clarify.
> > `init_training` is under development. It needs a special state set up to use that. Troublesome.
> > The protected constructor contains too many other parts (compute_context, model_backend) that potentially will make the test fail if we call `init_train` later. Isolation is the key part of the unittest and this approach contradicts the unittest idea.
> > ```
> >   drawing_classifier(
> >       const std::map<std::string, variant_type>& initial_state,
> >       std::unique_ptr<neural_net::model_spec> nn_spec,
> >       std::unique_ptr<neural_net::compute_context> training_compute_context,
> >       std::unique_ptr<data_iterator> training_data_iterator,
> >       std::unique_ptr<neural_net::model_backend> training_model)
> >       : nn_spec_(std::move(nn_spec)),
> >         training_data_iterator_(std::move(training_data_iterator)),
> >         training_compute_context_(std::move(training_compute_context)),
> >         training_model_(std::move(training_model)) {
> >     add_or_update_state(initial_state);
> > ```
> > 
> > 
> > I **ONLY** need one simple line to test everything, which is `nn_spec_ = init_model()`. :-)
> 
> You can simply pass `nullptr` for any state you don't need to initialize.

I understand, but I don't want to call a virtual function (`init_model`) in my constructor. 

I think `clone` is good design. I can use the constructor. `init_model_spec` helper function is neat than the constructor. Just a small function, why I need to call `init_model` from one instance and use the constructor to do it?
```
```

File: src/toolkits/drawing_classifier/drawing_classifier.hpp [hunk 13]
```diff
@@ -23,17 +23,28 @@ namespace drawing_classifier {
...
+   */
 
   void init_options(const std::map<std::string, flexible_type>& opts) override;
   /* Commented out for the purpose of a skeleton. */
-  // size_t get_version() const override;
-  // void save_impl(oarchive& oarc) const override;
-  // void load_version(iarchive& iarc, size_t version) override;
 
-  /* Interface exposed via Unity server */
+  // void init_options(const std::map<std::string, flexible_type>& opts) override;
```
nickjong: Do we still need these comments?

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -21,9 +21,18 @@ namespace {
...
  public:
-  std::unique_ptr<neural_net::model_spec> my_init_model() {
+  std::unique_ptr<neural_net::model_spec> get_model_spec() const {
     return init_model();
   }
+
+  drawing_classifier_mock() = default;
+
+  drawing_classifier_mock(std::unique_ptr<neural_net::model_spec> ms)
+      : drawing_classifier({}, std::move(ms), nullptr, nullptr, nullptr) {}
```
nickjong: Let's have constructors precede other methods in the class definition

https://google.github.io/styleguide/cppguide.html#Declaration_Order

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -176,8 +185,211 @@ BOOST_AUTO_TEST_CASE(test_export_coreml) {
                    target + "Probability");
 }
 
+BOOST_AUTO_TEST_CASE(test_load_version) {
```
nickjong: In what way does this test `load_version`?

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -176,8 +185,211 @@ BOOST_AUTO_TEST_CASE(test_export_coreml) {
...
+   * The test calls `save_impl` first and write serialized initial
+   * `nn_spec_` to disk, let's say `file_1`.
+   *
+   * Then we load the previously stored spec file and update
+   * all params to the model spec. Then we store what's loaded
+   * to `file_2`.
+   *
+   * By comparing `file_1` and `file_2`, we can verify that the
+   * both model spec instances have the same content and same content
+   * is written to the disk during the serialization.
```
nickjong: This only tests that the two saved files are the same, but it doesn't test that `save` works properly. For example, implementing `save` in a way that writes nothing at all would yield equivalent (incorrect) archives.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -176,8 +185,211 @@ BOOST_AUTO_TEST_CASE(test_export_coreml) {
...
+      std::perror(ss.str().c_str());
+    }
+    if (std::remove(my_file_loaded)) {
+      std::stringstream ss;
+      ss << "fail to remove file:" << my_file_loaded << std::endl;
+      std::perror(ss.str().c_str());
+    }
+
+    // compare weights in memory
+    auto original_spec = dc.get_model_spec_copy();
```
nickjong: Can we remove the file comparison above and just compare the in-memory representations as here?

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -176,8 +185,211 @@ BOOST_AUTO_TEST_CASE(test_export_coreml) {
...
+  std::vector<int> is_same;
+
+  for (const auto& entry : view1) {
+    auto& weights2 = view2.at(entry.first);
+    auto& weights1 = entry.second;
+    TS_ASSERT(weights1.size() > 0);
+    TS_ASSERT_EQUALS(weights1.size(), weights2.size());
+    is_same.push_back(
+        std::memcmp(weights1.data(), weights2.data(), weights1.size()));
+  }
```
nickjong: Why is there a layer-by-layer comparison here and in the implementation of `load_save_compare`? All of this feels unnecessarily complex

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: I should remove this.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: I think it's also good for us to compare the file written to disk.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: Yeah, that's why we also check the nn_spec in memory as well.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
nickjong: Why?

Jarvi-Izana: okay, that's right. we cannot verify whether save works or not by comparing the file.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
Jarvi-Izana: because I want to make sure the initial`nn_spec` of 2 mock instances are different.

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: no, we don't

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -34,7 +34,7 @@ class EXPORT drawing_classifier : public ml_model_base {
   void init_options(const std::map<std::string, flexible_type>& opts) override;
   /* Commented out for the purpose of a skeleton. */
```
nickjong: Let's delete this comment

src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -34,7 +34,7 @@ class EXPORT drawing_classifier : public ml_model_base {
...
 
-  // void init_options(const std::map<std::string, flexible_type>& opts) override;
+  void init_options(const std::map<std::string, flexible_type>& opts) override;
```
nickjong: And let's delete this, which is redundant with line 34

src/toolkits/drawing_classifier/drawing_classifier.hpp
Jarvi-Izana: just about to push! Done.

PR #2487, opened 2019-10-23
Additional arguments for OD's predict and evaluate function

jakesabathia2:
Change several things in this PR:
1.Add confidence_threshold, iou_threshold arguments for predict() and evaluate().
2.Enable different input data types (sframe, array, image) for predict().
3.Enable different output data types (dict, sframe) for evaluate().
4.Change the output attribute "bounding box" to "coordinate" for predict() in order to pass the python unit test.
5.Add output attribute "type" for predict() in order to pass the python unit test. 
6.Change several lines in the unit test for OD regarding the additional arguments for perform_evaluation()

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/python/turicreate/toolkits/object_detector/object_detector.py
@@ -1644,7 +1644,7 @@ class ObjectDetector_beta(_Model):
         return self.__proxy__.export_to_coreml(filename, options)
 
 
-    def predict(self, dataset, confidence_threshold=0.25, iou_threshold=None):
+    def predict(self, dataset, confidence_threshold=0.25, iou_threshold=0.45):
         """
         Predict object instances in an SFrame of images.
 
@@ -1693,11 +1693,10 @@ class ObjectDetector_beta(_Model):
         """
         options = {}
         options["confidence_threshold"] = confidence_threshold
-        if iou_threshold != None:
-            options["iou_threshold"] = iou_threshold
+        options["iou_threshold"] = iou_threshold
         return  self.__proxy__.predict(dataset, options)
 
-    def evaluate(self, dataset, metric='auto', output_type='dict', confidence_threshold = None, iou_threshold = None):
+    def evaluate(self, dataset, metric='auto', output_type='dict', confidence_threshold = 0.001, iou_threshold = 0.45):
         """
         Evaluate the model by making predictions and comparing these to ground
         truth bounding box annotations.
@@ -1743,8 +1742,6 @@ class ObjectDetector_beta(_Model):
         mAP: 43.2%
         """
         options = {}
-        if confidence_threshold != None:
-            options["confidence_threshold"] = confidence_threshold
-        if iou_threshold != None:
-            options["iou_threshold"] = iou_threshold
+        options["confidence_threshold"] = confidence_threshold
+        options["iou_threshold"] = iou_threshold
         return self.__proxy__.evaluate(dataset, metric, output_type, options)
--- a/src/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/python/turicreate/toolkits/object_detector/object_detector.py
@@ -1644,7 +1644,7 @@ class ObjectDetector_beta(_Model):
         return self.__proxy__.export_to_coreml(filename, options)
 
 
-    def predict(self, dataset):
+    def predict(self, dataset, confidence_threshold=0.45, iou_threshold=0.45):
         """
         Predict object instances in an SFrame of images.
 
@@ -1691,8 +1691,23 @@ class ObjectDetector_beta(_Model):
             # Visualize predictions by generating a new column of marked up images
             >>> data['image_pred'] = turicreate.object_detector.util.draw_bounding_boxes(data['image'], data['predictions'])
         """
-        return self.__proxy__.predict(dataset)
-
+        dataset, unpack = self._canonize_input(dataset)
+        stacked_pred = self.__proxy__.predict(dataset, confidence_threshold, iou_threshold)
+        from . import util
+        return unpack(util.unstack_annotations(stacked_pred, num_rows=len(dataset)))
+    def _canonize_input(self, dataset):
+        """
+        Takes input and returns tuple of the input in canonical form (SFrame)
+        along with an unpack callback function that can be applied to
+        prediction results to "undo" the canonization.
+        """
+        unpack = lambda x: x
+        if isinstance(dataset, _tc.SArray):
+            dataset = _tc.SFrame({self.feature: dataset})
+        elif isinstance(dataset, _tc.Image):
+            dataset = _tc.SFrame({self.feature: [dataset]})
+            unpack = lambda x: x[0]
+        return dataset, unpack
     def evaluate(self, dataset, metric='auto'):
         """
         Evaluate the model by making predictions and comparing these to ground
--- a/src/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/python/turicreate/toolkits/object_detector/object_detector.py
@@ -445,7 +445,6 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
         }
         model = _tc.extensions.object_detector()
         model.train(data=dataset, annotations_column_name=annotations, image_column_name=feature, options=tf_config)
-        print(annotations)
         return ObjectDetector_beta(model_proxy=model, name="object_detector")
 
     else:  # Use MxNet
@@ -1645,7 +1644,7 @@ class ObjectDetector_beta(_Model):
         return self.__proxy__.export_to_coreml(filename, options)
 
 
-    def predict(self, dataset, confidence_threshold=0.45, iou_threshold=0.45):
+    def predict(self, dataset, confidence_threshold=0.25, iou_threshold=0.45):
         """
         Predict object instances in an SFrame of images.
 
@@ -1692,10 +1691,9 @@ class ObjectDetector_beta(_Model):
             # Visualize predictions by generating a new column of marked up images
             >>> data['image_pred'] = turicreate.object_detector.util.draw_bounding_boxes(data['image'], data['predictions'])
         """
-        print("shit")
         return  self.__proxy__.predict(dataset, confidence_threshold, iou_threshold)
 
-    def evaluate(self, dataset, metric='auto'):
+    def evaluate(self, dataset, metric='auto', confidence_threshold = 0.001, iou_threshold = 0.45):
         """
         Evaluate the model by making predictions and comparing these to ground
         truth bounding box annotations.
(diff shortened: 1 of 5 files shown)


src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1644,7 +1644,7 @@ def export_coreml(self, filename,
         return self.__proxy__.export_to_coreml(filename, options)
 
 
-    def predict(self, dataset):
+    def predict(self, dataset, confidence_threshold=0.25, iou_threshold=0.45):
```
srikris: Where did we get these defaults from?

jakesabathia2: I follow the default values from Python's implementation.

jakesabathia2: But if you are asking where did the values come from in the beginning,
I think they are just empirical results.

src/toolkits/object_detection/object_detector.hpp
```diff
@@ -40,8 +40,12 @@ class EXPORT object_detector: public ml_model_base {
...
              std::string image_column_name, variant_type validation_data,
              std::map<std::string, flexible_type> opts);
-  variant_map_type evaluate(gl_sframe data, std::string metric);
-  gl_sarray predict(gl_sframe data);
+  variant_type evaluate(gl_sframe data, std::string metric,
+                        std::string output_type = "dict",
+                        float confidence_threshold = 0.001,
+                        float iou_threshold = 0.45);
+  gl_sarray predict(variant_type data, float confidence_threshold,
+   float iou_threshold);
```
nickjong: I think the new arguments should probably be in an `options` dictionary, like what we do for `train`. That also makes it easier to define defaults that can change over time, if needed.

src/python/turicreate/toolkits/object_detector/object_detector.py
jakesabathia2: It seems like for the `evaluate()`, the default value for `confidence_threshold` and `iou_threshold` are both `None`, and for the `predict()`, the default values are 0.25 and `None`, respectively. hmmmm

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -522,12 +521,44 @@ variant_map_type object_detector::evaluate(
...
+    final_result = to_variant(result_map);
+  } else if (output_type == "sframe") {
+    gl_sframe sframe_result = gl_sframe({{"label", class_labels}});
+    auto add_score_list = [=](gl_sframe& sframe_result,
+                              std::string metric_name) {
+      flex_list score_list;
+      auto it = result_map.find(metric_name);
+      if (it != result_map.end()) {
+        flex_dict dict = variant_get_value<flex_dict>(it->second);
+        for (auto& label_score_pair : dict) {
```
nickjong: Please use `const auto &`

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1691,9 +1691,9 @@ def predict(self, dataset):
             # Visualize predictions by generating a new column of marked up images
             >>> data['image_pred'] = turicreate.object_detector.util.draw_bounding_boxes(data['image'], data['predictions'])
         """
-        return self.__proxy__.predict(dataset)
+        return  self.__proxy__.predict(dataset, confidence_threshold, iou_threshold)
 
-    def evaluate(self, dataset, metric='auto'):
+    def evaluate(self, dataset, metric='auto', output_type='dict', confidence_threshold = 0.001, iou_threshold = 0.45):
```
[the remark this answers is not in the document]
jakesabathia2: hmmm good catch!

src/python/turicreate/toolkits/object_detector/object_detector.py
TobyRoseman: I think `0.45` is the way to go. Explicit is definitely the pythonic way.

nickjong: Okay, let's go with that for now, since it's a better way to express the actual functionality supported by 5.8. But worth considering: the current Python implementation of the toolkit unofficially supports setting the `iou_threshold` once during model creation, and then defaulting to that threshold for all future calls to `predict` and `evaluate`. With this approach, a user who wants a different threshold will never be able to encode that into the model itself.

nickjong: A relevant question: is the IOU threshold for non-maximum suppression, which is _only_ relevant for inference, something that belongs in the trained model itself?

jakesabathia2: @nickjong This is a good point,
somehow in the `init_training()`,
we set the default value for noms to 0.45 :P .
```
  add_or_update_state({
      { "annotations", annotations_column_name },
      { "feature", image_column_name },
      { "model", "darknet-yolo" },
      { "non_maximum_suppression_threshold",
        DEFAULT_NON_MAXIMUM_SUPPRESSION_THRESHOLD },
  })
```
And we never use it if we change the `predict` and `evaluate` function to have input for the you_threshold :P .
Do you think we should delete this state?

jakesabathia2: Note that in `export_to_coreml()`,
it also enable use to set both thresholds :P hmmm.
So I feel like they don't belong to model itself :P .

src/toolkits/object_detection/object_detector.hpp
jakesabathia2: :)

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -98,7 +98,6 @@ constexpr float DEFAULT_NON_MAXIMUM_SUPPRESSION_THRESHOLD = 0.45f;
 
 // Predictions with confidence scores below this threshold will be discarded
```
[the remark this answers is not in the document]
jakesabathia2: :)

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -522,12 +521,44 @@ variant_map_type object_detector::evaluate(
     result_map.erase(MAP);
   }
 
-  return result_map;
+  // Handle different output types here
+  // If output_type = "dict", just return the result_map.
+  // If output_type = "sframe", construct a sframe,
+  // whose rows indicate class labels, and columns denote different metric
+  // scores. Note that the "sframe" output only shows AP or AP50.
```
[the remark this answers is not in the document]
jakesabathia2: :)

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -522,12 +521,44 @@ variant_map_type object_detector::evaluate(
...
+  variant_type final_result;
+
+  if (output_type == "dict") {
+    final_result = to_variant(result_map);
+  } else if (output_type == "sframe") {
+    gl_sframe sframe_result = gl_sframe({{"label", class_labels}});
+    auto add_score_list = [=](gl_sframe& sframe_result,
+                              std::string metric_name) {
```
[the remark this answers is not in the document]
jakesabathia2: :)

src/python/turicreate/toolkits/object_detector/object_detector.py
nickjong: Let's file a followup issue to delete this state as superfluous. Thinking more about it, I think we want to encourage a workflow where somebody trains a model and then uses some kind of model-understanding process to choose the thresholds. So they shouldn't be established by training itself.

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1644,7 +1644,7 @@ def export_coreml(self, filename,
         return self.__proxy__.export_to_coreml(filename, options)
 
 
-    def predict(self, dataset):
+    def predict(self, dataset, confidence_threshold=0.25, iou_threshold=None):
```
nickjong: Didn't we decide (offline) to use explicit defaults on the Python side?

src/toolkits/object_detection/object_detector.hpp
```diff
@@ -207,6 +214,8 @@ class EXPORT object_detector: public ml_model_base {
   flex_int get_training_iterations() const;
   flex_int get_num_classes() const;
 
+  variant_type convert_map_to_types(variant_map_type& result_map, std::string output_type);
```
nickjong: Could this be a static method of the class? (Or a file-local global helper function in the .cpp file?) I don't think this should depend on any instance data members

src/toolkits/object_detection/object_detector.hpp
```diff
@@ -180,14 +185,16 @@ class EXPORT object_detector: public ml_model_base {
...
+                                          std::string output_type,
+                                          float confidence_threshold,
+                                          float iou_threshold);
 
   void perform_predict(
       gl_sframe data,
       std::function<void(const std::vector<neural_net::image_annotation>&,
                          const std::vector<neural_net::image_annotation>&)>
-          consumer);
+          consumer, float confidence_threshold = 0.001, float iou_threshold = 0.45);
```
nickjong: Let's not use defaults here. Calls to `perform_predict` can be explicit. This way these magic numbers appear in fewer redundant places

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -522,45 +538,123 @@ variant_map_type object_detector::evaluate(
...
+    std::vector<flexible_type> image_vector {image_data};
+    std::map<std::string, std::vector<flexible_type>> image_map ={{image_column_name, image_vector}};
+    sframe_data = gl_sframe(image_map);
+  } else if (variant_is<gl_sarray>(data)){
+    gl_sarray sarray_data = variant_get_value<gl_sarray>(data);
+    std::map<std::string, gl_sarray> sarray_map = {{image_column_name, sarray_data}};
+    sframe_data = gl_sframe(sarray_map);
+  } else {
+    log_and_throw("Invalid data type for predict()! Expect Sframe, Sarray, or flexible_type!");
+  }
```
nickjong: Could you pull this out into a helper function?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -522,45 +538,123 @@ variant_map_type object_detector::evaluate(
     result_map.erase(MAP);
   }
 
-  return result_map;
+  return convert_map_to_types(result_map, output_type);
+}
+
+variant_type object_detector::convert_map_to_types(variant_map_type& result_map, std::string output_type){
```
nickjong: Why do we need a mutable reference for `result_map`?

Could we pass `output_type` by (`const`) reference?

src/python/turicreate/toolkits/object_detector/object_detector.py
jakesabathia2: https://github.com/apple/turicreate/issues/2506

jakesabathia2:
Thanks for everyone's review!

PR #2439, opened 2019-10-11
init/export model spec for DC

Jarvi-Izana:
model spec follows CoreML neural network spec format which describes ANN architecture to be used.

__MySuggestion__
You can skip reviewing `drawing_classifier.hpp` since it only has minor code style change, which is irrelevant to this topic.

__MyPlan__
- [X] `init_model`
- [x] `export_coreml`
- [x] basic unittest for `init_model`
- [x] basic unittest for `export_coreml`
 
__MySummary__
* [Y] adding unittest[Y/N]?
* [N] breaking current unittest[Y/N]?

PR #2491, opened 2019-10-24
Fixed up state in DC C++, and fixed layer names for PyBind consumption

shantanuchhabra:
Some miscellaneous bugfixes from running train end-to-end, like fixing missing/inconsistent state variables, and adding functionality to strip _fwd in the layer names.


src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -217,26 +226,64 @@ std::unique_ptr<data_iterator> drawing_classifier::create_iterator(
...
+   */
   return create_iterator(data_params);
 }
 
+float_array_map drawing_classifier::get_model_params() const {
+
+  float_array_map raw_model_params = nn_spec_->export_params_view();
+
+  // Strip the substring "_fwd" from any parameter names, for compatibility with
+  // the compute backend. (We preserve the substring in nn_spec_ for inclusion
```
Jarvi-Izana: why `_fwd` maters?

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -217,26 +226,64 @@ std::unique_ptr<data_iterator> drawing_classifier::create_iterator(
 
   data_params.is_train = is_train;
   data_params.target_column_name = read_state<flex_string>("target");
-  data_params.feature_column_name = read_state<flex_string>("feature");
+  const flex_list &features_values = read_state<flex_list>("features");
+  DASSERT_GE(features_values.size(), 1);
+  data_params.feature_column_name = features_values[0].get<flex_string>();
```
Jarvi-Izana: since the feature_list only contains one element, why bother using a list?

shantanuchhabra: I think there's an argument to be made both for and against using a feature list but from https://github.com/apple/turicreate/pull/2439/files#r334137025, I thought we agreed that we'll stick with a single-element feature list. This would also give us the flexibility to add more features down the road if we want to. I do agree that it is a little awkward since the data_params only has one feature_column_name as a string and not a vector.

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Similar to OD/AC, the Python TF code is written with the assumption that the layer names do not have the `_fwd` string.

nickjong: Could we just remove the `_fwd` from the model spec then?

Jarvi-Izana: we can remove it, easily.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: There's one philosophy called [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

Jarvi-Izana:
remove _fwd needs more work. I can put up a pr to remove it first.

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Haha okay I was always convinced to have just one feature like I mentioned in the comment thread on #2439. I'll make the software change to take care of that in this PR.

shantanuchhabra: Sorry, I take that back. Changing to a single feature name would require a reasonable code change in init_model and export_coreml functionality and tests. I think we should table that change for convergence, and not as part of this PR.

PR #2495, opened 2019-10-24
Move MPS loss multiplier out of OD C++ implementation

shreyajain17:


src/ml/neural_net/mps_graph_cnnmodule.mm
```diff
@@ -115,6 +115,7 @@ - (instancetype)initWithCommandBuffer:(id<MTLCommandBuffer>)commandBuffer {
 
   @autoreleasepool {
 
+  constexpr float MPS_LOSS_MULTIPLIER = 8;
```
nickjong: Nit: `8.f`

src/ml/neural_net/mps_graph_cnnmodule.mm
```diff
@@ -163,7 +164,7 @@ - (instancetype)initWithCommandBuffer:(id<MTLCommandBuffer>)commandBuffer {
       std::vector<float> loss(loss_size);
       auto loss_it = loss.begin();
       for (MPSCNNLossLabels *lossState in batch.lossState) {
-        *loss_it = sumImage([lossState lossImage]);
+        *loss_it = sumImage([lossState lossImage]) / MPS_LOSS_MULTIPLIER;
```
nickjong: Could you add an explanatory comment either here or where the constant is defined?

src/ml/neural_net/mps_compute_context.mm
```diff
@@ -88,12 +88,45 @@
 std::unique_ptr<model_backend> mps_compute_context::create_object_detector(
     int n, int c_in, int h_in, int w_in, int c_out, int h_out, int w_out,
     const float_array_map& config, const float_array_map& weights) {
-
+  float_array_map updated_config;
+  constexpr float MPS_LOSS_MULTIPLIER = 8;
```
nickjong: Let's add a comment here. Also, we should be able to define the constant in only one place, probably in `mps_graph_cnn_module.h`

Can probably just define `static constexpr float MPS_OD_LOSS_MULTIPLIER = 8.f;` somewhere

src/ml/neural_net/mps_compute_context.mm
```diff
@@ -88,12 +88,45 @@
...
+  if (config.count("learning_rate") > 0) {
+    updated_config["learning_rate"] =
+      shared_float_array::wrap(*config.at("learning_rate").data() / MPS_LOSS_MULTIPLIER);
+  }
```
nickjong: Could we define a quicker helper function or lambda here?

Also, prefer this pattern:
```
auto config_iter = config.find("learning_rate");
if (config_iter != config.end()) {
  float old_value = config_iter->second.data()[0];
  float new_value = old_value / MPS_LOSS_MULTIPLIER;
  config_iter->second = shared_float_array::wrap(new_value);
}

File: src/ml/neural_net/mps_compute_context.mm [hunk 5]
```diff
@@ -20,6 +20,25 @@
 
 namespace {
 
+float_array_map multiply_mps_od_loss_multiplier(float_array_map config,
+                                                std::vector<std::string> update_keys) {
```
nickjong: Nit: pass `update_keys` by `const` reference

src/ml/neural_net/mps_compute_context.mm
```diff
@@ -20,6 +20,25 @@
...
+  for (std::vector<std::string>::const_iterator i = update_keys.begin(); i != update_keys.end();
```
nickjong: This would be more readable as:
```
for (const std::string& key : update_keys) {
```

PR #2496, opened 2019-10-24
StopIteration changes

### LINKED ISSUE #647: use try and except block around StopIteration or avoid StopIteration completely
`StopIteration` is deprecated from python3.6+, https://docs.python.org/3.6/whatsnew/3.6.html#deprecated-python-behavior 

As per https://www.python.org/dev/peps/pep-0479/ we should not use `raise StopIteration` on generators.
See #636

dhivyaaxim:
This PR resolves #647 
use try and except block around StopIteration or avoid StopIteration completely

### DIFF
--- a/src/python/turicreate/toolkits/style_transfer/_sframe_loader.py
+++ b/src/python/turicreate/toolkits/style_transfer/_sframe_loader.py
@@ -125,9 +125,9 @@ class SFrameSTIter(_mx.io.DataIter):
         return self._provide_data
 
     def _next(self):
-        
         if self.cur_epoch == self.num_epochs or len(self.sframe) == 0:
             import sys
+            #In python3.7, no support for StopIteration to raise as an exception. 
             if(sys.version_info.major == 3 and sys.version_info.minor == 7):
                 raise RuntimeError("StopIteration")
             else:
--- a/src/python/turicreate/toolkits/style_transfer/_sframe_loader.py
+++ b/src/python/turicreate/toolkits/style_transfer/_sframe_loader.py
@@ -125,9 +125,13 @@ class SFrameSTIter(_mx.io.DataIter):
         return self._provide_data
 
     def _next(self):
-
+        
         if self.cur_epoch == self.num_epochs or len(self.sframe) == 0:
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/style_transfer/_sframe_loader.py
```diff
@@ -125,9 +125,13 @@ def provide_data(self):
         return self._provide_data
 
     def _next(self):
-
+
```
shantanuchhabra: Let's not add leading whitespace.

src/python/turicreate/toolkits/style_transfer/_sframe_loader.py
```diff
@@ -125,9 +125,13 @@ def provide_data(self):
         return self._provide_data
 
     def _next(self):
-
+        
         if self.cur_epoch == self.num_epochs or len(self.sframe) == 0:
-            raise StopIteration
+            import sys
+            if(sys.version_info.major == 3 and sys.version_info.minor == 7):
```
shantanuchhabra: What is the significance of 3 and 7? Could we add a comment explaining the magic numbers or make those into variables?

TobyRoseman: It seems `StopIteration` is still included in Python 3.7. It just get automatically converted to a RuntimeError.
From https://docs.python.org/3.7/library/exceptions.html -
```
Changed in version 3.7: Enable PEP 479 for all code by default: a StopIteration error raised in a generator is transformed into a RuntimeError.
```
So I don't think this change is necessary. @dhivyaaxim - are we missing something?

dhivyaaxim: @TobyRoseman  Thanks for your suggestion!

TobyRoseman: @dhivyaaxim - The new comment is not accurate. Python 3.7 does support raising a `StopIteration`. It just converts it to a `RuntimeError`.

PR #2500, opened 2019-10-24
Fixes Load model for AC and OD

fareeha:
`_native_name()` implementation in both Python and C++ classes set such that when using the C++ codepath the Python model will not be registered and vice versa. This will enable the model to be wrapped in the correct class.

### DIFF
--- a/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -356,6 +356,8 @@ class ActivityClassifier_beta(_Model):
 
     @classmethod
     def _native_name(cls):
+        if USE_CPP:
+            return "activity_classifier"
         return None
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -356,6 +356,8 @@ def __init__(self, model_proxy=None, name=None):
 
     @classmethod
     def _native_name(cls):
+        if USE_CPP:
+            return "activity_classifier"
         return None
```
jakesabathia2: Just curious what happens when returning None?

fareeha: Good question! So the class doesn't get registered (via `RegistrationMetaClass`) and placed in `MODEL_NAME_MAP`. We can use this to our advantage by returning `None` from `_native_name` when we don't want that class to be registered. This way we control if the Python or C++ class is registered upon importing turicreate. This fixes the bug of wrapping a C++ model with the Python class wrapper.

jakesabathia2: Great!!Thanks!

PR #2501, opened 2019-10-24
[WIP] Fixes AC `list_fields()` and adapted python test

fareeha:
`list_fields()` now has attributes `num_sessions`, `verbose`, `training_time` and `num_examples`. AC python test is adapted so that `test__list_fields()` queries and validates model attributes specific to C++ and Python given the codepath chosen.

[WIP] because I need to remove the load model fix changes.
TODO: Replicate changes for `list_fields()` in OD

### DIFF
--- a/src/python/turicreate/test/test_activity_classifier.py
+++ b/src/python/turicreate/test/test_activity_classifier.py
@@ -15,10 +15,11 @@ import numpy as np
 from numbers import Number
 from . import util as test_util
 import pytest
-from turicreate.toolkits._internal_utils import _mac_ver
+from turicreate.toolkits._internal_utils import _mac_ver, _read_env_var_cpp
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
 import uuid
 
+USE_CPP = _read_env_var_cpp('TURI_AC_USE_CPP_PATH')
 
 def _load_data(self, num_examples = 1000, num_features = 3, max_num_sessions = 4,
                randomize_num_sessions = True, num_labels = 9, prediction_window = 5,
@@ -257,9 +258,16 @@ class ActivityClassifierTest(unittest.TestCase):
             'classes': lambda x: sorted(x) == sorted(self.data[self.target].unique())
         }
         self.exposed_fields_ans = list(self.get_ans.keys())
-        self.fields_ans = self.exposed_fields_ans + ['_recalibrated_batch_size',
-                '_pred_model', '_id_target_map',
-                '_predictions_in_chunk', '_target_id_map']
+        if USE_CPP:
+            self.fields_ans = self.exposed_fields_ans + ['training_report_by_class',
+                'training_iterations', 'random_seed',
+                'training_precision', 'training_confusion_matrix',
+                'use_data_augmentation', 'training_f1_score',
+                'training_auc', 'training_roc_curve', 'training_recall']
+        else:
+            self.fields_ans = self.exposed_fields_ans + ['_recalibrated_batch_size',
+                    '_pred_model', '_id_target_map',
+                    '_predictions_in_chunk', '_target_id_map']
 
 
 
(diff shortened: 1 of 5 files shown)


src/python/turicreate/test/test_object_detector.py
```diff
@@ -114,30 +115,58 @@ def setUpClass(self):
...
-           'training_epochs': lambda x: x >= 0,
-           'num_bounding_boxes': lambda x: x > 0,
-           'num_examples': lambda x: x > 0,
-           'training_loss': lambda x: x > 0,
-           'annotations': lambda x: x == self.annotations,
-           'num_classes': lambda x: x == len(_CLASSES),
-        }
+        if USE_CPP:
+            self.get_ans = {
+               'annotation_position': lambda x: isinstance(x, str),
```
fareeha: @nickjong please let me know if lambda checks for `annotation_position`, `annotation_scale`, `annotation_origin` , `training_average_precision_50`, `training_mean_average_precision_50`, `grid_height`, `grid_width`, `training_mean_average_precision`, `random_seed` and `training_average_precision` seem sufficient.

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -190,6 +190,8 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
         options['prediction_window'] = prediction_window
         options['batch_size'] = batch_size
         options['max_iterations'] = max_iterations
+        options['verbose'] = verbose
+        options['num_examples'] = len(dataset)
```
nickjong: I don't think it makes sense the pass `num_examples` explicitly. Isn't this implicit in `dataset`? The C++ side can record this field itself from `dataset`.

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -187,6 +188,25 @@ void activity_classifier::init_options(
...
+      0, // should it be?
+      std::numeric_limits<int>::max(),
+      true);
+  options.create_integer_option(
+      "num_sessions",
+      "Number of sessions.",
+      FLEX_UNDEFINED,
+      0,
+      std::numeric_limits<int>::max(),
+      true);
```
nickjong: These should be inferred from the data

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -26,6 +26,7 @@
 #include <toolkits/object_detection/od_yolo.hpp>
 #include <toolkits/supervised_learning/automatic_model_creation.hpp>
 #include <toolkits/util/training_utils.hpp>
+#include <timer/timer.hpp>
```
nickjong: Let's keep these sorted

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -410,14 +411,22 @@ void object_detector::train(gl_sframe data,
...
+  
+  turi::timer time_object;
+  time_object.start();
 
   // Perform all the iterations at once.
   while (get_training_iterations() < get_max_iterations()) {
     iterate_training();
   }
-
+
```
nickjong: Please don't introduce trailing whitespace

src/python/turicreate/test/test_activity_classifier.py
```diff
@@ -305,7 +313,10 @@ def test_export_coreml(self):
 
         if _mac_ver() >= (10, 13):
             w = self.prediction_window
-            labels = list(map(str, sorted(self.model._target_id_map.keys())))
+            if USE_CPP:
+                labels = list(map(str, sorted(self.model.classes)))
```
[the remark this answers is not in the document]
fareeha: They are the same but strangely I get error `AssertionError: After model save and load, method test_export_coreml has failed with error: '0.0'` if I use `labels = list(map(str, sorted(self.model.classes)))` on the Python codepath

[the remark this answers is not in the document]
fareeha: It is the only test breaking on the Python codepath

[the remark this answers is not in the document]
nickjong: Hmm, strange. But let's not worry about it now, I'm fine with keeping these distinct if that gets the test passing

PR #2507, opened 2019-10-25
Refactor Code to use Boost instead of CXX

abhishekpratapa:
- Refactored code to use Boost Test instead to fix compilation issues for timer test.

### DIFF
--- a/src/toolkits/style_transfer/style_transfer.hpp
+++ b/src/toolkits/style_transfer/style_transfer.hpp
@@ -36,6 +36,8 @@ class EXPORT style_transfer : public ml_model_base {
   void train(gl_sarray style, gl_sarray content,
              std::map<std::string, flexible_type> opts);
 
+  gl_sarray predict(variant_type data, std::map<std::string, flexible_type> opts);
+
   virtual void init_train(gl_sarray style, gl_sarray content,
                           std::map<std::string, flexible_type> opts);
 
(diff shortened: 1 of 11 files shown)


test/timer/timer_test.cxx
```diff
@@ -37,4 +37,4 @@ BOOST_AUTO_TEST_CASE(test_timer) {
 BOOST_AUTO_TEST_CASE(test_lowres_timer) {
   timer_test::test_lowres_timer();
 }
-BOOST_AUTO_TEST_SUITE_END()
+BOOST_AUTO_TEST_SUITE_END()
```
shantanuchhabra: Let's add the newline back.

nickjong:
What's the relationship between these changes and #2478? @hoytak 

Any reason not just to change the definition of the macro, instead of the call sites?

src/toolkits/style_transfer/style_transfer.hpp
```diff
@@ -36,6 +36,8 @@ class EXPORT style_transfer : public ml_model_base {
   void train(gl_sarray style, gl_sarray content,
              std::map<std::string, flexible_type> opts);
 
+  gl_sarray predict(variant_type data, std::map<std::string, flexible_type> opts);
```
hoytak: I think this is the only actual change here...

abhishekpratapa:
> What's the relationship between these changes and #2478? @hoytak
> 
> Any reason not just to change the definition of the macro, instead of the call sites?

Genius. Yeah.

PR #2508, opened 2019-10-26
Python frontend for style transfer

shreyajain17:
This PR does the following -
- Prints device name its training on (added in `style_transfer.cpp`)
- Pretrained models hosted on developer.apple.com are being pulled in the frontend and being passed to C++
- Added calculation `num_styles` also to python side
- Introduced a environment variable like for AC/OD for style_transfer
- Added the style_transfer_beta class

### DIFF
--- a/src/python/turicreate/toolkits/_pre_trained_models.py
+++ b/src/python/turicreate/toolkits/_pre_trained_models.py
@@ -224,10 +224,41 @@ class Vgg16():
     def get_model_path(self):
         return self.model_path
 
+class StyleTransferTransformerCoreML():
+
+    def __init__(self):
+        self.name = 'transformer'
+        self.source_url = _urlparse.urljoin(MODELS_URL_ROOT, 'transformer.mlmodel')
+        self.source_md5 = ''
+        self.model_path = _download_and_checksum_files([
+            (self.source_url, self.source_md5)
+        ], _get_cache_dir())[0]
+
+
+    def get_model_path(self):
+        return self.model_path
+
+
+class Vgg16CoreML():
+
+    def __init__(self):
+        self.name = 'vgg16'
+        self.source_url = _urlparse.urljoin(MODELS_URL_ROOT, 'vgg16.mlmodel')
+        self.source_md5 = ''
+        self.model_path = _download_and_checksum_files([
+            (self.source_url, self.source_md5)
+        ], _get_cache_dir())[0]
+
+
+    def get_model_path(self):
+        return self.model_path
+
 
 STYLE_TRANSFER_BASE_MODELS = {
     'resnet-16': StyleTransferTransformer,
-    'Vgg16': Vgg16
+    'Vgg16': Vgg16,
+    'resnet_mlmodel': StyleTransferTransformerCoreML,
+    'vgg_mlmodel': Vgg16CoreML
 }
 
 
--- a/src/python/turicreate/toolkits/_pre_trained_models.py
+++ b/src/python/turicreate/toolkits/_pre_trained_models.py
@@ -205,10 +205,11 @@ class StyleTransferTransformer():
 
     def get_model_path(self, format):
         assert(format in ('coreml', 'mxnet'))
-        if (format in 'coreml') :
-            url =  _urlparse.urljoin(MODELS_URL_ROOT, 'transformer.mlmodel')
+        if(format == 'coreml'):
+            filename = self.name + '.mlmodel'
         else:
-            url = _urlparse.urljoin(MODELS_URL_ROOT, 'resnet-16.params')
+            filename = self.name + '.params'
+        url = _urlparse.urljoin(MODELS_URL_ROOT, filename)
         checksum = self.source_md5[format]
         model_path = _download_and_checksum_files([(url, checksum)], _get_cache_dir())[0]
         return model_path
@@ -227,9 +228,10 @@ class Vgg16():
     def get_model_path(self, format):
         assert(format in ('coreml', 'mxnet'))
         if (format in 'coreml') :
-            url = _urlparse.urljoin(MODELS_URL_ROOT, 'vgg16.mlmodel')
+            filename = 'vgg16-conv1_1-4_3.mlmodel'
         else:
-            url = _urlparse.urljoin(MODELS_URL_ROOT, 'vgg16-conv1_1-4_3.params')
+            filename = 'vgg16-conv1_1-4_3.params'
+        url = _urlparse.urljoin(MODELS_URL_ROOT, filename)
         checksum = self.source_md5[format]
         model_path = _download_and_checksum_files([(url, checksum)], _get_cache_dir())[0]
         return model_path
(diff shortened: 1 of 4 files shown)


src/python/turicreate/toolkits/_pre_trained_models.py
```diff
@@ -224,10 +224,41 @@ def __init__(self):
     def get_model_path(self):
         return self.model_path
 
+class StyleTransferTransformerCoreML():
+
+    def __init__(self):
+        self.name = 'transformer'
```
abhishekpratapa: Underscore `_coreml`? Perhaps could be a problem.

src/python/turicreate/toolkits/_pre_trained_models.py
```diff
@@ -224,10 +224,41 @@ def __init__(self):
...
+
+
+    def get_model_path(self):
+        return self.model_path
+
+
+class Vgg16CoreML():
+
+    def __init__(self):
+        self.name = 'vgg16'
```
abhishekpratapa: Same ^

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -634,6 +661,154 @@ def _raise_error_if_not_training_sframe(dataset, context_column):
...
         raise _ToolkitError("Context Image column must contain images")
 
+class StyleTransfer_beta(_Model):
+    """
+    A trained model using C++ implementation that is ready to use for classification or export to
+    CoreML.
+
+    This model should not be constructed directly.
+    """
+    _CPP_ACTIVITY_CLASSIFIER_VERSION = 1
```
abhishekpratapa: Change `_CPP_ACTIVITY_CLASSIFIER_VERSION` to `_CPP_STYLE_TRANSFER_VERSION `

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -634,6 +661,154 @@ def _raise_error_if_not_training_sframe(dataset, context_column):
...
+
+    def __init__(self, model_proxy=None, name=None):
+        self.__proxy__ = model_proxy
+        self.__name__ = name
+
+    @classmethod
+    def _native_name(cls):
```
abhishekpratapa: In custom model import we need a name here. Perhaps "style_transfer_beta" or the same name as above?

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -634,6 +661,154 @@ def _raise_error_if_not_training_sframe(dataset, context_column):
...
+
+        Returns
+        -------
+        out : string
+            A description of the model.
+        """
+        return self.__class__.__name__
+
+    def _get_version(self):
+        return self._CPP_ACTIVITY_CLASSIFIER_VERSION
```
abhishekpratapa: Replace with `_CPP_STYLE_TRANSFER_VERSION `

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -634,6 +661,154 @@ def _raise_error_if_not_training_sframe(dataset, context_column):
...
+
+        See Also
+        --------
+        save
+
+        Examples
+        --------
+        >>> model.export_coreml('StyleTransfer.mlmodel')
+        """
+        return self.__proxy__.export_to_coreml(filename, image_shape, include_flexible_shape)
```
abhishekpratapa: Make it as options as in c++.

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -634,6 +661,154 @@ def _raise_error_if_not_training_sframe(dataset, context_column):
...
+        |   0    |   3   | Height: 256 Width: 256 |
+        |   1    |   0   | Height: 640 Width: 648 |
+        |   1    |   1   | Height: 640 Width: 648 |
+        |   1    |   2   | Height: 640 Width: 648 |
+        |   1    |   3   | Height: 640 Width: 648 |
+        +--------+-------+------------------------+
+        [8 rows x 3 columns]
+        """
+
+        return self.__proxy__.predict(images, style=None, verbose=True, max_size=800, batch_size = 4)
```
abhishekpratapa: Follow the opts model defined in c++.

src/python/turicreate/toolkits/_pre_trained_models.py
shreyajain17: Yes.

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -160,6 +163,11 @@ def create(style_dataset, content_dataset, style_feature=None,
     _raise_error_if_not_training_sframe(content_dataset, content_feature)
     _tkutl._handle_missing_values(style_dataset, style_feature, 'style_dataset')
     _tkutl._handle_missing_values(content_dataset, content_feature, 'content_dataset')
+
+    pretrained_resnet_model = _pre_trained_models.STYLE_TRANSFER_BASE_MODELS['resnet_mlmodel']()
+    pretrained_resnet_model_path = pretrained_resnet_model.get_model_path()
+    pretrained_vgg16_model = _pre_trained_models.STYLE_TRANSFER_BASE_MODELS['vgg_mlmodel']()
+    pretrained_vgg16_model_path = pretrained_vgg16_model.get_model_path()
```
nickjong: Shouldn't we avoid downloading this if we're not using the CPP path?

src/python/turicreate/toolkits/style_transfer/style_transfer.py
nickjong: Please consult with what @fareeha did for OD and AC to make this work for both settings of the feature flag

src/python/turicreate/toolkits/_pre_trained_models.py
```diff
@@ -199,30 +199,40 @@ class StyleTransferTransformer():
...
+        self.source_md5 = { 'mxnet' : 'ac232afa6d0ead93a8c75b6c455f6dd3',
+                            'coreml': 'e0f3adaa9952ecc7d96f5e4eefb0d690' }
 
 
-    def get_model_path(self):
-        return self.model_path
+    def get_model_path(self, format):
+        assert(format in ('coreml', 'mxnet'))
+        if (format in 'coreml') :
+            url =  _urlparse.urljoin(MODELS_URL_ROOT, 'transformer.mlmodel')
```
nickjong: Should I worry that the base names are different between CoreML and MXNet? Is this the only transformer network we'll ever publish here?

src/python/turicreate/toolkits/_pre_trained_models.py
```diff
@@ -199,30 +199,40 @@ class StyleTransferTransformer():
...
 
 
-    def get_model_path(self):
-        return self.model_path
+    def get_model_path(self, format):
+        assert(format in ('coreml', 'mxnet'))
+        if (format in 'coreml') :
+            url = _urlparse.urljoin(MODELS_URL_ROOT, 'vgg16.mlmodel')
+        else:
+            url = _urlparse.urljoin(MODELS_URL_ROOT, 'vgg16-conv1_1-4_3.params')
```
nickjong: What's the significance of `-conv1_1-4_3` in the MXNet case?

abhishekpratapa: It's specifying the subset of layers we use in VGG16. We only use the first 12 or so layers of the VGG16 network. May be worth renaming.

PR #2510, opened 2019-10-28
flip two lines in image_similarity documentation

lunardog:
The documentation for `create` in `image_similarity` didn't make sense. I flipped two lines to make a legible sentence. Also, added the word _one_ to clarify that exactly one column of the Image type is expected.

## BEFORE:

```
   feature : string
        indicates that the SFrame has only column of Image type and that will
        Name of the column containing the input images. 'None' (the default)
        be used for similarity.
```

## AFTER:

```
   feature : string
        Name of the column containing the input images. 'None' (the default)
        indicates that the SFrame has only one column of Image type and that will
        be used for similarity.
```

### DIFF
--- a/src/python/turicreate/toolkits/image_similarity/image_similarity.py
+++ b/src/python/turicreate/toolkits/image_similarity/image_similarity.py
@@ -41,8 +41,8 @@ def create(dataset, label = None, feature = None, model = 'resnet-50', verbose =
         identify reference dataset rows when the model is queried.
 
     feature : string
-        indicates that the SFrame has only column of Image type and that will
         Name of the column containing the input images. 'None' (the default)
+        indicates that the SFrame has only column of Image type and that will
         be used for similarity.
 
     model: string, optional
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -41,8 +41,8 @@ def create(dataset, label = None, feature = None, model = 'resnet-50', verbose =
         identify reference dataset rows when the model is queried.
 
     feature : string
-        indicates that the SFrame has only column of Image type and that will
         Name of the column containing the input images. 'None' (the default)
+        indicates that the SFrame has only column of Image type and that will
```
shantanuchhabra: Should this be "SFrame has only **one** column of Image type"?

lunardog: Probably the word *one* is needed here, since it calls `_find_only_image_column`, which finds a column of type `Image` and raises and exception if there is more than one. Changing.

TobyRoseman:
@lunardog - thanks for fixing this.

PR #2512, opened 2019-10-28
Added style_transfer predict

abhishekpratapa:
# Overview
- Refactored Data iterator to have a train parameter
- Added Predict to the style_transfer wrapper class

### DIFF
--- a/src/toolkits/style_transfer/style_transfer.cpp
+++ b/src/toolkits/style_transfer/style_transfer.cpp
@@ -23,6 +23,7 @@ using turi::coreml::MLModelWrapper;
 using turi::neural_net::compute_context;
 using turi::neural_net::float_array_map;
 using turi::neural_net::float_scalar;
+using turi::neural_net::model_backend;
 using turi::neural_net::shared_float_array;
 
 namespace {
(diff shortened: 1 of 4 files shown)


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -49,8 +50,44 @@ void prepare_images(const image_type& image,
...
+    const shared_float_array& contents, size_t index, size_t batch_size,
+    size_t width, size_t height) {
+  constexpr size_t channels = 3;
+
+  std::vector<std::pair<flex_int, flex_image>> result;
+  result.reserve(batch_size);
+
+  size_t image_size = height * width * channels;
+
+  ASSERT_TRUE(contents.size() == image_size * batch_size);
```
nickjong: Nit: can use `ASSERT_EQ` here

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -49,8 +50,44 @@ void prepare_images(const image_type& image,
...
+  for (size_t index = 0; index < batch_size; index++) {
+    std::vector<uint8_t> image_data;
+    image_data.reserve(image_size);
+
+    size_t start_offset = image_size * index;
+    size_t end_offset = start_offset + image_size;
+
+    std::transform(start_ptr + start_offset, start_ptr + end_offset,
+                   std::back_inserter(image_data), [](float val) {
+                     return (uint8_t)(std::round(val * 255.f));
```
nickjong: Nit: use `static_cast` instead of C-style cast

Also, would be wise to clamp the result before casting to the range [0.f, 255.f]

If we had C++17, you could use `std::clamp` https://en.cppreference.com/w/cpp/algorithm/clamp

For now, I would just re-implement that:
```
float clamp(float v, float low, float high) {
  return (v < low) ? low : (high < v) ? high : v;
}
```

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -49,8 +50,44 @@ void prepare_images(const image_type& image,
...
+                   });
+
+    image_type img((char*)image_data.data(), height, width, channels,
```
nickjong: Nit: `reinterpret_cast<char*>`

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -188,6 +236,125 @@ void style_transfer::infer_derived_options() {
...
+
+  std::unique_ptr<compute_context> ctx = create_compute_context();
+  if (ctx == nullptr) {
+    log_and_throw("No neural network compute context provided");
+  }
+
+  float_array_map weight_params = m_resnet_spec->export_params_view();
+
+  // A value of `0` to indicate prediction
+  shared_float_array st_train(std::make_shared<float_scalar>(0));
```
nickjong: FYI you also write this as:
```
shared_float_array st_train = shared_float_array::wrap(0.f);
```
using this factory function: https://github.com/apple/turicreate/blob/65ef3e972953776d186c8d6a12f3e23c64a4d55b/src/ml/neural_net/float_array.hpp#L140

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -188,6 +236,125 @@ void style_transfer::infer_derived_options() {
...
+      // get shared float array from prediction result
+      shared_float_array out_shared_float_array = result_batch.at("output");
+
+      // populate gl_sframe_writer
+      std::vector<std::pair<flex_int, flex_image>> processed_batch =
+          process_output(out_shared_float_array, i, actual_batch_size,
+                         image_width, image_height);
+
+      // Write result to gl_sframe_writer
+      for (auto row : processed_batch) {
```
nickjong: Use `const auto& row` to avoid a copy

src/toolkits/style_transfer/style_transfer.hpp
```diff
@@ -114,6 +121,8 @@ class EXPORT style_transfer : public ml_model_base {
 
   std::unique_ptr<table_printer> training_table_printer_;
 
+  gl_sarray convert_types_to_sarray(const variant_type& data);
```
nickjong: I believe this could be a static method (doesn't depend on any instance variables)

PR #2514, opened 2019-10-28
End to end model.train for Drawing Classifier C++

shantanuchhabra:
Currently, works with no validation data, for one iteration (until `predict` is called in drawing_classifier.cpp:570)

Repro steps:

```python
import turicreate as tc
import turicreate.toolkits.libtctensorflow
data = tc.SFrame('train.sframe/')
model = tc.extensions.drawing_classifier()
model.train(data, 'label', 'image', tc.SFrame())
```

### DIFF
--- a/src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
@@ -28,8 +28,6 @@ class DrawingClassifierTensorFlowModel(TensorFlowModel):
         for key in net_params.keys():
             net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
 
-        # import pdb; pdb.set_trace()
-
         _tf.reset_default_graph()
 
         self.num_classes = num_classes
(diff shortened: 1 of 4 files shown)


src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
```diff
@@ -175,99 +206,99 @@ def export_weights(self):
...
-                        "time": _time.time() - start_time}
-            if validation_set is not None:
-                kwargs["validation_accuracy"] = "{:.3f}".format(val_acc)
-            table_printer.print_row(**kwargs)
-
-    final_val_accuracy = val_acc if validation_set else None
-    total_train_time = _time.time() - start_time
-
-    return final_train_accuracy, final_val_accuracy, final_train_loss, total_train_time
+    # def evaluate(self, train_loader):
```
shreyajain17: I suggest we remove this portion completely as this was needed when tensorflow worked with Python implementation.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -450,10 +457,13 @@ void drawing_classifier::iterate_training() {
     // below should be concurrent with the neural net inference for that batch.
     pop_until_size(1);
 
+    std::cout << "has next batch" << std::endl;
```
shreyajain17: Should remove debugging print statements before merging to master.

shreyajain17:
I left feedback without realizing it is WIP.

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -184,6 +184,12 @@ data_iterator::batch simple_data_iterator::next_batch(size_t batch_size) {
 
   // Wrap the buffers as float_array values.
   data_iterator::batch result;
+
+  if (real_batch_size < batch_size) {
+     auto pend = std::begin(batch_drawings) + real_batch_size * image_data_size;
+     batch_drawings.erase(pend, batch_drawings.end());
+   }
+
```
nickjong: Doesn't this just duplicate lines 200-203?

shantanuchhabra: Fixed, thanks!

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -187,7 +187,8 @@ std::unique_ptr<model_spec> drawing_classifier::init_model() const {
       /* input               */ input_name,
       /* num_output_channels */ 128,
       /* num_input_channels  */ 64 * 3 * 3,
-      /* weight_init_fn      */ initializer);
+      /* weight_init_fn      */ initializer,
+      /* bias_init_fn        */ zero_weight_initializer());
```
Jarvi-Izana: Nice catch! In protobuf it's implicitly zero-initialized. But explicit is better!

shantanuchhabra:
Internal tests passed.

PR #2515, opened 2019-10-28
OD API breaks fixed

jakesabathia2:
In this PR I did several things:
1. Fix the bug that the predict() function needs data with ground truth annotations.
2. Enable the train() function to accept `class_labels`, when the invalid labels are met, the data iterator will ignore it.
3. Fix the bug that the previous one will crash if evaluate() has input data without annotations.
4. Change the output format of predict() when the input is a single image (on Python side tho).
5. In the previous implementation, the OD model assumes the model input has name "image", which cause couple of crashed when exporting to cormel. Change the input name to be user specified to fix this issue.
6. Enable the annotation to be a list of dictionary or a single dictionary.
7. Change couple of test case in order to deal with the fixed issues.


src/ml/neural_net/model_spec.hpp
```diff
@@ -395,6 +395,13 @@ class model_spec {
    */
   void add_preprocessing(const std::string& feature_name,
                          const float image_scale);
+  void change_preprocessing_name(const std::string& old_name,
+                                 const std::string& new_name);
+
+  /** Set input name for given layer
+   */
+  void set_layer_input(const std::string& layer_name,
+                       const std::string& input_name, const size_t index);
```
nickjong: It feels strange to me that we should need these.... I'll follow up below where they are used, but it seems weird to need to modify the spec after we've built it. Ideally we can build it correctly the first time, as necessary

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -444,7 +444,7 @@ def update_progress(cur_loss, iteration):
 
         }
         model = _tc.extensions.object_detector()
-        model.train(data=dataset, annotations_column_name=annotations, image_column_name=feature, options=tf_config)
+        model.train(data=dataset, annotations_column_name=annotations, image_column_name=feature, class_labels=classes, options=tf_config)
```
nickjong: Please add `class_labels` to options. I doesn't feel like a top-level argument to me

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -141,8 +141,9 @@ void set_image_feature_size_range(ImageFeatureType* image_feature,
 ImageFeatureType* set_image_feature(
     FeatureDescription* feature_desc, size_t image_width, size_t image_height,
     std::string description = "",
-    ImageFeatureType::ColorSpace image_type = ImageFeatureType::RGB) {
-  feature_desc->set_name("image");
+    ImageFeatureType::ColorSpace image_type = ImageFeatureType::RGB,
+    std::string input_name = "image") {
```
nickjong: Let's not encode specific strings as defaults here. The caller should specify every time. I think `input_name` should be earlier in the argument list, too. It's more primary than e.g. `image_type`

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -161,6 +162,13 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
     size_t image_height, size_t num_classes, size_t num_predictions,
     flex_dict user_defined_metadata, flex_list class_labels,
     std::map<std::string, flexible_type> options) {
+  // Get the name of image input
+  std::string input_name;
+  for (auto& it : user_defined_metadata) {
+    if (it.first == "feature") {
+      input_name = it.second.get<flex_string>();
+    }
+  }
```
nickjong: Could we just require the caller to specify this, as another argument, instead of assuming that this is in the `user_defined_metadata`?

src/toolkits/object_detection/object_detector.hpp
```diff
@@ -39,6 +39,7 @@ class EXPORT object_detector: public ml_model_base {
 
   void train(gl_sframe data, std::string annotations_column_name,
              std::string image_column_name, variant_type validation_data,
+             std::vector<std::string> class_labels,
```
nickjong: As discussed above, let's just include this `options`. Almost everything optional should be in options.

src/ml/neural_net/model_spec.hpp
jakesabathia2: It is weird, the reason I do here is in the `init_model()`, the `nn_spec_` is read from a pretrain model file:
```
std::unique_ptr<model_spec> nn_spec(new model_spec(pretrained_mlmodel_path));
```
which fixed the input name to "image", which might not be the feature column name the user gives.
Currently I don't have a better solutions for this weird usage.
But yes for the `import_from_custom()` I can directly construct a correct one.

nickjong: Let's set the names during export, not in the main toolkit. That code can use the underlying proto code if needed, not the `model_spec` abstraction.

We need to keep this abstraction clean, so it's more obvious how to use it when defining a model from scratch

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -990,6 +1001,42 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
...
+          iterator_params.annotations_column_name)) {
+    std::vector<flexible_type> dummy;
+    for (size_t i = 0; i < iterator_params.data.size(); i++) {
+      dummy.push_back(flex_list());
+    }
+    gl_sarray empty_ann(dummy, flex_type_enum::LIST);
+    iterator_params.data.add_column(empty_ann,
+                                    iterator_params.annotations_column_name);
+    iterator_params.has_annotations = false;
+  }
```
nickjong: Why is this needed? Shouldn't it be an error if the annotations column is missing?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -990,6 +1001,42 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
...
+      list_type_annotations.push_back(flex_list());
+    } else {
+      log_and_throw(
+          "Invalid annotation! Expect list of dictionary, a singel "
+          "dictionary!");
+    }
+  }
+  gl_sarray ann_sarray(list_type_annotations, flex_type_enum::LIST);
+  iterator_params.data.replace_add_column(
+      ann_sarray, iterator_params.annotations_column_name);
```
nickjong: Could we handle this on the Python side? I think it makes the most sense for the C++ code to be more strict about types.

Also, we shouldn't be using random access to seek into SArray or SFrame values. This is extremely inefficient! Talk to @Jarvi-Izana or @TobyRoseman about using `apply` or other parallel methods. You should also be able to query the column's type upfront, not query each individual element

nickjong:
In general, I recommend using different commits for different issues. Then it would be trivial to cherry pick those into different branches for review. At worst, you can use a soft `git reset` to un-stage all your cumulative changes, then use `git add -i` to parse them into separate commits first.

jakesabathia2:
Okay I'll do it! Thanks!

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: This is not relative to `evaluate()`, which I check the existence of annotation in the very beginning.
The change is for `predict()` with no annotations data.
Basically, the current OD data iterator assumes there "MUST" be an annotation column, or lots of function will crash. So what I did here is for the case for which the input data for `predict()` doesn't;t has ground truth, then I append a dummy one.
But yes I will break it down to several PRs.
Thanks!

nickjong: Ah, then let's just make the iterator robust to the `annotations_column_name` being empty, in which case there are no annotations. We shouldn't create an empty column just for this purpose

jakesabathia2: Okay for sure :D
This might be a much bigger change tho.

nickjong: It shouldn't be. Make sure you understand how the data iterator works. We already support the predictions column name being empty.

jakesabathia2: Alright I'll take a look!

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1698,7 +1698,10 @@ def predict(self, dataset, confidence_threshold=0.25, iou_threshold=0.45):
         options = {}
         options["confidence_threshold"] = confidence_threshold
         options["iou_threshold"] = iou_threshold
-        return  self.__proxy__.predict(dataset, options)
+        prediction = self.__proxy__.predict(dataset, options)
+        if isinstance(dataset, _tc.Image):
```
Jarvi-Izana: Do you need to check the output type of `predict` as well?

PR #2516, opened 2019-10-29
Handle error message with path info

### LINKED ISSUE #789: Default value used for TURI_CACHE_FILE_LOCATIONS if that path does not exist

dhivyaaxim:
This PR resolves #789 
Default value used for TURI_CACHE_FILE_LOCATIONS if that path does not exist

### DIFF
--- a/src/core/storage/fileio/fileio_constants.cpp
+++ b/src/core/storage/fileio/fileio_constants.cpp
@@ -85,8 +85,10 @@ static bool check_cache_file_location(std::string val) {
   if (paths.size() == 0)
     throw std::string("Value cannot be empty");
   for (std::string path: paths) {
-    if (!boost::filesystem::is_directory(path))
+    if (!boost::filesystem::is_directory(path)){
+      std::cerr<<"Could not write to the turicreate file cache, which is currently set to '" + path + "' Default path is set to : "<<std::endl;
       throw std::string("Directory: ") + path + " does not exist";
+    }
   }
   return true;
 }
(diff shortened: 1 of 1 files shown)


src/core/storage/fileio/fileio_constants.cpp
```diff
@@ -85,8 +85,10 @@ static bool check_cache_file_location(std::string val) {
   if (paths.size() == 0)
     throw std::string("Value cannot be empty");
   for (std::string path: paths) {
-    if (!boost::filesystem::is_directory(path))
+    if (!boost::filesystem::is_directory(path)){
+      std::cerr<<"Could not write to the turicreate file cache, which is currently set to '" + path + "' Default path is set to : "<<std::endl;
```
TobyRoseman: Rather than `Default path is set to : `, I think it would be clearer to say `Using path:`.

PR #2517, opened 2019-10-29
Added Fixes For single Scale value in CoreML

abhishekpratapa:
- The scale used in OD and Style Transfer are slightly different. To handle theses different use_cases, and match the CoreML specs a new `add_scale` is proposed.

### DIFF
--- a/src/ml/neural_net/model_spec.hpp
+++ b/src/ml/neural_net/model_spec.hpp
@@ -306,17 +306,6 @@ public:
 
 
   /**
-   * Appends a layer that scales an entire batch with a single value
-   *
-   * \param name The name of the layer and its output
-   * \param input The name of the layer's input
-   * \param shape_c The shape of the channel
-   * \param weight_initializer_fn Callback used to initialize the weights
-   */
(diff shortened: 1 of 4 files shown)


jakesabathia2:
Cool!

src/ml/neural_net/model_spec.hpp
```diff
@@ -305,6 +305,17 @@ class model_spec {
...
+  /**
+   * Appends a layer that scales an entire batch with a single value
+   *
+   * \param name The name of the layer and its output
+   * \param input The name of the layer's input
+   * \param shape_c The shape of the channel
+   * \param weight_initializer_fn Callback used to initialize the weights
+   */
+  void add_scale(const std::string& name, const std::string& input,
+                 weight_initializer scale_initializer_fn);
```
nickjong: Could we instead just generalize the existing `add_scale` to take an arbitrary shape vector?

nickjong: Just have it take `const std::vector<size_t>& scale_shape` or something like that?

PR #2518, opened 2019-10-29
OD fix bug for missing annotation for evaluate.

jakesabathia2:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -470,6 +470,11 @@ void object_detector::finalize_training() {
 variant_type object_detector::evaluate(gl_sframe data, std::string metric,
                                        std::string output_type,
                                        std::map<std::string, flexible_type> opts) {
+  // check if data has ground truth annotation
+  if (!data.contains_column(read_state<flex_string>("annotations"))) {
+    log_and_throw("No annotation is found in data for evaluate!!");
```
nickjong: Two exclamation points seems like overkill. There's no need for the user to panic! Could we just keep the old error message? https://github.com/apple/turicreate/blob/2c2a206c8b1d3eb86916fd22677c3de751fe0f43/src/python/turicreate/toolkits/object_detector/object_detector.py#L94

PR #2521, opened 2019-10-29
Style transfer - Load Custom Model

abhishekpratapa:
Added load custom model for Style Transfer.


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -36,6 +36,95 @@ constexpr size_t DEFAULT_WIDTH = 256;
 
 constexpr size_t DEFAULT_BATCH_SIZE = 1;
 
+const std::map<std::string, std::string> CUSTOM_MODEL_NAMING_MAP = {
```
nickjong: Today is the day you learn about some of the intricacies of static initialization. https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

Anything global variable you define like this is initialized early in the program, effectively before `main` runs. It's often important to keep program loading very cheap and efficient, so it's generally best practice to initialize non-trivial global/static variables lazily.

You can make things lazy by wrapping this in a function. So instead of
```
const std::map<std::string, std::string> CUSTOM_MODEL_NAMING_MAP = ...
```
prefer
```
const std::map<std::string, std::string>& get_custom_model_naming_map() {
  static const std::map<std::string, std::string> CUSTOM_MODEL_NAMING_MAP = ...;
  return CUSTOM_MODEL_NAMING_MAP;
}
```

Even then, this lazily initialized variable inside the function will be deconstructed when the program ends, essentially right after `main` returns. So best practice (thanks, C++), is actually something like:
```
const std::map<std::string, std::string>& get_custom_model_naming_map() {
  static const auto *CUSTOM_MODEL_NAMING_MAP = new std::map<std::string, std::string>(...);
  return *CUSTOM_MODEL_NAMING_MAP;
}
```
Because raw pointers (like `auto*`) don't do anything when they are "deconstructed". They "leak" the memory, but that's okay, since the process is ending anyway....

I hope this makes some sense. The main idea is that we (a) want to delay doing the work of allocating this map until we need it, and (b) less importantly, it's nice to skip going through the work of deallocating this map when the program ends.

PR #2522, opened 2019-10-29
OD annotation conversion

jakesabathia2:
The OD can use data containing annotations in either list or dictionary type.


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -959,6 +959,13 @@ std::shared_ptr<MLModelWrapper> object_detector::export_to_coreml(
 std::unique_ptr<data_iterator> object_detector::create_iterator(
     gl_sframe data, std::vector<std::string> class_labels, bool repeat) const
 {
+  // Check annotations type. If it is a single dictionary, put it into a list.
+  const std::string& annotations_column_name =
+      read_state<flex_string>("annotations");
+  if (data.contains_column(annotations_column_name)) {
+    change_annotations_type(data, annotations_column_name);
```
nickjong: Functions that modify their arguments are not as obvious/intuitive to read. Could we move the call to `replace_add_column` to the front, so that this line reads something more like:
```
data[annotations_column_name] = canonicalize_annotations(data[annotations_column_name]);
```
Then it's more immediately obvious what's happening here

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -1347,5 +1354,31 @@ void object_detector::update_model_metrics(gl_sframe data,
   add_or_update_state(metrics);
 }
 
+void object_detector::change_annotations_type(
+    gl_sframe& data, const std::string& annotations_name) {
+  auto change_annotations_type = [=](const flexible_type& annotation) {
```
nickjong: Why use a capture list of `[=]` instead of the simpler `[]`?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -1347,5 +1354,31 @@ void object_detector::update_model_metrics(gl_sframe data,
...
+        break;
+      case flex_type_enum::DICT:
+        annotation_list.push_back(annotation);
+        break;
+      case flex_type_enum::UNDEFINED:
+        break;
+      default:
+        log_and_throw(
+            "Invalid annotation! Expect list of dictionary, a singel "
+            "dictionary!");
```
nickjong: This error message contains a typo. Could we just use whatever we were printing before, in the Python implementation?

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: 🧐 Somehow in the previous code I used a outside local value and forget to delete it after I changed my implementation.
Thanks!

src/toolkits/object_detection/object_detector.hpp
```diff
@@ -219,6 +219,8 @@ class EXPORT object_detector: public ml_model_base {
                                            const std::string& output_type);
   static gl_sframe convert_types_to_sframe(const variant_type& data,
                                            const std::string& column_name);
+  static gl_sarray canonicalize_annotations(gl_sarray& data);
```
nickjong: Let's pass by value. It could be unexpected that this function can mutate its argument. (https://google.github.io/styleguide/cppguide.html#Reference_Arguments)

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -959,6 +959,15 @@ std::shared_ptr<MLModelWrapper> object_detector::export_to_coreml(
...
 {
+  // Check annotations type. If it is a single dictionary, put it into a list.
+  const std::string& annotations_column_name =
+      read_state<flex_string>("annotations");
+  if (data.contains_column(annotations_column_name)) {
+    gl_sarray annotations = data[annotations_column_name];
+    data.replace_add_column(canonicalize_annotations(annotations),
+                            annotations_column_name);
+  }
+
```
nickjong: I think logic might be better suited to live in `od_data_iterator.cpp`, which is where we currently reason about the data format.

jakesabathia2:
I have put the conversion into data_iterator.cpp::get_data

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -27,7 +48,13 @@ flex_image get_image(const flexible_type& image_feature) {
   }
 }
 
-gl_sframe get_data(const data_iterator::parameters& params) {
+gl_sframe get_data(data_iterator::parameters params) {
+  // Check annotations type. If it is a single dictionary, put it into a list.
+  if (params.data.contains_column(params.annotations_column_name)) {
+    gl_sarray raw_annotations = params.data[params.annotations_column_name];
+    params.data.replace_add_column(canonicalize_annotations(raw_annotations),
```
nickjong: The rest of the code in `get_data` takes some care not to mutate the user's original SFrame. Could we integrate this below, so that we don't call destructive (mutating) methods on the input SFrame, such as `replace_add_column`?

PR #2480, opened 2019-10-22
dc predict

Jarvi-Izana:
add reviewers later.

reorganized the mock interface for code reuse.

__mY mANIFEST__
- [x] perform_infernece
- [x] predict
- [x] predict_topk

__My Summary__
- [x] provide unittest.
  - [x] perfrom_inference
  - [x] predict
  - [x] predict topk
- [x] pass the test pipeline.

PR #2523, opened 2019-10-29
fix for test break

Jarvi-Izana:
a small fix for test break of `test_drawing_classifier.cxx` to unblock #2480.

model state `feature` type of string is changed to state `features` with type list. We should make sure all existing test cases pass in master so that we won't block other people's pipeline validation process.

one suspicious part is that there's a model state called `classes`. I'm not sure it's used. @shantanuchhabra may spend some time investigating.


Jarvi-Izana:
cherry-picked from #2480

PR #2524, opened 2019-10-29
Adding support for warmstart CoreML in DC

fareeha:
Support loading warmstart CoreML model and plumbing through to C++ `train()`.


shantanuchhabra:
What feedback are you looking for, with this WIP PR?

fareeha:
> What feedback are you looking for, with this WIP PR?

Nothing yet. As I mentioned this is just to checkpoint where I am on this effort. Once the relevant pieces go in, I can test these changes. (In the meantime I'll context switch to help with SC)

nickjong:
For future reference, there's no need to open a PR if you just want to have a branch with your WIP

nickjong:
Please feel free to add reviewers when you want someone to review this

fareeha:
> solve here. Could you merge master and then sanity check what happens when you train with the option enabled? And what is the state of the unit tests after this change?

Yes just merged master and resolved conflicts. I'll run the unit tests and check.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -349,9 +349,24 @@ void drawing_classifier::init_training(
...
 
-  // Initialize the neural net. Note that this depends on statistics computed
-  // by the data iterator.
-  nn_spec_ = init_model();
+  // If warm_start enabled, an 'mlmodel_path' is passed in options.
+  // Else, call init_model() to initialize the neural net using xavier.
+  if (opts.find("mlmodel_path") != opts.end() ) {
+    // Initialize the neural net with warm start model weights.
+    auto mlmodel_path_iter = opts.find("mlmodel_path");
+    if (mlmodel_path_iter == opts.end()) {
```
Jarvi-Izana: this logic is redundant.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -349,9 +349,24 @@ void drawing_classifier::init_training(
...
+      log_and_throw("Expected option \"mlmodel_path\" not found.");
+    }
+    const std::string mlmodel_path = mlmodel_path_iter->second;
```
Jarvi-Izana: no need to const. just use `const &`. Don't erase it from the option. It's good for debugging.

Jarvi-Izana: I mean it's better to keep `opts` intact since it's passed from the user layer.

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: Yeah, we should just initialize `mlmodel_path_iter` first and test that in line 354

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: In this particular case, we should delete it because we don't want to surface it in the model state. This isn't something the user can specify manually for now

Using `const&` to avoid a copy would still be nice, though.

Jarvi-Izana: I believe we don't surface that in our metadata section unless we explicitly want to expose that. Correct me if I'm wrong.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -298,6 +304,17 @@ void drawing_classifier::init_training(
...
   });
 
+  auto warm_start = opts.at("warm_start");
+  std::string mlmodel_path;
+  if ( warm_start == true ) {
+    auto mlmodel_path_iter = opts.find("mlmodel_path");
+    if (mlmodel_path_iter == opts.end()) {
+      log_and_throw("Expected option \"mlmodel_path\" not found.");
+    }
+    mlmodel_path = mlmodel_path_iter->second.to<std::string>();
```
Jarvi-Izana: to me, you don't need to introduce `mlmodel_path` to the outer scope. Putting all relevant logic into the right decision branch is better to group your code.
```
if ( warm_start == true ) {
``` 
Please don't feel strange if I'm too polite today.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -860,7 +886,7 @@ std::shared_ptr<coreml::MLModelWrapper> drawing_classifier::export_to_coreml(
       {"features", features_string},
       {"max_iterations", read_state<flex_int>("max_iterations")},
       // TODO: Uncomment as part of #2524
-      // {"warm_start", read_state<flex_int>("warm_start")},
+      {"warm_start", read_state<flex_int>("warm_start")},
```
Jarvi-Izana: I believe we don't expose opts unless we specify it at here.

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: Everything we put into the `state` object will be exposed to the user in Python, via this mechanism: https://github.com/apple/turicreate/blob/b5e16c58d75913ad2a2dd4d4a05b631145fdff4c/src/model_server/lib/extensions/ml_model.cpp#L24

Jarvi-Izana: glad to know this path! I didn't notice this hidden part.

Jarvi-Izana: if @fareeha wants to use the reference of the string, make sure you defer to delete the iterator after you use the string reference variable. Or you can move itr->second to your string and then delete it immediately.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -235,6 +235,12 @@ void drawing_classifier::init_options(
       FLEX_UNDEFINED,
       std::numeric_limits<int>::min(), std::numeric_limits<int>::max());
 
+  options.create_boolean_option(
+      "warm_start",
+      "Enable warm start model initialization for training.",
+      false,
+      true);
```
nickjong: What we save in the model state should probably reflect what the user-visible API is: https://github.com/apple/turicreate/blob/2ebb43b1dbe10b369f57ad22c3432eaf62148d28/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py#L97

My guess is that this should be a string value, either "quickdraw_245_v0" or the empty string as a proxy for `None`.

(Someday we might be able to handle the model downloading in C++ and sidestep the use of the internal "mlmodel_path" option.)

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: At this stage in the release, let's avoid changing the order of when initialize `nn_spec_` below, which depends on the value of `mlmodel_path`

nickjong: Honestly, instead of looking at `warm_start`, we could just go whether or not there is a `mlmodel_path` key. Trust the Python side to do the right thing for now, aligning the value of `mlmodel_path` with the value of `warm_start`

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -349,9 +366,18 @@ void drawing_classifier::init_training(
   std::vector<std::string> gpu_names = training_compute_context_->gpu_names();
   print_training_device(gpu_names);
 
-  // Initialize the neural net. Note that this depends on statistics computed
-  // by the data iterator.
-  nn_spec_ = init_model();
+  // If warm_start enabled, use warmstart CoreML model to
+  // initialize the neural net. Else, call init_model() to
+  // initialize the neural net using xavier.
```
nickjong: Don't we still need to initialize the final layers? We need to append the class-specific layers, like we do for OD

src/toolkits/drawing_classifier/drawing_classifier.cpp
fareeha: I'm not sure I follow..

`mlmodel_path` is referenced below so it needs to be defined outside any one scope.

src/toolkits/drawing_classifier/drawing_classifier.cpp
fareeha: Yup, working on that right now

Jarvi-Izana: that's a good point. I think we only need the weights.

Jarvi-Izana: I forgot to mention this while I was cooking just now.

fareeha: @nickjong right now I'm deciding between (a) repeating dense1 onwards init code from `init_model()` in the loop on line 372. Or (b) move the coldstart + warmstart model definition inside `init_model()` but this would require passing `warm_start` as an argument and it effects `load_version()` which calls `init_model()`

Jarvi-Izana: it's better just call init_model first and copy weights from warm start to avoid any incompatible issues of different spec versions.

src/toolkits/drawing_classifier/drawing_classifier.cpp
fareeha: For simplicity I used a boolean because in the API we have:
```
"auto": Uses quickdraw_245_v0
"quickdraw_245_v0": Uses a model trained on 245 of the 345 classes in the
                         Quick, Draw! dataset.
None: No Warm Start"
```
It appears that auto and quickdraw_245_v0 both point to the same warmstart model. So it is simply an option to use a warmstart model or not, hence I captured that as True or False.

fareeha:
```
turicreate/test/test_drawing_classifier.py ............xxxx....................xxxx....................xxxx........ 
```

src/toolkits/drawing_classifier/drawing_classifier.cpp
fareeha: I'm not sure I follow, can you explain what you mean?

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -171,8 +180,12 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         options = dict()
         options["batch_size"] = batch_size
         options["max_iterations"] = max_iterations
-        # Enable the following line when #2524 is merged
-        # options["warm_start"] = warm_start
+        options["warm_start"] = False if warm_start==None else True
```
nickjong: Let’s pass either `None`or “quickdraw_245_v0” here

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: Since this option is encoded into the state, we should make this more future proof. Also, if the user specifies “quickdraw_245_v0” explicitly during model creation, that value should appear in the model state too

fareeha: @nickjong After the change now, we replace 'auto' with whichever model version is currently our default, i.e. 

- if user passes 'auto' or 'quickdraw_245_v0' in `warm_start`, opts `warm_start` stores “quickdraw_245_v0”
- if user passes 'quickdraw_245_v[X]' in `warm_start`, opts `warm_start` stores “quickdraw_245_v[X]” as we add/support more versions in the future
- if user passes None in `warm_start`, opts `warm_start` stores "None"

src/toolkits/drawing_classifier/drawing_classifier.cpp
srikris: @Jarvi-Izana This is for export-coreml metadata, not a user option.

nickjong: Isn't "warm_start" a string option?

fareeha: Good catch! Ty. This typo was the reason tests failed.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -171,8 +182,12 @@ def create(input_dataset, target, feature=None, validation_set='auto',
...
         options["batch_size"] = batch_size
         options["max_iterations"] = max_iterations
-        # Enable the following line when #2524 is merged
-        # options["warm_start"] = warm_start
+        if warm_start:
+            # Load CoreML warmstart model
+            pretrained_mlmodel = _pre_trained_models.DrawingClassifierPreTrainedMLModel()
+            options["mlmodel_path"] = pretrained_mlmodel.get_model_path()
+
+        options["warm_start"] = "None" if warm_start==None else warm_start
```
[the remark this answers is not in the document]
fareeha: done

src/toolkits/drawing_classifier/drawing_classifier.cpp
fareeha: done

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -353,6 +373,13 @@ void drawing_classifier::init_training(
   // by the data iterator.
   nn_spec_ = init_model();
 
+  if ( enable_warmstart ) {
+    // Initialize the neural net with warm start model weights.
+    model_spec warmstart_model(mlmodel_path);
+    float_array_map trained_weights = warmstart_model.export_params_view();
```
Jarvi-Izana: the 6.0 layer name might be different from the warm_start model.

fareeha: different how?

fareeha: Also specifically what change are you requesting here?

Jarvi-Izana: the layer name before has suffix `_fwd` but this release doesn't have. You may check the implementation of update_weights.

fareeha: the coreml model does not have the suffix `fwd`

fareeha: I manually removed it and uploaded the model again. A number of cleanups were done on the CoreML model (pasting from my previous standup in case you missed):

> I uploaded the Drawing Classifier CoreML model with the valid author and license metadata and the license txt file is uploaded and CoreML model now contains only the relevant layers (i.e. removed dense1, softmax, preprocessing etc), edited model output field to match dense0 output (removing ClassProbabilites and ClassLabels) and converted from neuralNetworkClassifier to neuralNetwork since it is no longer a classification network. Added the supporting code to fetch the MLmodel path from Python and load in C++. Added warm_start as a user option in C++. My focus is to ensure the warm start model correctly reaches the Python TF train(). Removed the additional fwd suffix in the CoreML model layer names which is not expected by the TF implementation of DC.

Jarvi-Izana: Good. That's good enough for this release. Let's do this pragmatically next time.

fareeha: What wasn't pragmatic about this?

fareeha: I don't think I follow what you mean to say -  what isn't good enough? and what can be done more pragmatically next time?

Jarvi-Izana: Sorry I didn't make myself clearer. Think about you have more than 10 warm start models to try on, and you don't want to manually do it for each of them (even though, you may want to have a pipeline to process them all). You can have a function to match the warm start layer name with the current layer name format in DC. This will loosely decouple your dependency on the **exact** naming format. If we want to add the `_fwd` suffix back or add some extra info, we don't need to modify the warm start model and upload the modified copy again. A potential solution is to use `regex`.

fareeha: Yeah that's a great point! Toby shared a script with me that helped me do a lot of the processing on the CoreML model, so we can set up a script to do the CoreML model pre-processing for DC warmstart models. The idea was also to remove any additional meta data or layers that we do not need and make the uploaded model file as compact as possible.

For name matching in the warm start model and the layer name, we could introduce name cleanup in the code. My options to deal with the suffix was either do it in code or do it as part of the CoreML model cleanup so I went with the latter because I didn't want to expose a user with layer names that were not consistent with the layer names we define in TF implementation. Ideally if a user opens the warmstart model in coreml, they should see the same layers names as ones we use in our code.

Jarvi-Izana: excellent!

PR #2525, opened 2019-10-30
Implementing TF class for Sound Classifier

fareeha:
Added model definition, `use_tensorflow` flag, `train()`, `predict()`, `export_weights()`, `evaluate()`


src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -0,0 +1,233 @@
...
+from .._tf_model import TensorFlowModel
+import turicreate.toolkits._tf_utils as _utils
+
+import tensorflow.compat.v1 as _tf
+_tf.disable_v2_behavior()
+
+
+class SoundClassifierTensorFlowModel(TensorFlowModel):
+
+    def __init__(self, batch_size, num_classes):
```
shreyajain17: I think the network might not need `batch_size` as input at all.

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -0,0 +1,233 @@
...
+        """
+        #for key in net_params.keys():
+        #    net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
+
+        _tf.reset_default_graph()
+
+        self.num_classes = num_classes
+        self.batch_size = batch_size
+
+        self.x = _tf.placeholder("float", [None, 1, 12288])
```
shreyajain17: Maybe better to use `_tf.float32`

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -0,0 +1,233 @@
...
+        self.y = _tf.placeholder("float", [None, self.num_classes])
+
+        # Weights
+        weights = {
+        'sound_dense0_weight'  : _tf.Variable(_tf.random.uniform([12288, 100]), name='sound_dense0_weight'),
```
shreyajain17: Are these weights going to be overwritten by MXNet? I realized when doing style transfer tensorflow graph that argument `initial_value` is extremely useful.

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -0,0 +1,233 @@
...
+import numpy as _np
+import time as _time
+from .._tf_model import TensorFlowModel
+import turicreate.toolkits._tf_utils as _utils
+
+import tensorflow.compat.v1 as _tf
+_tf.disable_v2_behavior()
+
+
+class SoundClassifierTensorFlowModel(TensorFlowModel):
```
shreyajain17: If SoundClassifierTensorFlowModel inherits from TensorFlowModel, you will also have to implement `set_learning_rate()`

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -0,0 +1,233 @@
...
+                                self.x: feed_dict['x'],
+                                self.y: feed_dict['y']
+                            })
+        result = {'accuracy' : final_accuracy, 'predictions' : pred_probs}
+        return result
+
+
+    def export_weights(self):
+        """
+        Retrieve weights from the TF model, convert to the format MXNET
```
shreyajain17: In general, export_weights is a function that converts the weights from existing format to CoreML format.

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
fareeha: That is a part I'm still figuring out how I want to do this. Since we are removing MXNET pieces I didn't want to rely on getting weights from MXNET

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
shreyajain17: Yeah sorry I thought it was high level stuff.

nickjong:
This is a great start! I think you've assembled most of the pieces on the TF side we'll need. My recommendation would be to integrate them in a few stages: I think we want to avoid duplicating this much of the existing logic between the `use_tensorflow` path and the old path. We can discuss in more detail offline, but I would recommend:
1. Encapsulate all the existing MXNet NN training code (it's fine for now to leave the data iterator and metrics) into a class, like  what you've done for a TF class
2. Update the toolkit to use this encapsulated class instead of MXNet code directly
3. Define your TF class with the same interface
4. Just use `use_tensorflow` to select which class to instantiate

In this way, we use this class abstraction/encapsulation to define rigorously the boundary between the old and new paths.

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
fareeha: There are still some pieces to be done: learning rate, data iterator etc

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -219,11 +232,9 @@ def create(dataset, target, feature, max_iterations=10,
                                     label=train_data['labels'].to_numpy(),
                                     batch_size=training_batch_size, shuffle=True)
 
-    custom_NN = SoundClassifier._build_custom_neural_network(feature_extractor.output_length, num_labels, custom_layer_sizes)
+    from ._mx_sound_classifier import SoundClassifierMXNETModel
     ctx = _mxnet_utils.get_mxnet_context()
```
nickjong: Could we move the MXNet context into `SoundClassifierMXNETModel`, too? I think it's too tied to the NN training, as an implementation detail. There won't be any effective way to use this with TensorFlow.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -237,40 +248,29 @@ def create(dataset, target, feature, max_iterations=10,
...
     if validation_data:
         validation_metric = mx.metric.Accuracy()
-    softmax_cross_entropy_loss = mx.gluon.loss.SoftmaxCrossEntropyLoss()
+
     for i in range(max_iterations):
         # TODO: early stopping
 
         for batch in train_data:
             data = mx.gluon.utils.split_and_load(batch.data[0], ctx_list=ctx, batch_axis=0, even_split=False)
             label = mx.gluon.utils.split_and_load(batch.label[0], ctx_list=ctx, batch_axis=0, even_split=False)
```
nickjong: This `split_and_load` stuff probably also belongs in the encapsulated class. Probably the `train` method can just take `batch.data` and `batch.label`, or perhaps `batch.data[0]` and `batch.label[0]`

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -289,7 +289,7 @@ def create(dataset, target, feature, max_iterations=10,
 
     state = {
         '_class_label_to_id': class_label_to_id,
-        '_custom_classifier': custom_NN,
```
nickjong: We'll also need changes where `SoundClassifier` uses this state, since `state._custom_classifier` is now an instance of `SoundClassifierMXNETModel`, not `mx.nn.Sequential` or whatever

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
```diff
@@ -0,0 +1,51 @@
...
+# Use of this source code is governed by a BSD-3-clause license that can
+# be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+
+
+import mxnet as mx
+import time
+import turicreate as _tc
+
+
+class SoundClassifierMXNETModel():
```
nickjong: Think bigger. There isn't anything here specific to sound classification. Maybe this can just be `MultiLayerPerceptionMXNetModel`

(Note: should be `MXNet` not `MXNET`, per the stylization I see on their web site: https://mxnet.incubator.apache.org (e.g. in the legal text at the bottom)

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+	def __init__(self, feature_extractor, num_labels, custom_layer_sizes, verbose, ctx):
+		from .sound_classifier import SoundClassifier
+		from .._mxnet import _mxnet_utils
+		
+		self.verbose = verbose
+		self.custom_NN = SoundClassifier._build_custom_neural_network(feature_extractor.output_length, num_labels, custom_layer_sizes)
```
nickjong: `_build_custom_neural_network` should be moved too, since it includes MXNet code

fareeha: Moved to where? It is moved from the main file to the SC MXNet class

fareeha: Ohhh do you mean move it from the CustomModel class over to this one?

nickjong: Yup, the long-term goal is to get all of the MXNet code out of the CustomModel class. For the first PR, my current guess would be to move everything except for the data iterator and metric

nickjong: (Since the data iterator and metric are compatible with a TF code path. The current implementation of `_build_custom_neural_network` is not compatible/usable by the TF code path

fareeha: Ahhh ok. Now I understand better what our plan is

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
nickjong: There is no strict need to implement `set_learning_rate` now anyway. We're not going to call this from C++ in the immediate future

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
nickjong: For now, since we are not using this from C++, I think we want to match the serialization format currently used by the sound classification toolkit

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -219,11 +232,8 @@ def create(dataset, target, feature, max_iterations=10,
                                     label=train_data['labels'].to_numpy(),
                                     batch_size=training_batch_size, shuffle=True)
 
-    custom_NN = SoundClassifier._build_custom_neural_network(feature_extractor.output_length, num_labels, custom_layer_sizes)
-    ctx = _mxnet_utils.get_mxnet_context()
-    custom_NN.initialize(mx.init.Xavier(), ctx=ctx)
-
-    trainer = mx.gluon.Trainer(custom_NN.collect_params(), 'nag', {'learning_rate': 0.01, 'momentum': 0.9})
+    from ._mx_sound_classifier import SoundClassifierMXNetModel
+    mxnet_model = SoundClassifierMXNetModel(feature_extractor, num_labels, custom_layer_sizes, verbose)
```
nickjong: Why do we need to pass the entire feature extractor?

nickjong: Also, let's not call this `mxnet_model`. The entire point of this is that this doesn't have to be MXNet.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -237,40 +247,25 @@ def create(dataset, target, feature, max_iterations=10,
...
-            trainer.step(batch.data[0].shape[0])
+            mxnet_model.train(batch)
         train_data.reset()
 
         # Calculate training metric
         for batch in train_data:
-            data = mx.gluon.utils.split_and_load(batch.data[0], ctx_list=ctx, batch_axis=0, even_split=False)
-            label = mx.gluon.utils.split_and_load(batch.label[0], ctx_list=ctx, batch_axis=0, even_split=False)
-            outputs = [custom_NN(x) for x in data]
+            data, label = mxnet_model.batch_process(batch)
```
nickjong: Let's move this inside `train` / `predict`. We don't want anything specific to MXNet backends here.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -289,7 +284,7 @@ def create(dataset, target, feature, max_iterations=10,
 
     state = {
         '_class_label_to_id': class_label_to_id,
-        '_custom_classifier': custom_NN,
+        '_custom_classifier': mxnet_model.custom_NN,
```
nickjong: This exposes MXNet to the toolkit. Let's just store the abstract wrapper class here, and then change all the places that use `_custom_classifier` to work with that

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Done

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: The reason I left this here was because we don't want to move metric in the mxnet class and `label` is used by `train_metric`

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Do you mean we pass object from the class here i.e. `mxnet_model` or the class itself i.e. `MultiLayerPerceptronMXNetModel()`?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Both good points. Changing now

fareeha:
Model has customizable layers and correctly trains, predicts, saves, and loads.

@nickjong  Currently we are not using `TF iterator` or `TF metric`. TF `evaluate` returns accuracy which is used for the progress table (MXNet metric implementation and TF `evaluate` method give matching values of accuracy).
 
TODO: 1) Numpy data iterator still needs shuffle. 2) Numpy data iterator in CustomModel predict() only gives one batch even when more are expected. Temporary patch is to use MXNet data iterator there instead.

With this current state the `test_audio_functionality` test gives:
```
turicreate/test/test_audio_functionality.py .........F.......F.......F........ 
```

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -105,14 +106,47 @@ def reset(self):
...
+        self.num_batches = int(_ceil(len(data) / float(self.batch_size)))
+
+    def __iter__(self):
+        if self.batch_idx < self.num_batches:
+            data = self.data_slices[0][self.batch_size*self.batch_idx:self.batch_size*(self.batch_idx+1)]
+            label = None
+            if len(self.data_slices)>1:
+                label = self.data_slices[1][self.batch_size*self.batch_idx:self.batch_size*(self.batch_idx+1)]
+            self.batch_idx += 1
+            yield data, label if label is not None else data
```
nickjong: Might need to return a tuple in the else case, `data,` instead of `data`

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -105,14 +106,47 @@ def reset(self):
...
+    def __init__(self, data, label=None, batch_size=1, shuffle=False):
+
+        # Always pass a tuple, so that the impl's built-in iterator returns a
+        # tuple.
+        self.data_slices = (data, label) if label is not None else (data,)
+        self.batch_size = batch_size
+        self.batch_idx = 0
+        self.num_batches = int(_ceil(len(data) / float(self.batch_size)))
+
+    def __iter__(self):
```
nickjong: This looks like an implementation of `__next__` for an iterator, not an implementation of `__iter__` itself...

Shouldn't `__iter__` just return an iterator? Or self if we're building the iterator logic into the main class?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -392,36 +433,22 @@ def create(dataset, target, feature, max_iterations=10,
...
-        for data, label in train_data:
-            custom_NN.train(data, label)
-        train_data.reset()
-
-        # Calculate training metric
         for data, label in train_data:
-            outputs = custom_NN.predict(data)
-            train_metric.update(label, outputs)
+            result = custom_NN.train(data, label)
+            train_accuracy = result['accuracy']
```
nickjong: How are we aggregating the accuracy across all the batches in the epoch?

fareeha: I replicated what we were doing previously in the MXNet case. i.e. `tf.reduce_means` returns the exact same output as `mx.metric.Accuracy()` on each iteration. i.e. the progress table is the same whether running the MXNet metric or TF's accuracy. 

If we want to aggregate accuracies I can add that feature.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: It automatically converts the return in the else case as a tuple. Which is why predict needs to extract `data[0]`

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -0,0 +1,205 @@
...
+        layer_dict = {}
+        for var, val in zip(layer_names, layer_weights):
+            layer_dict[var.name] = val
+
+        layers = []
+        for i, name in enumerate(self.names_of_layers):
+            weight_name = 'sound_{}_weight:0'.format(name)
+            bias_name = 'sound_{}_weight:0'.format(name)
+            layer={}
+            layer['weight'] = layer_dict[weight_name]#.transpose(1, 0)#.asnumpy() #### wait we don't need this!
```
TobyRoseman: Probably want to delete the commented out code here and on the next line.

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -0,0 +1,205 @@
...
+                self.sess.run(_tf.assign(_tf.get_default_graph().get_tensor_by_name(new_layer_name+":0"),
+                    net_params['data'][layer_name]))
+            else:
+                curr_shape = [int(x) for x in net_params['shapes'][layer_name]]
+                self.sess.run(_tf.assign(_tf.get_default_graph().get_tensor_by_name(new_layer_name+":0"),
+                                        net_params['data'][layer_name].reshape(curr_shape)))#.transpose(1, 0)))
+
+
+    def get_layer_activations(self, data, label):
+        #TODO: remove this
```
TobyRoseman: Can we remove it now?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -936,8 +962,11 @@ class as a vector. Label ordering is dictated by the ``classes``
             batch_size = len(deep_features)
 
         y = []
-        for data, in _create_data_iterator(deep_features['deep features'].to_numpy(), None, batch_size=batch_size):
-            y += self._custom_classifier.predict(data).tolist()
+        data_gen = _MXNetDataIterator(deep_features['deep features'].to_numpy(), None, batch_size=batch_size)
+
+        #for data in _create_data_iterator(deep_features['deep features'].to_numpy(), None, batch_size=batch_size):
```
TobyRoseman: Delete this comment out code too?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -105,14 +106,47 @@ def reset(self):
...
+        # this method because their __iter__ function returns the NDArrayIter
+        # instance itself.
+        self.batch_idx = 0
+
+
 def _create_data_iterator(data, label=None, batch_size=1, shuffle=False):
+
     if USE_TF:
-        return _TFDataIterator(data, label=label, batch_size=batch_size,
+        print("******** Using Numpy data iterator for TF")
```
srikris: Might want to remove this.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -374,14 +414,15 @@ def create(dataset, target, feature, max_iterations=10,
...
-    training_batch_size = min(len(train_data), batch_size)
-    train_data = _create_data_iterator(train_data['deep features'].to_numpy(),
-                                       train_data['labels'].to_numpy(),
-                                       batch_size=training_batch_size,
-                                       shuffle=True)
 
-    from ._mx_sound_classifier import MultiLayerPerceptronMXNetModel
-    custom_NN = MultiLayerPerceptronMXNetModel(feature_extractor.output_length, num_labels, custom_layer_sizes, verbose)
+    if USE_TF:
+        print("------ Using TensorFlow ...")
```
srikris: Does the user need to see this?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -11,6 +11,7 @@
 from __future__ import absolute_import as _
 
 import logging as _logging
+from math import ceil as _ceil
```
srikris: I don't see tests being un-marked as x-fail on this PR. Am I missing something?

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
fareeha: Yup, this is for debugging purposes.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Yes, debugging print statements will be removed.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: I will if I'm unable to fix the data iterator issue in CustomModel `predict()` within this PR.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Yes, great catch! Thank you

fareeha:
> Model has customizable layers and correctly trains, predicts, saves, and loads.
> 
> @nickjong Currently we are not using `TF iterator` or `TF metric`. TF `evaluate` returns accuracy which is used for the progress table (MXNet metric implementation and TF `evaluate` method give matching values of accuracy).
> 
> TODO: 1) Numpy data iterator still needs shuffle. 2) Numpy data iterator in CustomModel predict() only gives one batch even when more are expected. Temporary patch is to use MXNet data iterator there instead.
> 
> With this current state the `test_audio_functionality` test gives:
> 
> ```
> turicreate/test/test_audio_functionality.py .........F.......F.......F........ 
> ```

The data iterator works correctly now and the issue using data iterator in CustomModel predict() is resolved.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
nickjong: Yes, please adjust the tests so that after we merge, there are only `.` and `x`, not `X` or `F`

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
nickjong: In general I recommend committing all debugging code to a separate commit that you don't include in review branches. This keeps reviews a bit cleaner and unambiguous. There's some git tricks I can show you to make this easier....

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -174,9 +212,26 @@ def reset(self):
...
 
+class _NumPyAccuracy(_Accuracy):
+
+    def __init__(self):
+        self.acc = 0.0
+        self.num_batches = 0.0
+
+    def update(self, accuracy):
+        self.acc += accuracy
+        self.num_batches += 1.0
```
nickjong: This doesn't have the same interface as the other accuracy metrics. If the last batch is a different size, it doesn't weight the samples equally.

Please look at how the MXNet and TF versions are implemented. In practice, these effectively tally (a) how many correct examples there are and (b) how many total examples there are.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -405,14 +467,13 @@ def create(dataset, target, feature, max_iterations=10,
...
+            result = custom_NN.evaluate(data, label)
+            train_metric.update(result['accuracy'])
         train_data.reset()
 
-        # Calculate validation metric
         for data, label in validation_data:
-            outputs = custom_NN.predict(data)
-            validation_metric.update(label, outputs)
+            result = custom_NN.evaluate(data, label)
+            validation_metric.update(result['accuracy'])
```
nickjong: Could we stick with the old interface for the accuracy metric, which was chosen to avoid the kind of error you're introducing here?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -937,7 +997,7 @@ class as a vector. Label ordering is dictated by the ``classes``
 
         y = []
         for data, in _create_data_iterator(deep_features['deep features'].to_numpy(), None, batch_size=batch_size):
-            y += self._custom_classifier.predict(data).tolist()
+            y += self._custom_classifier.predict(data)[0].tolist()
```
nickjong: Why do we need to subscript here?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -405,14 +467,13 @@ def create(dataset, target, feature, max_iterations=10,
 
         # Calculate training metric
         for data, label in train_data:
-            outputs = custom_NN.predict(data)
-            train_metric.update(label, outputs)
+            result = custom_NN.evaluate(data, label)
```
nickjong: Why did we change from `predict` to `evaluate`? Doesn't that break compatibility with the MXNet implementation (which we haven't deleted yet?)

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Yes I changed the interface because I can fetch accuracy from TensorFlow by calling `self.accuracy` from the same session which is running `predict` on the batch. It's more cost effective then repeating the same computation outside in `sound_classifier.py`. 

Do we really care about matching the interface of the MXNet and TF metrics? When we are going to remove them anyway.

fareeha: I'll see if I can solve the unequal weigh problem, otherwise I'll switch over to matching interface with MXNet metric and manually compute accuracy.

nickjong: We want the computation of the metric to be correct, so if we change the interfaces we should be sure we understand precisely how they work and what we need to do to preserve correctness. I'm worried that we're creating more work by trying to reinvent interfaces than it would have taken just to follow them and build upon the prior effort to sort out correct patterns

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: To remove an extra outer bracket

nickjong: Could we just do this on the TF side? I guess I don't see why we should break compatibility with the MXNet implementation if we don't need to. If you think it makes more sense for predict to return a list, could we add a list to the MXNet side, to keep it in sync?

nickjong: But I think you should ask yourself why you think `predict` should return a list or whatever it is that we need to subscript

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: `predict` returns prediction probabilities. `evaluate` returns prediction probabilities and accuracies.

This will break with MXNet since we don't have `evaluate()` in `MultiLayerPerceptronMXNetModel()`

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: @nickjong yes please!

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: I wanted to fix what I can before marking xfail. Added `xfail`s in the PR

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Removed the subscript since the outer bracket was added because I was running `self.sess.run([self.predictions]` instead of `self.sess.run(self.predictions`. This is fixed now.

The list conversion is happening here to update the `y` list in this code. IMO we shouldn't change `predict` return type based on one specific usage. If we are doing `tolist()` in more than one place I would agree to change the return type of `predict`.

nickjong: Makes sense, I agree!

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Done

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
fareeha: Yes I agree. Thanks for pointing this out. Now my NumPy metric implementation matches MXNet implementation output exactly.

fareeha:
> Model has customizable layers and correctly trains, predicts, saves, and loads.
> 
> @nickjong Currently we are not using `TF iterator` or `TF metric`. TF `evaluate` returns accuracy which is used for the progress table (MXNet metric implementation and TF `evaluate` method give matching values of accuracy).
> 
> TODO: 1) Numpy data iterator still needs shuffle. 2) Numpy data iterator in CustomModel predict() only gives one batch even when more are expected. Temporary patch is to use MXNet data iterator there instead.
> 
> With this current state the `test_audio_functionality` test gives:
> 
> ```
> turicreate/test/test_audio_functionality.py .........F.......F.......F........ 
> ```

```
turicreate/test/test_audio_functionality.py .................................. 
```

NumPy data iterator fixed.

src/python/turicreate/test/test_audio_functionality.py
```diff
@@ -394,7 +387,6 @@ def test_model(self):
             self.assertNotEqual(r['probability'], None)
 
 
-@unittest.skipIf(_mac_ver() < (10,14), 'Custom models only supported on macOS 10.14+')
```
srikris: You need to unskip this.

src/python/turicreate/test/test_audio_functionality.py
```diff
@@ -224,7 +222,6 @@ def test_save_and_load(self):
         for a, b in zip(old_model_probs, new_model_probs):
             np.testing.assert_array_almost_equal(a, b, decimal=6)
 
-    @unittest.skipIf(_mac_ver() < (10,14), 'Custom models only supported on macOS 10.14+')
```
srikris: You need to revert this.

src/python/turicreate/test/test_audio_functionality.py
fareeha: Oh good catch! Thank you! done.

PR #2535, opened 2019-10-30
Sound Classifier: Feature Extraction using TensorFlow

TobyRoseman:


TobyRoseman:
> Since we're changing this unconditionally, should we now also make Turi Create formally depend on TensorFlow? (To the best of my knowledge, we haven't done this yet, right?)

@nickjong - That's a really good point. Yes, I believe we haven't done that yet. I was actually thinking of putting up a seperate PR tomorrow that removes the MXNet dependency and adds the TensorFlow dependency. I could make that part of this PR if you prefer.

nickjong:
Is it safe to depend on both temporarily, while we still have other toolkits that default to MXNet? If so, maybe we can have that as a separate PR that goes in before this one

TobyRoseman:
> Is it safe to depend on both temporarily, while we still have other toolkits that default to MXNet? If so, maybe we can have that as a separate PR that goes in before this one

It should be safe to have both TensorFlow and MXNet as dependencies; I smoked tested that a few weeks ago. 

Pull request to add TensorFlow: #2541

src/python/turicreate/toolkits/sound_classifier/_audio_feature_extractor.py
```diff
@@ -73,39 +60,13 @@ def _preprocess_data(audio_data, verbose=True):
...
         vggish_model_file = VGGish()
 
         if _mac_ver() < (10, 14):
-            # Use MXNet
-            model_path = vggish_model_file.get_model_path(format='mxnet')
-            self.vggish_model = VGGishFeatureExtractor._build_net()
-            net_params = self.vggish_model.collect_params()
-            self.ctx = _mxnet_utils.get_mxnet_context()
-            net_params.load(model_path, ctx=self.ctx)
+	    # Use TensorFlow/Keras
```
shantanuchhabra: Not a big deal but incorrectly indented comment here, I believe.

TobyRoseman: Thanks. Fixed.

srikris:
@TobyRoseman Seems like you might want to rebase against master.

TobyRoseman:
> @TobyRoseman Seems like you might want to rebase against master.

Thanks. Done.

srikris:
@TobyRoseman Seems like there are build failures unrelated to this. Should be good to merge.

PR #2537, opened 2019-10-30
OD iterator fixed: missing annotations and invalid labels.

jakesabathia2:
Modify OD data iterator in order to read data with no annotations.
And also skip the data which has invalid labels.

### DIFF
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -980,8 +980,7 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
   // Check if data has annotations column
   std::string annotations_column_name = read_state<flex_string>("annotations");
   if (data.contains_column(annotations_column_name)) {
-    iterator_params.annotations_column_name =
-        read_state<flex_string>("annotations");
+    iterator_params.annotations_column_name = annotations_column_name;
   }
 
   iterator_params.data = std::move(data);
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -635,7 +635,6 @@ variant_type object_detector::predict(
   } else {
     final_result = to_variant(result_sarray[0]);
   }
-
   return final_result;
 }
 
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -635,6 +635,7 @@ variant_type object_detector::predict(
   } else {
     final_result = to_variant(result_sarray[0]);
   }
+
   return final_result;
 }
 
@@ -976,9 +977,15 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
     gl_sframe data, std::vector<std::string> class_labels, bool repeat) const
 {
   data_iterator::parameters iterator_params;
+
+  // Check if data has annotations column
+  std::string annotations_column_name = read_state<flex_string>("annotations");
+  if (data.contains_column(annotations_column_name)) {
+    iterator_params.annotations_column_name =
+        read_state<flex_string>("annotations");
+  }
+
   iterator_params.data = std::move(data);
-  iterator_params.annotations_column_name =
-      read_state<flex_string>("annotations");
   iterator_params.image_column_name = read_state<flex_string>("feature");
   iterator_params.class_labels = std::move(class_labels);
   iterator_params.repeat = repeat;
--- a/src/toolkits/object_detection/od_data_iterator.cpp
+++ b/src/toolkits/object_detection/od_data_iterator.cpp
@@ -218,8 +218,6 @@ std::pair<gl_sarray, size_t> get_annotation_info(const gl_sarray& annotations) {
 
   // Construct an SFrame with one row per bounding box.
   gl_sframe instances;
-  gl_sarray classes;
-  size_t num_instances;
 
   if (annotations.dtype() == flex_type_enum::LIST) {
     gl_sframe unstacked_instances({{"annotations", annotations}});
@@ -234,13 +232,7 @@ std::pair<gl_sarray, size_t> get_annotation_info(const gl_sarray& annotations) {
                                {flex_type_enum::STRING},
                                /* na_value */ FLEX_UNDEFINED, {"label"});
 
-  // Determine the list of unique class labels,
-  classes = instances["label"].unique().sort();
-
-  // Record the number of labeled bounding boxes.
-  num_instances = instances.size();
-
-  return std::make_pair(classes, num_instances);
+  return std::make_pair(instances["label"].unique().sort(), instances.size());
 }
 
 }  // namespace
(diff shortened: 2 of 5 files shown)


src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -39,14 +38,18 @@ gl_sframe get_data(const data_iterator::parameters& params) {
...
   if (!params.predictions_column_name.empty()) {
     result[params.predictions_column_name] =
         params.data[params.predictions_column_name];
   }
 
+  if (!params.annotations_column_name.empty()) {
+    result[params.annotations_column_name] =
+        params.data[params.annotations_column_name];
+  }
+
```
nickjong: Nit: let's handle the annotations case before the predictions case, to parallel the order in the parameters (and the order of importance)

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -73,10 +75,12 @@ std::vector<image_annotation> parse_annotations(
       const flex_string& key = kv.first.get<flex_string>();
 
       if (key == "label") {
-
-        annotation.identifier =
-            class_to_index_map.at(kv.second.get<flex_string>());
-        has_label = true;
+        // If the label in invalid (not in class_to_index_map) then ignore it.
```
nickjong: s/in invalid/is invalid/

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -244,6 +253,13 @@ simple_data_iterator::compute_properties(
       result.class_to_index_map[label] = i++;
     }
   } else {
+    // Make sure the expected labels are unique and sorted.
+    std::sort(expected_class_labels.begin(), expected_class_labels.end());
+    std::vector<std::string>::iterator it;
+    it =
+        std::unique(expected_class_labels.begin(), expected_class_labels.end());
+    expected_class_labels.resize(
+        std::distance(expected_class_labels.begin(), it));
```
nickjong: Why do the labels need to be sorted? The developer can't choose what label is first?

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -215,24 +219,29 @@ simple_data_iterator::compute_properties(
...
-                               { flex_type_enum::STRING },
-                               /* na_value */ FLEX_UNDEFINED, { "label" });
+    // Determine the list of unique class labels,
+    classes = instances["label"].unique().sort();
 
-  // Determine the list of unique class labels,
-  gl_sarray classes = instances["label"].unique().sort();
+    // Record the number of labeled bounding boxes.
+    result.num_instances = instances.size();
+  }
```
nickjong: Could we extract this into a helper function that, given the annotations SArray, returns the list of unique labels and the number of bounding boxes?

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -307,14 +320,16 @@ std::vector<labeled_image> simple_data_iterator::next_batch(size_t batch_size) {
...
+    flexible_type annotations = FLEX_UNDEFINED;
     if (predictions_index_ >= 0) {
       preds = row[predictions_index_];
     }
-    raw_batch.emplace_back(row[image_index_], row[annotations_index_], preds);
-
+    if (annotations_index_ >= 0) {
+      annotations = row[annotations_index_];
+    }
+    raw_batch.emplace_back(row[image_index_], annotations, preds);
```
nickjong: Nit: here, too, let's match the order of the fields in the struct

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -976,9 +976,15 @@ std::unique_ptr<data_iterator> object_detector::create_iterator(
     gl_sframe data, std::vector<std::string> class_labels, bool repeat) const
 {
   data_iterator::parameters iterator_params;
+
+  // Check if data has annotations column
+  std::string annotations_column_name = read_state<flex_string>("annotations");
+  if (data.contains_column(annotations_column_name)) {
+    iterator_params.annotations_column_name =
+        read_state<flex_string>("annotations");
```
nickjong: Reuse `annotations_column_name` here

src/toolkits/object_detection/od_data_iterator.cpp
jakesabathia2: okay

src/toolkits/object_detection/od_data_iterator.cpp
jakesabathia2: I don't know, I follow in implementation in Python side:
https://github.com/apple/turicreate/blob/62ea3f7276ec987ce94956ef52ee217f407d4e33/src/python/turicreate/toolkits/object_detector/object_detector.py#L333

And it seems like in the python side they already sort it and take the unique value of it?
Do I need to do it again in the C++ side?

src/toolkits/object_detection/od_data_iterator.cpp
jakesabathia2: sounds like a good idea

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: Oh yes I forgot :(

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -207,17 +211,16 @@ std::vector<image_annotation> parse_annotations(
...
-  annotation_properties result;
+std::pair<gl_sarray, size_t> get_annotation_info(const gl_sarray& annotations) {
+  if (annotations.size() == 0) {
+    return std::make_pair(gl_sarray(), 0);
+  }
 
   // Construct an SFrame with one row per bounding box.
   gl_sframe instances;
+  gl_sarray classes;
+  size_t num_instances;
```
nickjong: Why separate the declaration of these variables from their initialization?

src/toolkits/object_detection/od_data_iterator.cpp
nickjong: Let's not duplicate that work for now

src/toolkits/object_detection/od_data_iterator.hpp
```diff
@@ -172,7 +172,7 @@ class simple_data_iterator: public data_iterator {
       std::vector<std::string> expected_class_labels);
 
   gl_sframe data_;
-  const size_t annotations_index_;
+  const ssize_t annotations_index_;
```
Jarvi-Izana: good!

src/toolkits/object_detection/od_data_iterator.cpp
jakesabathia2: hmmmm ill fix this

test/unity/toolkits/object_detection/test_od_data_iterator.cxx
```diff
@@ -282,8 +282,9 @@ BOOST_AUTO_TEST_CASE(test_simple_data_iterator_with_unexpected_classes) {
   params.class_labels = { "bar" };
 
   // The data contains the label "foo", which is not among the expected class
-  // labels.
-  TS_ASSERT_THROWS_ANYTHING(simple_data_iterator unused_var(params));
+  // labels. SHOULDN'T throw anything.
+  simple_data_iterator it_unexpected(params);
+  TS_ASSERT_EQUALS(it_unexpected.class_labels().size(), 1);
```
shantanuchhabra: What should the one element inside of `it_unexpected.class_labels()`? Can we assert that it is actually there?

PR #2538, opened 2019-10-30
Fix the test case for OD test_get

jakesabathia2:
Fix the 'training_average_precision' to dictionary type.

### DIFF
--- a/src/python/turicreate/test/test_object_detector.py
+++ b/src/python/turicreate/test/test_object_detector.py
@@ -150,7 +150,9 @@ class ObjectDetectorTest(unittest.TestCase):
             self.get_ans['grid_width'] = lambda x: x > 0
             self.get_ans['training_mean_average_precision'] = lambda x: True
             self.get_ans['random_seed'] = lambda x: True
-            self.get_ans['training_average_precision'] = lambda x: isinstance(x, dict)
+            self.get_ans['training_average_precision'] = lambda x: (isinstance(x, dict) and
+                set(x.keys()) == set(_CLASSES) and all([isinstance(y,float) for y in x.values()])
+            )
             del self.get_ans['_model']
             del self.get_ans['_class_to_index']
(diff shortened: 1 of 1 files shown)


src/python/turicreate/test/test_object_detector.py
```diff
@@ -150,7 +150,7 @@ def setUpClass(self):
             self.get_ans['grid_width'] = lambda x: x > 0
             self.get_ans['training_mean_average_precision'] = lambda x: True
             self.get_ans['random_seed'] = lambda x: True
-            self.get_ans['training_average_precision'] = lambda x: x >= 0
+            self.get_ans['training_average_precision'] = lambda x: isinstance(x, dict)
```
Jarvi-Izana: give any explanation? the 2 lambdas are functionally different.

jakesabathia2: Yes, it seems like the correct type of x should be a dictionary instead of a single value.

jakesabathia2: 'training_average_precision' contains ap for each class (dict).
'training_mean_average_precision' is the mean of each ap which should be a single value.

nickjong: Could we also check that the keys are the class labels, and that the values are numeric?

Jarvi-Izana: so you mean the data representation is wrong. but how could a `dict` compare with int? How could it pass the test pipeline? What happens in our test pipeline? Do we need to hire `the real ghostbusters` to clear the ghost in our test machines? Should I say happy Halloween?
```
>>> dict() > 0
TypeError: '>' not supported between instances of 'dict' and 'int'
```

jakesabathia2: sound like a good idea will add it

PR #2541, opened 2019-10-31
Add TensorFlow Dependency + Remove MXNet Dependency

TobyRoseman:


shreyajain17:
Are we going to wait till we take MXNet out of our requirements or we are going to just have both from now on?

TobyRoseman:
> Are we going to wait till we take MXNet out of our requirements or we are going to just have both from now on?

We'll have both TensorFlow and MXNet as dependencies for a short period of time before we ship.

nickjong:
Damn, I remember seeing that Travis failure the last time we looked at this. I believe trying to load both MXNet and TF is exhausting some limit on the amount of thread-local storage being used by Turi Create (and its dependencies). Unless there's an easy workaround, it might be simplest to switch all in one go....

PR #2542, opened 2019-10-31
Add OD input classes argument.

jakesabathia2:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -1069,8 +1075,11 @@ void object_detector::init_training(gl_sframe data,
                                          read_state<int>("random_seed"));
 
   // Bind the data to a data iterator.
-  training_data_iterator_ = create_iterator(
-      training_data_, /* expected class_labels */ {}, /* repeat */ true);
+  std::vector<std::string> class_labels =
+      read_state<std::vector<std::string>>("classes");
+  training_data_iterator_ =
```
Jarvi-Izana: should it be flex_list? or it can implicitly convert to std::vector<std::string>?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -269,6 +269,12 @@ void object_detector::init_options(
       /* default_value     */ "center",
       /* allowed_values    */ {flexible_type("center"), flexible_type("top_left"), flexible_type("bottom_left")},
```
Jarvi-Izana: you can run formatter on this line as well.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -269,6 +269,12 @@ void object_detector::init_options(
...
       /* allowed_overwrite */ false);
+  options.create_flexible_type_option(
+      /* name              */ "classes",
+      /* description       */
```
Jarvi-Izana: better put in one liner.

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: I have tried to get<flex_List> and then construct a vector<string> from it.
Either method works here,
so I pick a concise one.

Jarvi-Izana: That's really weird. Probably some conversion. I cannot assign vec<str> to vec<flexible>

Jarvi-Izana: `vec str` to `vec flexible_type`

jakesabathia2: I guess I mean
```
flex_list flex_classes = read_state<flex_list>("classes");
std::vector<std::string> class_labels (flex_classes.begin(), flex_classes.end());
```

jakesabathia2: Yes I think it does some conversion :P

Jarvi-Izana:
I think we should warn the user if the `class` they pass in is not contained in the training labels (inferred unique labels).

If you have annotation [dog, cat, bird], and you set `class` to [catt], due to a typo, you should warn them early. 

Also, if you pass [horse], it will learn nothing and you should let the user know in advance that the model won't work for horse detection. And they will realize it's inappropriate to request od to do something it can't do.

src/toolkits/object_detection/object_detector.cpp
Jarvi-Izana: oh well, it must do.

I mean
```
vector<string> blah = {"1", "1"};
flex_list foo = blah; // will error out
```

jakesabathia2:
=

PR #2543, opened 2019-10-31
Fixes Predict for Style Transfer C++ implementation 

shreyajain17:
It fixes predict in C++ which was breaking. 
It also has the fixes for two unit tests on python side. 
I also used the PR to add the frontend for `get_styles()`.
I have also passed all the input arguments from python to C++. All arguments passed to C++ don't function correctly though.


nickjong:
In the future, please err on the side of not putting unrelated changes into the same PR.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -268,7 +268,10 @@ void style_transfer::init_options(
   options.create_integer_option(
       "num_styles", "The number of styles present in the model", FLEX_UNDEFINED,
       1, std::numeric_limits<int>::max());
-
+  options.create_boolean_option(
+      "verbose", "When set to true, verbose is printed", true, true);
+  options.create_string_option("content_feature", "Name of the content column", "image", true);
+  options.create_string_option("style_feature", "Name of the style column", "image", true);
```
nickjong: Why are these needed? I thought we were passing the data as two SArrays now, not as SFrames with column names.

What happens if both of these are left as the default? Isn't that a namespace collision of sorts?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -367,9 +370,26 @@ gl_sframe style_transfer::predict(variant_type data,
       case flex_type_enum::VECTOR:
         style_idx = std::move(flex_style_idx.get<flex_vec>());
         break;
+      case flex_type_enum::LIST:
+      {
+        auto list = flex_style_idx.get<flex_list>();
```
nickjong: Could we use `const auto&` to avoid a copy?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -367,9 +370,26 @@ gl_sframe style_transfer::predict(variant_type data,
...
+        style_idx.resize(list.size());
+        std::transform(list.begin(), list.end(),
+                   std::back_inserter(style_idx), [](flexible_type val) {
+                     return static_cast<double>(val.get<flex_float>());
+                   });
```
nickjong: Why do we want to append to `style_idx` after resizing to `list.size()`? The result will be a sequence twice as long as `list`, where the first half of the elements are indeterminate

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -367,9 +370,26 @@ gl_sframe style_transfer::predict(variant_type data,
...
+        break;
+
+      }
+      case flex_type_enum::UNDEFINED: {
+        int num_styles = read_state<flex_int>("num_styles");
```
nickjong: The type of the variable should be consistent with the type used to instantiate the `read_state` template

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -382,10 +402,10 @@ void style_transfer::perform_predict(gl_sarray data, gl_sframe_writer& result,
...
   if (data.size() == 0) return;
 
-  flex_int batch_size = read_state<flex_int>("batch_size");
-  flex_int num_styles = read_state<flex_int>("num_styles");
-  flex_int image_width = read_state<flex_int>("image_width");
-  flex_int image_height = read_state<flex_int>("image_height");
+  int batch_size = read_state<flex_int>("batch_size");
+  int num_styles = read_state<flex_int>("num_styles");
+  int image_width = read_state<flex_int>("image_width");
+  int image_height = read_state<flex_int>("image_height");
```
nickjong: Why are these being changed? Before it was consistent, now it is inconsistent

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: This is for the python side where the api still takes SFrames. Should we modify the api in Python as well?

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Good catch! Just read up on `back_inserter` didn't know it did that. Would:

```cpp
style_idx.resize(list.size());
std::transform(list.begin(), list.end(), style_idx.begin(), [](flexible_type val) {
   return static_cast<double>(val.get<flex_float>());
});
```

be better than:

```cpp
std::transform(list.begin(), list.end(), std::back_inserter(style_idx), [](flexible_type val) {
   return static_cast<double>(val.get<flex_float>());
});
```

In terms of mallocs and general guidelines?

nickjong: For the latter, you would want to `clear()` and possibly `reserve()` first. The former probably makes more sense in this case, since resizing is cheap for simple types. (Where the default constructor is cheap).

src/toolkits/style_transfer/style_transfer.cpp
shreyajain17: > In the future, please err on the side of not putting unrelated changes into the same PR.

I will take care next time.

src/toolkits/style_transfer/style_transfer.cpp
shreyajain17: Python frontend is not changed so that still accepts SFrames for content and style. I am passing the relevant column as SArray to C++ but then since user passed content/style as SFrame, they would want to know the column names after loading the model.

PR #2544, opened 2019-10-31
OD export_coreml breaks fixed

jakesabathia2:


src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -140,9 +141,9 @@ void set_image_feature_size_range(ImageFeatureType* image_feature,
 
 ImageFeatureType* set_image_feature(
     FeatureDescription* feature_desc, size_t image_width, size_t image_height,
-    std::string description = "",
+    std::string input_name, std::string description = "",
     ImageFeatureType::ColorSpace image_type = ImageFeatureType::RGB) {
-  feature_desc->set_name("image");
+  feature_desc->set_name(input_name);
```
Jarvi-Izana: probably I miss some context here. is the input name passed by user side?

jakesabathia2: the input_name will be inferred from the column of the feature column of the dataset.
It is not included in arguments of export_to_coreml() so the users don't pass it.

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -185,20 +186,30 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
   model_nn->mutable_neuralnetwork()->MergeFrom(nn_spec.get_coreml_spec());
   ASSERT_GT(model_nn->neuralnetwork().layers_size(), 1);
 
+  // Set the name of preprocessing layer to input_name as well.
+  // Since in the test case somehow the model doesn't have a preprocessing
+  // layer, so we need to check the size first...
+  if (model_nn->mutable_neuralnetwork()->mutable_preprocessing()->size() > 0) {
```
nickjong: Do not use `mutable` accessors here, since they will create these fields if they didn't exist before. Just replace e.g. `mutable_neuralnetwork()->` with `neuralnetwork().`

jakesabathia2: This is a really good point. Thanks!

PR #2546, opened 2019-10-31
Ac break fixed

jakesabathia2:
This PR fixed all AC breaks except the ones relative to export_coreml.


src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -259,6 +240,27 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
...
+
+        model = _turicreate.extensions.activity_classifier()
+        options = {}
+        options['prediction_window'] = prediction_window
+        options['batch_size'] = batch_size
+        options['max_iterations'] = max_iterations
+        options['verbose'] = verbose
+
+        model.train(dataset, target, session_id, validation_set, options)
+        return ActivityClassifier_beta(model_proxy=model, name=name)
```
nickjong: Why did this move?

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -549,6 +551,103 @@ def evaluate(self, dataset, metric='auto'):
         """
         return self.__proxy__.evaluate(dataset, metric)
 
+    def predict_topk(self, dataset, output_type='probability', k=3, output_frequency='per_row'):
```
nickjong: This looks like an implementation of `predict_topk`, not a fix for a breakage. What was broken?

jakesabathia2: The unit test test_predict_topk for AC was throwing en error : 
AttributeError: 'ActivityClassifier' object has no attribute 'predict_topk'.
Sl the fix is just to add the function to the beta class I guess.

jakesabathia2: Same for the test_classify as well.

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
jakesabathia2: hmmmm actually I changed the move 
the failing of `test_create_missing_value` can be solved by 
https://github.com/apple/turicreate/blob/d18f03a04eeeb9c3de71a134d1a8682e6c86f766/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py#L191

and the breaks which relative to auto split train/val data `test_create_auto_validation_set_small`, `test_create_auto_validation_set_string_session_id`, `test_create_auto_validation_set_typical` are solved in 
https://github.com/apple/turicreate/blob/d18f03a04eeeb9c3de71a134d1a8682e6c86f766/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py#L179

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -176,6 +176,23 @@ def create(dataset, session_id, target, features=None, prediction_window=100,
             raise _ToolkitError('Unknown advanced parameters: {}'.format(unsupported))
         params.update(kwargs['_advanced_parameters'])
 
+    if isinstance(validation_set, str) and validation_set == 'auto':
```
shreyajain17: Is this for an auto split? C++ has that functionality.

jakesabathia2: Yes it does, thanks for pointing it out.
The thing is it seems like the implementation might be different?
@nickjong is it better to done it in Python side or C++ for this case :).

shreyajain17: It is better to use C++ as far as I think. The minimum sessions needed for split are different. In python, it is 100 while in C++ it is 50. I think you would have to modify the tests to accommodate this change.

jakesabathia2:
hmmmmm So it turns out the failure it not due to the implementation of random_split hmmmmm.
So will just leave the random_split and the test case the same,
and fix the bugs.

jakesabathia2:
Edit:
So the failures is caused by the random_split ratio and a bug in c++ hmmm.
1. change the expected_ratio in test case according to USE_CPP flag.
2. model.num_sessions should return the number of session of the 'training' data, in the previous implementation this value is inferred from all data including train and val, which cause the test failure.

PR #2548, opened 2019-11-01
Appending Empty SArray Should Ignore Type

### LINKED ISSUE #2390: Appending Empty SArray Should Ignore Type
Constructing new SArrays and appending them to an existing SArray is a very common usage pattern.  Sometimes these new SArrays end up being empty, in which case the type is assumed to be float. This causes a problem if the SArray being appending to is not also of type float.

```python

In [2]: existing_sa = tc.SArray([1,2,3])                                                                                                                                                                                                                                                                                                                                  

In [3]: new_sa = tc.SArray()                                                                                                                                                                                                                                                                                                                                              

In [4]: existing_sa.append(new_sa)                                                                                                                                                                                                                                                                                                                                        
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-cd34ef9a601c> in <module>
----> 1 existing_sa.append(new_sa)

~/anaconda3/envs/prod-py36/lib/python3.6/site-packages/turicreate/data_structures/sarray.py in append(self, other)
   2891 
   2892         if self.dtype != other.dtype:
-> 2893             raise RuntimeError("Data types in both SArrays have to be the same")
   2894 
   2895         with cython_context():

RuntimeError: Data types in both SArrays have to be the same
```

I suggest we ignore types when appending an empty SArray. It should just be a no-op.

dhivyaaxim:
This PR resolves #2390 
Appending Empty SArray Should Ignore Type

### DIFF
--- a/src/python/turicreate/data_structures/sarray.py
+++ b/src/python/turicreate/data_structures/sarray.py
@@ -2902,17 +2902,19 @@ class SArray(object):
         Rows: 6
         [1, 2, 3, 4, 5, 6]
         """
-        if len(other) == 0:
-            other=other.astype(self.dtype)
-
-        if len(self) == 0:
-            self=self.astype(other.dtype)
 
         if type(other) is not SArray:
             raise RuntimeError("SArray append can only work with SArray")
 
         if self.dtype != other.dtype:
-            raise RuntimeError("Data types in both SArrays have to be the same")
+            if len(other) == 0:
+                other=other.astype(self.dtype)
+
+            elif len(self) == 0:
+                self=self.astype(other.dtype)
+
+            else:
+                raise RuntimeError("Data types in both SArrays have to be the same")
 
         with cython_context():
             return SArray(_proxy = self.__proxy__.append(other.__proxy__))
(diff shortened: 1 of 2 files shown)


TobyRoseman:
Thanks for fixing this.

Please write unit tests for both cases (appending to an empty SArray and appending an empty SArray).

src/python/turicreate/data_structures/sarray.py
```diff
@@ -2902,6 +2902,12 @@ def append(self, other):
         Rows: 6
         [1, 2, 3, 4, 5, 6]
         """
+        if len(other) == 0:
```
hoytak: Let's move these under the the  `if self.dtype != other.dtype:` statement.  The reason for this is that len(...) may trigger the materialization of an unmaterialized SArray, potentially expensive operation.  Thus let's only do it if we have to.

dhivyaaxim: @hoytak  @TobyRoseman  Please resolve the conflicting files issues in master
 File name : src/python/turicreate/test/test_sarray.py .

PR #2549, opened 2019-11-01
Moved MXNet implementation from Sound Classifier into a new class

fareeha:
Removed MXNet impl from Sound Classifier `create()` and `CustomModel` class into an MXNet specific class `MultiLayerPerceptronMXNetModel`. The changes also support save and load from old impl to the new one. 

Remaining MXNet not replaced yet are: MXNet data iterator and metric calculator.

### DIFF
--- a/src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
+++ b/src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2019 Apple Inc. All rights reserved.
+#
+# Use of this source code is governed by a BSD-3-clause license that can
+# be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+
+
+import mxnet as mx
+import time
+import turicreate as _tc
+
+from .._mxnet import _mxnet_utils
+
+class MultiLayerPerceptronMXNetModel():
+
+
+	def __init__(self, feature_output_length, num_labels, custom_layer_sizes, verbose):
+
+		self.ctx = _mxnet_utils.get_mxnet_context()
+		self.verbose = verbose
+		self.custom_NN = self._build_custom_neural_network(feature_output_length, num_labels, custom_layer_sizes)
+		self.custom_NN.initialize(mx.init.Xavier(), ctx=self.ctx)
+
+		self.trainer = mx.gluon.Trainer(self.custom_NN.collect_params(), 'nag', {'learning_rate': 0.01, 'momentum': 0.9})
+		self.softmax_cross_entropy_loss = mx.gluon.loss.SoftmaxCrossEntropyLoss()
+
+
+	def train(self, data, label):
+		# Inside training scope
+		data_shape = data.shape[0]
+		data = mx.gluon.utils.split_and_load(data, ctx_list=self.ctx, batch_axis=0, even_split=False)
+		label = mx.gluon.utils.split_and_load(label, ctx_list=self.ctx, batch_axis=0, even_split=False)
+		with mx.autograd.record():
+		    for x, y in zip(data, label):
+		        z = self.custom_NN(x)
+		        # Computes softmax cross entropy loss.
+		        loss = self.softmax_cross_entropy_loss(z, y)
+		        # Backpropagate the error for one iteration.
+		        loss.backward()
+		# Make one step of parameter update. Trainer needs to know the
+		# batch size of data to normalize the gradient by 1/batch_size.
+		self.trainer.step(data_shape)
+
+
+	def predict(self, data):
+		data = mx.gluon.utils.split_and_load(data, ctx_list=self.ctx, batch_axis=0, even_split=False)
+		outputs = [self.custom_NN(x).asnumpy() for x in data]
+		return outputs
+
+	@staticmethod
+	def _build_custom_neural_network(num_inputs, num_labels, layer_sizes):
+	    from mxnet.gluon import nn
+
+	    net = nn.Sequential(prefix='custom_')
+	    with net.name_scope():
+	        for i, cur_layer_size in enumerate(layer_sizes):
+	            prefix = "dense%d_" % i
+	            if i == 0:
+	                in_units = num_inputs
+	            else:
+	                in_units = layer_sizes[i-1]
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
```diff
@@ -0,0 +1,84 @@
...
+import mxnet as mx
+import time
+import turicreate as _tc
+
+from .._mxnet import _mxnet_utils
+
+class MultiLayerPerceptronMXNetModel():
+
+
+	def __init__(self, feature_output_length, num_labels, custom_layer_sizes, verbose):
```
nickjong: Please don't use tabs for indentation. Let's follow the conventions elsewhere in our codebase, and use four spaces for indentation in Python code

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
```diff
@@ -0,0 +1,84 @@
...
+		self.ctx = _mxnet_utils.get_mxnet_context()
+		self.verbose = verbose
+		self.custom_NN = self._build_custom_neural_network(feature_output_length, num_labels, custom_layer_sizes)
+		self.custom_NN.initialize(mx.init.Xavier(), ctx=self.ctx)
+
+		self.trainer = mx.gluon.Trainer(self.custom_NN.collect_params(), 'nag', {'learning_rate': 0.01, 'momentum': 0.9})
+		self.softmax_cross_entropy_loss = mx.gluon.loss.SoftmaxCrossEntropyLoss()
+
+
+	def train(self, data, label):
```
nickjong: Could we document the expected types and shapes of these arguments? And for the other methods, too?

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
```diff
@@ -0,0 +1,84 @@
...
+		# Inside training scope
+		data_shape = data.shape[0]
```
nickjong: What is a typical value of `data_shape`? Isn't this really just the batch size?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -289,7 +273,7 @@ def create(dataset, target, feature, max_iterations=10,
 
     state = {
         '_class_label_to_id': class_label_to_id,
-        '_custom_classifier': custom_NN,
+        '_custom_classifier': sc_model,
```
TobyRoseman: I would prefer to keep the variable name `custom_NN`. I think that's a more descriptive name. It also matches language used in the user guide.

fareeha: We refer to our Sequential model as `custom_NN` in the MXNet class so I didn't want to make it confusing. Any thoughts @nickjong ?

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
fareeha: It can be <=`batch_size` depending on the size of data passed. If input data is less than batch size, `data_shape` would have a value less than batch size.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
nickjong: I think `custom_NN` probably is the clearer name, since this object only contains the neural network classifier, not the audio preprocessing and feature extraction that are also part of the overall Sound Classifier model

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
```diff
@@ -0,0 +1,108 @@
...
+        with mx.autograd.record():
+            for x, y in zip(data, label):
+                z = self.custom_NN(x)
+                # Computes softmax cross entropy loss.
+                loss = self.softmax_cross_entropy_loss(z, y)
+                # Backpropagate the error for one iteration.
+                loss.backward()
+        # Make one step of parameter update. Trainer needs to know the
+        # batch size of data to normalize the gradient by 1/batch_size.
+        self.trainer.step(data_shape)
```
nickjong: This comment describes the information being passed to `self.trainer.step` as the batch size. Why not just call the variable `batch_size`, since by construction we know it is a scalar value, the first element of the data's shape?

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
```diff
@@ -0,0 +1,108 @@
...
+
+        self.trainer = mx.gluon.Trainer(self.custom_NN.collect_params(), 'nag', {'learning_rate': 0.01, 'momentum': 0.9})
+        self.softmax_cross_entropy_loss = mx.gluon.loss.SoftmaxCrossEntropyLoss()
+
+
+    def train(self, data, label):
+        """
+        Parameters
+        ----------
+        data : NumPy Array
```
nickjong: Could you document the meaning of the parameter, not just the expected type?

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -825,11 +788,8 @@ class as a vector. Label ordering is dictated by the ``classes``
             if batch.pad != 0:
                 batch_data = batch_data[:-batch.pad]    # prevent batches looping back
 
-            batch_data = mx.gluon.utils.split_and_load(batch_data, ctx_list=ctx, batch_axis=0, even_split=False)
-
-            for x in batch_data:
-                forward_output = self._custom_classifier.forward(x)
-                y += mx.nd.softmax(forward_output).asnumpy().tolist()
+            forward_output = self._custom_classifier.predict(batch_data)
+            y += mx.nd.softmax(mx.nd.array(forward_output[0])).asnumpy().tolist()
```
nickjong: This leaves some MXNet code outside of the MXNet encapsulation. Could we move the application of `softmax` inside `predict`?

src/python/turicreate/toolkits/sound_classifier/_mx_sound_classifier.py
fareeha: Sure I can make it `batch_size = min(batch_size, data.shape[0])` since `data.shape[0]` can be smaller than batch_size if examples it holds are fewer.

PR #2552, opened 2019-11-01
Added stroke support for predict

shantanuchhabra:
Still need to make the unit tests work for the `is_bitmap_based` option set to `false`. That will come in a different PR.


src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -661,6 +661,23 @@ gl_sframe drawing_classifier::perform_inference(data_iterator* data) const {
   return writer.close();
 }
 
+gl_sframe infer_feature_column_and_prepare_data(gl_sframe &data) {
```
nickjong: Please move free (non-class) helper functions to the top of the file and put them in an anonymous namespace (or otherwise specify only local linkage)

Also, prefer not to mutate arguments. Could this return a new SFrame, without mutating the input?

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -661,6 +661,23 @@ gl_sframe drawing_classifier::perform_inference(data_iterator* data) const {
...
+  std::string feature_column_name;
+
+  for (std::string column_name: data.column_names()) {
+    if (data[column_name].dtype() == flex_type_enum::LIST
+      || data[column_name].dtype() == flex_type_enum::IMAGE) {
+      feature_column_name = column_name;
```
nickjong: What happens if there's more than one matching column? (What do other toolkits do in similar scenarios?)

shantanuchhabra: Good point. Following what we were doing earlier in https://github.com/apple/turicreate/blob/master/src/python/turicreate/toolkits/_internal_utils.py#L122, looks like we error out in all scenarios except when there is exactly one candidate feature column.

However, looking at the `predict` docstring, I see:
```
            If dataset is an SFrame, it must have a column with the same name
            as the feature column during training. Additional columns are
            ignored.
```

So should we save the `feature_column_name_` as a private/public variable during training and just make sure the data passed in for testing has that same column? It does bring into question what our strategy would be if the user passed in an SArray. In order to make a `predict` call from the Python layer, we would want to convert the SArray to an SFrame at the Python layer and hence, know the feature column name at the Python layer. Alternatively, we could have a check in the C++ layer for a one-column SFrame in which case we rename the SFrame before continuing further processing.

What are your thoughts @nickjong ?

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Done.

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: Let's have the final result match the behavior we had in the original implementation. It's pretty typical to expect the prediction SFrame to use the same column name as the training data.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
```diff
@@ -261,31 +262,34 @@ BOOST_AUTO_TEST_CASE(test_drawing_classifier_perform_inference) {
   log_for_debug("test_drawing_classifier_perform_inference");
 
   for (auto& entry : TEST_CASES) {
-    size_t batch_size = std::get<0>(entry);
-    size_t num_of_rows = std::get<1>(entry);
-    size_t num_of_classes = std::get<2>(entry);
+    for (bool is_bitmap_based : {true}) {
```
Jarvi-Izana: you can edit the `TEST_CASES` to have a tuple of 4 elements.

Jarvi-Izana: this line is a bad practice since it removes the non-bitmap-based test.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
```diff
@@ -316,7 +320,8 @@ BOOST_AUTO_TEST_CASE(test_drawing_classifier_predict_rank) {
       }
     };
 
-    prediction_test_driver(batch_size, num_of_rows, num_of_classes, runner);
+    prediction_test_driver(/* is_bitmap_based */ true,
```
Jarvi-Izana: why you need to hard code true?

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
Jarvi-Izana: to make myself clear, tuple of <flag_of_bitmap, batch_size, num_of_rows, num_of_classes>. which means you need to double the size of `TEST_CASES`.

Jarvi-Izana: or a more simple way
```
prediction_test_driver(true, batch_size, num_of_rows, num_of_classes, runner);
prediction_test_driver(false, batch_size, num_of_rows, num_of_classes, runner);
```
Now I think I understand what you are going to do
```
for (bool is_bitmap_based : {true, false}) {
```
That's also okay, kind of a overkill.

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Fixed.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
shantanuchhabra: Addressed.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
shantanuchhabra: Not doing that anymore.

src/toolkits/drawing_classifier/data_preparation.cpp
```diff
@@ -448,9 +448,10 @@ gl_sframe _drawing_classifier_prepare_data(const gl_sframe &data,
     for (const auto& row: data.range_iterator()) {
         const flexible_type &strokes = row[column_index_map[feature]];
         flex_list current_stroke_based_drawing = strokes.to<flex_list>();
-        bitmaps.push_back(convert_stroke_based_drawing_to_bitmap(
-            current_stroke_based_drawing, row_number)
-        );
+        flex_image bitmap = convert_stroke_based_drawing_to_bitmap(
+            current_stroke_based_drawing, row_number);
+        bitmaps.push_back(bitmap);
+        std::cout << row_number;
```
Jarvi-Izana: might consider suppressing the print.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -680,6 +686,12 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
                   "Expected one of: probability, probability_vector, rank");
   }
 
+  std::string feature_column_name = read_state<flex_string>("feature_column_name");
+  if (!(data.contains_column(feature_column_name))) {
```
Jarvi-Izana: consider using ASSERT_MSG for a sanity check.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
```diff
@@ -99,7 +99,8 @@ gl_sframe set_up_perform_inference(
     test_drawing_classifier& mock_model,
     std::unique_ptr<mock_model_backend>& mock_backend,
     std::unique_ptr<mock_compute_context>& mock_context, size_t batch_size,
-    const size_t num_of_rows, const size_t num_of_classes) {
+    const size_t num_of_rows, const size_t num_of_classes,
+    const std::string feature_column_name) {
```
Jarvi-Izana: I don't know why you need this variable regarding the fact that you already have `feature_list` with size one. If you need this variable, why you change the original single feature variable to a `feature_list`?

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
```diff
@@ -225,25 +236,19 @@ void prediction_test_driver(size_t batch_size, size_t num_of_rows,
...
-  mock_model.create_iterator_calls_.push_back([=](data_iterator::parameters) {
-    data_iterator::parameters my_params;
-    my_params.data = my_data;
-    my_params.is_train = false;
-    my_params.feature_column_name = feature_name;
-    my_params.target_column_name = target_name;
-    return std::unique_ptr<data_iterator>(new simple_data_iterator(my_params));
-  });
+  mock_model.create_iterator_calls_.push_back(
+      [=](data_iterator::parameters my_params) {
```
Jarvi-Izana: Why remove lines for setting up params? I intend to explicitly set params with specific values for test purposes.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
Jarvi-Izana: please remove the outter loop ` for (bool is_bitmap_based : {true}) {` completely. 

You already have `is_bitmap_based` taken care of
```
for (bool is_bitmap_based : {true, false}) {
       prediction_test_driver(batch_size, num_of_rows, num_of_classes, runner,
                              is_bitmap_based);
     }
```

src/core/storage/fileio/fileio_constants.cpp
```diff
@@ -85,8 +85,10 @@ static bool check_cache_file_location(std::string val) {
   if (paths.size() == 0)
     throw std::string("Value cannot be empty");
   for (std::string path: paths) {
-    if (!boost::filesystem::is_directory(path))
+    if (!boost::filesystem::is_directory(path)){
+      std::cerr<<"Could not write to the turicreate file cache, which is currently set to '" + path + "' Using path : "<<std::endl;
```
Jarvi-Izana: A log with a warning should be better. The code quality is not even ok. @TobyRoseman

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -1069,8 +1069,9 @@ class as a vector. Label ordering is dictated by the ``classes``
             Rows: 10
             [3, 4, 3, 3, 4, 5, 8, 8, 8, 4]
         """
+        feature_column = self.__proxy__.get_state()["feature_column_name"]
```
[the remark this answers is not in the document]
shantanuchhabra: How would we access a private member variable from the Python layer?

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -283,18 +289,24 @@ void drawing_classifier::init_training(
...
-  // Read user-specified options.
-  init_options(opts);
+  if (!(data.contains_column(target_column_name))) {
+    log_and_throw(target_column_name + "not found. Data passed in " + 
+      "does not contain the target column model was trained with");
+  }
 
   add_or_update_state({
-      {"training_iterations", 0},
+    {"feature_column_name", feature_column_name},
```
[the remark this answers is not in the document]
shantanuchhabra: We don't really need the target column name in the state right now, but I guess it wouldn't hurt so I'll add it.

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Why should we have an assert message instead of throwing an error with a message to the user?

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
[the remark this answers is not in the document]
nickjong: I believe this can just be `self.feature_column_name` without any changes elsewhere, due to the magic of Python and `ExposeAttributesFromProxy` (https://github.com/apple/turicreate/blob/209c8a64be2c8aea6c278acc2f15bb5790d88d8b/src/python/turicreate/toolkits/_model.py#L308)

src/toolkits/drawing_classifier/drawing_classifier.cpp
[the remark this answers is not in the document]
nickjong: How are we naming the output feature in the CoreML model? In 5.8 and now?

src/toolkits/drawing_classifier/data_preparation.cpp
nickjong: Let's just revert this file. Consider staging debugging changes under a separate commit that you don't push to GH for review

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
shantanuchhabra: Okay sure, we can just use the feature column name from the features list and not add another state variable. 

We never changed the original single feature variable to a feature list. We started with a feature list (see https://github.com/apple/turicreate/pull/2439#discussion_r334133796). I agree that at some point, we should change this to be just a `feature_column_name`, but that would require a sizable code change on the `init_model` and `export_to_coreml` side of things which is not related to this PR. I believe that should come as a separate PR altogether.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
shantanuchhabra: Because `my_data` here can be stroke-based data, whereas the data iterator can only take in bitmap-based data.

Since we wrote this code with an intention to use the real data iterator, we would have to set up the data (by calling the real `create_iterator` where the stroke to bitmap conversion happens) appropriately for consumption by the data iterator.

src/toolkits/drawing_classifier/drawing_classifier.cpp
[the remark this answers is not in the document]
Jarvi-Izana: then no need to use redundant var `target_column_name`. we use var `target`.

src/toolkits/drawing_classifier/data_preparation.cpp
shantanuchhabra: Thanks.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
Jarvi-Izana: Be brief. Then stick to `feature_list`. Adding a new var referring to the first element of the list adds more complexity for reading the code. I'd better ensure the state is tight and simple.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
nickjong: It's very strange for the callback to set the values for `my_params`. For mocks, these callbacks are intended to be the (mock) implementation, and to make assertions about what the parameters are.

For example, it would be very weird if `drawing_classifier::create_iterator` blindly overwrote the parameters you pass it as an argument

src/toolkits/drawing_classifier/drawing_classifier.cpp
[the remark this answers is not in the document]
Jarvi-Izana: https://github.com/apple/turicreate/blob/209c8a64be2c8aea6c278acc2f15bb5790d88d8b/src/toolkits/coreml_export/neural_net_models_exporter.cpp#L465

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
shantanuchhabra: ^Yes, I agree.

test/unity/toolkits/drawing_classifier/test_dc_prediction.cxx
Jarvi-Izana: That's by design to override the `params`. There's no way to override inside of predict call. This is designed to bypass the infinite loop bug in dc. And also pass extra column name info into params. I don't remember why I need this workaround. You can run the test to see if it passes. If it passed, then you are good to go.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -1069,7 +1069,7 @@ class as a vector. Label ordering is dictated by the ``classes``
             Rows: 10
             [3, 4, 3, 3, 4, 5, 8, 8, 8, 4]
         """
-        feature_column = self.__proxy__.get_state()["feature_column_name"]
+        feature_column = self.features[0]
```
nickjong: Per offline discussion, let's preserve the structure of the model state from 5.8

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -289,18 +289,12 @@ void drawing_classifier::init_training(
...
-  if (!(data.contains_column(feature_column_name))) {
-    log_and_throw(feature_column_name + "not found. Data passed in " + 
-      "does not contain the feature column model was trained with");
-  }
-
-  if (!(data.contains_column(target_column_name))) {
-    log_and_throw(target_column_name + "not found. Data passed in " + 
-      "does not contain the target column model was trained with");
-  }
+  ASSERT_MSG(data.contains_column(feature_column_name),
```
nickjong: Actually, `log_and_throw` is the correct mechanism for validating user inputs

`ASSERT_MSG` is for correcting programmer errors

Jarvi-Izana:
```
#define ASSERT_MSG(condition, fmt, ...)                                  \
  do {                                                                   \
    if (__builtin_expect(!(condition), 0)) {                             \
      auto throw_error = [&]() GL_GCC_ONLY(GL_COLD_NOINLINE_ERROR) {     \
        logstream(LOG_ERROR) << "Check failed: " << #condition << ":\n"; \
        std::ostringstream ss;                                           \
        ss << "Assertion Failure: " << #condition << ": " << fmt;        \
        logger(LOG_ERROR, fmt, ##__VA_ARGS__);                           \
        __print_back_trace();                                            \
        TURI_LOGGER_FAIL_METHOD(ss.str().c_str());                       \
        ASSERT_UNREACHABLE();                                            \
      };                                                                 \
      throw_error();                                                     \
      TURI_BUILTIN_UNREACHABLE();                                        \
    }                                                                    \
  } while (0)
```
The impl is same for log_and_throw. It doesn't really hault the program.

Jarvi-Izana:
```
#define log_and_throw(message)                        \
  do {                                                \
    auto throw_error = [&]() GL_COLD_NOINLINE_ERROR { \
      logstream(LOG_ERROR) << (message) << std::endl; \
      throw(std::string(message));                    \
    };                                                \
    throw_error();                                    \
  } while (0)
```
log_and_throw. Ah, slightly different. I would agree `ASSERT_MSG` is for programmers. Thanks for pointing out.

nickjong: I can't speak to the implementation differences, but maybe @hoytak can.

The intent is for `log_and_throw` to be user-facing. Assertions are for developers and can contain more technical information, such as backtraces. We also want to make it easy to tell at a glance what error strings are meant to be user-facing

Jarvi-Izana: yeah, that's great. The reason I use asser_msg is that it's neater. Since they have different semantics, I should differentiate them.

test/unity/toolkits/drawing_classifier/test_dc_serialization.cxx
```diff
@@ -40,14 +40,14 @@ BOOST_AUTO_TEST_CASE(test_dc_init_model) {
   // states
   constexpr unsigned int num_classes = 10;
   const std::string target = "target";
-  const std::vector<std::string> features = {"0", "1"};
+  const std::vector<std::string> features = {"0"};
```
Jarvi-Izana: I will change this later. merge first.

shantanuchhabra:
Relevant internal tests passed.

PR #2553, opened 2019-11-01
Fixed style transfer export in style transfer

abhishekpratapa:
- Added opts to export to coreml in ST

### DIFF
--- a/src/toolkits/coreml_export/neural_net_models_exporter.cpp
+++ b/src/toolkits/coreml_export/neural_net_models_exporter.cpp
@@ -400,15 +400,13 @@ std::shared_ptr<coreml::MLModelWrapper> export_style_transfer_model(
 
   FeatureDescription* model_input = model_desc->add_input();
   ImageFeatureType* input_feat = set_image_feature(model_input, image_width, image_height, content_feature, "Input image");
-  model_input->set_name(content_feature);
 
   set_array_feature(
       model_desc->add_input(), "index",
       "Style index array (set index I to 1.0 to enable Ith style)", {1});
 
   FeatureDescription* model_output = model_desc->add_output();
-  ImageFeatureType* style_feat = set_image_feature(model_output, image_width, image_height, "stylizedImage", "Stylized image");
-  model_output->set_name(style_feature);
+  ImageFeatureType* style_feat = set_image_feature(model_output, image_width, image_height, style_feature, "Stylized image");
 
   /**
    * The -1 indicates no upper limits for the image size
@@ -418,10 +416,18 @@ std::shared_ptr<coreml::MLModelWrapper> export_style_transfer_model(
     set_image_feature_size_range(style_feat, 64, -1, 64, -1);
   }
 
-  model.mutable_neuralnetwork()->MergeFrom(nn_spec.get_coreml_spec());
+  CoreML::Specification::NeuralNetwork* nn = model.mutable_neuralnetwork();
+  nn->MergeFrom(nn_spec.get_coreml_spec());
 
-  // content_feature
-  // style_feature
+  /*
+    Change input to first and last layers to match input and output feature names.
+  */
+  int last_layer_index = nn->layers_size() - 1;
+  NeuralNetworkLayer* first_layer = nn->mutable_layers(0);
+  NeuralNetworkLayer* last_layer = nn->mutable_layers(last_layer_index);
+
+  first_layer->set_input(0, content_feature);
+  last_layer->set_output(0, style_feature);
 
   auto model_wrapper =
       std::make_shared<MLModelWrapper>(std::make_shared<CoreML::Model>(model));
--- a/src/toolkits/coreml_export/neural_net_models_exporter.cpp
+++ b/src/toolkits/coreml_export/neural_net_models_exporter.cpp
@@ -18,6 +18,7 @@ using CoreML::Specification::ImageFeatureType;
 using CoreML::Specification::ImageFeatureType_ImageSizeRange;
 using CoreML::Specification::ModelDescription;
 using CoreML::Specification::NeuralNetworkLayer;
+using CoreML::Specification::NeuralNetworkPreprocessing;
 using CoreML::Specification::SizeRange;
 using turi::coreml::MLModelWrapper;
 
@@ -140,9 +141,9 @@ void set_image_feature_size_range(ImageFeatureType* image_feature,
 
 ImageFeatureType* set_image_feature(
     FeatureDescription* feature_desc, size_t image_width, size_t image_height,
-    std::string description = "",
+    std::string input_name, std::string description = "",
     ImageFeatureType::ColorSpace image_type = ImageFeatureType::RGB) {
-  feature_desc->set_name("image");
+  feature_desc->set_name(input_name);
   if (!description.empty()) feature_desc->set_shortdescription(description);
 
   ImageFeatureType* image_feature =
@@ -160,8 +161,8 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
     const neural_net::model_spec& nn_spec, size_t image_width,
     size_t image_height, size_t num_classes, size_t num_predictions,
     flex_dict user_defined_metadata, flex_list class_labels,
+    const std::string& input_name,
     std::map<std::string, flexible_type> options) {
-
   // Set up Pipeline
   CoreML::Specification::Model model_pipeline;
   model_pipeline.set_specificationversion(3);
@@ -174,7 +175,7 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
   NeuralNetworkLayer* first_layer =
       model_nn->mutable_neuralnetwork()->add_layers();
   first_layer->set_name("_divscalar0");
-  first_layer->add_input("image");
+  first_layer->add_input(input_name);
   first_layer->add_output("_divscalar0");
   first_layer->mutable_scale()->add_shapescale(1);
   first_layer->mutable_scale()->mutable_scale()->add_floatvalue(1 / 255.f);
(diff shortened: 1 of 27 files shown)


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -296,7 +296,7 @@ void style_transfer::load_version(iarchive& iarc, size_t version) {
 
 std::unique_ptr<compute_context> style_transfer::create_compute_context()
     const {
-  return compute_context::create();
+  return compute_context::create_tf();
```
shreyajain17: Change it back to `create`

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -390,33 +390,31 @@ std::shared_ptr<MLModelWrapper> export_activity_classifier_model(
...
-      set_image_feature(model_desc->add_input(), image_width, image_height,
-                        "image", "Input image");
-
-  /**
-   * The -1 indicates no upper limits for the image size
-   */
-  set_image_feature_size_range(input_feat, 64, -1, 64, -1);
+  FeatureDescription* model_input = model_desc->add_input();
+  ImageFeatureType* input_feat = set_image_feature(model_input, image_width, image_height, "Input image");
+  model_input->set_name("image");
```
nickjong: Why can't we set the name "image" as part of the call to `set_image_feature`?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -616,23 +615,27 @@ void style_transfer::train(gl_sarray style, gl_sarray content,
 
 std::shared_ptr<MLModelWrapper> style_transfer::export_to_coreml(
     std::string filename, std::map<std::string, flexible_type> opts) {
-  flex_int image_width = read_state<flex_int>("image_width");
-  flex_int image_height = read_state<flex_int>("image_height");
+  const flex_int image_width =
+      read_opts<flex_int>(opts, "image_width", /* delete item */ true);
```
nickjong: Why are we deleting these keys from the dictionary?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -616,23 +615,27 @@ void style_transfer::train(gl_sarray style, gl_sarray content,
...
-      {"training_iterations", read_state<flex_int>("training_iterations")},
+      {"max_iterations", std::to_string(read_state<flex_int>("max_iterations"))},
+      {"training_iterations", std::to_string(read_state<flex_int>("training_iterations"))},
       {"type", "StyleTransfer"},
-      {"content_feature", read_state<flex_string>("content_feature")},
-      {"style_feature", read_state<flex_string>("style_feature")},
-      {"num_styles", read_state<flex_string>("num_styles")},
-      {"version", get_version()},
+      {"content_feature", "image"},
+      {"style_feature", "image"},
```
nickjong: Why are we storing the hard-coded "image" here instead of the column name from the user?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -616,23 +615,27 @@ void style_transfer::train(gl_sarray style, gl_sarray content,
...
+  const flex_int image_height =
+      read_opts<flex_int>(opts, "image_height", /* delete item */ true);
+  const flex_int include_flexible_shape =
+      read_opts<flex_int>(opts, "include_flexible_shape", /* delete item */ true);
 
   flex_dict user_defined_metadata = {
       {"model", read_state<flex_string>("model")},
-      {"max_iterations", read_state<flex_int>("max_iterations")},
-      {"training_iterations", read_state<flex_int>("training_iterations")},
+      {"max_iterations", std::to_string(read_state<flex_int>("max_iterations"))},
```
nickjong: Why are we storing these as strings?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -642,37 +645,14 @@ std::shared_ptr<MLModelWrapper> style_transfer::export_to_coreml(
...
-
-  auto model_type_iter = model_data.find("model");
-  if (model_type_iter == model_data.end()) {
-    log_and_throw("The loaded turicreate model must contain 'model_iter'!\n");
-  }
-  const std::string model_type =
-      variant_get_value<flex_string>(model_type_iter->second);
-  model_data.erase(model_type_iter);
+  const flex_dict& model =
+      read_opts<flex_dict>(model_data, "_model", /* delete item */ true);
```
nickjong: Why are we deleting these keys from the dictionary?

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: I guess because we don't need them anymore so I delete them.

src/toolkits/coreml_export/neural_net_models_exporter.cpp
abhishekpratapa: At the time I coded it the set_image_feature didn't have an option for an image name and there were too many conflicts to resolve so I thought this was the best solution. Now that implementation has changed so I'll refactor to use `set_image_feature`.

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: CoreML expects this metadata as string data so if we don't cast this user_defined_metadata, `export_core_ml` throws an exception.

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: I guess because we don't need them anymore so I delete them, is that bad?

src/toolkits/style_transfer/style_transfer.cpp
nickjong: What exception? The implementation is intended to handle this, and to convert to string automatically: https://github.com/apple/turicreate/blob/4a8bbf96c6dfd1b302fc264e58ddc2c04f398431/src/toolkits/coreml_export/coreml_export_utils.cpp#L40

src/toolkits/style_transfer/style_transfer.cpp
nickjong: It's confusing, and therefore makes the code less readable and maintainable.

These will get deleted anyway when the dictionary goes out of scope.

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Since we used a predefined model_spec to generate the layers, it seemed like it would be difficult to change the input and output layer names at the time. Ehh I'll fix it!

srikris:
@abhishekpratapa Is this good to be merged. Seems like all comments have been addressed.

nickjong:
> @abhishekpratapa Is this good to be merged. Seems like all comments have been addressed.

I don't see any commits pushed since the last set of requested changes.... Probably best form not to resolve comments until the resolution has been pushed to the branch

abhishekpratapa:
@nickjong forgot to push my local branch! Good catch. I'll resolve after I've pushed next time.

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Might have conflated two different errors. I reverted and everything worked!

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Matched behaviors to the original python export to CoreML

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -390,36 +390,39 @@ std::shared_ptr<MLModelWrapper> export_activity_classifier_model(
...
-      set_image_feature(model_desc->add_input(), image_width, image_height,
-                        "image", "Input image");
-
-  /**
-   * The -1 indicates no upper limits for the image size
-   */
-  set_image_feature_size_range(input_feat, 64, -1, 64, -1);
+  FeatureDescription* model_input = model_desc->add_input();
+  ImageFeatureType* input_feat = set_image_feature(model_input, image_width, image_height, content_feature, "Input image");
+  model_input->set_name(content_feature);
```
nickjong: Is this needed? Doesn't `set_image_feature` do this?

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -390,36 +390,39 @@ std::shared_ptr<MLModelWrapper> export_activity_classifier_model(
...
   set_array_feature(
       model_desc->add_input(), "index",
       "Style index array (set index I to 1.0 to enable Ith style)", {1});
 
-  ImageFeatureType* style_feat =
-      set_image_feature(model_desc->add_output(), image_width, image_height,
-                        "image", "Stylized image");
+  FeatureDescription* model_output = model_desc->add_output();
+  ImageFeatureType* style_feat = set_image_feature(model_output, image_width, image_height, "stylizedImage", "Stylized image");
+  model_output->set_name(style_feature);
```
nickjong: What if this doesn't match "stylizedImage"?

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -390,36 +390,39 @@ std::shared_ptr<MLModelWrapper> export_activity_classifier_model(
...
+  if (include_flexible_shape) {
+    set_image_feature_size_range(input_feat, 64, -1, 64, -1);
+    set_image_feature_size_range(style_feat, 64, -1, 64, -1);
+  }
 
   model.mutable_neuralnetwork()->MergeFrom(nn_spec.get_coreml_spec());
 
+  // content_feature
+  // style_feature
+
```
nickjong: What are these comments meant to communicate?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -616,23 +615,26 @@ void style_transfer::train(gl_sarray style, gl_sarray content,
...
+  
   flex_dict user_defined_metadata = {
       {"model", read_state<flex_string>("model")},
       {"max_iterations", read_state<flex_int>("max_iterations")},
       {"training_iterations", read_state<flex_int>("training_iterations")},
       {"type", "StyleTransfer"},
-      {"content_feature", read_state<flex_string>("content_feature")},
-      {"style_feature", read_state<flex_string>("style_feature")},
-      {"num_styles", read_state<flex_string>("num_styles")},
+      {"content_feature", content_feature},  // TODO: refactor to take content name and style name
```
nickjong: Not sure what the TODO is trying to communicate. Aren't you already taking `content_feature` and `style_feature`?

src/toolkits/style_transfer/style_transfer.hpp
```diff
@@ -119,6 +119,33 @@ class EXPORT style_transfer : public ml_model_base {
...
+  }
+
+  template <typename T>
+  T read_opts(std::map<std::string, turi::flexible_type>& opts,
+              const std::string& key, bool remove_item = false) const {
+    auto iter = _read_iter_opts<turi::flexible_type>(opts, key);
+    T value = iter->second.get<T>();
+    if (remove_item) opts.erase(iter);
+    return value;
+  }
```
nickjong: Can we remove the `remove_item` argument? It seems incompatible with these being `const` methods anyway

src/toolkits/coreml_export/neural_net_models_exporter.cpp
abhishekpratapa: That I should have written some code there.

PR #2554, opened 2019-11-02
Provide TF alternative to MXNet accuracy metric during SC training

nickjong:
We just need to do the same for the data iterator, alongside Fareeha's changes for the actual NN training, and we should be in good shape to flip the switch for SC.


src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -19,6 +19,74 @@
...
+    def reset(self):
+        self.impl.reset()
+
+    def get(self):
+        _, result = self.impl.get()
+        return result
+
+class _TFAccuracy(_Accuracy):
+    def __init__(self):
+        import tensorflow.compat.v1 as tf
```
TobyRoseman: Can you use`tensorflow.keras.metrics.Accuracy` not `tensorflow.compat.v1.keras.metrics.Accuracy`?

nickjong: Ah, good point. That change appears to work fine!

PR #2555, opened 2019-11-02
Drawing Classifier: Added USE_CPP flag + bugfixes throughout

shantanuchhabra:
Also, skipped DC tests that are throwing some error when used with the C++ implementation. Will resolve those in future PRs.


src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -345,8 +348,9 @@ void drawing_classifier::init_training(
   }
 
   // Report to the user what GPU(s) is being used.
-  std::vector<std::string> gpu_names = training_compute_context_->gpu_names();
-  print_training_device(gpu_names);
+  /* The following two lines segfault!!! */
+  // std::vector<std::string> gpu_names = training_compute_context_->gpu_names();
+  // print_training_device(gpu_names);
```
shantanuchhabra: We should discuss whether this change should go in or not. Locally, I get a Segmentation Fault, but commenting this line out gets rid of it.

shreyajain17: We should figure out this one. I don't get this segmentation fault.

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -40,44 +40,52 @@ void add_drawing_pixel_data_to_batch(float* next_drawing_pointer,
...
-      result.class_to_index_map[label] = i++;
-    }
-  } else {
-    // Construct the class-to-index map from the expected labels.
-    result.classes = std::move(expected_class_labels);
-    int i = 0;
-    for (const std::string& label : result.classes) {
-      result.class_to_index_map[label] = i++;
-    }
+  if (data.contains_column(target_column_name)) {
```
nickjong: Instead of indenting essentially the entire body of this function inside this `if` statement, could we instead have an early return?

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: What is the backtrace? Which compute context is affected?

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -648,7 +652,9 @@ gl_sframe drawing_classifier::perform_inference(data_iterator* data) const {
 
     // Send the inputs to the model.
     std::map<std::string, shared_float_array> results =
-        backend->predict({{"input", result_batch.data_info.drawings}});
+        backend->predict({{"input", result_batch.data_info.drawings},
+                          {"num_samples", shared_float_array::wrap(result_batch.data_info.num_samples)}
+                         });
```
nickjong: It seems strange for me to have `num_samples` be a top-level input to the network. Typically this is captured in the first dimension of the input, or perhaps via a weight tensor (if the batch size is fixed in the backend). If all we support is TF for now, can't we just pass a smaller tensor?

Can follow up in a separate PR. But please leave a TODO: I think having `num_samples` here is ultimately technical debt

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: @nickjong `tf_compute_context`.

Backtrace as below:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xa0)
    frame #0: 0x000000010709285c Python`list_dealloc + 28
Python`list_dealloc:
->  0x10709285c <+28>: movl   0xa0(%r14), %eax
    0x107092863 <+35>: cmpl   $0x31, %eax
    0x107092866 <+38>: jg     0x107092880               ; <+64>
    0x107092868 <+40>: incl   %eax
Target 0: (python) stopped.
(lldb) up
frame #1: 0x0000000149320816 libtctensorflow.so`pybind11::handle::dec_ref(this=0x00007ffee8ba35e0) const & at pytypes.h:197:39
   194             preferable to use the `object` class which derives from `handle` and calls
   195             this function automatically. Returns a reference to itself.
   196         \endrst */
-> 197         const handle& dec_ref() const & { Py_XDECREF(m_ptr); return *this; }
   198     
   199         /** \rst
   200             Attempt to cast the Python object into the given C++ type. A `cast_error`
(lldb) up
frame #2: 0x0000000149320765 libtctensorflow.so`pybind11::object::~object(this=0x00007ffee8ba35e0) at pytypes.h:240:17
   237         /// Move constructor; steals the object from ``other`` and preserves its reference count
   238         object(object &&other) noexcept { m_ptr = other.m_ptr; other.m_ptr = nullptr; }
   239         /// Destructor; automatically calls `handle::dec_ref()`
-> 240         ~object() { dec_ref(); }
   241     
   242         /** \rst
   243             Resets the internal pointer to ``nullptr`` without without decreasing the
(lldb) up
frame #3: 0x000000014931eba5 libtctensorflow.so`pybind11::object::~object(this=0x00007ffee8ba35e0) at pytypes.h:240:15
   237         /// Move constructor; steals the object from ``other`` and preserves its reference count
   238         object(object &&other) noexcept { m_ptr = other.m_ptr; other.m_ptr = nullptr; }
   239         /// Destructor; automatically calls `handle::dec_ref()`
-> 240         ~object() { dec_ref(); }
   241     
   242         /** \rst
   243             Resets the internal pointer to ``nullptr`` without without decreasing the
(lldb) up
frame #4: 0x000000014931f227 libtctensorflow.so`turi::neural_net::tf_compute_context::gpu_names(this=0x00007fbe7504a2a0) const at tf_compute_context.cpp:283:1
   280       });
   281     
   282       return gpu_device_names;
-> 283     }
   284     
   285     
   286     std::unique_ptr<model_backend> tf_compute_context::create_object_detector(
```

nickjong: What if we move the `gpu_devices` local variable here
https://github.com/apple/turicreate/blob/25eab55fd492f355f8667ddd2cc60e26697d8681/src/ml/neural_net/tf_compute_context.cpp#L270
inside the lambda?

I wonder if we can't release the GIL before destroying it on function return

shantanuchhabra:
Internal tests passed.

shantanuchhabra:
Waiting on #2561 before merging this.

shantanuchhabra:
Internal tests passed.

Issue #954, opened 2018-08-09
Dictionary input to count_words(....)


TobyRoseman:
The doctoring for `turicreate.text_analytics.count_words(...)` contains the following example:
```python
# Run count_words with dictionary input
>>> sa = turicreate.SArray([{'alice bob': 1, 'Bob alice': 0.5},
                                                {'a dog': 0, 'a dog cat': 5}])
>>> turicreate.text_analytics.count_words(sa)
dtype: dict
Rows: 2
[{'bob': 1.5, 'alice': 1.5}, {'a': 5, 'dog': 5, 'cat': 5}]
```
However the actual output is:
```python
dtype: dict
Rows: 2
[{'bob': 2, 'alice': 2}, {'cat': 1, 'dog': 2, 'a': 2}]
```
It's determining count by looking at only occurrence in the keys. However the doctoring (both the example and the description) claim it should sum the values for keys that contain the word.

This is a bug as things are not working as described, but a bigger question is why do we want to support this use case? Why do we want be able to tokenize strings/keys and add up their values for each token?

nickjong:
Yeah, I'm not sure I understand the use case, but the documented behavior makes the most sense given the shape of the API. If you treat each key as an atomic unit, then there's no really need for the API (and there's no notion of "word" involved). If you just count the number of keys containing each word, then the dictionary values are irrelevant and these should just be lists.

dhivyaaxim:
@TobyRoseman  Please assign this ticket to me as I would like to try fixing it and contribute. Thanks in advance!

PR #2557, opened 2019-11-04
Wrong behavior for nan converting from float to int

### LINKED ISSUE #1761: Wrong behavior for nan converting from float to int
When changing the dtype of an SArray from float to int, `nan` get converted to very large number:
```python

In [2]: sa = tc.SArray([1.0,2.0,float('nan'),3.0])

In [3]: sa
Out[3]: 
dtype: float
Rows: 4
[1.0, 2.0, nan, 3.0]

In [4]: sa.astype(int)
Out[4]: 
dtype: int
Rows: 4
[1, 2, -9223372036854775808, 3]
```
I'm not certain what the best behavior would be here; it seems `nan` is a floating point only representation. It would probably be best to turn the `nan` into `None`. Raising an exception is also probably reasonable.

dhivyaaxim:
This PR resolves #1761 
Wrong behavior for nan converting from float to int

### DIFF
--- a/src/python/turicreate/test/test_sarray.py
+++ b/src/python/turicreate/test/test_sarray.py
@@ -252,6 +252,11 @@ class SArrayTest(unittest.TestCase):
         self.assertEqual(len(g), len(tlist))
         for i in range(len(tlist)):
           self.assertEqual(g[i], tlist[i])
+    
+    def test_nan_to_none_type_conversion(self):
+        sa=SArray([1.0,2.0,float('nan'),3.0])
+        sa=sa.astype(int)
+        self.assertTrue(None in sa)
 
     def test_list_with_array_creation(self):
         import array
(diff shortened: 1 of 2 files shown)


src/python/turicreate/test/test_sarray.py
```diff
@@ -252,6 +252,11 @@ def test_list_with_none_creation(self):
         self.assertEqual(len(g), len(tlist))
         for i in range(len(tlist)):
           self.assertEqual(g[i], tlist[i])
+    
+    def test_nan_to_none_type_conversion(self):
+        sa=SArray([1.0,2.0,float('nan'),3.0])
+        sa=sa.astype(int)
+        self.assertTrue(None in sa)
```
TobyRoseman: This is good but it's better to to check that there is only one `None` and that it's where it's suppose to be. Also good to check the other elements are what we expect too. 

I would change this line to:
```python
expected = SArray([1, 2, None, 3])
for a, b in zip(sa, expected):
    self.assertEqual(a, b)
```

dhivyaaxim:
@TobyRoseman We had added changes from previous fork turicreate master and run unit test for python and C, it's got failed. Pulled master changes and integrate my changes, then run unit test, it's passed. Any possible to sync latest master changes and integrate my changes in this commit? Please advise. Thanks in advance!

TobyRoseman:
@dhivyaaxim - These error are unrelated to your change. I'll just merge this now.

dhivyaaxim:
@TobyRoseman Thanks for your prompt response! Please merge and close the below two issues too have similar build issue
**#954** and **#2390**

PR #2558, opened 2019-11-04
Dictionary input to count_words

### LINKED ISSUE #954: Dictionary input to count_words(....)

dhivyaaxim:
This PR resolves #954 
Dictionary input to count_words

### DIFF
--- a/src/python/turicreate/test/test_text_analytics.py
+++ b/src/python/turicreate/test/test_text_analytics.py
@@ -257,7 +257,7 @@ class FeatureEngineeringTest(unittest.TestCase):
 
     def test_count_words_dict_type(self):
         sa = tc.SArray([{'Alice bob mike': 1, 'Bob Alice Sue': 0.5, '': 100},
-            {'a dog cow': 0, 'a dog cat ': 5, 'mice dog': -1, 'mice cat': 2}])
+             {'a dog cow': 0, 'a dog cat ': 5, 'mice dog': -1, 'mice cat': 2}])
         result = tc.text_analytics.count_words(sa)
         expected = [{'bob': 1.5, 'mike': 1.0, 'sue': 0.5, 'alice': 1.5}, {'a': 5.0, 'mice': 1.0, 'dog': 4.0, 'cow': 0.0, 'cat': 7.0}]
         self.assertEqual(result.dtype, dict)
--- a/src/python/turicreate/test/test_text_analytics.py
+++ b/src/python/turicreate/test/test_text_analytics.py
@@ -255,6 +255,15 @@ class FeatureEngineeringTest(unittest.TestCase):
         assert len(empty_dict_sf) == 1
         assert len(empty_dict_sf.apply(lambda x: len(x) == 0)) == 1
 
+    def test_count_words_dict_type(self):
+        sa = tc.SArray([{'alice bob': 1, 'Bob alice': 0.5},
+                            {'a dog': 0, 'a dog cat': 5}])
+        result = tc.text_analytics.count_words(sa)
+        expected = [{'bob': 1.5, 'alice': 1.5}, {'a': 5.0, 'dog': 5.0, 'cat': 5.0}]
+        self.assertEqual(result.dtype, dict)
+        self.sframe_comparer._assert_sarray_equal(result, expected)
+
+
 class RandomWordSplitTest(unittest.TestCase):
     """
     Test that the random split utility in the text analytics toolkit works
--- a/src/python/turicreate/test/test_text_analytics.py
+++ b/src/python/turicreate/test/test_text_analytics.py
@@ -256,10 +256,10 @@ class FeatureEngineeringTest(unittest.TestCase):
         assert len(empty_dict_sf.apply(lambda x: len(x) == 0)) == 1
 
     def test_count_words_dict_type(self):
-        sa = tc.SArray([{'alice bob': 1, 'Bob alice': 0.5},
-                            {'a dog': 0, 'a dog cat': 5}])
+        sa = tc.SArray([{'Alice bob mike': 1, 'Bob Alice Sue': 0.5, '': 100},
+            {'a dog cow': 0, 'a dog cat ': 5, 'mice dog': -1, 'mice cat': 2}])
         result = tc.text_analytics.count_words(sa)
-        expected = [{'bob': 1.5, 'alice': 1.5}, {'a': 5.0, 'dog': 5.0, 'cat': 5.0}]
+        expected = [{'bob': 1.5, 'mike': 1.0, 'sue': 0.5, 'alice': 1.5}, {'a': 5.0, 'mice': 1.0, 'dog': 4.0, 'cow': 0.0, 'cat': 7.0}]
         self.assertEqual(result.dtype, dict)
         self.sframe_comparer._assert_sarray_equal(result, expected)
 
--- a/src/toolkits/feature_engineering/word_counter.cpp
+++ b/src/toolkits/feature_engineering/word_counter.cpp
@@ -18,7 +18,9 @@ namespace feature_engineering {
  * Ported from original count_words implementation.
  */
 void word_count_delimiters_update(flexible_type f, flex_list delimiter_list, bool to_lower,
-                           std::unordered_map<flexible_type, float>& ret_count, flex_type_enum run_mode, flexible_type f_type = 0) {
+                                  std::unordered_map<flexible_type, float>& ret_count, 
+			          flex_type_enum run_mode, 
+			          flexible_type f_type = 0) {
   std::set<char> delimiters;
   for (auto it = delimiter_list.begin(); it != delimiter_list.end(); ++it) {
     // iterate through flexible_types storing the delimiters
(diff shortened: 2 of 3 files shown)


src/toolkits/feature_engineering/word_counter.cpp
```diff
@@ -18,9 +18,7 @@ namespace feature_engineering {
  * Ported from original count_words implementation.
  */
 void word_count_delimiters_update(flexible_type f, flex_list delimiter_list, bool to_lower,
-                                           std::unordered_map<flexible_type, size_t>& ret_count) {
-
-
+                           std::unordered_map<flexible_type, float>& ret_count, flex_type_enum run_mode, flexible_type f_type = 0) {
```
Jarvi-Izana: please also pay attention to your code style. Currently, we encourage google style for CPP change. https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions

src/toolkits/feature_engineering/word_counter.cpp
```diff
@@ -125,7 +129,10 @@ flexible_type word_counter_apply_with_manual(const flexible_type& input,
         if (kvp.second.get_type() != flex_type_enum::INTEGER
             && kvp.second.get_type() != flex_type_enum::FLOAT)
           log_and_throw("Invalid type. Dictionary input to WordCounter must have integer or float values.");
-        word_count_delimiters_update(kvp.first.get<flex_string>(), delimiter_list, to_lower, ret_count);
+	if (kvp.second.get_type() == flex_type_enum::INTEGER)
+            word_count_delimiters_update(kvp.first.get<flex_string>(), delimiter_list, to_lower, ret_count, run_mode, kvp.second.get<flex_int>());
```
Jarvi-Izana: formatting for function calls - https://google.github.io/styleguide/cppguide.html#Function_Calls

src/python/turicreate/test/test_text_analytics.py
```diff
@@ -255,6 +255,15 @@ def test_tf_idf(self):
         assert len(empty_dict_sf) == 1
         assert len(empty_dict_sf.apply(lambda x: len(x) == 0)) == 1
 
+    def test_count_words_dict_type(self):
+        sa = tc.SArray([{'alice bob': 1, 'Bob alice': 0.5},
+                            {'a dog': 0, 'a dog cat': 5}])
+        result = tc.text_analytics.count_words(sa)
+        expected = [{'bob': 1.5, 'alice': 1.5}, {'a': 5.0, 'dog': 5.0, 'cat': 5.0}]
```
TobyRoseman: For each dictionary in `expected` the values are all the same. So this unit test is only really checking one possible code path. When testing it is best to cover as many possible code path combinations as possible. 

Try setting `sa` to something like:
```
tc.SArray([{'Alice bob mike': 1, 'Bob Alice Sue': 0.5, '': 100},
                             {'a dog cow': 0, 'a dog cat': 5, 'mice dog': -1}])
```
Then figure out what `expected` should be and see if this test passes.

Jarvi-Izana: for word count, the count number shouldn't be negative.

TobyRoseman: @Jarvi-Izana - that's a very good point. You're quite right.

Negative counts and the empty string as a key would probably be good error/edge case tests but if you're only adding only one test it probably doesn't make sense to add either of these.

TobyRoseman:
@Jarvi-Izana - please re-review. Have your issues been resolved? If so, you need to approve this change.

src/toolkits/feature_engineering/word_counter.cpp
```diff
@@ -18,9 +18,9 @@ namespace feature_engineering {
  * Ported from original count_words implementation.
  */
 void word_count_delimiters_update(flexible_type f, flex_list delimiter_list, bool to_lower,
-                                           std::unordered_map<flexible_type, size_t>& ret_count) {
-
-
+                                  std::unordered_map<flexible_type, float>& ret_count, 
+			          flex_type_enum run_mode,
```
Jarvi-Izana: not aligned with previous line.

src/toolkits/feature_engineering/word_trimmer.cpp
```diff
@@ -310,7 +311,8 @@ flexible_type word_trimmer_apply(const flexible_type& input,
         if (kvp.first.get_type() != flex_type_enum::STRING)
           log_and_throw("Invalid type. Dictionary input to RareWordTrimmer must have string-typed keys.");
 
-        if (kvp.second.get_type() != flex_type_enum::INTEGER)
+        if (kvp.second.get_type() != flex_type_enum::INTEGER
+	   && kvp.second.get_type() != flex_type_enum::FLOAT)
```
Jarvi-Izana: let's do this
```
    if (kvp.second.get_type() != flex_type_enum::INTEGER &&
        kvp.second.get_type() != flex_type_enum::FLOAT) {

    }
```

src/python/turicreate/test/test_text_analytics.py
```diff
@@ -255,6 +255,15 @@ def test_tf_idf(self):
         assert len(empty_dict_sf) == 1
         assert len(empty_dict_sf.apply(lambda x: len(x) == 0)) == 1
 
+    def test_count_words_dict_type(self):
```
Jarvi-Izana:
```
def test_count_words_dict_type(self):
    sa = tc.SArray([{'Alice bob mike': 1, 'Bob Alice Sue': 0.5, '': 100},
                    {'a dog cow': 0, 'a dog cat ': 5, 'mice dog': -1, 'mice cat': 2}])
    result = tc.text_analytics.count_words(sa)
    expected = [{'bob': 1.5, 'mike': 1.0, 'sue': 0.5, 'alice': 1.5}, {
        'a': 5.0, 'mice': 1.0, 'dog': 4.0, 'cow': 0.0, 'cat': 7.0}]
    self.assertEqual(result.dtype, dict)
    self.sframe_comparer._assert_sarray_equal(result, expected)
```
let's use pep8. Please use spaces instead of tabs for indentation.

src/toolkits/feature_engineering/word_counter.cpp
```diff
@@ -18,9 +18,9 @@ namespace feature_engineering {
  * Ported from original count_words implementation.
  */
 void word_count_delimiters_update(flexible_type f, flex_list delimiter_list, bool to_lower,
-                                           std::unordered_map<flexible_type, size_t>& ret_count) {
-
-
+                                  std::unordered_map<flexible_type, float>& ret_count, 
+			          flex_type_enum run_mode, 
+			          flexible_type f_type = 0) {
```
Jarvi-Izana: `f_type` is a bad name. consider `delta`

src/toolkits/feature_engineering/word_counter.cpp
```diff
@@ -125,7 +131,18 @@ flexible_type word_counter_apply_with_manual(const flexible_type& input,
         if (kvp.second.get_type() != flex_type_enum::INTEGER
             && kvp.second.get_type() != flex_type_enum::FLOAT)
           log_and_throw("Invalid type. Dictionary input to WordCounter must have integer or float values.");
-        word_count_delimiters_update(kvp.first.get<flex_string>(), delimiter_list, to_lower, ret_count);
+	if (kvp.second.get_type() == flex_type_enum::INTEGER){
+            word_count_delimiters_update(
+		       kvp.first.get<flex_string>(), delimiter_list, 
+		       to_lower, ret_count, 
+		       run_mode, (float)kvp.second.get<flex_int>());
```
Jarvi-Izana: why do you need to cast to float? flexible_type can hold an integer value.

src/toolkits/feature_engineering/word_counter.cpp
```diff
@@ -103,7 +109,7 @@ flexible_type word_counter_apply_with_manual(const flexible_type& input,
   // Tokenize all string inputs according to delimiters and to_lower options,
   // then accumulate counts and return as dictionary
   flexible_type output;
-  std::unordered_map<flexible_type, size_t> ret_count;
+  std::unordered_map<flexible_type, float> ret_count;
```
Jarvi-Izana: flex_float is using double. why not using double? Otherwise, you will lose precision.

src/toolkits/feature_engineering/word_counter.cpp
Jarvi-Izana: the logic for `run_mode == INTEGER` and `run_mode == FLOAT` is almost same. I bet flexible<flex_int> can be converted to flexible<flex_float>. You can try to make your logic more precise.

Jarvi-Izana:
@TobyRoseman do you think we should treat negative word count as a valid input?

TobyRoseman:
> @TobyRoseman do you think we should treat negative word count as a valid input?

I don't think the dictionary values are intended to be word counts. The original example which we are trying to enable has non-integer values. As I said in the original issue (#954) the usage for this functionality is not clear to me.

src/toolkits/feature_engineering/word_counter.cpp
```diff
@@ -109,7 +109,7 @@ flexible_type word_counter_apply_with_manual(const flexible_type& input,
   // Tokenize all string inputs according to delimiters and to_lower options,
   // then accumulate counts and return as dictionary
   flexible_type output;
-  std::unordered_map<flexible_type, float> ret_count;
+  std::unordered_map<flexible_type, double> ret_count;
```
Jarvi-Izana: again. if you change double to flex_float, it would be better.

src/toolkits/feature_engineering/word_trimmer.cpp
```diff
@@ -188,9 +188,10 @@ void word_trimmer_topk_index_mapping(const gl_sarray& src,
             if (kvp.first.get_type() != flex_type_enum::STRING)
           log_and_throw("Invalid type. Dictionary input to RareWordTrimmer must have string-typed keys.");
 
-            if (kvp.second.get_type() != flex_type_enum::INTEGER
-	       && kvp.second.get_type() != flex_type_enum::FLOAT)
+            if (kvp.second.get_type() != flex_type_enum::INTEGER &&
+	        kvp.second.get_type() != flex_type_enum::FLOAT) {
```
Jarvi-Izana: if you align `k`s, that would be great.

Jarvi-Izana:
remember to merge your nice work. ;-)

dhivyaaxim:
@Jarvi-Izana Thanks! We don't merge rights. Can you please merge it 
Thanks in advance.

Jarvi-Izana:
@dhivyaaxim I see. Merged. Nice work!

PR #2559, opened 2019-11-04
Style transfer get_styles c++

abhishekpratapa:
- Implemented `get_styles` in c++
- Output `SArray` from c++ to maintain consistency with `predict` and `train`


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -523,7 +527,10 @@ void style_transfer::init_train(gl_sarray style, gl_sarray content,
 
   infer_derived_options();
 
-  add_or_update_state({{"model", "resnet-16"}});
+  add_or_update_state({
+    {"model", "resnet-16"},
+    {"style_sarray", style},
```
nickjong: The Python implementation stores this as an SFrame named "styles". Since we must support that representation for any already existing models, why not just adopt that here?

https://github.com/apple/turicreate/blob/25eab55fd492f355f8667ddd2cc60e26697d8681/src/python/turicreate/toolkits/style_transfer/style_transfer.py#L475

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -63,7 +63,7 @@ def _vgg16_data_prep(batch):
 
 def create(style_dataset, content_dataset, style_feature=None,
         content_feature=None, max_iterations=None, model='resnet-16',
-        verbose=True, batch_size = 6, **kwargs):
+        verbose=True, batch_size = 1, **kwargs):
```
TobyRoseman: Why change the default `batch_size` to 1?

abhishekpratapa: When we tested style_transfer earlier this year, a batch_size of 1 produced a better result. Shreya and I discussed and decided that it was proper to change this here. (Plus, MPS defaults to a batch size of 1 currently so that makes things easier).

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -237,6 +237,13 @@ flex_int estimate_max_iterations(flex_int num_styles, flex_int batch_size) {
   return static_cast<flex_int>(num_styles * 10000.0f / batch_size);
 }
 
+void check_style_index(int idx, int num_styles) {
+  std::cout << idx << std::endl;
+  std::cout << num_styles << std::endl;
```
nickjong: Debugging code?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -347,8 +355,65 @@ void style_transfer::infer_derived_options() {
...
+  index_list.resize(num_styles);
+  std::iota(index_list.begin(), index_list.end(), 0);
+
+  return gl_sframe({
+      {"index", index_list},
+      {"style", styles},
+  });
+}
+
+gl_sarray style_transfer::convert_variant_to_filter(const variant_type& data) {
```
nickjong: What does this function do? Please add comments

src/python/turicreate/test/test_style_transfer.py
```diff
@@ -161,6 +161,7 @@ def test_stylize_fail(self):
             with self.assertRaises(_ToolkitError):
                 model.stylize(self.content_sf[0:1], style=style)
 
+    @unittest.skip("Fix in a Later PR, known failure")
```
TobyRoseman: Could you be more descriptive here? When will it be fixed? Pre 6.0 release candidate? 

Same question for line 275.

src/python/turicreate/test/test_style_transfer.py
```diff
@@ -161,7 +161,7 @@ def test_stylize_fail(self):
             with self.assertRaises(_ToolkitError):
                 model.stylize(self.content_sf[0:1], style=style)
 
-    @unittest.skip("Fix in a Later PR, known failure")
+    @unittest.skip("Fix before 6.0 release candidate, known failure")
```
TobyRoseman: If you rebase this file, you can do `@pytest.mark.xfail(IS_PRE_6_0_RC, reason='Requires MXNet')`

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -296,7 +303,7 @@ void style_transfer::load_version(iarchive& iarc, size_t version) {
 
 std::unique_ptr<compute_context> style_transfer::create_compute_context()
     const {
-  return compute_context::create();
```
nickjong: Why can't we use MPS?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -347,6 +353,92 @@ void style_transfer::infer_derived_options() {
...
+  return gl_sframe({
+      {"index", index_list},
+      {"style", styles},
+  });
+}
+
+/**
+ * convert_variant_to_filter
+ *
+ * This function takes a variant type and converts it into a boolean filter. The
```
nickjong: How does it convert to a Boolean filter? What is the input? How is the output to be used? If I'm just looking at this function, how do I have any sense of what it does without reading through the entire thing?

nickjong: It might help to start with a name that indicates the purpose, or how this function fits into the overall class

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: It's a mistake, thanks for the catch!

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Changed the name!

src/python/turicreate/test/test_style_transfer.py
abhishekpratapa: Thanks for the heads up! I don't want to risk rebasing right now. I added TODO's for the next PR after this one goes in!

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -347,6 +353,92 @@ void style_transfer::infer_derived_options() {
...
+}
+
+/**
+ * convert_style_indicies_to_filter
+ *
+ * This function takes a variant type and converts it into a boolean filter. The
+ * elements at the indicies we want to keep are set to a value of `1`, the
+ * elements we don't want to keep are set to a value of `0`.
+ */
+gl_sarray style_transfer::convert_style_indicies_to_filter(const variant_type& data) {
```
nickjong: The plural of "index" is "indices" (or "indexes", as in some Apple APIs)

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -347,6 +353,92 @@ void style_transfer::infer_derived_options() {
...
+                      case flex_type_enum::FLOAT: {
+                        int idx = static_cast<int>(ft.get<flex_float>());
+                        check_style_index(idx, num_styles);
+                        index_filter[idx] = 1;
+                        break;
+                      }
+                      default:
+                        log_and_throw(
+                            "Invalid data type! List should contain either "
+                            "flex_float or flex_int values!");
```
nickjong: I don't think we want `flex_float` or `flex_int` to appear in the error messages being surfaced to users (when validating their inputs). What error message do we have for the old Python implementation?

abhishekpratapa: The error we have is `"the `style` variable cannot be parsed"` which I feel is a horrible message. Haha.

src/toolkits/style_transfer/style_transfer.cpp
abhishekpratapa: Thanks for catching the spelling mistakes!

src/toolkits/style_transfer/style_transfer.cpp
nickjong: Honestly, it’s better than exposing C++ type names to users who have no idea what “flex_float” or “flex_int” even are!

abhishekpratapa: True, what do you think about the newer message?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -347,6 +353,94 @@ void style_transfer::infer_derived_options() {
...
+    index_filter[idx] = 1;
+    return index_filter;
+  } else if (variant_is<flex_undefined>(data)) {
+    // If undefined set the value to all of the styles in the sframe to 1. Or \
+    // run stylize on all elements of the `SFrame`
+    return std::vector<flexible_type>(num_styles, 1);
+  } else {
+    log_and_throw(
+        "Invalid data type! Expect flex_list, flex_vec, flex_int, or "
+        "flex_undefined!");
```
nickjong: We're still using internal implementation details here in user-facing error messages

abhishekpratapa: Missed that, thanks!

PR #2561, opened 2019-11-04
OD tf fixed get_gpu_name.

jakesabathia2:
Fix the crash for invalid conversion from Python side to vector<string>.

### DIFF
--- a/src/ml/neural_net/tf_compute_context.cpp
+++ b/src/ml/neural_net/tf_compute_context.cpp
@@ -292,8 +292,7 @@ std::unique_ptr<model_backend> tf_compute_context::create_object_detector(
       // Make an instance of python object
       pybind11::object object_detector = tf_od_backend.attr(
           "ODTensorFlowModel")(h_in, w_in, n, c_out, weights, config);
-      result = std::unique_ptr<tf_model_backend>(
-          new tf_model_backend(object_detector));
+      result.reset(new tf_model_backend(object_detector));
     });
   return result;
 }
@@ -313,8 +312,7 @@ std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
     // Make an instance of python object
     pybind11::object activity_classifier = tf_ac_backend.attr(
         "ActivityTensorFlowModel")(weights, n, c_in, c_out, pw, w_out);
-    result = std::unique_ptr<tf_model_backend>(
-        new tf_model_backend(activity_classifier));
+    result.reset(new tf_model_backend(activity_classifier));
   });
   return result;
 }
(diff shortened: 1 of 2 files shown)


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -261,17 +261,16 @@ tf_compute_context::tf_compute_context() = default;
...
   return 4294967296lu;
 }
 
 std::vector<std::string> tf_compute_context::gpu_names() const {
-  pybind11::object gpu_devices;
   std::vector<std::string> gpu_device_names;
 
   call_pybind_function([&]() {
-    
+    pybind11::object gpu_devices;
```
jakesabathia2: There is the change.

src/python/turicreate/toolkits/_tf_utils.py
```diff
@@ -16,11 +16,11 @@ def get_gpu_names():
 	"""
 	import tensorflow as tf
 	gpu_names = tf.config.experimental.list_physical_devices('GPU')
-	return gpu_names
-	
+	return [str(gpu_name) for gpu_name in gpu_names]
```
jakesabathia2: And there.

jakesabathia2:
And my editor just auto correct all last blanks.
There are only two places which are really modified.

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: Can you add it like `pybind11::object gpu_devices = tf_gpu_devices.attr("get_gpu_names")();`

nickjong: In addition, let's fix the other instances in this file. Let's only declare `pybind11::object` variables inside the lambda being passed to `call_pybind_function`

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -285,17 +284,18 @@ std::vector<std::string> tf_compute_context::gpu_names() const {
...
   call_pybind_function([&]() {
       pybind11::module tf_od_backend = pybind11::module::import(
           "turicreate.toolkits.object_detector._tf_model_architecture");
 
       // Make an instance of python object
-      object_detector = tf_od_backend.attr("ODTensorFlowModel")(h_in, w_in, n, c_out, weights, config);
+      pybind11::object object_detector = tf_od_backend.attr(
+          "ODTensorFlowModel")(h_in, w_in, n, c_out, weights, config);
+      result = std::unique_ptr<tf_model_backend>(
+          new tf_model_backend(object_detector));
```
nickjong: This can be simplified to `result.reset(new tf_model_backend(...))`

PR #2565, opened 2019-11-05
Provide TF alternative to MXNet data iterator during SC training

nickjong:


src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -21,6 +21,82 @@
...
+        self.impl = tf.data.Dataset.from_tensor_slices(tensor_slices)
+        self.impl = self.impl.batch(batch_size)
+        if shuffle:
+            self.impl = self.impl.shuffle(data.shape[0])
+
+    def __iter__(self):
+        return self.impl.__iter__()
+
+    def reset(self):
+        pass
```
TobyRoseman: Do you want to raise an exception here? If not, it would be great to comment on why a `pass` here is ok.

nickjong:
> I noticed that you added the functionality to shuffle the data. I'm curious why. Do you think shuffling the training data might result in a better model?

Actually, I was just carrying forward what I thought was the original behavior: https://github.com/apple/turicreate/blob/dc86bfbc06b892f2a515dcc7d6a0ae725dc56ab3/src/python/turicreate/toolkits/sound_classifier/sound_classifier.py#L220

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
nickjong: Good point. It turns out that TF (at least in 2.0) returns a fresh iterator every time for `__iter__()`, unlike MXNet, which uses the main instance itself as an iterator

PR #2566, opened 2019-11-05
Setting TURI_NUM_GPUS=0 should disable MPS backend

nickjong:
This should simplify testing TensorFlow implementations on Macs (albeit slowly, with CPU-only computation)

### DIFF
--- a/src/ml/neural_net/mps_compute_context.mm
+++ b/src/ml/neural_net/mps_compute_context.mm
@@ -9,6 +9,7 @@
 #import <ml/neural_net/mps_device_manager.h>
 
 #include <core/logging/logger.hpp>
+#include <core/storage/fileio/fileio_constants.hpp>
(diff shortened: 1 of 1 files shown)


shreyajain17:
Can you also just remove `create_tf` in `compute_context` files, which was a work around that exists in master?

nickjong:
> Can you also just remove `create_tf` in `compute_context` files, which was a work around that's existed in master!

Good idea!

nickjong:
> Can you also just remove `create_tf` in `compute_context` files, which was a work around that existed in master!

On closer inspection, this is still needed by DC. Since there's no MPS implementation yet, using the normal factory function will usually instantiate the MPS compute context on Macs, and attempting to produce a DC model from the MPS context will fail. (DC in particular should continue to force TF always)

PR #2570, opened 2019-11-06
OD tf reset graph.

jakesabathia2:
Several unit test failure are due to every time a compute context is called,
duplicated models are create, lead to unexpected behavior.

### DIFF
--- a/src/python/turicreate/test/test_object_detector.py
+++ b/src/python/turicreate/test/test_object_detector.py
@@ -91,7 +91,7 @@ def _get_data(feature, annotations):
     return data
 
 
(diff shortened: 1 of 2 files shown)


nickjong:
Does this resolve any unit test failures? If so, do we need to mark those tests as no-longer-xfail?

jakesabathia2:
tbh,
most of the failures is caused by couples issue,
which I separate in different PRs hmmmm

jakesabathia2:
But particularly,
it resolves ObjectDetectorTest::test_extra_classes,
ObjectDetectorTest::test_export_coreml,
ObjectDetectorTest::test_export_coreml_with_non_maximum_suppression.

PR #2571, opened 2019-11-06
OD tf flexible batch size.

jakesabathia2:


jakesabathia2:
ObjectDetectorTest::test_single_image, ObjectDetectorTest::test_dict_annotations.
Basically the code crash whenever the number of example isn't a multiply of the batch size (or smaller than batch size).
Like batch size 32, num_example = 100 will crash.
batch size 2, num_example = 1 (single image) will crash.
Note that I also change the C++ code in order to fix those crashes.

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -171,8 +171,7 @@ image_augmenter::result float_array_image_augmenter::prepare_images(
   // Decode a batch of images to raw format (shared_float_arrays)
   // Also convert annotations and predictions per batch of images
   // to vectors of shared_float_arrays.
-  for (size_t i = 0; i < n; i++) {
-    const labeled_image& source = source_batch[i];
+  for (const labeled_image& source : source_batch) {
```
jakesabathia2: Here is previous `n` is fixed to the predefined batch size,
we change it to the actual data batch size.

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -199,7 +198,7 @@ image_augmenter::result float_array_image_augmenter::prepare_images(
 
   // Convert augmented_data to the data structure needed
   res.image_batch = augmented_data.images;
-  for (size_t a = 0; a < n; a++) {
+  for (size_t a = 0; a < source_batch.size(); a++) {
```
jakesabathia2: Here too.

src/ml/neural_net/image_augmentation.cpp
Jarvi-Izana: I wonder this is a very simple logical error, why it is only revealed by now? I should fail the unit test very easily.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -41,10 +40,10 @@ def __init__(self, input_h, input_w, batch_size, output_size, init_weights, conf
         self.is_train = is_train  # Set flag for training or val
 
         # Create placeholders for image and labels
-        self.images = _tf.placeholder(_tf.float32, [self.batch_size, input_h,
+        self.images = _tf.placeholder(_tf.float32, [None, input_h,
                                                               input_w, 3], name='images')
         self.labels = _tf.placeholder(_tf.float32,
-                                [self.batch_size, self.grid_shape[0], self.grid_shape[1],
+                                [None, self.grid_shape[0], self.grid_shape[1],
```
Jarvi-Izana: Just curious, if you fix batch size, why you need to default it to `None`?

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -380,7 +380,7 @@ def train(self, feed_dict):
         """
         for key in feed_dict.keys():
             feed_dict[key] = _utils.convert_shared_float_array_to_numpy(feed_dict[key])
-        feed_dict['labels'] = feed_dict['labels'].reshape(self.batch_size, self.grid_shape[0], self.grid_shape[1],self.num_anchors, self.num_classes + 5)
+        feed_dict['labels'] = feed_dict['labels'].reshape(-1, self.grid_shape[0], self.grid_shape[1],self.num_anchors, self.num_classes + 5)
```
Jarvi-Izana: not sure why here is not using `None`. what does `-1` mean? It's better if you add a line comment.

Jarvi-Izana:
> ObjectDetectorTest::test_single_image, ObjectDetectorTest::test_dict_annotations.
> Basically the code crash whenever the number of example isn't a multiply of the batch size (or smaller than batch size).
> Like batch size 32, num_example = 100 will crash.
> batch size 2, num_example = 1 (single image) will crash.
> Note that I also change the C++ code in order to fix those crashes.

Okay, the test case should be covered. I did some similar tests for DC, and you should be able to find it in `test_dc_predictions.cxx`. Feel free to reuse some of the test drivers.

I'd also recommend you check what's inside OD data iterator and make sure it returns the actual size instead of the batch size.

Jarvi-Izana:
one more thing, could we use `pep8` format for code we adding in? We cannot enfore previous code to be pep8 style, but we can enfore the present code we are working on. I see prs from the community users, they don't give a care to coding style since they follow our style. That's a bad feedback loop.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
jakesabathia2: So here I changed it to be a dynamic batch size, by declaring the shape dimension to be None.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
jakesabathia2: If we set the dimension to -1, the number will be automatically inferred.
Say you have a matrix which size [2,6]
if you reshape it to [-1,4],
the first element will be 3.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
Jarvi-Izana: Oh well, I think you can still pass batch_size but in your backend, you use actual size. This means the inference for the right size is mandatory and you don't need to explicitly enable it.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
shantanuchhabra: Yeah, could we add a comment explaining this line?

Jarvi-Izana: great. just make sure the behavior is consistent. You can use -1 for placeholder too if you think that's a trivial change to make.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -337,7 +337,7 @@ def loss_layer(self, predict, labels):
         kr_obj_ij = _tf.stop_gradient(resp_box * gt_conf)
 
         kr_noobj_ij = 1 - kr_obj_ij
-        s = 1 / (self.batch_size * self.grid_shape[0] * self.grid_shape[1])
+        s = 1 / (batch_size * self.grid_shape[0] * self.grid_shape[1])
```
shantanuchhabra: What is `s`?

jakesabathia2: It is a normalization constant for the loss function I guess ;)

PR #2572, opened 2019-11-06
Enable C++ implementations of deep-learning toolkits by default

nickjong:


src/python/turicreate/toolkits/_internal_utils.py
```diff
@@ -32,10 +32,15 @@
               UnityGraphProxy: (lambda x: _SGraph(_proxy=x))}
 
 def _read_env_var_cpp(var_name):
+    """
+    Reads environment variables used to enable/disable the new C++
+    implementations of existing Python toolkits.
+    """
+    # Default to using the C++ implementation for toolkits that have one.
+    # TODO: Remove the old implementations, and this function, once the dust has
```
srikris: Can you file a GitHub issue for this todo :)

nickjong: Sorry, I don't believe in GH issues that are pure code-cleanup. They don't get done. Seeing a reminder in the code is more useful, IMHO 😛

srikris: I wasn't suggesting you remove the comment :)

nickjong:
Merging now to unblock testing, since the affected tests are all marked as skipped or xfail anyway

PR #2577, opened 2019-11-06
AC unit tests passing turned on with TF

shreyajain17:


src/python/turicreate/test/test_activity_classifier.py
```diff
@@ -310,6 +305,7 @@ def test_predict(self):
             expected_len = self._calc_expected_predictions_length(self.data)
             self.assertEqual(len(preds), expected_len)
 
+    @unittest.skipIf(IS_PRE_6_0_RC, 'Requires MXNet')
```
nickjong: Why skip and not `xfail`? And this doesn't actually require MXNet, does it?

shreyajain17: I will make it xfail. It doesn't require MXNet but only passes in MXNet.

nickjong: I'm not sure it's accurate to say it passes in MXNet... It passes if you manually install MXNet and manually disable the CPP path. The message here definitely doesn't capture/reflect that!

PR #2578, opened 2019-11-06
OD unit test passing turned on with TF

jakesabathia2:
Get rid of MXNET from OD to make it testable.


src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -345,7 +352,27 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
...
+            'grid_height': params['grid_shape'][0],
+            'grid_width': params['grid_shape'][1],
+            'max_iterations': num_iterations,
+            'mlmodel_path' : params['mlmodel_path'],
+            'classes' : classes
+
+        }
+        model = _tc.extensions.object_detector()
+        model.train(data=dataset, annotations_column_name=annotations, image_column_name=feature, options=tf_config)
+        return ObjectDetector_beta(model_proxy=model, name="object_detector")
```
nickjong: Why can't we move this earlier and avoid more of the `if not USE_CPP` cases above?

src/python/turicreate/test/test_object_detector.py
```diff
@@ -199,6 +198,7 @@ def test_create_with_empty_dataset(self):
         with self.assertRaises(_ToolkitError):
             tc.object_detector.create(self.sf[:0])
 
+    @pytest.mark.xfail(IS_PRE_6_0_RC, reason='Requires MXNet')
```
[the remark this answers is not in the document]
jakesabathia2: Okay I think it doesn't, I just want to mark out those test cases which aren't passes now.

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -182,12 +182,17 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
...
-    from ._mx_detector import YOLOLoss as _YOLOLoss
-    from ._model import tiny_darknet as _tiny_darknet
-    from ._sframe_loader import SFrameDetectionIter as _SFrameDetectionIter
-    from ._manual_scheduler import ManualScheduler as _ManualScheduler
-    import mxnet as _mx
-    from .._mxnet import _mxnet_utils
+
+    if not USE_CPP:
+        # Use MxNet
+        print('shit')
```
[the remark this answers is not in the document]
jakesabathia2: Sorry I will make sure all debug message are deleted next time.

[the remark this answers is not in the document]
Jarvi-Izana: there are a bunch of profanity filter git plugins to make sure your commit is clean. ;-).

jakesabathia2:
1. Fix batch size issue for OD (pad each data batch to full batch_size).
fix: ObjectDetectorTest::test_dict_annotations, ObjectDetectorTest::test_single_image
2. Fix the crash when an image doesn't have annotations.
fix: ObjectDetectorTest::test_dict_annotations
3. Reset tf graph when create model 
fix: ObjectDetectorTest::test_export_coreml, ObjectDetectorTest::test_export_coreml_with_non_maximum_suppression

jakesabathia2:
Now the OD should run without any crashes and failures.

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -35,6 +35,11 @@ shared_float_array convert_to_shared_float_array(
 
 std::vector<image_annotation> convert_to_image_annotation(
     shared_float_array augmented_annotation) {
+  // Check if the annotation is empty
+  if (augmented_annotation.size() == 1) {
```
nickjong: Please check for `.dim()` being 0 instead. Size equal 1 doesn’t read empty in the same way

src/ml/neural_net/image_augmentation.cpp
```diff
@@ -198,8 +204,15 @@ image_augmenter::result float_array_image_augmenter::prepare_images(
   float_array_result augmented_data = prepare_augmented_images(input_to_tf_aug);
 
   // Convert augmented_data to the data structure needed
-  res.image_batch = augmented_data.images;
-  for (size_t a = 0; a < n; a++) {
+  std::vector<float> result_array(n * h * w * c);
+  size_t image_size = augmented_data.images.size();
+  const float* start_address = augmented_data.images.data();
+  const float* end_address = (float*)(start_address + image_size);
```
nickjong: The `float *` should be unnecessary 

For future reference, don’t use C-style casts in C++ code anyway

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -300,7 +324,6 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
     # to be problematic to run independently of a MXNet-powered neural network
     # in a separate thread. For this reason, we restrict IO threads to when
     # the neural network backend is MPS.
-    io_thread_buffer_size = params['io_thread_buffer_size'] if use_mps else 0
```
nickjong: Why did this move? (And without its comment?)

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -182,12 +182,16 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
...
+
+    if not USE_CPP:
+        # Use MxNet
+        from ._model import tiny_darknet as _tiny_darknet
+        import mxnet as _mx
+        from .._mxnet import _mxnet_utils
+        from ._mx_detector import YOLOLoss as _YOLOLoss
+        from ._sframe_loader import SFrameDetectionIter as _SFrameDetectionIter
+        from ._manual_scheduler import ManualScheduler as _ManualScheduler
+
```
nickjong: Could we just move these down until they are needed? It seems weird to import them conditionally here

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -267,6 +271,28 @@ def create(dataset, annotations=None, feature=None, model='darknet-yolo',
         'mlmodel_path': pretrained_model_path,
     }
 
+    if batch_size < 1:
+        batch_size = 32  # Default if not user-specified
```
nickjong: The C++ implementation defines the default itself. If the user does not specify a value, pass no batch size to use the heuristic for computing batch size defined in C++

src/ml/neural_net/image_augmentation.cpp
Jarvi-Izana: it seems like empty annotation is not covered by our python unit test.

jakesabathia2: @Jarvi-Izana I think it is covered:
```
for j in range(rs.randint(max_num_boxes_per_image)):
            left, right = np.sort(rs.randint(0, img_shape[1], size=2))
            top, bottom = np.sort(rs.randint(0, img_shape[0], size=2))
```
the bounding box number is generated randomly,
so that why I found this bug ;) .

src/ml/neural_net/image_augmentation.cpp
jakesabathia2: Good learn ;)

Jarvi-Izana: besides, you don't need to cast type for pointer arithmetic.

src/ml/neural_net/image_augmentation.cpp
Jarvi-Izana: lol, good

src/python/turicreate/test/test_object_detector.py
```diff
@@ -329,6 +328,7 @@ def test_export_coreml(self):
 
     @unittest.skipIf(_mac_ver() < (10, 14),
         "Non-maximum suppression is only supported on MacOS 10.14+.")
+
```
nickjong: Why are we adding a blank line here? Does it affect the scope of the annotation?

jakesabathia2: I'll delete it

Issue #1974, opened 2019-05-29
Drawing classifier evaluation object has additional keys that it should not have


srikris:
There seems to be more in this dictionary that is needed. 

```
In [23]: evals.keys()
Out[23]:
['f1_score',
 'roc_curve',
 'labels',
 'precision',
 'label_metrics',
 'sorted_labels',
 'training_time',
 'conf_mat',
 'num_test_examples',
 'confidence_threshold',
 'num_classes',
 'max_iterations',
 'confidence_metric_for_threshold',
 'training_loss',
 'auc',
 'confusion_matrix',
 'feature',
 'test_data',
 'num_examples',
 'recall',
 'hesitant_threshold',
 'model_name',
 'accuracy']
```

znation:
@srikris Which ones should it have, and which ones should it not have?

srikris:
I think it should only have these:

```
['f1_score',
 'roc_curve',
 'precision',
 'auc',
 'confusion_matrix',
 'recall',
 'accuracy']
```

TobyRoseman:
This is related to #1785. Some of those additional keys/values are very costly to generate, particularly when there are a large number of classes.

nickjong:
I think the main issue here is that many of these keys are not actually metrics, whereas previously the caller could iterate over the keys of the evaluation dictionary and assume that each key was an evaluation metric

syoutsey:
To verify after implementing in C++.

srikris:
Verified fix for 6.0

```
In [16]: metrics.keys()
Out[16]:
['f1_score',
 'auc',
 'recall',
 'precision',
 'log_loss',
 'roc_curve',
 'confusion_matrix',
 'accuracy']
```

PR #2579, opened 2019-11-07
API breaking changes for 6.0

### LINKED ISSUE #1759: Sketch.num_na() is broken for float
According to documentation, Sketch.num_na() should return the exact number of missing values (https://apple.github.io/turicreate/docs/api/generated/turicreate.Sketch.html#turicreate-sketch), however, when `dtype` is float,  it always returns 0.

```
sf = tc.SFrame({'A':[1.0, 2.0] + [float('nan')]*100})
print sf['A'].summary().num_na() 
print len(sf['A']) - len(sf['A'].dropna()) 
```

```
sf = tc.util.generate_random_sframe(10000, 'R', random_seed=11)
print sf['X1-R'].dtype
print sf['X1-R'].summary().num_na() 
print len(sf) - len(sf.dropna())
print sum(sf['X1-R'].apply(lambda x : math.isnan(x)))
```

TobyRoseman:
Actually make the API changes we warned about. Most of these warnings where add in #2321. The other warnings have been around even longer, in some cases **much** longer.

These changes fix #1759 and  #1974.


srikris:
@TobyRoseman Given that this only fails the 3 unit tests you should consider merging this.

nickjong:
Yes, given that the three test failures are handled by #2581

TobyRoseman:
@srikris & @nickjong - Thanks for the info. Merging now. 

It's interesting, none of those tests fail for me locally.

PR #2580, opened 2019-11-07
Disabling metrics calculation after train in OD

shreyajain17:


src/toolkits/object_detection/object_detector.hpp
```diff
@@ -58,7 +58,7 @@ class EXPORT object_detector: public ml_model_base {
   virtual void resume_training(gl_sframe data, variant_type validation_data);
   virtual void iterate_training();
   virtual void synchronize_training();
-  virtual void finalize_training();
+  virtual void finalize_training(bool use_update_metrics);
```
nickjong: This method is exposed via the C-API registration below, and I think we want it to continue to work with no arguments. For now, could you update the registration and supply a default that does compute the metrics?

In addition, I think the name `use_update_metrics` could be more clear. How about something like `disable_metrics` or `compute_final_metrics`?

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -413,6 +413,13 @@ void object_detector::train(gl_sframe data,
                             variant_type validation_data,
                             std::map<std::string, flexible_type> opts)
 {
+  auto use_update_metrics_iter = opts.find("use_update_metrics");
+  if (use_update_metrics_iter == opts.end()) {
+    log_and_throw("Expected option \"use_update_metrics\" not found.");
+  }
```
nickjong: Let's not require the option. We can have a usable default value here, unlike `mlmodel_path`

jakesabathia2:
Please note that for the python unit test,
they also check for the model fields,
this changes might lead to some failures,
make sure you run/modify those test:
`test__list_field` and `test_get` might b e the first two you would like to check ;)

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -413,6 +413,15 @@ void object_detector::train(gl_sframe data,
                             variant_type validation_data,
                             std::map<std::string, flexible_type> opts)
 {
+  auto compute_final_metrics_iter = opts.find("compute_final_metrics");
+  bool compute_final_metrics;
+  if (compute_final_metrics_iter == opts.end()) {
+    compute_final_metrics = true;
+  } else {
+    compute_final_metrics = compute_final_metrics_iter->second;
+  }
```
nickjong: Let's simplify this to e.g.
```
bool foo = true;
if (foo_iter != opts.end()) {
  foo = foo_iter->second;
}

File: src/toolkits/object_detection/object_detector.cpp [hunk 4]
```diff
@@ -414,13 +414,10 @@ void object_detector::train(gl_sframe data,
                             std::map<std::string, flexible_type> opts)
 {
   auto compute_final_metrics_iter = opts.find("compute_final_metrics");
-  bool compute_final_metrics;
-  if (compute_final_metrics_iter == opts.end()) {
-    compute_final_metrics = true;
-  } else {
+  bool compute_final_metrics = true;;
```
nickjong: Let's remove the extraneous semicolon, too

PR #2582, opened 2019-11-07
enable evaluate test

Jarvi-Izana:
#2533 is in.


Jarvi-Izana:
The test passed without stoke-based data input.

PR #2583, opened 2019-11-07
Marked drawing tests xfail

shantanuchhabra:


Jarvi-Izana:
#2582 I enable the evaluate for all.

Jarvi-Izana:
okay, I see I still need stroke model. I will leave it there.

PR #2584, opened 2019-11-07
OSOD unit tests turned on

shreyajain17:


shreyajain17:
> Have you tested this without MXNet installed? How does it interact with Henry's PR?
Yes this is verified against Henry's changes.

PR #2586, opened 2019-11-08
Ac export coreml file name

jakesabathia2:
It turns out the export coreml function first checks for the stream is good or not.
It looks like the temporary file is invalid.


src/python/turicreate/test/test_activity_classifier.py
```diff
@@ -482,10 +481,9 @@ def test_gpu_save_load_export(self):
                                     session_id=self.session_id)
                 with test_util.TempDirectory() as filename:
                     model.save(filename)
-                    tc.config.set_num_gpus(out_gpus)
                     model = tc.load_model(filename)
 
-                with test_util.TempDirectory() as filename:
```
TobyRoseman: I think I like the old way better. Is there a reason you need to change these two lines of code?

jakesabathia2: In C++ the export_coreml check the outfield is good() nor not,
and it seems like with `test_util.TempDirectory()` is treated as a bad osstream ? 

So I follow the code here instead:
https://github.com/apple/turicreate/blob/26a7716fe9bec9d2267a5f42fc5ef633cad3bc50/src/python/turicreate/test/test_activity_classifier.py#L314

srikris:
@jakesabathia2 The only failing tests are drawing classifier which has been fixed in master so you should be good to merge this.

PR #2587, opened 2019-11-08
Enable skipped tests in `test_audio_functionality`

fareeha:


fareeha:
> This doesn't require any changes from #2525? What is the result of running this test suite with these changes?

Right now on the #2525 branch I see 6 test failures that look like requires changes on the TF side, rather than the test. I propose letting the tests be active and if changes are needed to made to the tests, we can do it in a separate PR after #2525 is complete and merged.

TobyRoseman:
If stuff shouldn't be skipped any more, I think we should just delete those lines of code rather than comment them out.

nickjong:
If the tests you're un-skipping are failing, they need to be marked `xfail` instead of skipped.

nickjong:
And to be clear, this PR should reflect the test state on master, not with #2525. That PR should update the tests accordingly, after this goes in

srikris:
I don't think we need this PR anymore. I think @TobyRoseman already enabled the skipped tests in #2535 which is now in master.

fareeha:
Tests working. Skipping no longer needed.

PR #2588, opened 2019-11-08
Enabling TC to install manylinux2010 dependencies + Dockerfile for CentOS 6

shantanuchhabra:


.gitlab-ci.yml
```diff
@@ -205,7 +205,7 @@ build_cpp_linux_no_remotefs:
     DOCKER_DRIVER: overlay2
   script:
     - apt-get -y update
-    - apt-get -y install apt-transport-https ca-certificates gnupg software-properties-common wget bzip2 build-essential xxd
+    - apt-get -y install apt-transport-https ca-certificates gnupg software-properties-common wget bzip2 build-essential xxd libbz2-dev
```
znation: Why is this change needed? Nothing else we changed should be related to this part of the build.

scripts/install_on_docker.sh
```diff
@@ -0,0 +1,199 @@
+# vim: set ft=dockerfile:
```
znation: This script seems redundant with the Dockerfile for CentOS 6, and doesn't seem to be used anywhere. Can we remove it?

.gitlab-ci.yml
abhishekpratapa: My bad, I'll revert and start a new branch

scripts/install_on_docker.sh
shantanuchhabra: Done!

PR #2590, opened 2019-11-11
Added default argument to get test_object_detector to compile.

abhishekpratapa:
@nickjong As per @jakesabathia2 suggestion this default value is set to false


nickjong:
Please just add the appropriate argument explicitly

srikris:
Fixes #2599

PR #2596, opened 2019-11-11
Load model from s3 path

### LINKED ISSUE #2595: Can not load saved model with S3 path
We have the ability to load save SFrames and SArrays using S3 paths. However we can not load saved model using S3 paths. We get the error: `ValueError: Unsupported protocol `.

We should support loading model with S3 paths.

TobyRoseman:
Fixes #2595

### DIFF
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -68,7 +68,7 @@ def load_model(location):
         dir_archive_exists = file_util.exists(os.path.join(model_path, 'dir_archive.ini'))
     else:
         model_path = location
-        if protocol in ['http', 'https']:
+        if protocol in ['http', 'https', 's3']:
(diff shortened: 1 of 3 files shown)


src/python/turicreate/toolkits/_model.py
```diff
@@ -68,7 +68,7 @@ def load_model(location):
         dir_archive_exists = file_util.exists(os.path.join(model_path, 'dir_archive.ini'))
     else:
         model_path = location
-        if protocol in ['http', 'https']:
+        if protocol in ['http', 'https', 's3']:
```
srikris: Is there an easy way to test this? With boto maybe?

TobyRoseman: Sadly, no. Right now we don't have any S3 unit test.

PR #2598, opened 2019-11-12
Userguide updates and fixes

syoutsey:
Make updates to paths in userguide, make consistent our model naming scheme.

### DIFF
--- a/userguide/style_transfer/README.md
+++ b/userguide/style_transfer/README.md
@@ -34,10 +34,10 @@ test_images = tc.load_images('test/')
 stylized_images = model.stylize(test_images)
 
 # Save the model for later use in Turi Create
-model.save('mymodel.model')
+model.save('style-transfer.model')
 
 # Export for use in Core ML
-model.export_coreml('MyStyleTransfer.mlmodel')
+model.export_coreml('MyCustomStyleTransfer.mlmodel')
 ```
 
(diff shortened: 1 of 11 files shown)


userguide/style_transfer/README.md
```diff
@@ -34,10 +34,10 @@ test_images = tc.load_images('test/')
 stylized_images = model.stylize(test_images)
 
 # Save the model for later use in Turi Create
-model.save('mymodel.model')
+model.save('style-transfer.model')
 
 # Export for use in Core ML
-model.export_coreml('MyStyleTransfer.mlmodel')
+model.export_coreml('MyCustomStyleTransfer.mlmodel')
```
TobyRoseman: The export to Core ML page for style transfer makes reference to `MyStyleTransfer.mlmodel`. Do we want to update that section as well?

syoutsey: I didn't want to change it there because the screenshots reference that title, I thought consistency across toolkits in the `Getting Started` section was better than exactly matching the `Advanced` section. What do you think @TobyRoseman?

TobyRoseman: `MyStyleTransfer.mlmodel` isn't that bad of a name. So I'd probably choose to have the style transfer section of the user guide be consistent with itself, but I think either way is reasonable.

PR #2601, opened 2019-11-12
Enhancement_SFrame_plus_operator_should_append

### LINKED ISSUE #2569: SFrame plus operator should append
Given two SFrames:
```python
sf1 = tc.SFrame({'a': [1,2,3]})                                                                                                                     
sf2 = tc.SFrame({'a': [6,7]})                                                                                                                       
```
This does not work: `sf1 + sf2`. However this does work `sf1.append(sf2)`. 

Expected: `sf1 + sf2` should work and it should be equivalent to calling`sf1.append(sf2)`. This can be achieved by implementing `SFrame.__add__` and having it call `self.append`.

dhivyaaxim:
This PR resolves #2569 
SFrame plus operator should append

### DIFF
--- a/src/python/turicreate/data_structures/sframe.py
+++ b/src/python/turicreate/data_structures/sframe.py
@@ -1958,7 +1958,7 @@ class SFrame(object):
     
     def __add__(self,other):
         """
-        Return sum of frames
+        Return append one frames to other
         """
         self=self.append(other)
         return self
--- a/src/python/turicreate/data_structures/sframe.py
+++ b/src/python/turicreate/data_structures/sframe.py
@@ -1955,7 +1955,13 @@ class SFrame(object):
         Just make it match default behavior in Python 2.
         '''
         return id(self) // 16
-
+    
+    def __add__(self,other):
+        """
+        Return sum of frames
+        """
+        self=self.append(other)
+        return self
 
     def __repr__(self):
         """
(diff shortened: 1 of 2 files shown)


src/python/turicreate/test/test_sframe.py
```diff
@@ -704,6 +704,13 @@ def test_transform(self):
 
         sa = sf.apply(lambda x: x['int_data'] + x['float_data'], float)
         self.__assert_sarray_equal(sf['int_data'] + sf['float_data'], sa)
+    
+    def test_add(self):
+        sf1 = SFrame({'a': [1,2,3]})
+        sf2 = SFrame({'a': [6,7]})
+        sf1=sf1+sf2
+        self.assertTrue({'a': 7} in sf1)
```
srikris: You probably want a stronger comparison than this. Can you also make the unit test work with 2 columns just in case.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -1955,7 +1955,13 @@ def __hash__(self):
         Just make it match default behavior in Python 2.
         '''
         return id(self) // 16
-
+    
+    def __add__(self,other):
+        """
+        Return sum of frames
```
TobyRoseman: It's not a `sum`. You're returning one SFrame append to the other.

src/python/turicreate/test/test_sframe.py
TobyRoseman: @dhivyaaxim - you should also check that `sf1` is exactly what you expect not just that it contains some elements you expect.

PR #2603, opened 2019-11-12
Return to CorerML should be a void function.

srikris:
- Fixed the return for export_coreml


abhishekpratapa:
copypasta. LMAO. Good catch.

PR #2618, opened 2019-11-12
PyTest failures for `export_to_coreml` updated. 

abhishekpratapa:
- Added prefix to style image to avoid naming conflicts
- Added `num_styles` to index to add support for multiple styles

### DIFF
--- a/src/toolkits/coreml_export/neural_net_models_exporter.cpp
+++ b/src/toolkits/coreml_export/neural_net_models_exporter.cpp
@@ -6,6 +6,7 @@
 
 #include <toolkits/coreml_export/neural_net_models_exporter.hpp>
 
+#include <cctype>
 #include <sstream>
 
 #include <core/logging/assertions.hpp>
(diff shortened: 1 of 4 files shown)


src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -391,21 +392,35 @@ std::shared_ptr<coreml::MLModelWrapper> export_style_transfer_model(
...
 
   set_array_feature(
       model_desc->add_input(), "index",
-      "Style index array (set index I to 1.0 to enable Ith style)", {1});
+      "Style index array (set index I to 1.0 to enable Ith style)", {num_styles});
+  /*
+   * prefix style with stylized and capitalize the following identifier, this
+   * avoids name clashes with the `content_feature` for exporting to CoreML.
+   */
+  style_feature[0] = toupper(style_feature[0]);
```
nickjong: Nit: could we use `std::toupper` to make it more clear this is from the standard library?

abhishekpratapa: Yup!

PR #2621, opened 2019-11-13
Pytest `test_stylize_success` removed xfail

abhishekpratapa:
- Added a batch size of one to predict due to Tensorflow ND array limitations.

### DIFF
--- a/src/toolkits/style_transfer/style_transfer.cpp
+++ b/src/toolkits/style_transfer/style_transfer.cpp
@@ -216,24 +216,16 @@ float_array_map prepare_batch(std::vector<st_example>& batch, size_t width,
 
   float_array_map map;
 
-  shared_float_array content_wrap = shared_float_array::wrap(
+  map["input"] = shared_float_array::wrap(
       std::move(content_array), {batch_size, height, width, channels});
-  map.emplace("input", content_wrap);
+  map["index"] = shared_float_array::wrap(std::move(index_array), {batch_size});
 
-  shared_float_array index_wrap =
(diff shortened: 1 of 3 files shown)


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -224,6 +224,12 @@ float_array_map prepare_batch(std::vector<st_example>& batch, size_t width,
       shared_float_array::wrap(std::move(index_array), {batch_size});
   map.emplace("index", index_wrap);
 
+  shared_float_array image_width_wrap = shared_float_array::wrap(width);
+  map.emplace("width", image_width_wrap);
```
nickjong: This is fine, but FYI I think this should also be perfectly acceptable:
```
map["width"] = shared_float_array::wrap(width);
```
Technically this form might be slightly less efficient, since it default constructs the value for the key before overwriting it with the RHS, but generally I err on the side of clarity and simple code

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -233,6 +239,22 @@ float_array_map prepare_batch(std::vector<st_example>& batch, size_t width,
   return map;
 }
 
+// Preparing Batch for prediction. Since the tensorflow implementation of style
+// transfer doesn't have multiple batches supported in predict this function
+// takes exactly one st_example as an argument.
+float_array_map prepare_predict(st_example& example) {
```
nickjong: Looks like this could be `const st_example&`

PR #2597, opened 2019-11-11
fix inconsistency in TF and CoreML inference outcome

Jarvi-Izana:
This pr includes change:
1. adding a preprocessing (scale factor of 1/255.f) layer for CoreML spec. @jakesabathia2 found this issue.
2. fix extremely slow convergence of TF training by adding data normalization in `dc_data_iterator.cpp`, where each pixel is divided by 255. 
3. fix the issue that CoreML doesn't work consistently with TF since TF is using non-normalized input data for training whereas CoreML normalizes input data before training.
4. put @fareeha 's NN graph conversion from TF to CoreML back.

Issues to be solved in the next step:
1. src/python/turicreate/toolkits/_tf_utils.py is problematic.
2. the TF trains but doesn't update the weights view, as I discovered with @fareeha. This means the initial weights view is the same as the eventual weights view exported to our Cpp backend. That's why we see the prediction probability to be a random number around .5.

__MY SUMMARY__
- [x] model spec test needs to be modified to accommodate change brought from adding a preprocessing layer.
- [x] pass relevant tests of aligning the behavior of CoreML and TF.
- [x] fix inconsistent preprocessing
- [x] fix incorrect weights export
- [x] align pooling padding with TC 5.8
- [x] fix the hardcoded input name in export_coreml
- [x] populate changes to data iterator test
- [x] pass **ALL** relevant DC Cpp test

PR #2623, opened 2019-11-13
bug fix for 2552 change

Jarvi-Izana:
#2552 changes the feature state from the plural to single.

The master fails on the python test because it used to use state `features` instead of `feature`. This change will unblock #2597.


Jarvi-Izana:
run `pytest src/python/turicreate/test/test_drawing_classifier.py` and you will find the anwser.

Jarvi-Izana:
False alarm. The change is already checked in by #2597.

PR #2627, opened 2019-11-13
DC: accept integer labels + fix up predict_topk

shantanuchhabra:


src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -700,12 +699,16 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
     result = result.apply(max_prob_label, class_labels.front().get_type());
 
   } else if (output_type == "probability") {
+
+    if (read_state<flex_int>("num_classes") > 2) {
```
Jarvi-Izana: Nah, I think this is fundamentally wrong. `num_classes` means the size of probability space. You predict the label with the largest probability in that probability space. It's sitll single class prediction. Correct me if I'm wrong. @nickjong

src/toolkits/drawing_classifier/dc_data_iterator.hpp
```diff
@@ -51,10 +52,8 @@ class data_iterator {
      *
      * If empty, then the labels will be inferred from the data. If non-empty,
      * an exception will be thrown upon encountering an unexpected label.
-     *
-     * \TODO: This should be a flex_list to accomodate integer labels!
```
Jarvi-Izana: okay, you already said, it's a feature change. The underlying representation of class labels is changed.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -760,30 +763,40 @@ gl_sframe drawing_classifier::predict_topk(gl_sframe data,
...
+    for (flexible_type class_index : topk_class_index_list) {
+      topk_class.push_back(class_labels[class_index]);
     }
     return topk_class;
   };
 
   dc_predictions.add_column(
       dc_predictions.apply(get_class_name, flex_type_enum::LIST), "class");
 
+  auto generate_ranks = [=](const sframe_rows::row& row) {
```
Jarvi-Izana: yeah, this is helpful for visualization. Good.

Jarvi-Izana:
Again, this is a known todo feature change of label representation, not a fix for predict. Predict works as expected after reviewing the change. Not being able to receive an integer label is caused by label data representation not predict. `fix predict` confuses me.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: after I think twice, this is redundant and waste of storage. You can easily find the start of a rank interval by id % k == 0

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: This is an effort to match our behavior in 5.8: https://github.com/apple/turicreate/blob/master/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py#L941

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: This is to match the output with what we used to have in 5.8.
https://github.com/apple/turicreate/blob/master/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py#L798

Note that `id` is not like a primary key here, which is why I changed the implementation around here.

src/toolkits/drawing_classifier/dc_data_iterator.hpp
shantanuchhabra: Not being able to support integer targets was surfaced by adding the stroke-based model to our Python unit tests (the stroke-based data we generated used integer targets). Since those changes were not that code-heavy, I made those changes as part of this PR. Since we supported integer targets in 5.8, we must also support them in 6.0.

shantanuchhabra:
> Again, this is a known todo feature change of label representation, not a fix for predict. Predict works as expected after reviewing the change. Not being able to receive an integer label is caused by label data representation not predict. `fix predict` confuses me.

Please see inline responses.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: id is the primary key for the ranking result. Thanks.

Jarvi-Izana: If it's that case, you are still not doing the same thing.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: Read the line below. Think twice.

shantanuchhabra: Sorry, can you elaborate?

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Could you explain how it's not doing the same thing?

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: https://github.com/apple/turicreate/blob/96a2c5c4035f3177e0d9b2e5d87fd575d816a8a7/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py#L935

At first glance, `self.classes` is the column name used to retrieve the predicted results.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: output id like [0001112222]; what you need to do is to change the primary key of the result by just simply divide k. That's the simplest change.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -793,21 +806,17 @@ gl_sframe drawing_classifier::predict_topk(gl_sframe data,
...
 
   // stack data into a column with single element for each row
   // stack the labels first
-  gl_sframe stacked_class =
-      gl_sframe({{"class", dc_predictions["class"]}}).stack("class", "class");
-
-  result.add_column(stacked_class["class"], "class");
-  result.add_column(gl_sarray::from_sequence(0, stacked_class.size()),
-                    "id");
+  result = result.stack("class", "class");
```
shantanuchhabra: Reference for discussion

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana:
```
 auto max_prob = [=](const flexible_type& ft) {
       const flex_vec& prob_vec = ft.get<flex_vec>();		       const flex_vec& prob_vec = ft.get<flex_vec>();
       auto max_it = std::max_element(prob_vec.begin(), prob_vec.end());		       auto max_it = std::max_element(prob_vec.begin(), prob_vec.end());
       return *max_it;		       return *max_it;
     };		     };

```
does the single class prediction.

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: How is that different from https://github.com/apple/turicreate/pull/2627/files/e35672aba1602f8b457bfbf1a744724a94daaf61#r346077530 ?

Jarvi-Izana: I guess ` std::iota(ranks.begin(), ranks.end(), 0);` inside of your index generation.

shantanuchhabra: `generate_ranks` is generating the "rank" column whereas `add_row_number` followed by the `stack` in https://github.com/apple/turicreate/pull/2627/files/e35672aba1602f8b457bfbf1a744724a94daaf61#r346077530 is responsible for the "id" column.

Jarvi-Izana: besides, the previous desgin is bad. How can I index on rows? The previous predict_topk is aligned with AC. And we should keep consistent among toolkits.

Jarvi-Izana: Okay, now I know what you are doing. You stack all of them at once in the end. You should stack both rank and class (or probability) together. print out the result first and see the results.

shantanuchhabra: Why should we align with AC and not with DC in 5.8?

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Would removing this `log_and_throw` be consistent with our behavior in 5.8? How do you recommend we stay consistent with our behavior in 5.8?

Jarvi-Izana: I don't know what to say already. This will be my last comment on this topic. An even simple and silly question, if we assume this logic was correct, why we wouldn't add the same thing for topk? You haven't figured out the semantic difference of `self.classes` and `num_classes`.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: because previous behavior is not good. You should at least provide a primary key for the user to index on sframe rows. I will stop here. Ask Nick.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -760,30 +763,40 @@ gl_sframe drawing_classifier::predict_topk(gl_sframe data,
   // store the index in a column "rank"
```
nickjong: Stale comment....

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: I think I understand this code now. I suggest simplifying it dramatically, into a single apply followed by a single stack.

For example, suppose the class labels are ["alpha", "bravo", "charlie"]. Then we want to apply each probability vector as follows:
1. Compute the arg-sort top k indices.
2. Given the k class indices in order, produce a `flex_dict` as follows:
```
flex_dict result;
for (int rank = 0; rank < k; ++k) {
  flex_int class_index = top_k_indices[rank];
  flexible_type value;
  if (output_type=="probability"]) {
    value = prob_vec[class_index];
  } else {
    value = rank;
  }
  result.emplace_back(classes[class_index], value);
}

Then unstack the result

Comment (nickjong) [reply src/toolkits/drawing_classifier/drawing_classifier.cpp]: (But yes, I don't think we have time to design a new output format for `predict_topk`. What we're implementing here is what is used in the existing supervised learning toolkits, and is the established precedent for now)

File: src/toolkits/drawing_classifier/dc_data_iterator.cpp [hunk 6]
```diff
@@ -59,21 +60,21 @@ simple_data_iterator::compute_properties(
     int i = 0;
     for (const flexible_type& label : classes.range_iterator()) {
       result.classes.push_back(label);
-      result.class_to_index_map[label] = i++;
+      result.class_to_index_map.push_back(std::make_pair(label, i++));
```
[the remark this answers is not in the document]
shantanuchhabra: Fixed.

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -59,21 +60,21 @@ simple_data_iterator::compute_properties(
...
     }
 
-    // Use the map to verify that we only encountered expected labels.
+    const flex_dict_view class_to_index_map(result.class_to_index_map);
+
     for (const flexible_type& ft : classes.range_iterator()) {
-      std::string label(ft);  // Ensures correct overload resolution below.
-      if (result.class_to_index_map.count(label) == 0) {
-        log_and_throw("Targets contained unexpected class label " + label);
+      if (!class_to_index_map.has_key(ft)) {
```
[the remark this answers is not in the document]
shantanuchhabra: Fixed.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -700,12 +699,16 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
...
 
   } else if (output_type == "probability") {
+
+    if (read_state<flex_int>("num_classes") > 2) {
+      log_and_throw("Use probability_vector in case of multi-class classification.");
+    }
+
     auto max_prob = [=](const flexible_type& ft) {
       const flex_vec& prob_vec = ft.get<flex_vec>();
       auto max_it = std::max_element(prob_vec.begin(), prob_vec.end());
```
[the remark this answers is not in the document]
shantanuchhabra: Thanks for pointing this out @nickjong . PR to fix this coming soon!

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Fixed.

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -136,7 +129,7 @@ data_iterator::batch simple_data_iterator::next_batch(size_t batch_size) {
 
   float* next_drawing_pointer = batch_drawings.data();
   size_t real_batch_size = 0;
-  const flex_dict_view class_to_index_map(target_properties_.class_to_index_map);
+  const flex_list classes(target_properties_.classes);
```
nickjong: Let's avoid copying here. This can be
```
const flex_list& classes = target_properties_.classes;
```

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -760,65 +760,39 @@ gl_sframe drawing_classifier::predict_topk(gl_sframe data,
...
-  result.add_column(stacked_rank["rank"], "rank");
+  // Get a result column that we can stack, from the probability vector
+  result.add_column(dc_predictions["preds"].apply(compute_result_column,
+                                                  flex_type_enum::DICT),
+                    "result");
+  result = result.add_row_number();
+  std::vector<std::string> new_column_names;
+  new_column_names.push_back("class");
+  new_column_names.push_back("rank");
+  result = result.stack("result", new_column_names);
```
nickjong: Could this just be
```
result = result.stack("result", { "class", "rank" });
```
?

shantanuchhabra: @nickjong I tried that initially, but I got the following strange compile error:

```
/Users/schhabra/workspace/turicreate/src/toolkits/drawing_classifier/drawing_classifier.cpp:792:19: error: call to member function 'stack' is ambiguous
  result = result.stack("result", {"class", "rank"});
           ~~~~~~~^~~~~
/Users/schhabra/workspace/turicreate/src/core/data/sframe/gl_sframe.hpp:2310:13: note: candidate function
  gl_sframe stack(const std::string& column_name,
            ^
/Users/schhabra/workspace/turicreate/src/core/data/sframe/gl_sframe.hpp:2383:13: note: candidate function
  gl_sframe stack(const std::string& column_name,
            ^
```

This was odd since both candidate functions have different signatures. One takes in a `std::string& new_column_name` to stack a column of dtype array, whereas the other takes in `std::vector<std::string>& new_column_names` to stack a column of dtype dictionary.

src/toolkits/drawing_classifier/dc_data_iterator.cpp
shantanuchhabra: Done, thanks!

src/toolkits/drawing_classifier/drawing_classifier.cpp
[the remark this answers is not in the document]
shantanuchhabra: PR up at #2649

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: per offline discussion, it sounds reasonable to align with the previous behavior. I wonder if we also want to add an output type called `probability_vector`.

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: Weird. You should at least be able to do:
```
result = result.stack("result", std::vector<std::string>{"class", "rank"});
```
(Helping the compiler out by specifying the types more explicitly)

PR #2630, opened 2019-11-14
Removing duplicates based on a subset of columns

### LINKED ISSUE #2540: Removing duplicates based on a subset of columns
It would be useful to remove duplicate rows of an SFrame based on a subset of columns, similar to pandas' `drop_duplicates`.

In pandas:
```python
df = pd.DataFrame({'A': ['a', 'b', 'a'], 'B': ['b', 'a', 'c'], 'C': [1, 2, 3]})
df.drop_duplicates(subset=['A'])
```
Results in:
```
   A  B  C
0  a  b  1
1  b  a  2
```

As a workaround I used the following but I am not sure if `SELECT_ONE` preserves its choice across columns
```python
sf = tc.SFrame({'A': ['a', 'b', 'a'], 'B': ['b', 'a', 'c'], 'C': [1, 2, 3]})
sf.groupby('A', {col: tc.aggregate.SELECT_ONE(col) for col in sf.column_names() if col != 'A'})
```

dhivyaaxim:
This PR resolves #2540 
Removing duplicates based on a subset of columns

### DIFF
--- a/src/python/turicreate/data_structures/sframe.py
+++ b/src/python/turicreate/data_structures/sframe.py
@@ -2189,7 +2189,36 @@ class SFrame(object):
                     printed_sf[t] = self[t].astype(str)
         return printed_sf.head(num_rows)
     
-    def drop_duplicates(self, subset): 
+    def drop_duplicates(self, subset):
+        """
+        Return SFrame with duplicate rows removed, optionally only considering certain columns.
+
+        Parameters
+        ----------
+        subset : column label or sequence of labels, optional
+            Only consider certain columns for identifying duplicates, by default use all of the columns.
+
+        Examples
+        --------
+        >>> import turicreate as tc
+        >>> sf =tc.SFrame({'A': ['a', 'b', 'a','C'], 'B': ['b', 'a', 'b','D'], 'C': [1, 2, 1,8]})
+        >>> sf.drop_duplicates(subset=["A","B"])
+        Columns:
+	        A	str
+	        B	str
+	        C	int
+        Rows: 3
+        Data:
+        +---+---+---+
+        | A | B | C |
+        +---+---+---+
+        | b | a | 2 |
+        | C | D | 8 |
+        | a | b | 1 |
+        +---+---+---+
+        [3 rows x 3 columns]
+
+        """
         result =  all(elem in self.column_names()  for elem in subset)
         if result :
             return (self.groupby(subset, {col: aggregate.SELECT_ONE(col) for col in self.column_names() if col not in subset}))
--- a/src/python/turicreate/data_structures/sframe.py
+++ b/src/python/turicreate/data_structures/sframe.py
@@ -2190,8 +2190,11 @@ class SFrame(object):
         return printed_sf.head(num_rows)
     
     def drop_duplicates(self, subset): 
-        if subset[0] in self.column_names(): 
-            return self.groupby(subset[0], {col: aggregate.SELECT_ONE(col) for col in self.column_names() if col != subset[0]}) 
+        result =  all(elem in self.column_names()  for elem in subset)
+        if result :
+            return (self.groupby(subset, {col: aggregate.SELECT_ONE(col) for col in self.column_names() if col not in subset}))
+        else:
+            raise TypeError("subset  not in  sframe column")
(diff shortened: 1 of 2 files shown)


src/python/turicreate/data_structures/sframe.py
```diff
@@ -2189,6 +2189,12 @@ def _imagecols_to_stringcols(self, num_rows=10):
                     printed_sf[t] = self[t].astype(str)
         return printed_sf.head(num_rows)
 
+    def drop_duplicates(self, subset=None, keep='first', inplace=False): 
+        self=pandas.DataFrame(self) 
+        self=self.drop_duplicates(subset=subset, keep=keep, inplace=inplace) 
+        self=load_sframe(self) 
+        return self
+
```
hoytak: This would work, but we should have this function wrap the  groupby SELECT_ONE operator to perform this.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -2188,7 +2188,10 @@ def _imagecols_to_stringcols(self, num_rows=10):
                 if t in image_column_names:
                     printed_sf[t] = self[t].astype(str)
         return printed_sf.head(num_rows)
-
+    
+    def drop_duplicates(self, subset): 
+        if subset[0] in self.column_names(): 
+            return self.groupby(subset[0], {col: aggregate.SELECT_ONE(col) for col in self.column_names() if col != subset[0]})
```
hoytak: I believe this only works with duplicates in the first column.  I believe that this should be: 

```
self.groupby(subset, {col: aggregate.SELECT_ONE(col) for col in self.column_names()})
```

src/python/turicreate/test/test_sframe.py
```diff
@@ -216,6 +216,12 @@ def test_auto_parse_csv(self):
             self.assertEqual(sf.dtype, [float, int, str])
             self.__test_equal(sf, df)
 
+    def test_drop_duplicates(self): 
+        sf = SFrame({'A': ['a', 'b', 'a'], 'B': ['b', 'a', 'b'], 'C': [1, 2, 1]}) 
+        df1=sf.drop_duplicates(subset=['A']) 
+        sf1 = pd.DataFrame({'A': ['a','b'], 'B': ['b','a'], 'C': [1,2]}) 
+        self.__test_equal(df1,sf1)
```
hoytak: Let's add a couple of random tests here too.  Take a look at tc.util.generate_random_sframe; I think this test doesn't cover all the functionality (e.g. multiple columns in the subset.  

I'd create a function that uses pandas to test equality, but you also need to e.g. sort them to make sure they are equal.  

E.g. test `sorted(list(pandas_result)) == sorted(list(turi_result))`.  Group_by doesn't always preserve order.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -2188,7 +2188,13 @@ def _imagecols_to_stringcols(self, num_rows=10):
                 if t in image_column_names:
                     printed_sf[t] = self[t].astype(str)
         return printed_sf.head(num_rows)
-
+    
+    def drop_duplicates(self, subset): 
+        result =  all(elem in self.column_names()  for elem in subset)
```
TobyRoseman: Please add a docstring comment. This should explain what `drop_duplicates` does and also the meaning of `subset`.

hoytak: Once we get a doc string in here (ideally with a few examples), this is good to merge.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -2188,7 +2188,42 @@ def _imagecols_to_stringcols(self, num_rows=10):
...
                     printed_sf[t] = self[t].astype(str)
         return printed_sf.head(num_rows)
+    
+    def drop_duplicates(self, subset):
+        """
+        Return SFrame with duplicate rows removed, optionally only considering certain columns.
+
+        Parameters
+        ----------
+        subset : column label or sequence of labels, optional
```
TobyRoseman: This parameter isn't actually optional. You need to specify it. It doesn't by default use all the columns.

PR #2638, opened 2019-11-15
Osod save and load

jakesabathia2:


src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -248,8 +249,11 @@ def _load_version(cls, state, version):
         assert(version == cls._PYTHON_ONE_SHOT_OBJECT_DETECTOR_VERSION)
         # we need to undo what we did at save and turn the proxy object
         # back into a Python class
-        state['detector'] = _ObjectDetector._load_version(
-            state['detector'], state["_detector_version"])
+        if USE_CPP:
+            state['detector'] = _ObjectDetector_beta(state['detector'])
+        else:
+            state['detector'] = _ObjectDetector._load_version(
```
nickjong: Let’s not depend on the old OD implementation here, since we want to delete it. We should migrate the model instead, as we do for straight OD models (in _model.py)

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -248,8 +249,11 @@ def _load_version(cls, state, version):
         assert(version == cls._PYTHON_ONE_SHOT_OBJECT_DETECTOR_VERSION)
         # we need to undo what we did at save and turn the proxy object
         # back into a Python class
-        state['detector'] = _ObjectDetector._load_version(
-            state['detector'], state["_detector_version"])
+        if USE_CPP:
+            state['detector'] = _ObjectDetector_beta(state['detector'])
```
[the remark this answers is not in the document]
srikris: Before we merge this PR, it would be good to test it locally on a 5.8 saved model.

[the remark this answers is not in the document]
jakesabathia2: sound good :)

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
jakesabathia2: This is a good catch

jakesabathia2:
Should be able to load 5.8 or older version now.

src/python/turicreate/toolkits/_model.py
```diff
@@ -115,7 +115,17 @@ def load_model(location):
                     model = _extensions.style_transfer()
                     model.import_from_custom_model(model_data, model_version)
                     return cls(model)
-                    
+
+                if name=='one_shot_object_detector' and OD_USE_CPP:
+                    od_cls = MODEL_NAME_MAP['object_detector']
```
nickjong: What class is `od_cls`? Why do we need something different from `cls`?

src/python/turicreate/toolkits/_model.py
```diff
@@ -115,7 +115,17 @@ def load_model(location):
...
+                    if 'detector_model' in model_data['detector']:
+                        model_data['detector'] = od_cls(model_data['detector']['detector_model'])
+                    else:
+                        model = _extensions.object_detector()
```
nickjong: We should always be using extensions. We should never use the old Python class. We are going to delete it.

src/python/turicreate/toolkits/_model.py
jakesabathia2: Oh `cls` here is an one-shot-object_detector model. It contains a object_detector model inside the class, which I refer to `od_cls` here.

src/python/turicreate/toolkits/_model.py
jakesabathia2: Oh it is not a python model,
`if 'detector_model' in model_data['detector']` means the od model is saved as native model,
`else` means that the ob model is saved as version 5.8 or older.
But `od_cls` is always an `Object_detector_beta` object.

src/python/turicreate/toolkits/_model.py
jakesabathia2: `od_cls` is always an `Object_detector_beta` object.

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -238,7 +244,7 @@ def _get_native_state(self):
         # We don't know how to serialize a Python class, hence we need to
         # reduce the detector to the proxy object before saving it.
         if USE_CPP:
-            pass
+            state['detector'] = {'detector_model':state['detector'].__proxy__}
```
nickjong: Why do we need this?

jakesabathia2: https://github.com/apple/turicreate/blob/e91c16f012c7bb96106209b655502208b48a9c48/src/python/turicreate/toolkits/_model.py#L122
Here we need to know the OD model is saved in version 5.8 (or older) or in the 6.0 version.

nickjong: Ah, got it. I keep getting confused between the OSOD wrapper class and the inner OD class

PR #2643, opened 2019-11-15
Added 3.7 to travis.yaml

abhishekpratapa:
- added 3.7 to the travis build.


.travis.yml
```diff
@@ -9,6 +9,7 @@ python:
   - "2.7"
   - "3.5"
   - "3.6"
+  - "3.7"
```
srikris: Seems like 3.7 is not officially supported on Travis yet. You might need to do (https://github.com/travis-ci/travis-ci/issues/9815)

abhishekpratapa: Ah thanks!

PR #2644, opened 2019-11-15
Align numpy memory view when exporting weights from TF to TC

Jarvi-Izana:
TL;DR, `ascontiguousarray` is preferrable over `copy`.

---

~This is [WIP] because I want to leave my debug trace for people who want to repro and have a deep understanding of the issue is.~

The root cause for inconsistency of weights exported from TF to TC is:
```
net_params.update({var.name.replace(":0", ""): _np.transpose(val, (3, 2, 0, 1))})
```

The numpy `ufunc` transpose doesn't change the memory layout of numpy array. When we load the python object to TC CPP backend, we use this logic here
```
      pybind11::buffer_info buf = kv.second.request();
      turi::neural_net::shared_float_array value =
          turi::neural_net::shared_float_array::copy(
              static_cast<float*>(buf.ptr),
              std::vector<size_t>(buf.shape.begin(), buf.shape.end()));
```
The shape is correct but the `buf.ptr` points to the actually memory storage of numpy array, which has the layout of the original layout before being transposed.

It's similar to the lazy evaluation of our SFrame. You can apply different operations on the SFrame but the underlying storage layout won't change unless you call to `materialize`. 

The solution is to call `copy()` to enfore numpy to materialize the new memory view of applying lazy-evaluated `ufunc`s.
```
net_params.update({var.name.replace(":0", ""): _np.transpose(val, (3, 2, 0, 1)).copy()})
```

This is a critical bug and it's worthy of our attention.

If you want. to repro, check out hash `51b7b2dcb42`. Without solidating to new memory layout (original layout with C order, transposed, TC interpretation of memory layout of numpy), the output is
```
[[0. 1. 2. 3. 4.]
 [5. 6. 7. 8. 9.]]
[[0. 5.]
 [1. 6.]
 [2. 7.]
 [3. 8.]
 [4. 9.]]
dummy
(5, 2, )
0
1
2
3
4
5
6
7
8
9
```
After solidating by calling `copy`,
```
dummy
(5, 2, )
0
5
1
6
2
7
3
8
4
9
```

### DIFF
--- a/src/ml/neural_net/tf_compute_context.cpp
+++ b/src/ml/neural_net/tf_compute_context.cpp
@@ -155,7 +155,14 @@ float_array_map tf_model_backend::export_weights() const {
   float_array_map result;
   call_pybind_function([&]() {
     // Call export_weights method on the TensorFLowModel
-    pybind11::object exported_weights = model_.attr("export_weights")();
+    pybind11::dict exported_weights = model_.attr("export_weights")();
+
+    // it should be trivial to call this if we use the same interpreter process
+    pybind11::module np = pybind11::module::import("numpy");
+    for (auto& kv : exported_weights) {
+      exported_weights[kv.first] = np.attr("ascontiguousarray")((kv.second));
+    }
+
     std::map<std::string, pybind11::buffer> buf_output =
         exported_weights.cast<std::map<std::string, pybind11::buffer>>();
 
--- a/src/ml/neural_net/tf_compute_context.cpp
+++ b/src/ml/neural_net/tf_compute_context.cpp
@@ -160,11 +160,9 @@ float_array_map tf_model_backend::export_weights() const {
     // it should be trivial to call this if we use the same interpreter process
     pybind11::module np = pybind11::module::import("numpy");
     for (auto& kv : exported_weights) {
-      // defensively call ascontiguousarray to force the
+      // defensively call ascontiguousarray to force to reorganize
       // underlying numpy memory layout using the real strides
-      // after this, stride should be vector of 1s
-      exported_weights[kv.first] = np.attr("ascontiguousarray")(
-          pybind11::reinterpret_borrow<pybind11::object>(kv.second));
+      exported_weights[kv.first] = np.attr("ascontiguousarray")(kv.second);
     }
 
     std::map<std::string, pybind11::buffer> buf_output =
--- a/src/ml/neural_net/tf_compute_context.cpp
+++ b/src/ml/neural_net/tf_compute_context.cpp
@@ -160,7 +160,11 @@ float_array_map tf_model_backend::export_weights() const {
     // it should be trivial to call this if we use the same interpreter process
     pybind11::module np = pybind11::module::import("numpy");
     for (auto& kv : exported_weights) {
-      exported_weights[kv.first] = np.attr("ascontiguousarray")((kv.second));
+      // defensively call ascontiguousarray to force the
+      // underlying numpy memory layout using the real strides
+      // after this, stride should be vector of 1s
+      exported_weights[kv.first] = np.attr("ascontiguousarray")(
+          pybind11::reinterpret_borrow<pybind11::object>(kv.second));
     }
 
     std::map<std::string, pybind11::buffer> buf_output =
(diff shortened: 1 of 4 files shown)


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -155,7 +155,18 @@ float_array_map tf_model_backend::export_weights() const {
...
+    pybind11::dict exported_weights = model_.attr("export_weights")();
+
+    // it should be trivial to call this if we use the same interpreter process
+    pybind11::module np = pybind11::module::import("numpy");
+    for (auto& kv : exported_weights) {
+      // defensively call ascontiguousarray to force the
+      // underlying numpy memory layout using the real strides
+      // after this, stride should be vector of 1s
+      exported_weights[kv.first] = np.attr("ascontiguousarray")(
+          pybind11::reinterpret_borrow<pybind11::object>(kv.second));
```
Jarvi-Izana: refers from here https://github.com/pybind/pybind11/issues/1201

Jarvi-Izana: for `pybind11::reinterpret_borrow<pybind11::object>`.

Jarvi-Izana: Per discussion with @hoytak, pybind should handle py::handle to py::object conversion if needed.

Jarvi-Izana: So I will remove `pybind11::reinterpret_borrow<pybind11::object>`

Jarvi-Izana:
```
In [4]: c = np.ascontiguousarray(b)
In [5]: %timeit  np.ascontiguousarray(c)
790 ns ± 4.12 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
applying `ascontiguousarray` on array with transformed memory layour is a no-op.

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -155,7 +155,16 @@ float_array_map tf_model_backend::export_weights() const {
...
     // Call export_weights method on the TensorFLowModel
-    pybind11::object exported_weights = model_.attr("export_weights")();
+    pybind11::dict exported_weights = model_.attr("export_weights")();
+
+    // it should be trivial to call this if we use the same interpreter process
+    pybind11::module np = pybind11::module::import("numpy");
+    for (auto& kv : exported_weights) {
+      // defensively call ascontiguousarray to force to reorganize
+      // underlying numpy memory layout using the real strides
+      exported_weights[kv.first] = np.attr("ascontiguousarray")(kv.second);
```
shreyajain17: If we are doing it here, I think we should remove it from export weights from AC/OD and any other toolkits that have it in `export_weights` in `tf_model_architecture` file.

Jarvi-Izana: Could you explain what we should remove?

shreyajain17: So I have had the same problem earlier. I was tackling it by calling it here https://github.com/apple/turicreate/blob/aedc7a9868ffe9df4509cd0c5adbb7949774f8bd/src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py#L289
Similarly, for OD also I was calling it in the python file. Your fix is better as it will do it for all toolkits together. So can you also remove these files along with this change since this change is better than what I was doing?

Jarvi-Izana: Ha, you don't have to worry about it. IMO, your change is good (be able to track down this issue is really excellent). If I had known you met this issue before, I would ask you and add `ascontiguousarray` into python file without any hesitation.

Back to the point, this is a defensive programming paradigm just to protect us from people forget to call `ascontiguousarray` accidentally. But this doesn't mean we don't need to call `ascontiguousarray` explicitly.

We should advocate adding `ascontiguousarray` eagerly in our python implementation to explicitly document to other contributors what we are doing here. It's better to let other contributors also not to worry about keeping `ascontiguousarray` in mind when exporting. Even though they forget to have it called, they won't jump into the quagmire of debugging the mismatched weights as we did before.

shreyajain17: Okay. I think you ran into it when I was away. Maybe should have documented it better. Good job tracking down the issue too.

Jarvi-Izana: yeap! Nice to have you around ;-)

nickjong: Could this line be simplified to:
```
kv.second = np.attr("ascontiguousarray")(kv.second);
```
?

Jarvi-Izana: nope. kv is a type of `pair<py::handle, py::handle>`. The py::handle is a pointer to an object without any ownership. We need py::object here so that why I'm using `operator[](py::handle)`.

reference:
[handle](https://pybind11.readthedocs.io/en/stable/reference.html#_CPPv46handle)
[index operator]https://pybind11.readthedocs.io/en/stable/reference.html#_CPPv4NK10object_apiixEPKc

nickjong: Ah, I see, we need to assign into the result of `operator[]` to trigger the call to Python-side `__setitem__`. Thanks for the link!

shreyajain17:
fixes #2692

PR #2647, opened 2019-11-15
added #ifdef MacOS 10.15

abhishekpratapa:
- build error on MacOS 10.15


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -317,7 +321,13 @@ void style_transfer::load_version(iarchive& iarc, size_t version) {
 
 std::unique_ptr<compute_context> style_transfer::create_compute_context()
     const {
-  return compute_context::create();
+  #ifdef __APPLE__
+  if (neural_net::mps_compute_context::has_style_transfer()) {
```
nickjong: Let's just add a comment here explaining what the heck is going on

PR #2649, opened 2019-11-16
DC: Fixed behavior of `predict` with `output_type=probability`

shantanuchhabra:
Thanks to @nickjong for pointing out that our implementation of `output_type="probability"` has been wrong all this while in `predict`.

The `output_type="probability"` option is exclusively meant for binary classifiers. This option is meant to provide the probability of the True class in binary classifiers.
For example, if one is building a binary classifier for "cat" vs "not cat", output_type="probability" should output the probability of the data point being "cat", not the probability of the predicted class.

This PR is an effort to fix this behavior (that we had misunderstood until now), and also adds a C++ unit test.

### DIFF
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -718,8 +718,7 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
 
     auto true_class_probability = [=](const flexible_type& ft) {
       const flex_vec& prob_vec = ft.get<flex_vec>();
-      size_t last_index = num_classes - 1;  // either 0 or 1
-      return *(prob_vec.begin() + last_index);
+      return prob_vec.back();
     };
 
     result = result.apply(true_class_probability, flex_type_enum::FLOAT);
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -45,7 +45,6 @@ using padding_type = model_spec::padding_type;
 // anonymous helper sections
 
 struct result {
-  shared_float_array loss_info;
   shared_float_array accuracy_info;
   data_iterator::batch data_info;
 };
@@ -235,6 +234,13 @@ void drawing_classifier::init_options(
       FLEX_UNDEFINED,
       std::numeric_limits<int>::min(), std::numeric_limits<int>::max());
 
+  options.create_string_option(
+      "warm_start",
+      "Record warm start model version used. If no warmstart used"
+      " None is assigned by default."
+      "",
+      true);
+
   // Validate user-provided options.
   options.set_options(opts);
 
@@ -255,7 +261,7 @@ std::unique_ptr<data_iterator> drawing_classifier::create_iterator(
 
 std::unique_ptr<data_iterator> drawing_classifier::create_iterator(
     gl_sframe data, bool is_train,
-    std::vector<std::string> class_labels) const {
+    flex_list class_labels) const {
   data_iterator::parameters data_params;
 
   std::string feature_column_name = read_state<flex_string>("feature");
@@ -298,6 +304,19 @@ void drawing_classifier::init_training(
     {"training_iterations", 0}
   });
 
+  // Capture Core ML model path from options,
+  // if provided by Python.
+  std::string mlmodel_path;
+  bool enable_warmstart = false;
+  auto mlmodel_path_iter = opts.find("mlmodel_path");
+  if ( mlmodel_path_iter != opts.end()) {
+    mlmodel_path = mlmodel_path_iter->second.to<std::string>();
+    // Remove `mlmodel_path` from options as
+    // it is not a user-defined option.
+    opts.erase(mlmodel_path_iter);
+    enable_warmstart = true;
+  }
+
   // Read user-specified options.
   init_options(opts);
 
@@ -322,11 +341,10 @@ void drawing_classifier::init_training(
       create_iterator(training_data_,
                       /* is_train */ true, /* class labels */ {});
 
-  const std::vector<std::string>& classes =
-      training_data_iterator_->class_labels();
+  const flex_list& classes = training_data_iterator_->class_labels();
 
   add_or_update_state({
-      {"classes", flex_list(classes.begin(), classes.end())},
+      {"classes", classes},
       {"num_classes", classes.size()},
   });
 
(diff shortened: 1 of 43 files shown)


Jarvi-Izana:
I don't have the full context of the behavior of 5.8. I guess you still want to support `probability_vector` when the number of classification for more than 2 classes.

shantanuchhabra:
> I don't have the full context of the behavior of 5.8. I guess you still want to support `probability_vector` when the number of classification for more than 2 classes.

Yes, we still support `probability_vector` as an output type. We first check to make sure the output type provided by the user is one of "class", "probability", and "probability_vector". The output from `perform_inference` is already aligned with what we want to output for `probability_vector` so we only check for `class` and `probability` and modify the result accordingly. Does that make sense?

Jarvi-Izana:
> > I don't have the full context of the behavior of 5.8. I guess you still want to support `probability_vector` when the number of classification for more than 2 classes.
> 
> Yes, we still support `probability_vector` as an output type. We first check to make sure the output type provided by the user is one of "class", "probability", and "probability_vector". The output from `perform_inference` is already aligned with what we want to output for `probability_vector` so we only check for `class` and `probability` and modify the result accordingly. Does that make sense?

makes sense.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -700,14 +700,29 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
...
+    if (num_classes > 2) {
+      log_and_throw(
+          "Use probability_vector in case of multi-class classification.");
+    }
+
+    auto true_class_probability = [=](const flexible_type& ft) {
       const flex_vec& prob_vec = ft.get<flex_vec>();
-      auto max_it = std::max_element(prob_vec.begin(), prob_vec.end());
-      return *max_it;
+      size_t last_index = num_classes - 1;  // either 0 or 1
```
Jarvi-Izana: Could you explain why only the last probability is returned?

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -700,14 +700,29 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
...
+      return *(prob_vec.begin() + last_index);
```
nickjong: Could we just return `prob_vec.back()` here?

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: With True and False labels, i.e. "cat" and "not cat", we assume that the False label has index 0 and the True label has index 1. I'm not sure where we broadcast this to users (probably somewhere).

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: Done, thanks!

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -700,14 +700,28 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
...
+
+    size_t num_classes = read_state<flex_int>("num_classes");
+    DASSERT_GT(num_classes, 0);
+
+    if (num_classes > 2) {
+      log_and_throw(
+          "Use probability_vector in case of multi-class classification.");
+    }
+
+    auto true_class_probability = [=](const flexible_type& ft) {
```
nickjong: Leave an empty capture list when not capturing anything with a lambda expression

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: The relative order of `cat` and `not-cat` should depend on the label from the training data. set, i.e, data iterator may return label vector of <cat, not-cat> or <cat, not-cat>. You may. check where `num_classes` is introduced in `init_training`.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -700,14 +700,28 @@ gl_sarray drawing_classifier::predict(gl_sframe data, std::string output_type) {
...
       const flex_vec& prob_vec = ft.get<flex_vec>();
-      auto max_it = std::max_element(prob_vec.begin(), prob_vec.end());
-      return *max_it;
+      return prob_vec.back();
```
Jarvi-Izana: I think we need to align with the label order generated from the training dataset.

nickjong: Could you sync with @hoytak? We should align with what all the supervised-learning toolkits do.

srikris: This should align with the label order in the training set that are sorted according to class name. So for example:

```
model.classes -> ['cat', 'dog', 'foosa']
```
and predictions should be in that order
```
[p_cat, p_dog, p_foosa
```

Sadly the binary classifier is a little annoying because all other packages have the convention of using a "reference" class to minimize multi-collinearity and have fewer parameters while this isn't much of a concept in deep learning based models.

shantanuchhabra: @hoytak , @Jarvi-Izana , and I discussed this at length and here are our thoughts:

1. For `output_type="probability"` in a binary classifier setting, all our supervised learning toolkits currently follow the approach of sorting labels lexicographically, and returning the probability of the last label.
2. This approach is kind of arbitrary, since the labels could be "cat" and "not cat", in which case the user will end up with probabilities for the label, "not cat".
3. After some discussion, we came to the conclusion that for TC 6.0, we should modify this behavior across Drawing Classifier, Logistic Classifier, SVMs, and other supervised learning toolkits to the following: `output_type="probability"` should only apply if the target values in the data are `0` and `1`. This issue is tracked in #2667 .

Here is what I suggest w.r.t this PR:

1. Let's merge this PR in its current state where it matches what we currently have for all supervised learning toolkits.
2. Let's address #2667 before TC 6.0 and make a code change in this logic (and elsewhere) then.

@nickjong @srikris @Jarvi-Izana @hoytak Thoughts?

Jarvi-Izana: As long as it's book-kept, I'm good.

nickjong: I don't think we have the time to adequately design, develop, and test this kind of change for 6.0.

PR #2650, opened 2019-11-16
DC: Sanitize validation_data value passed in to C++

### LINKED ISSUE #2605: Drawing classifier validation set to None should work
Setting validation set to None should work
```
model = tc.drawing_classifier.create(sf, 'label', validation_set=None)
```

But it return the following error. 
```
ToolkitError: Validation data parameter must be either "auto", an empty SFrame (no validation info is computed), or an SFrame with the same schema as the training data.. operator() from /Volumes/Data/srikris/workspace/turicreate/src/toolkits/supervised_learning/automatic_model_creation.cpp at 308
```

### LINKED ISSUE #2646: Drawing Classifier API Breaking Change
Seems the drawing classifier’s create function no longer allows `validation_set=None` which is an API breaking change from 5.8.

shantanuchhabra:
Since the C++ layer can only accept values = "auto", an SFrame, or an empty SFrame (so no validation info is computed), we sanitize the `validation_data` value passed in to the C++ layer from Python by translating a `None` on the Python side into an empty SFrame on the C++ side.

Fixes #2605

### DIFF
--- a/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -171,6 +171,8 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         options = dict()
         options["batch_size"] = batch_size
         options["max_iterations"] = max_iterations
+        if validation_set is None:
+            validation_set = _tc.SFrame()
         # Enable the following line when #2524 is merged
         # options["warm_start"] = warm_start
         model.train(input_dataset, target, feature, validation_set, options)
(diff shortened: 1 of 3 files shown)


src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -171,6 +171,8 @@ def create(input_dataset, target, feature=None, validation_set='auto',
         options = dict()
         options["batch_size"] = batch_size
         options["max_iterations"] = max_iterations
+        if validation_set is None:
```
srikris: Please add a unit test.

shantanuchhabra: Done, thanks.

PR #2653, opened 2019-11-16
DC: Enabled the x-fail drawing test.

srikris:
Enabled 2 passing tests that were accidentally marked xFailed.


srikris:
@shantanuchhabra Before I merge it, can you confirm that this was accidentally disabled?

PR #2656, opened 2019-11-17
Working CoreML export with metadata version number

### LINKED ISSUE #2562: Add Turi Create version to user-defined metadata in Core ML model
Turi Create should add a key to the [user-defined Core ML metadata](https://developer.apple.com/documentation/coreml/mlmodelmetadatakey/2923322-creatordefinedkey) to embed which version the model was built with: `com.github.apple.turicreate.version`.

For example:

Key | Value
----|------
com.github.apple.turicreate.version | 5.8

srikris:
- Fixes #2562 for all based toolkits.
- Added unit tests to check correctness
- Behavior is intended to match 5.8 as closely as possible

Each commit is separately reviewable to make this simpler to review. Commits are done toolkit by toolkit.

### DIFF
--- a/src/python/turicreate/test/test_drawing_classifier.py
+++ b/src/python/turicreate/test/test_drawing_classifier.py
@@ -284,6 +284,7 @@ class DrawingClassifierTest(unittest.TestCase):
         import coremltools
         import platform
         max_iters_ans = ['20', '1']
+        warm_start_ans = '' if self.warm_start is None else self.warm_start
         for i, model in enumerate(self.models):
             filename = _mkstemp("bingo.mlmodel")[1]
             model.export_coreml(filename)
@@ -296,7 +297,7 @@ class DrawingClassifierTest(unittest.TestCase):
                 'target': self.target,
                 'feature': self.feature,
                 'type': 'drawing_classifier',
-                'warm_start': '',
+                'warm_start': warm_start_ans,
                 'max_iterations': max_iters_ans[i],
                 'version': '2',
                 }, dict(coreml_model.user_defined_metadata)
--- a/src/python/turicreate/test/test_drawing_classifier.py
+++ b/src/python/turicreate/test/test_drawing_classifier.py
@@ -285,6 +285,7 @@ class DrawingClassifierTest(unittest.TestCase):
 
     def test_export_coreml(self):
         import coremltools
+        import platform
         for model in self.models:
             filename = _mkstemp("bingo.mlmodel")[1]
             model.export_coreml(filename)
@@ -293,6 +294,7 @@ class DrawingClassifierTest(unittest.TestCase):
             coreml_model = coremltools.models.MLModel(filename)
             self.assertDictEqual({
                 'com.github.apple.turicreate.version': _tc.__version__,
+                'com.github.apple.os.platform': platform.platform(),
                 'target': self.target,
                 'feature': self.feature,
                 'type': 'drawing_classifier',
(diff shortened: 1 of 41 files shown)


syoutsey:
Changes LGTM. Either as part of this PR or another for 6.0, it'd be good to also have the following keys:
```
com.github.apple.turicreate.os.name : Short name of OS (macOS, Ubuntu, etc.)
com.github.apple.turicreate.os.version : Version of OS (10.15.1, 18.04, etc.)
```

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1655,8 +1655,12 @@ def export_coreml(self, filename,
         options['include_non_maximum_suppression'] = include_non_maximum_suppression
         options['confidence_threshold'] = confidence_threshold
         options['iou_threshold'] = iou_threshold
+        options['iou_threshold'] = iou_threshold
```
nickjong: Copy and paste error?

srikris: Yes. I'll fix that.

srikris:
@syoutsey Good point! I'll make it that. Luckily with this change, it can be done in 1 place :)

src/python/turicreate/test/test_drawing_classifier.py
```diff
@@ -282,10 +282,28 @@ def test_save_and_load(self):
...
+                'com.github.apple.turicreate.version': _tc.__version__,
+                'com.github.apple.os.platform': platform.platform(),
+                'target': self.target,
+                'feature': self.feature,
+                'type': 'drawing_classifier',
+                'max_iterations': '20',
+                'version': '2',
+                }, dict(coreml_model.user_defined_metadata)
+            )
+            expected_result = 'Drawing classifier created by Turi Create (version %s)' \
```
Jarvi-Izana: I bet this will fail since we didn't add version info for DC. I will add it.

srikris: It's added as part of this PR.

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -726,8 +724,13 @@ std::shared_ptr<MLModelWrapper> activity_classifier::export_to_coreml(
       {"type", "activity_classifier"},
       {"version", 2},
   };
+  for(const auto& kvp : additional_user_defined) {
+       user_defined_metadata.emplace_back(kvp.first, kvp.second);
+  }
+
   model_wrapper->add_metadata({
-      {"user_defined", std::move(user_defined_metadata)}
+      {"short_description", short_desc},
```
Jarvi-Izana: could be moved. or we could have `const std::string& short_desc`. But to me, I get quite confused. The version should be anchored by the CPP side since all version info should come from CPP implementation. Otherwise, the python layer can pass the wrong version header back by the user, e.g., they are using 6.0 cpp backend but somehow 7.0 is passed from them. I bet user should not modify the `short_desc` on our python side.

But for me, I think it's better to provide a cpp backend version info. Even though we fast iterate the turicreate python implementation, the backend may not have that many changes and be stable for a long time. In which way it's more clear for user to identify backward compatibility.

src/python/turicreate/test/test_drawing_classifier.py
Jarvi-Izana: Ah, need to scroll down more. ;-)

src/python/turicreate/toolkits/object_detector/object_detector.py
srikris: Fixed.

src/toolkits/activity_classification/activity_classifier.cpp
srikris: The short description gets displayed to the user when the open a CoreML model in Xcode. The should only know the Python version they used to make the model.

Issue #2640, opened 2019-11-15
TF Style Transfer doesn’t take images of different sizes


abhishekpratapa:
`Cannot feed value of shape (1, 403, 311, 3) for Tensor u'Placeholder:0', which has shape '(?, 256, 256, 3)'`

PR #2657, opened 2019-11-18
Fixed Resizing issue in TensorFlow for Style Transfer

abhishekpratapa:
## Overview

- Added shape to float array in predict.
- removed explicit shape in `process_output`
- compute_contexts should take care of sizing themselves
- Added crop to TF (It was present previously in the MxNet implementation since padding is added in both the TF and MxNet implementations from the way Reflective Padding is done)

## Fixed
- The four failing Unit Tests in the TF implementation of ST.
- Tests Passing in both TF and MPS

### DIFF
--- a/src/ml/neural_net/style_transfer/mps_style_transfer.m
+++ b/src/ml/neural_net/style_transfer/mps_style_transfer.m
@@ -324,6 +324,19 @@
     imagesOut[key] = styleData;
   }
 
+  NSMutableData * imageChannelsData = [NSMutableData dataWithCapacity:0];
+  float imageChannels = 3.0;
+  [imageChannelsData appendBytes:&imageChannels length:sizeof(float)];
+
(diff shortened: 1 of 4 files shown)


shreyajain17:
fixes #2640

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -324,6 +324,19 @@ - (instancetype) initWithDev:(id<MTLDevice>) dev
     imagesOut[key] = styleData;
   }
 
+  NSMutableData * imageChannelsData = [NSMutableData dataWithCapacity:0];
```
nickjong: Let's simplify to:
```
NSData* imageChannelsData = [NSData dataWithBytes:&imageChannels length:sizeof(float)];
```

PR #2659, opened 2019-11-18
Implement "prediction_id" column for "per_window" prediction for AC

### LINKED ISSUE #1706: Functionality of prediction_id in predict of Activity Classifier
The functionality of prediction_id column given as output of predict when making prediction per window is not clear in documentation. The output given at the moment is the following. 

```
|      253      |   1    | [  4.63483302e-04   9.9950... |
|      254      |   1    | [  7.90477672e-04   9.9914... |
|      255      |   1    | [  4.41032015e-02   9.5374... |
|      256      |   2    | [  7.02369167e-03   2.4149... |
|      257      |   2    | [  5.60363929e-04   3.1854... |
|      258      |   2    | [  1.27506602e-04   6.1436... |
|      259      |   2    | [  6.75232295e-05   2.6096... |
```

nickjong:
Adds this column for `predict` and updates the semantics for `classify`. Unlike 5.8, this column now gives the index of the prediction within the session (instead of the index overall, which is just the row number in the result table)

Also fixes some places where the result SFrame should use the user's column name for the session ID, not the string "session_id"

Fixes #1706

### DIFF
--- a/src/toolkits/activity_classification/activity_classifier.cpp
+++ b/src/toolkits/activity_classification/activity_classifier.cpp
@@ -461,8 +461,10 @@ gl_sframe activity_classifier::predict_per_window(gl_sframe data,
 
   // Accumulate the class probabilities for each prediction window.
   gl_sframe raw_preds_per_window = perform_inference(data_it.get());
+  std::string session_id_column_name = read_state<std::string>("session_id");
   gl_sframe result =
-      gl_sframe({{"session_id", raw_preds_per_window["session_id"]},
+      gl_sframe({{session_id_column_name, raw_preds_per_window["session_id"]},
+                 {"prediction_id", raw_preds_per_window["prediction_id"]},
                  {"probability_vector", raw_preds_per_window["preds"]}});
 
   if (output_type == "class") {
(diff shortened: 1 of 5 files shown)


src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -660,10 +666,10 @@ gl_sframe activity_classifier::predict_topk(gl_sframe data,
     stacked_rank = stacked_rank.stack("rank", "rank");
     result.add_column(stacked_rank["rank"], "rank");
   } else {
-    // add "exp_id" and "prediction_id" if the output_frequency is per_window
-    result.add_column(raw_preds_per_window["session_id"], "exp_id");
-    result.add_column(gl_sarray::from_sequence(0, raw_preds_per_window.size()),
-                      "prediction_id");
+    // add session ID and prediction ID if the output_frequency is per_window
```
jakesabathia2: I think the 5.8 used "exp_id" instead?

nickjong: The 5.8 Python implementation used whatever string the user provided as the column name for the session ID. It just so happens that the test data you've been using uses "exp_id". Note that this string appears nowhere in the original Python implementation

jakesabathia2: Oh cool!

PR #2660, opened 2019-11-18
Bump the egg version

### LINKED ISSUE #2632: Upgrade the version number of the Python egg to 6.0

shreyajain17:
fixes #2632
```

In [2]: tc.__version__
Out[2]: '6.0rc0'
```

This turns the variable `IS_PRE_6_0_RC` in pytests False. This PR should be merged after it passes all unit tests or condition must be changed.

Also, in the PR -
Removed `IS_PRE_6_0_RC` wherever it is not needed anymore.
Changed it so the tests run and don't crash.


src/python/turicreate/test/test_sframe_sequence_iterator.py
```diff
@@ -12,7 +12,7 @@
 import array
 
 
-IS_PRE_6_0_RC = float(tc.__version__) < 6.0
```
srikris: @TobyRoseman Is there a more pythonic way to do it?

TobyRoseman: I don't understand. A more pythonic way of doing what?

src/python/turicreate/test/test_audio_functionality.py
```diff
@@ -28,7 +28,7 @@
 from turicreate.toolkits._internal_utils import _mac_ver
 
 
-IS_PRE_6_0_RC = float(tc.__version__) < 6.0
+IS_PRE_6_0_RC = float(tc.__version__[:3]) < 6.0
```
znation: Should we use [packaging.version.parse](https://packaging.pypa.io/en/latest/version/#packaging.version.parse)?

znation: Actually, now that this is 6.0rc0, shouldn't we just do `IS_PRE_6_0_RC = false` (and file a bug to clean it up later), or just remove the pre-6.0 code now?

nickjong:
It looks like the Travis build failed due to `test_sframe_sequence_iterator.py`. Let’s go ahead and just delete that test. It’s for a Python-based data iterator used by the old Python implementation of AC. The code it’s testing is no longer being used.

shreyajain17:
> It looks like the Travis build failed due to `test_sframe_sequence_iterator.py`. Let’s go ahead and just delete that test. It’s for a Python-based data iterator used by the old Python implementation of AC. The code it’s testing is no longer being used.

Okay.

shreyajain17:
> It looks like the Travis build failed due to `test_sframe_sequence_iterator.py`. Let’s go ahead and just delete that test. It’s for a Python-based data iterator used by the old Python implementation of AC. The code it’s testing is no longer being used.

I made a mistake. It should be True and that would skip these tests. So just skipping for now. This should pass travis.

src/python/turicreate/test/test_audio_functionality.py
```diff
@@ -28,7 +28,7 @@
 from turicreate.toolkits._internal_utils import _mac_ver
 
 # TODO: Remove for 6.0
-IS_PRE_6_0_RC = False
+IS_PRE_6_0_RC = True
```
znation: This no longer seems correct; `IS_PRE_6_0_RC` should now be false, since we are at 6.0.0rc0. Right?

shreyajain17: Yeah I was wrong. But whatI realized now is that we don't need the condition. I removed the test_sframe_sequence_iterator.py as it was testing old AC implementation. (as @nickjong mentioned above) And audio_functionality does not need it anymore after merging master. I removed it from one place in style transfer that needs to be checked in a machine with linux gpu!

srikris:
@shreyajain17 I'm gonna merge this to make it easier for us to do reression testing.

PR #2661, opened 2019-11-18
Changes padding for first five pooling layers TF OD

### LINKED ISSUE #2616: OD TensorFlow model should use the same padding as 5.8
The current TensorFlow model fix the padding to `SAME` for all pooling layers,
however for 5.8 implementation, the first five pooling layers use `VALID` padding instead.

shreyajain17:
fixes #2616

Passes all unit tests after the change (python and cpp)

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
+++ b/src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
@@ -260,12 +260,10 @@ class ODTensorFlowModel(object):
             if idx < 7:
                 if idx < 6:
                     strides = [1, 2, 2, 1]
+                    net = self.pooling_layer(net, pool_size=[1, 2, 2, 1], strides=strides, padding='VALID', name='pool%d_' % idx)
                 else:
                     strides = [1, 1, 1, 1]
-                if idx < 6 :
-                    net = self.pooling_layer(net, pool_size=[1, 2, 2, 1], strides=strides, padding='VALID', name='pool%d_' % idx)
-                else :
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -262,8 +262,10 @@ def tiny_yolo(self, inputs, output_size=125):
                     strides = [1, 2, 2, 1]
                 else:
                     strides = [1, 1, 1, 1]
-
-                net = self.pooling_layer(net, pool_size=[1, 2, 2, 1], strides=strides, name='pool%d_' % idx)
+                if idx < 6 :
```
jakesabathia2: Can we only use one `if else`? Set both `stride` and `padding` in the same `if else`.

jakesabathia2:
Oh and for cpp, the sixth pooling layer has an argument `set_avgpoolexcludepadding = true`, which might suggests that this layer is average pooling layer instead of max pooling layer:
https://github.com/apple/coremltools/blob/7f2b314f496072320e1088ad8c6f6f572063c454/coremltools/models/neural_network/builder.py#L1824
So maybe we want to use `tf.nn.avg_pooling` here instead @nickjong .

shreyajain17:
@jakesabathia2 Good catch. I will make changes to accommodate it.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
shreyajain17: Yeah makes sense.

Issue #2631, opened 2019-11-14
Tensorflow prints out strange text


srikris:
```
In [4]: model.export_coreml('tc-6.0-tc.mlmodel')
WARNING:root:TensorFlow version 2.0.0 detected. Last version known to be fully compatible is 1.13.1 .
2019-11-14 01:08:03.327787: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2019-11-14 01:08:03.370602: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x7fba90358970 executing computations on platform Host. Devices:
2019-11-14 01:08:03.370629: I tensorflow/compiler/xla/service/service.cc:175]   StreamExecutor device (0): Host, Default Version
WARNING:tensorflow:No training configuration found in save file: the model was *not* compiled. Compile it manually.
WARNING:tensorflow:No training configuration found in save file: the model was *not* compiled. Compile it manually.
```

srikris:
Not sure the fix for this is easy though.

srikris:
It might be worth identifying if we can hide all the logs for TF 6.0

TobyRoseman:
This line is from coremltools not TensorFlow:
```
WARNING:root:TensorFlow version 2.0.0 detected. Last version known to be fully compatible is 1.13.1 .
```

TobyRoseman:
I believe even that line getting printed is a regression from 5.8. It was not getting printed in 5.8 because we were not using TensorFlow in 5.8. We should probably silence output of that line too.

shreyajain17:
And I am supressing log from tensorflow and I had picked verbosity level to be error i.e. warnings should not be showing.

shreyajain17:
Can you tell me which python were you using?

srikris:
Python 3.6 and Python 2.7. Do you not see this on master?

shreyajain17:
```
In [1]: import turicreate as tc

In [2]: data = tc.SFrame('/Users/shreya_jain3/Desktop/breakfast-data.sframe/')[:2]

In [3]: model = tc.object_detector.create(data, 'annotation', max_iterations=1)
Using 'image' as feature column
[...]

PR #2665, opened 2019-11-18
Suppress warnings for tf

shreyajain17:
Some verification left. 

- [x] A line still seems to appear from tensorflow.
- [x] Python 2.7
- [x] Python 3.6

### DIFF
--- a/src/python/turicreate/toolkits/_tf_utils.py
+++ b/src/python/turicreate/toolkits/_tf_utils.py
@@ -16,17 +16,17 @@ def suppress_tensorflow_warnings():
 	import os
 	os.environ['TF_CPP_MIN_LOG_LEVEL'] = '1'
 
-	import tensorflow.compat.v1 as tf
-	tf.disable_v2_behavior()
-	tf.logging.set_verbosity(_tf.logging.ERROR)
-	tf.debugging.set_log_device_placement(False)
+	import tensorflow.compat.v1 as _tf
+	_tf.disable_v2_behavior()
+	_tf.logging.set_verbosity(_tf.logging.ERROR)
+	_tf.debugging.set_log_device_placement(False)
 
 def get_gpu_names():
 	"""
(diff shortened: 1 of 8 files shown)


src/python/turicreate/toolkits/_tf_utils.py
```diff
@@ -14,8 +14,9 @@ def get_gpu_names():
 	"""
 	Gets the available GPU names.
 	"""
-	import tensorflow as tf
-	gpu_names = tf.config.experimental.list_physical_devices('GPU')
+	import tensorflow.compat.v1 as _tf
+	_tf.disable_v2_behavior()
```
TobyRoseman: If we're writing new code, we should be using the new API.

Does the following work?
```
import TensorFlow as tf
tf.config.experimental.list_physical_devices
```

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -21,6 +21,8 @@ def __init__(self, input_h, input_w, batch_size, output_size, init_weights, conf
         _tf.reset_default_graph()
 
         # Suppresses verbosity to only errors
+        import os
```
srikris: Can we make this into a utility and call it everywhere.

src/python/turicreate/toolkits/_tf_utils.py
shreyajain17: Yeah the new one works. Yeah that does make sense which is why I had that earlier. I will revert back this change.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
shreyajain17: Yeah that makes sense.

srikris:
Fixes #2631

Issue #2648, opened 2019-11-16
Can not load SFrame from S3 on Linux


TobyRoseman:
This seems to happen with any SFrame but only on Linux.

syoutsey:
@TobyRoseman Does this also apply to loading any remote SFrame period, say like using one of the SFrames listed in our user guide?

TobyRoseman:
> @TobyRoseman Does this also apply to loading any remote SFrame period, say like using one of the SFrames listed in our user guide?

I believe it affects all SFrame. We don't believe we have any S3 SFrames in our user guide.

PR #2666, opened 2019-11-18
Fix for issue #2648, SSL cert issue.

hoytak:
The root cause of the S3 Cert issue is that during the compilation of curl, the path /etc/ssl/certs/ is only added as a search path for certificates if it is present on the machine at build time.  Without that, CURL does not build with that enabled as a search path.  When we switched our build environment to CentOS 6 to support the manylinux2010 standard, it didn't have this directory on the resulting build machine and so our resulting egg did not search that path.

### DIFF
--- a/scripts/Dockerfile-CentOS-6
+++ b/scripts/Dockerfile-CentOS-6
@@ -56,6 +56,7 @@ RUN make install
 WORKDIR /src
 RUN curl -O https://www.openssl.org/source/openssl-1.1.0j.tar.gz
 RUN tar xf openssl-1.1.0j.tar.gz
+RUN mkdir -p /etc/ssl/certs/
 WORKDIR /src/openssl-1.1.0j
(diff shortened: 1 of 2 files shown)


scripts/Dockerfile-CentOS-6
```diff
@@ -56,6 +56,7 @@ RUN make install
 WORKDIR /src
 RUN curl -O https://www.openssl.org/source/openssl-1.1.0j.tar.gz
 RUN tar xf openssl-1.1.0j.tar.gz
+RUN mkdir -p /etc/ssl/certs/
```
TobyRoseman: If we're adding it to the image, do we really also need to add this line in `.gitlab-ci.yml`?

hoytak: We can remove them from the .gitlab scripts after we regenerate all the docker images, but for now we need it in both.

Issue #2606, opened 2019-11-12
Toolkits are missing __repr__


srikris:
Currently, we get the following output with some toolkits in 6.0 master (using DC as an example). 
```
In [31]: model
Out[31]: DrawingClassifier_beta
```

Close this issue when all boxes are checked:

- [x] Drawing Classifier (#2669)
- [x] Object Detector (#2671)
- [x] Style Transfer (#2711)
- [x] Activity classifier (#2680)

srikris:
```
In [10]: model
Out[10]: ObjectDetector_beta
```
@nickjong Can be all fixed in one PR since it's probably the same template that needs to be moved from 5.8

shantanuchhabra:
This is a little more nuanced than just a copy-paste. State variables need to be added/modified, and the summary struct needs to be modified in some places too. I suggest breaking this up into separate issues (and PRs, albeit small). What do you think @srikris @nickjong ?

srikris:
Based on offline conversations, I think it might make sense to put it into a single PR. It should be a pretty small PR.

srikris:
On second thought, doing it by toolkit would actually be simpler and easier to review. It would also ensure we don't have any regressions.

shantanuchhabra:
Per offline discussion, updated this issue to track all toolkits and will put up PRs per toolkit.

PR #2669, opened 2019-11-19
Adds __repr__() to DrawingClassifier_beta class

shantanuchhabra:
This PR will help check off the "Drawing Classifier" box in #2606

### DIFF
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -93,6 +93,8 @@ def load_model(location):
         if name in MODEL_NAME_MAP:
             cls = MODEL_NAME_MAP[name]
             if 'model' in saved_state:
+                if name=='activity_classifier' or name=='object_detector' or name=='style_transfer':
+                    import turicreate.toolkits.libtctensorflow
                 # this is a native model
                 return cls(saved_state['model'])
             else:
@@ -102,16 +104,19 @@ def load_model(location):
                 del model_data['model_version']
 
                 if name=='activity_classifier' and AC_USE_CPP:
+                    import turicreate.toolkits.libtctensorflow
                     model = _extensions.activity_classifier()
                     model.import_from_custom_model(model_data, model_version)
                     return cls(model)
 
                 if name=='object_detector' and OD_USE_CPP:
+                    import turicreate.toolkits.libtctensorflow
                     model = _extensions.object_detector()
                     model.import_from_custom_model(model_data, model_version)
                     return cls(model)
 
                 if name=='style_transfer' and ST_USE_CPP:
+                    import turicreate.toolkits.libtctensorflow
                     model = _extensions.style_transfer()
                     model.import_from_custom_model(model_data, model_version)
                     return cls(model)
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -448,8 +448,8 @@ class Model(ExposeAttributesFromProxy):
         if field in self._list_fields():
             return self.__proxy__.get_value(field)
         else:
-            raise KeyError('Field \"%s\" not in model. Available fields are'
-                         '%s.') % (field, ', '.join(self.list_fields()))
+            raise KeyError(('Field \"%s\" not in model. Available fields are '
+                         '%s.') % (field, ', '.join(self._list_fields())))
 
     @classmethod
     def _native_name(cls):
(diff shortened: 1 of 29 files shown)


src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -508,6 +508,10 @@ void drawing_classifier::iterate_training() {
         {"validation_accuracy", average_val_accuracy},
         {"validation_log_loss", average_val_loss},
     });
+  } else {
+    add_or_update_state({
+      {"validation_accuracy", FLEX_UNDEFINED}
```
srikris: We should add `validation_log_loss` hear ideally.

shantanuchhabra: Done, thanks!

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -506,7 +506,10 @@ void drawing_classifier::iterate_training() {
   if (validation_data_iterator_) {
     add_or_update_state({
         {"validation_accuracy", average_val_accuracy},
-        {"validation_log_loss", average_val_loss},
+    });
+  } else {
+    add_or_update_state({
```
Jarvi-Izana: probably we don't want the user to be aware of the validation metrics if we don't have validation data set.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -599,6 +605,9 @@ void drawing_classifier::train(gl_sframe data, std::string target_column_name,
     }
   }
 
+  state_update["num_examples"] = data.size();
```
Jarvi-Izana: `num_of_training_example` might be more clear? To me, it's a bit ambiguous since it doesn't tell me whether it includes the validation set of not.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -1180,3 +1185,37 @@ def evaluate(self, dataset, metric='auto'):
...
+            ('Feature column', 'feature'),
+            ('Target column', 'target')
+        ]
+        training_fields = [
+            ('Training Iterations', 'max_iterations'),
+            ('Training Accuracy', 'training_accuracy'),
+            ('Validation Accuracy', 'validation_accuracy'),
+            ('Training Time', 'training_time'),
+            ('Number of Examples', 'num_examples'),
+            ('Final Loss (specific to model)', 'training_log_loss')
```
shantanuchhabra: On further thought, for backward compatibility reasons, I don't actually think we should change the name of the state variable from `training_loss` to `training_log_loss`. @nickjong @znation @srikris  What do you think?

src/toolkits/drawing_classifier/drawing_classifier.cpp
srikris: Sadly we have used `num_examples` everywhere :(. We have to change it everywhere if we do it here.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
srikris: We should leave it as `training_loss`. You might change the loss in the future.

src/toolkits/drawing_classifier/drawing_classifier.cpp
srikris: But it adds it to the API exposed via Python so it's better for the API to be consistent irrespective of the options that are used during the training process.

src/python/turicreate/toolkits/_model.py
```diff
@@ -513,6 +513,9 @@ def summary(self, output=None):
             return self.__class__.__name__
 
 
+    def _list_fields(self):
```
srikris: This seems strange. You are simply calling the same function by itself? Am I missing something here. Isn't this a recursive call?

shantanuchhabra: Yeah, this is an oversight. Taking it out. Will rebase and squash to update the base commit for all other `__repr__` PRs.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
shreyajain17: Btw I removed loss in #2662

PR #2671, opened 2019-11-19
Added __repr__() for ObjectDetection_beta

shantanuchhabra:
Checks off Object Detection box in #2606 .

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/python/turicreate/toolkits/object_detector/object_detector.py
@@ -1567,24 +1567,25 @@ class ObjectDetector_beta(_Model):
     def __str__(self):
         """
         Return a string description of the model to the ``print`` method.
+
         Returns
         -------
         out : string
-            A description of the model.
+            A description of the ObjectDetector.
         """
-        return self.__class__.__name__
+        return self.__repr__()
 
     def __repr__(self):
         """
-        Returns a string description of the model, including (where relevant)
-        the schema of the training data, description of the training data,
-        training statistics, and model hyperparameters.
-        Returns
-        -------
-        out : string
-            A description of the model.
+        Print a string description of the model when the model name is entered
+        in the terminal.
         """
-        return self.__class__.__name__
+
+        width = 40
+        sections, section_titles = self._get_summary_struct()
+        out = _tkutl._toolkit_repr_print(self, sections, section_titles,
+                                         width=width)
+        return out
 
     def _get_version(self):
         return self._CPP_OBJECT_DETECTOR_VERSION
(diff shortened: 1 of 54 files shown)


src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1761,3 +1762,36 @@ def evaluate(self, dataset, metric='auto', output_type='dict', confidence_thresh
...
+        section_titles: list
+            A list of section titles.
+              The order matches that of the 'sections' object.
+        """
+        model_fields = [
+            ('Model', 'model'),
+            ('Number of classes', 'num_classes'),
+        ]
+        training_fields = [
+            ('Training time', 'training_time'),
```
shantanuchhabra: I changed this from `_training_time_as_string` to `training_time`, after confirming that pre-6.0 models also had a `training_time` state variable. Is this okay?

The alternatives are as follows:
- Add logic on the C++ side to compute the training time as a string to mimic the behavior pre-6.0.
- Use existing Python code and create a state variable on the Python layer to maintain time output consistent with pre-6.0 Turi Create.

nickjong: The problem isn't clear to me. What appears in 5.8? What appears in 6.0 in each of the alternative you're enumerating. Please provide context

nickjong: I'm assuming `_training_time_as_string` existed to improve the formatting of the value. How is the formatting with your implementation?

shantanuchhabra: In 5.8, when `__repr__()` is called, the training time printed out is a string of the form, "1 d 23 h 57 m 32 s". This string value was also stored as a state variable called `_training_time_as_string`. There was also a `training_time` state variable that would store the number of seconds training took. 

As of this PR in its current state, we have a state variable called "training_time" that also stores the number of seconds training took. I think it's definitely not okay to print out the training time as number of seconds instead of the more informational way in which we were printing it out earlier.

Here are the alternatives along with my take on them.

1. Add logic on the C++ side to compute the training time as a string to match what we printed pre-6.0.
2. Use existing Python code (the `_seconds_to_string` function in `src/python/turicreate/toolkits/object_detection/object_detector.py` to create a state variable at the Python layer only for `__repr__` purposes)

I think Option 1 makes the most sense -- we should make a `training_time_seconds_to_string` function that lives in `src/toolkits/utils/` that we can call on the C++ side. @nickjong thoughts?

nickjong: Why can't we just keep the state `training_time` having an integer value in seconds and use `_training_time_as_string` in the implementation of `__repr__()`? I don't think we need to persist it, do we?

shantanuchhabra: Per offline discussion, we decided to persist `_training_time_as_string` to align with the situation that `_toolkit_repr_print` prints out model field values.

Since `_seconds_to_string` is not a very complicated function, we will replicate this functionality on the C++ side in `src/toolkits/utils/` so it can be used by both OD and ST. I will put up a PR with just that functionality shortly, which the OD and ST PRs can depend on. Does that sound good?

shantanuchhabra: Update: The table printer has a `progress_time()` struct that we can use.

src/toolkits/object_detection/object_detector.hpp
```diff
@@ -10,6 +10,7 @@
 #include <functional>
 #include <map>
 #include <memory>
+#include <sstream>
```
nickjong: Why are we adding a header to the `.hpp` file if it's not being used in the `.hpp` file?

src/python/turicreate/toolkits/object_detector/object_detector.py
nickjong: Shouldn't this use `_training_time_as_string` now?

(How are you testing? Could you include an example of the final output?)

src/toolkits/object_detection/object_detector.hpp
shantanuchhabra: Moved it to the cpp file.

src/python/turicreate/toolkits/object_detector/object_detector.py
shantanuchhabra: Sorry, that is my bad. Fixed now. Here is an example of the final output:

```
Class                                    : ObjectDetector_beta

Schema
------
Model                                    : darknet-yolo
Number of classes                        : 4

Training summary
----------------
Training time                            : 790.529ms
Training epochs                          : 0
Training iterations                      : 1
Number of examples (images)              : 100
Number of bounding boxes (instances)     : 477
Final loss (specific to model)           : 6.1412
```

PR #2684, opened 2019-11-19
support imports from saved CustomModel

Jarvi-Izana:
layers with weights in 5.8.
```
  1 drawing_conv0_fwd
  2 drawing_conv1_fwd
  3 drawing_conv2_fwd
  4 drawing_dense0_fwd
  5 drawing_dense1_fwd
```

6.0
```
  1 drawing_conv0
  2 drawing_conv1
  3 drawing_conv2
  4 drawing_dense0
  5 drawing_dense1
```

This will affect the weights loaded from the saved model prior to 6.0 when we call `load`.


Jarvi-Izana:
just realize the situation is more complicated than I thought, 
```
drawing_conv0_bias
drawing_conv0_weight
```
The weight name is constructed by appending layer name with `_bias` and `_weights`. I will strip the `_fwd` in the middle.

Jarvi-Izana:
> Why is this conversion happening in TF? For our other toolkits, we migrate the pre-6.0 weights to the 6.0 names in C++, when we migrate the 5.8 Custom Model serialization
> 
> Do we now already have that for DC?

no. DC `load_version` doesn't have. I can migrate to cpp side.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -234,11 +234,11 @@ void drawing_classifier::init_options(
       FLEX_UNDEFINED,
       std::numeric_limits<int>::min(), std::numeric_limits<int>::max());
 
-  options.create_string_option(
+  options.create_boolean_option(
       "warm_start",
       "Record warm start model version used. If no warmstart used"
-      " None is assigned by default."
-      "",
+      " None is assigned by default.",
```
Jarvi-Izana: I'm not sure I'm doing the right thing. do @fareeha you want to store the warm start model name at here?

shantanuchhabra: @Jarvi-Izana let's revert this change. I think we want this to be a string option but I'll let @fareeha confirm.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -863,5 +864,105 @@ std::shared_ptr<coreml::MLModelWrapper> drawing_classifier::export_to_coreml(
...
+      new_classes_list.reserve(classes_list.size());
+
+      std::transform(classes_list.begin(), classes_list.end(), std::back_inserter(new_classes_list),
+        [](flexible_type& ft) { return ft.to<flex_int>(); });
+      
+      model_data.erase("classes");
+      model_data.emplace("classes", std::move(new_classes_list));
+    
+    } else if (classes_list.size() == 0) {
+      log_and_throw("The loaded turicreate model must have at least one class!\n");
```
srikris: `Unable to load model. Model must contain at least one class`

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -863,5 +864,105 @@ std::shared_ptr<coreml::MLModelWrapper> drawing_classifier::export_to_coreml(
...
+    }
+  } else {
+    log_and_throw("The loaded turicreate model must contain 'classes'!\n");
```
srikris: `Unable to load model. Unable to find classes saved in the model`

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: yeah, modified. please have a look!

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -237,7 +237,7 @@ void drawing_classifier::init_options(
   options.create_string_option(
       "warm_start",
       "Record warm start model version used. If no warmstart used"
-      " None is assigned by default."
+      "Empty string is assigned by default.",
       "",
```
Jarvi-Izana: The default value is an empty string. I assume empty string will be used instead of UNDEFINED or "None".

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -371,11 +371,22 @@ void drawing_classifier::init_training(
...
 
-  if ( enable_warmstart ) {
+  if (mlmodel_path.size()) {
     // Initialize the neural net with warm start model weights.
     model_spec warmstart_model(mlmodel_path);
     float_array_map trained_weights = warmstart_model.export_params_view();
     nn_spec_->update_params(trained_weights);
+
+    size_t idx = 0;
+#ifdef _WIN32
```
Jarvi-Izana: not sure we support windows or not. @nickjong @shantanuchhabra

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -341,28 +341,8 @@ void object_detector::import_from_custom_model(variant_map_type model_data,
...
-  model_data.emplace("grid_width", width);
-
-  model_data.emplace("annotation_scale", "pixel");
-  model_data.emplace("annotation_origin", "top_left");
-  model_data.emplace("annotation_position", "center");
-
-  state.clear();
-  state.insert(model_data.begin(), model_data.end());
 
+  const flex_dict& model = variant_get_value<flex_dict>(model_iter->second);
```
Jarvi-Izana: reorganize the code to follow the guide of C++: when you use it, you declare it.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -371,11 +371,22 @@ void drawing_classifier::init_training(
...
+    idx = mlmodel_path.rfind('\\');
+#else
+    idx = mlmodel_path.rfind('/');
+#endif
+    if (idx == std::string::npos)
+      state["warm_start"] = mlmodel_path;
+    else
+      state["warm_start"] = mlmodel_path.substr(idx + 1);
```
nickjong: Isn't this option populated by Python? Let's avoid duplicating the logic here. This state should reflect what the user selected, not the path that is the implementation detail

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: Let's just avoid doing path manipulations here

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -863,5 +874,111 @@ std::shared_ptr<coreml::MLModelWrapper> drawing_classifier::export_to_coreml(
   return model_wrapper;
 }
 
+void drawing_classifier::import_from_custom_model(variant_map_type model_data,
+                                               size_t version) {
```
nickjong: Let's run `git clang-format` (just on your own changes, please)

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -863,5 +874,111 @@ std::shared_ptr<coreml::MLModelWrapper> drawing_classifier::export_to_coreml(
...
+      flex_list new_classes_list;
+      new_classes_list.reserve(classes_list.size());
+
+      std::transform(classes_list.begin(), classes_list.end(),
+                     std::back_inserter(new_classes_list),
+                     [](flexible_type& ft) { return ft.to<flex_int>(); });
+
+      model_data.erase("classes");
+      model_data.emplace("classes", std::move(new_classes_list));
+
```
nickjong: I think all this could be expressed more concisely as:
```
for (flexible_type& ft : classes_list) {
  ft = ft.to<flex_int>();
}
model_data["classes"] = std::move(classes_list);
```

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -863,5 +874,111 @@ std::shared_ptr<coreml::MLModelWrapper> drawing_classifier::export_to_coreml(
...
+    layer_weight.clear();
+    layer_shape.clear();
+    layer_weight.insert(layer_weight.end(), model_weight.begin(), model_weight.end());
+    layer_shape.insert(layer_shape.end(), model_shape.begin(), model_shape.end());
+
+    nn_params[layer_name] = shared_float_array::wrap(std::move(layer_weight),
+                                                     std::move(layer_shape));
+  }
+
+  // missing state from prior 6.0; use default value
```
nickjong: Why do we need to add these things? Are they required by any of the inference logic? I think this just misleads people. The point of the random seed for example is to allow people to create the model. But the random seed you're generating below doesn't actually help somebody reproduce the model you've loaded.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: Don't you have to store the `option` into `state`?

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: Oops, my bad. Didn't notice it.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: great! I will give that.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: because we need to call `init_model()`, which is required by that function call. The random state is just to make sure the call can be done successfully. The random state won't take any effect since weights will be overridden later.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: Here, I'm storing the model name into the `state`. Not manipulating the option. The `state` will be serialized into disk when we save model.

shantanuchhabra: I believe the option gets stored to the state by https://github.com/apple/turicreate/blob/master/src/toolkits/drawing_classifier/drawing_classifier.cpp#L248 .

Jarvi-Izana: ha, I confused [model_path](https://github.com/apple/turicreate/blob/58527ce50121455b1e201198c1a4fbe2e695cf14/src/toolkits/drawing_classifier/drawing_classifier.cpp#L311) with `wram_start`. good catch.

src/toolkits/drawing_classifier/drawing_classifier.cpp
shantanuchhabra: @fareeha @nickjong @Jarvi-Izana 

Are we changing the behavior of the toolkit to use "" (which I assume later translates to FLEX_UNDEFINED, and mimics what used to be `warm_start=None` behavior) as the default value instead of "auto"?

Jarvi-Izana: like python, the empty string is not None. The default value is passed from the python layer. It should be fine.

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: Yes but it will appear to the user in the python attributes and mislead them into thinking that it can be used to reproduce the model. Let's not put it in the state and instead make `init_model` more robust. Nothing should need random initialization for loading a saved model. The weights should be read from disk, not random. (Let's not pay the overhead of randomly initializing weights and then overwriting them, either.)

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: I think I original suggested that the empty string be used here, but in hindsight, I wonder if `FLEX_UNDEFINED` works here?

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: good.

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: It should work. Which way do we want to go?

src/toolkits/drawing_classifier/drawing_classifier.cpp
Jarvi-Izana: addressed.

Jarvi-Izana: True that. And I just want to make sure I understand the context here. For the loading saved model from the prior 6.0 version, we are not treating it as a checkpoint, which means the user can continue to train the model, but a completed model, where it can only be used for inferencing. Is my understanding correct?

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -236,9 +245,9 @@ void drawing_classifier::init_options(
 
   options.create_string_option(
       "warm_start",
-      "Record warm start model version used. If no warmstart used"
-      " None is assigned by default."
-      "",
+      "Record warm start model version used. If no warmstart used, "
+      "empty string is assigned by default.",
+      FLEX_UNDEFINED,
```
Jarvi-Izana: later you cannot assign `UNDEFINED` for key `warm_start` through `add_or_update_state` because it expects a string type input, not `None`.

Jarvi-Izana: I think if you cannot reset the value of the key to its default value, then what's the purpose of setting that unreachable value as default?

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: We do not support checkpointing at this time. No model we are loading from 5.8 or earlier will be a checkpoint.

src/toolkits/drawing_classifier/drawing_classifier.cpp
nickjong: I see other instances in the codebase where we assigned `FLEX_UNDEFINED` into the model state, e.g. in the `supervised_learning` toolkits. Maybe you just can't do it via `add_or_update_state`, but that may just be a flaw of that implementation.

Anyway, what's more important is the behavior the user sees. I think using `FLEX_UNDEFINED` here is more likely to give us robust behavior here, insofar as this value translating into what the user expects if they query the `warm_start` property on the model class.

Jarvi-Izana:
wait on #2722

srikris:
@Jarvi-Izana @shantanuchhabra Can we merge this to get an RC out?

PR #2701, opened 2019-11-20
OD invalid inputs

jakesabathia2:
Catch those invalid inputs.


jakesabathia2:
Hmmm for `grid_shape`,
the thing is the toolkit won't throw any error or crash when it is a list of `string`.
And also `'a' > 0` will produce `True` in Python.
So I think we should still need to make sure that it contains `int` ? 
Any thought?

jakesabathia2:
Or we can check it in c++.
I don't know which one is better :) .

jakesabathia2:
This is good. Actually we don't need to check > 0,
since it is check in c++.

PR #2702, opened 2019-11-20
Added Progress Printing to Stylization method

abhishekpratapa:
## Overview
- Added stylization to the predict method of style_transfer.cpp


abhishekpratapa:
I was going to use `table_printer`, but I saw that 5.8 didn't use it and had a different way of printing progress for st. To maintain how we were doing it earlier, I wrote my own table printer but I guess now's the time to migrate the code to be more consistent across all toolkits.

Issue #2705, opened 2019-11-20
OD predict() returns 'identifier' instead of 'label'


jakesabathia2:
5.8:
```
{'confidence': 0.5260603757353245, 'type': 'rectangle', 'coordinates': {'y': 199.10811523220985, 'x': 119.97527128740205, 'width': 239.9505425748041, 'height': 85.82003666804388}, 'label': 'carsgraz'}
```

6.0:
```
{'confidence': 0.41931280493736267, 'identifier': 1, 'type': 'rectangle', 'coordinates': {'y': 0.3170698881149292, 'x': -0.04356098175048828, 'height': 0.18256889283657074, 'width': 0.43753254413604736}}
```

PR #2706, opened 2019-11-21
OD change identifier to label.

jakesabathia2:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -608,15 +608,16 @@ variant_type object_detector::predict(
...
       flex_dict bb_dict = {{"x", each_row.bounding_box.x}, {"y", each_row.bounding_box.y},
                       {"width", each_row.bounding_box.width},
                       {"height", each_row.bounding_box.height}};
-      flex_dict each_annotation = {{"identifier", each_row.identifier},
-                                   {"type", "rectangle"},
-                                   {"coordinates", std::move(bb_dict)},
-                                   {"confidence", each_row.confidence}
-                                   };
+      flex_dict each_annotation = {
+          {"label", class_labels[each_row.identifier].get<flex_string>()},
```
nickjong: Let's use `.to<flex_string>` here. `get` is not used for conversions.

srikris:
Fixes #2705

PR #2710, opened 2019-11-21
Removed Unnecessary Ifdefs for MacOS 10.15

abhishekpratapa:


abhishekpratapa:
Should fix the issue, though it's worth discussing. I build the smaller bits of ST on MacOS 10.15 and lower. But I build the parent class of ST on 10.15 only.

So the encoding and decoding layers are built on 10.14 but all the pieces together are built in 10.15. Should I just `#ifdef` those smaller pieces to be only on 10.15 as well?

PR #2712, opened 2019-11-21
Sound Classifier Fixes

TobyRoseman:
This gets all the unit test to pass on Linux. It also fixes lots of issues that will happen if there are more than one instances of a sound classifier.

### DIFF
--- a/src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
+++ b/src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
@@ -6,12 +6,8 @@
 from __future__ import print_function as _
 from __future__ import division as _
 from __future__ import absolute_import as _
-from turicreate.util import _ProgressTablePrinter
-import tensorflow as _tf
-import numpy as _np
-import time as _time
+
 from .._tf_model import TensorFlowModel
-import turicreate.toolkits._tf_utils as _utils
 
 import tensorflow.compat.v1 as _tf
 _tf.disable_v2_behavior()
@@ -22,17 +18,16 @@ class SoundClassifierTensorFlowModel(TensorFlowModel):
     def __init__(self, num_inputs, num_classes, custom_layer_sizes):
         """
         Defines the TensorFlow model, loss, optimisation and accuracy.
-
         """
+        from uuid import uuid1
 
-        _tf.reset_default_graph()
-
+        self.instance_id = str(uuid1())
         self.num_classes = num_classes
 
         self.x = _tf.placeholder("float", [None, 12288])
         self.y = _tf.placeholder("float", [None, self.num_classes])
 
-        #xavier initialization
+        # Xavier initialization
         initializer = _tf.keras.initializers.glorot_uniform()
 
         weights = {}
@@ -41,8 +36,8 @@ class SoundClassifierTensorFlowModel(TensorFlowModel):
 
         # Create variables for customized layers
         for i, cur_layer_size in enumerate(custom_layer_sizes):
-            weight_name = 'sound_dense{}_weight'.format(i)
-            bias_name = 'sound_dense{}_bias'.format(i)
+            weight_name = '{}_sound_dense{}_weight'.format(self.instance_id, i)
+            bias_name = '{}_sound_dense{}_bias'.format(self.instance_id, i)
             self.names_of_layers.append('dense{}'.format(i))
             out_units = cur_layer_size
             if i==0:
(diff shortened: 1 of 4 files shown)


abhishekpratapa:
WAHOO!

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -22,17 +18,16 @@ class SoundClassifierTensorFlowModel(TensorFlowModel):
     def __init__(self, num_inputs, num_classes, custom_layer_sizes):
         """
         Defines the TensorFlow model, loss, optimisation and accuracy.
-
         """
+        from uuid import uuid1
 
-        _tf.reset_default_graph()
-
+        self.instance_id = str(uuid1())
```
Jarvi-Izana: this won't work since each time predict/predict_topk is called, a new python model object will be created. So every time you create a new id instead of reusing the previous id.

Jarvi-Izana: I think a class level static variable is already sufficient to provide a unique uuid for each ml model class.

TobyRoseman: A new python model is not created each time `predict` is called. I *think* that's how many of the other deep learning toolkits work, but that's not how the sound classifier works.

hoytak:
Loading a 5.8 model with this gives

```
In [3]: model = tc.load_model("/Users/hoytak/data/test_sc.model/")
WARNING:root:TensorFlow version 2.0.0 detected. Last version known to be fully compatible is 1.13.1 .
WARNING:tensorflow:From /Users/hoytak/venv_60b1/lib/python2.7/site-packages/tensorflow_core/python/compat/v2_compat.py:65: disable_resource_variables (from tensorflow.python.ops.variable_scope) is deprecated and will be removed in a future version.
Instructions for updating:
non-resource variables are not supported in the long term
WARNING:tensorflow:From /Users/hoytak/venv_60b1/lib/python2.7/site-packages/tensorflow_core/python/compat/v2_compat.py:65: disable_resource_variables (from tensorflow.python.ops.variable_scope) is deprecated and will be removed in a future version.
Instructions for updating:
non-resource variables are not supported in the long term
2019-11-21 13:44:27.914408: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2019-11-21 13:44:27.935343: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x7fe042c43570 executing computations on platform Host. Devices:
2019-11-21 13:44:27.935364: I tensorflow/compiler/xla/service/service.cc:175]   StreamExecutor device (0): Host, Default Version
KeyError                                  Traceback (most recent call last)
<ipython-input-3-54a956f15cfc> in <module>()
----> 1 model = tc.load_model("/Users/hoytak/data/test_sc.model/")

/Users/hoytak/venv_60b1/lib/python2.7/site-packages/turicreate/toolkits/_model.pyc in load_model(location)
    132                     return cls(model_data)
    133
--> 134                 return cls._load_version(model_data, model_version)
    135
    136         elif hasattr(_extensions, name):

/Users/hoytak/venv_60b1/lib/python2.7/site-packages/turicreate/toolkits/sound_classifier/sound_classifier.pyc in _load_version(cls, state, version)
    563             from ._mx_sound_classifier import MultiLayerPerceptronMXNetModel
    564             custom_NN = MultiLayerPerceptronMXNetModel(num_inputs, num_classes, custom_layer_sizes, 1)
--> 565         custom_NN.load_weights(state['_custom_classifier'])
    566         state['_custom_classifier'] = custom_NN
    567

/Users/hoytak/venv_60b1/lib/python2.7/site-packages/turicreate/toolkits/sound_classifier/_tf_sound_classifier.pyc in load_weights(self, net_params)
    190             else:
    191                 curr_shape = [int(x) for x in net_params['shapes'][layer_name]]
--> 192                 self.sess.run(_tf.assign(_tf.get_default_graph().get_tensor_by_name(new_layer_name+":0"),
    193                                         net_params['data'][layer_name].reshape(curr_shape).transpose(1,0)))
    194

/Users/hoytak/venv_60b1/lib/python2.7/site-packages/tensorflow_core/python/framework/ops.pyc in get_tensor_by_name(self, name)
   3784       raise TypeError("Tensor names are strings (or similar), not %s." %
   3785                       type(name).__name__)
-> 3786     return self.as_graph_element(name, allow_tensor=True, allow_operation=False)
   3787
   3788   def _get_tensor_by_tf_output(self, tf_output):

/Users/hoytak/venv_60b1/lib/python2.7/site-packages/tensorflow_core/python/framework/ops.pyc in as_graph_element(self, obj, allow_tensor, allow_operation)
   3608
   3609     with self._lock:
-> 3610       return self._as_graph_element_locked(obj, allow_tensor, allow_operation)
   3611
   3612   def _as_graph_element_locked(self, obj, allow_tensor, allow_operation):

/Users/hoytak/venv_60b1/lib/python2.7/site-packages/tensorflow_core/python/framework/ops.pyc in _as_graph_element_locked(self, obj, allow_tensor, allow_operation)
   3650           raise KeyError("The name %s refers to a Tensor which does not "
   3651                          "exist. The operation, %s, does not exist in the "
-> 3652                          "graph." % (repr(name), repr(op_name)))
   3653         try:
   3654           return op.outputs[out_n]

KeyError: "The name '178fc35e-0ca8-11ea-8c6f-8c8590740729:0' refers to a Tensor which does not exist. The operation, '178fc35e-0ca8-11ea-8c6f-8c8590740729', does not exist in the graph."

In [4]:
```

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
Jarvi-Izana: I see! I thought SC is written in Cpp. My bad. Will this namespace thing solve it? Do we have to disable all v2 functionalities for this release?

nickjong: Adding this UUID seems invasive to me. In resolving #2726 it seemed to suffice to use `tf.Graph.as_default()`, see #2728 

Could we use this PR just to address problem #1 that we discussed? The consistent disabling of v2 behavior?

TobyRoseman: @nickjong - I agree there is probably a better fix than this PR. It sounds like your basically suggesting option 2 from #2731. I agree that would be better. Although I think there may be some subtle issues with code that is actually using the V2 APIs.

nickjong: Agreed! For 6.0, I think only the SC case is P1. It's wrong that we're clobbering global state in the other toolkits, but only SC is actually broken as a result. Anyway, it looks like we can resolve all of the SC issues with #2741

TobyRoseman:
Closing this in favor of #2741

PR #2714, opened 2019-11-21
Never call tensorflow.compat.v1.reset_default_graph

TobyRoseman:
The default graph contains other instances of TuriCreate model as well as other TensorFlow models the user might have created.


nickjong:
I think I need more context to understand this change. What problem or bug does it resolve? If the default graph is something that contains instances of Turi Create models, don’t we need to reset it before using it for a different Turi Create model?

TobyRoseman:
This is not the right fix, at least not in isolation. We should not being `tensorflow.compat.v1.reset_default_grap()` but just removing those call breaks a lot of things. I've created #2731 to track this issue.

PR #2715, opened 2019-11-21
Fix IC/DC annotation on Linux

### LINKED ISSUE #2401: Classifier - add_image_features method is not defined
@znation @TobyRoseman  add_image_features method is not defined

**Error:**
import turicreate as tc
 images = tc.image_analysis.load_images("/home/aximsoft/Desktop/images")
 images = tc.image_classifier.annotate(images)
[15:54:15] src/nnvm/legacy_json_util.cc:209: Loading symbol saved by previous version v0.8.0. Attempting to upgrade...
[15:54:15] src/nnvm/legacy_json_util.cc:217: Symbol successfully upgraded!
  File "/home/aximsoft/.local/lib/python2.7/site-packages/turicreate/toolkits/image_classifier/_annotate.py", line 118, in annotate
    annotation_window.add_image_features(feature_sframe)
  File "/home/aximsoft/.local/lib/python2.7/site-packages/turicreate/extensions.py", line 299, in __getattr__
    raise AttributeError("no attribute " + name)
AttributeError: no attribute add_image_features

znation:
Removes call to `add_image_features` -- looks like this was perhaps prototyped but never made it into master, as the code on the server side does not exist. This fixes #2401.

Also in this PR:
* Moves an assertion into a case where it makes sense; empty SArray's dtype is reported as 0 (integer) but that's spurious (as soon as it contains something other than null the dtype will reflect that type).
* Adds the build directory to be marked as `linguist-generated` in `.gitattributes`. This should cause future PRs to ignore/hide code changes in this directory (as well as not counting its contents for language breakdown reported for the repo).


src/python/turicreate/toolkits/image_classifier/_annotate.py
```diff
@@ -119,10 +119,6 @@ def annotate(data, image_column=None, annotation_column='annotations'):
                             [image_column],
                             annotation_column
                         )
-    if __platform == "linux" or __platform == "linux2":
-        model = _image_feature_extractor._create_feature_extractor("squeezenet_v1.1")
-        feature_sframe = model.extract_features(data, image_column, verbose=False, batch_size=64)
-        annotation_window.add_image_features(feature_sframe)
```
znation: Note: since this doesn't appear to be implemented on the server side yet, I opened #2716 to add it.

PR #2721, opened 2019-11-21
Change has apple to HAS_MPS when using the MPS_Compute_Context

abhishekpratapa:
- Use MPS Compute Context when MPS exists not when Apple exists.

### DIFF
--- a/src/toolkits/style_transfer/style_transfer.cpp
+++ b/src/toolkits/style_transfer/style_transfer.cpp
@@ -18,9 +18,9 @@
 #include <toolkits/style_transfer/style_transfer_model_definition.hpp>
 #include <toolkits/util/training_utils.hpp>
 
-#ifdef __APPLE__
+#ifdef HAS_MPS
 #import <ml/neural_net/mps_compute_context.hpp>
-#endif  // __APPLE__
+#endif  // HAS_MPS
 
 namespace turi {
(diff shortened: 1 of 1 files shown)


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -335,11 +335,11 @@ std::unique_ptr<compute_context> style_transfer::create_compute_context()
 // for MacOS greater than 10.15 is performed. If it is then the Style Transfer
 // MPS implementation is used. On all other systems currently the TensorFlow
 // implementation is used.
```
shantanuchhabra: Comment needs to be updated based on the HAS_MPS change.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -18,9 +18,9 @@
 #include <toolkits/style_transfer/style_transfer_model_definition.hpp>
 #include <toolkits/util/training_utils.hpp>
 
-#ifdef __APPLE__
+#ifdef HAS_MPS
```
shantanuchhabra: Out of curiosity, where is `HAS_MPS` defined?

abhishekpratapa: Top Level CMakeLists.txt

PR #2724, opened 2019-11-21
Added printing to style transfer for verbosity on `max_iterations`

abhishekpratapa:
## Overview
- added verbosity for setting max_iterations in style transfer


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -377,6 +377,7 @@ void style_transfer::infer_derived_options() {
     flex_int max_iterations = estimate_max_iterations(
         read_state<flex_int>("num_styles"), read_state<flex_int>("batch_size"));
 
+    std::cout << "Setting max_iterations to be " << max_iterations << std::endl;
```
srikris: Please use logprogress_stream

PR #2728, opened 2019-11-22
Use a fresh tf.Graph for each data augmentation batch

### LINKED ISSUE #2726: TF image augmenter leaks memory
- OD in master crashes due to a memory leak

nickjong:
Fixes #2726

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
+++ b/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
@@ -27,20 +27,22 @@ def get_augmented_data(images, annotations, output_height, output_width, resize_
     # Suppresses verbosity to only errors
     tf.logging.set_verbosity(tf.logging.ERROR)
 
-    session = tf.Session()
-    output_shape = (output_height, output_width)
-    
-    if resize_only:
-        images = get_resized_images(images, output_shape)
-        resized_images = session.run(images)
-        resized_images = np.array(resized_images, dtype=np.float32)
-        return tuple((resized_images, len(resized_images)*[np.zeros(6)]))
-    else:
-        imgs, transformations = get_augmented_images(images, output_shape)
-        augmented_images, trans = session.run([imgs, transformations])
-        augmented_annotations = apply_bounding_box_transformation(images, annotations, trans, output_shape)
-        augmented_images = np.array(augmented_images, dtype=np.float32)
-        return tuple((augmented_images, augmented_annotations))
+    graph = tf.Graph()
+    with graph.as_default():
+        with tf.Session() as session:
+            output_shape = (output_height, output_width)
+
+            if resize_only:
+                images = get_resized_images(images, output_shape)
+                resized_images = session.run(images)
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,20 +22,27 @@
...
-    else:
+
+def get_augmented_data(images, annotations, output_height, output_width):
+
+    output_shape = (output_height, output_width)
+
+    # Suppresses verbosity to only errors
+    tf.logging.set_verbosity(tf.logging.ERROR)
+
+    with tf.Session() as session:
```
Jarvi-Izana: Could you have any explanation of this? It seems very rare that TF won't recycle unused sessions.

shreyajain17: On digging a bit, I have found resources that state that session leaks may occur if sessions are not closed properly. I am trying to understand a bit more by reproducing it and looking at the memory footprint.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -27,7 +27,14 @@ def get_augmented_data(images, annotations, output_height, output_width, resize_
     # Suppresses verbosity to only errors
     tf.logging.set_verbosity(tf.logging.ERROR)
 
-    session = tf.Session()
+    graph = tf.Graph()
+    with graph.as_default:
+        with tf.Session() as session:
```
shreyajain17: I just saw this. Wouldn't you have to mention `with tf.Session(graph=graph) as session:`

nickjong:
It turns out the underlying problem here is the same as the sound classifier issues: we keep reusing the default graph, which grows without bound as we add more and more operations. (Unlike the core NN models, we have to extend new graph nodes for each batch, since the precise operations being performed are randomly determined each time.)

The solution is to apply the `Graph` and `Session` types more correctly. See the introductory comments at https://www.tensorflow.org/api_docs/python/tf/Graph and https://www.tensorflow.org/api_docs/python/tf/compat/v1/Session

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
nickjong: I believe the `graph.as_default()` takes care of that: any attempt to use the default graph, including by the `tf.Session` initializer, will use our local `graph` instance instead

shreyajain17: I think it should be `graph.as_default()`

nickjong:
Ugh, my "cleaned up" version doesn't actually work. More functional patch forthcoming

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
nickjong: Yup, I got it working like that in some messed up code and then thought I could clean it up without testing before I pushed, sigh

Issue #2707, opened 2019-11-21
OD predict should return bounding box coordinate in normal image scale.


jakesabathia2:
`predict` for OD returns something like this
```
'coordinates': {'y': 0.2976797819137573, 'x': -0.0002831071615219116, 'height': 0.19749416410923004, 'width': 0.3819955289363861}
```
for bounding box which is in normalized coordinate.
Should change it to the normal coordinate as shown in the documentation.

PR #2729, opened 2019-11-22
od predict correct boundingbox.

jakesabathia2:
fix https://github.com/apple/turicreate/issues/2707


jakesabathia2:
Oh now the format of the predict() only match the bounding box definition for turicreate only (top_left for origin, center for position).
@nickjong should we return different format according to the model state:
`state['annotation_origin']`, `state['annotation_position']` ? 
(I actually don't know why we allow user to define the bounding box coordinate for annotation at the c++ level,
but I guess it is for createml).

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -48,10 +48,17 @@ struct image_box {
...
 
+  // Set image dimentsion
+  void set_image_dimension(float image_width, float image_height);
+
   float x = 0.f;
   float y = 0.f;
   float width = 0.f;
   float height = 0.f;
+
+  // Store true image width and height
```
nickjong: I don’t think we should store this state here. This type is meant to be closer to `CGRect` conceptually. See other comments below for other approaches to this bug

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -237,6 +244,8 @@ class image_augmenter {
      * each inner vector should have the same length as the corresponding input
      * image's annotations vector. */
     std::vector<std::vector<image_annotation>> annotations_batch;
+
+    std::vector<std::pair<float, float>> img_dimensions_batch;
```
nickjong: I’d rather not add this state here. It confuses the purpose of this type, which contains resized/ augmented images. The original image dimensions are not relevant here. 

We need to keep these types streamlined to optimize for future maintainability of the core image augmentation pipeline

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -610,9 +610,17 @@ variant_type object_detector::predict(
...
     flex_list class_labels = read_state<flex_list>("classes");
     for (const image_annotation& each_row : predicted_row) {
-      flex_dict bb_dict = {{"x", each_row.bounding_box.x}, {"y", each_row.bounding_box.y},
-                      {"width", each_row.bounding_box.width},
-                      {"height", each_row.bounding_box.height}};
+      float width_scale = each_row.bounding_box.img_width;
+      float height_scale = each_row.bounding_box.img_height;
+
+      flex_dict bb_dict = {
+          {"x", (each_row.bounding_box.x + each_row.bounding_box.width / 2.) *
```
nickjong: Let’s factor out this kind of logic to `od_data_iterator.hpp`, since that’s where lives the code for converting from the user-provided (`flex_dict`) data

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -779,6 +797,13 @@ void object_detector::perform_predict(gl_sframe data,
       input_batch[i].annotations.clear();
     }
 
+    // Store image dimentions
+    result_batch.img_dimensions_batch.resize(input_batch.size());
```
nickjong: Instead of storing this state I’m the `result_batch`, let’s create a new struct that contains the result batch and, say, the entire original vector of `labeled_image` instances. This struct can be defined in the header (in the private section of the class)

src/ml/neural_net/image_augmentation.hpp
jakesabathia2: good!

src/ml/neural_net/image_augmentation.hpp
jakesabathia2: good!

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: good!

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: the `consumer` is a function of two `std::vector<image_annotation`, so I stored the real image size in `image_annotation` instead.

nickjong: I think it would be clearer to change the `consumer` function to take the image's original dimensions

When in doubt, leave the lower-level code as simple as possible!

jakesabathia2:
good!

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: good!

src/ml/neural_net/image_augmentation.hpp
```diff
@@ -239,6 +239,13 @@ class image_augmenter {
     std::vector<std::vector<image_annotation>> annotations_batch;
   };
 
+  /** Data structure that contains result and labeled image. */
+  struct result_with_dimension {
```
nickjong: This type is not used or otherwise referenced anywhere in `image_augmenter.cpp`. Please move it to `object_detector` as a private struct

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -778,17 +796,27 @@ void object_detector::perform_predict(gl_sframe data,
...
+      result_with_dimension_batch.result_batch.annotations_batch[i] =
+          std::move(input_batch[i].annotations);
       input_batch[i].annotations.clear();
     }
 
+    // Store image dimentions
+    result_with_dimension_batch.image_dimensions_batch.resize(
+        input_batch.size());
+    for (size_t i = 0; i < input_batch.size(); ++i) {
+      result_with_dimension_batch.image_dimensions_batch[i] = std::make_pair(
```
nickjong: Nit: in most NN-specific code, height comes before width. Let’s preserve that ordering to minimize confusion, especially since we’re just using a tulle here, which doesn’t name its elements

src/ml/neural_net/image_augmentation.hpp
nickjong: Also worth considering having this inherit from `result` 

When moved to the OD code, though, could probably be called something like `inference_batch` or something like that. The name should give some indication of how it’s used. Here, the name `result_with_dimension` is not clear, since this type isn’t used as the result of any function defined in this file.

jakesabathia2: this is nice.

src/toolkits/object_detection/object_detector.cpp
jakesabathia2: good!

jakesabathia2:
Pass on gitlab. Will merge it.

PR #2730, opened 2019-11-22
Added crop functionality to MPS Objective C

abhishekpratapa:
## Major Fixes
- added crop to Objective C in MPS

## Minor Fixes
- moved progress printing table footer to the proper place
- added index to predict in MPS style transfer

### DIFF
--- a/src/ml/neural_net/style_transfer/mps_style_transfer.m
+++ b/src/ml/neural_net/style_transfer/mps_style_transfer.m
@@ -41,6 +41,26 @@
 
 @property (nonatomic) id<MTLDevice> dev;
 @property (nonatomic) id<MTLCommandQueue> commandQueue;
+
++ (NSData *) cropImageWidth:(NSData *) data
+                     height:(NSUInteger) height
+                 inputWidth:(NSUInteger) inputWidth
+                outputWidth:(NSUInteger) outputWidth
+                numChannels:(NSUInteger) numChannels;
+
++ (NSData *) cropImageHeight:(NSData *) data
+                 inputHeight:(NSUInteger) inputHeight
+                outputHeight:(NSUInteger) outputHeight
+                       width:(NSUInteger) width
+                 numChannels:(NSUInteger) numChannels;
+
++ (NSData *) topLeftCropImage:(NSData *) data
+                  inputHeight:(NSUInteger) inputHeight
+                   inputWidth:(NSUInteger) inputWidth
+                 outputHeight:(NSUInteger) outputHeight
+                  outputWidth:(NSUInteger) outputWidth 
+                  numChannels:(NSUInteger) numChannels;
+
 @end
 
 @implementation TCMPSStyleTransfer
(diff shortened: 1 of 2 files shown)


src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -251,6 +271,60 @@ - (instancetype) initWithDev:(id<MTLDevice>) dev
...
+  NSUInteger outputLength = height * outputWidth * numChannels;
+  NSMutableData *mutableData = [NSMutableData dataWithLength:outputLength];
+
+  NSUInteger xOffset = inputWidth - outputWidth;
+
+  for (NSUInteger idx = 0; idx < height; idx = idx + 1) {
+    NSUInteger startIndex = idx * inputWidth * numChannels;
+    NSUInteger endIndex = startIndex + (numChannels * (inputWidth - xOffset));
+
+    NSRange range = NSMakeRange(startIndex, endIndex);
```
nickjong: There should be a `sizeof(float)` in here somewhere, no?

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -251,6 +271,60 @@ - (instancetype) initWithDev:(id<MTLDevice>) dev
...
+                 outputHeight:(NSUInteger) outputHeight
+                  outputWidth:(NSUInteger) outputWidth 
+                  numChannels:(NSUInteger) numChannels {
+  NSData *croppedHeightData = [TCMPSStyleTransfer cropImageHeight:data
+                                                      inputHeight:inputHeight
+                                                     outputHeight:outputHeight
+                                                            width:inputWidth
+                                                      numChannels:numChannels];
+
+  return [TCMPSStyleTransfer cropImageWidth:croppedHeightData
```
nickjong: This incurs two copies of the data. Let’s do it with only one. For each contiguous portion of the output, figure out what to copy from the input. 

For clarity, I recommend specifying the arguments to this function as the desired offset and output height/width

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: yeah, I ran into issues testing it, I added size of float here.

src/ml/neural_net/style_transfer/mps_style_transfer.m
abhishekpratapa: I did it with one copy!

src/ml/neural_net/style_transfer/mps_style_transfer.m
```diff
@@ -251,6 +259,24 @@ - (instancetype) initWithDev:(id<MTLDevice>) dev
...
+                  numChannels:(NSUInteger) numChannels {
+  NSUInteger dataSize = sizeof(float) * numChannels * outputWidth * outputHeight;
+  NSMutableData *mutableData = [NSMutableData dataWithLength:dataSize];
+  for (NSUInteger idx = 0; idx < outputHeight; idx = idx + 1) {
+    NSUInteger rowSize = sizeof(float) * numChannels * outputWidth;
+    NSUInteger rowOffset = rowSize * idx;
+    NSUInteger startIndex = sizeof(float) * idx * numChannels * inputWidth;
+    NSRange range = NSMakeRange(startIndex, rowSize);
+    [data getBytes:mutableData.mutableBytes + rowOffset range:range];
+  }
```
nickjong: Let's try to adopt more of the patterns we see in e.g. NumPy and other various multi-array types we have, as well as in Apple APIs. Think in terms of strides. Something closer to:
```
NSUInteger outputRowStride = sizeof(float) * numChannels * outputWidth;
NSUInteger inputRowStride = sizeof(float) * numChannels * inputWidth;
for (NSUInteger idx = 0; idx < outputHeight; idx = idx + 1) {
  NSRange range = NSMakeRange(inputRowStride * idx, outputRowStride);
  [data getBytes:mutableData.mutableBytes + outputRowStride * idx range:range];
}
```
The above code should match the behavior of what you have here. But just to be clear, the desired behavior is to exclude only pixels on the bottom and right?

By reasoning about strides more explicitly here, we'll make it easier to reduce this to zero copies once `float_array` supports strides too. As an even more general principle, tend to factor out of loops variables than can be computed once beforehand.

PR #2733, opened 2019-11-22
od invalid threshold.

jakesabathia2:


src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1718,6 +1718,8 @@ def predict(self, dataset, confidence_threshold=0.25, iou_threshold=0.45, verbos
             # Visualize predictions by generating a new column of marked up images
             >>> data['image_pred'] = turicreate.object_detector.util.draw_bounding_boxes(data['image'], data['predictions'])
         """
+        _numeric_param_check_range('confidence_threshold', confidence_threshold, 0.0, 1.0)
```
srikris: Might be worth checking this via a simple set of unit tests that check that assertions are thrown.

jakesabathia2: good!

Issue #2720, opened 2019-11-21
Style Transfer crashes on stylize


shreyajain17:
Crashes if content is SFrame/SArray.
```
model.stylize(content)                                                                                   
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
~/turicreate/debug/src/python/turicreate/extensions.py in __run_class_function(self, fnname, args, kwargs)
    273         try:
--> 274             ret = self._tkclass.call_function(fnname, argument_dict)
    275         except RuntimeError as exc:

cy_model.pyx in turicreate._cython.cy_model.UnityModel.call_function()

cy_model.pyx in turicreate._cython.cy_model.UnityModel.call_function()

RuntimeError: Assertion failed: (/Users/shreya_jain3/turicreate/src/toolkits/style_transfer/style_transfer.cpp:252): 3==example.content_image.m_channels  [3 == 1]
. operator() from /Users/shreya_jain3/turicreate/src/toolkits/style_transfer/style_transfer.cpp at 252```

syoutsey:
BRB: could be bad error message.

abhishekpratapa:
That's an image channel issue. It's a bug that needs to be fixed.

PR #2740, opened 2019-11-22
Added resizing to channels to get them to match for ST

abhishekpratapa:
## Overview
- Style Transfer requires all of the channels to be of size 3 as the network depends on this.

## Solution
- Calling the resize method in image_utils get's us this functionality
- It was added upstream in the data loader because that's where the data for the style transfer model is prepared.

### DIFF
--- a/src/toolkits/style_transfer/style_transfer_data_iterator.cpp
+++ b/src/toolkits/style_transfer/style_transfer_data_iterator.cpp
@@ -18,10 +18,17 @@ namespace style_transfer {
 namespace {
 
 flex_image get_image(const flexible_type& image_feature) {
+  flex_image image;
   if (image_feature.get_type() == flex_type_enum::STRING) {
-    return read_image(image_feature, /* format_hint */ "");
+    image = read_image(image_feature, /* format_hint */ "");
   } else {
-    return image_feature;
+    image = image_feature;
+  }
+
+  if (image.m_channels != 3) {
+    return image_util::resize_image(image, image.m_width, image.m_height, 3, true);
+  } else {
+    return image;
   }
 }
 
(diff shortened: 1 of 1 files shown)


abhishekpratapa:
Fixes #2720

src/toolkits/style_transfer/style_transfer_data_iterator.cpp
```diff
@@ -18,10 +18,17 @@ namespace style_transfer {
...
+    image = read_image(image_feature, /* format_hint */ "");
   } else {
-    return image_feature;
+    image = image_feature;
+  }
+
+  if (image.m_channels != 3) {
+    return image_util::resize_image(image, image.m_width, image.m_height, 3, true);
+  } else {
+    return image;
```
nickjong: The output of `read_image` is typically encoded. As is, the images you are yielding are always RGB, but they may or may not be encoded depending on whether they were originally RGB.

To keep things simple, I suggest choosing upfront whether you want them to be encoded or decoded. Just always call `resize_image` with the desired flag. It'll be a no-op if the image is already in the desired state

nickjong:
> This won't work for ARGB, since `resize_image` only supports pair-wise conversions among Grayscale, RGB, and RGBA (https://github.com/apple/turicreate/blob/master/src/core/data/image/image_util_impl.cpp#L55). While not a 6.0 blocker, could you:
> 
> 1. Open an issue tracking the lack of support for ARGB image format,
> 2. Add guidance in our Userguide for (at least) Style Transfer on what users should do if they have an ARGB image (maybe give them a code snippet or a task-based procedure to convert it to a format we support)

I'm not sure what you mean. The function does convert RGBA to RGB, and we don't support ARGB layouts anywhere.

Issue #2607, opened 2019-11-12
Classifier evaluation should have an explore in 6.0


srikris:
Seems like a regression from 5.8. The evaluation should have an explore 
```
evals = model.evaluate(data)
evals.explore()
```

PR #2742, opened 2019-11-22
Added model evaluate to 6.0 matching behavior from 5.8

abhishekpratapa:
## Overview
Previously Drawing Classifier explore was unimplemented in 6.0

## Fixed
- Parity of `.explore` with 5.8 in 6.0
- `is` to using `==` for key comparison for robustness. Compare value rather than identity.

### DIFF
--- a/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -1204,7 +1204,69 @@ class DrawingClassifier(_Model):
           >>> results = model.evaluate(data)
           >>> print results['accuracy']
         """
-        return self.__proxy__.evaluate(dataset, metric)
+        class_label = self.__proxy__.predict(dataset, "class")
+        probability_vector = self.__proxy__.predict(dataset, "probability_vector")
+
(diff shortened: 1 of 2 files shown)


abhishekpratapa:
Fixes #2607

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -1204,7 +1204,69 @@ def evaluate(self, dataset, metric='auto'):
...
           >>> print results['accuracy']
         """
-        return self.__proxy__.evaluate(dataset, metric)
+        class_label = self.__proxy__.predict(dataset, "class")
+        probability_vector = self.__proxy__.predict(dataset, "probability_vector")
+
+        predicted  = _tc.SFrame({"label": class_label, "probability": probability_vector})
+        labels = self.classes
+
+        evaluation_result = self.__proxy__.evaluate(dataset, 'auto')
```
nickjong: Let's add a TODO here to avoid doing three passes through the test data

nickjong:
I'm confused now. What does IC do? In 5.8 and 6.0? Is there a path where we have `explore` working without those keys?

Issue #2723, opened 2019-11-21
Stylize on one image with multiple styles does not seem to work (no crash)


shreyajain17:
```
result = model.stylize(content['image'][0])
result.explore()
```

syoutsey:
@shreyajain17, what is the expected and actual behavior here?

abhishekpratapa:
Its a silly bug where the first style is always used to stylize the images in predict. The correct behavior would be to set the correct style number in predict before stylizing the image.

Since we didn't define a way of changing the index in MPS this is the behavior we have. The fix would be to set this index variable before the call to predict in MPS.

shreyajain17:
Could it be done `style_transfer.cpp` before MPS? I can look into it if that is the case.

PR #2746, opened 2019-11-23
Got Multiple Styles to work in MPS

abhishekpratapa:
- I was doing bad things.

### DIFF
--- a/src/ml/neural_net/mps_layer_instance_norm_data_loader.h
+++ b/src/ml/neural_net/mps_layer_instance_norm_data_loader.h
@@ -63,9 +63,11 @@ API_AVAILABLE(macos(10.14))
 
 - (void) loadBeta:(float *)beta;
 - (float *) beta;
+- (float *) betaWeights;
 
 - (void) loadGamma:(float *)gamma;
 - (float *) gamma;
+- (float *) gammaWeights;
 
 - (MPSCNNNormalizationGammaAndBetaState *)updateGammaAndBetaWithCommandBuffer:(id<MTLCommandBuffer>)commandBuffer 
                                               instanceNormalizationStateBatch:(MPSCNNInstanceNormalizationGradientStateBatch *)instanceNormalizationStateBatch;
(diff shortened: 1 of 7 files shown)


abhishekpratapa:
Fixes #2723

abhishekpratapa:
## Overview
Multiple styles wasn't working in TuriCreate because of this silly problem.

## Issue
- The instance norm layers which determine the stylistic feature of an image are supplied by a class called the `InstanceNormDataLoader`. This data loader takes care of all of the `beta` and `gamma` weights during inference as well as the `velocity` and `momentum` buffers during training.

- Two methods exist that need to be overridden on this data loader: the `gamma` and the `beta` method.

- The instanceNorm layer in mps only supports one style meaning one `gamma` and `beta` pair.

- To circumvent this limitation, in the InstanceNormDataLoader class we created `NSArrays` to store the different `gamma` and `beta` values.

- We used an indexing variable `styleIndex` to select the different `gamma` and `beta` values for different styles.

- To export the weights after training I added a concat to the `gamma` and `beta` methods so that all of the `gamma` and `beta` weights for all styles are concatenated into one contagious block of `NSData`

- This worked during export because the part of the style transfer application TuriCreate owned was aware that these were concatenated `gamma` and `beta` values.

- this didn't work during `predict` because these concatenated values were truncated by MPS since they had no awareness that these exported `gamma` and `beta` values were for multiple styles.

- Since the gamma and beta values were contagious the truncation happened to the `gamma` and `beta` values for the first index only.

## The Solution

- Separating the methods that exported the `gamma` and `beta` weights and the methods that gave the current `gamma` and current `beta` weights for the selected style, solved this issue.

- calling `gammaWeights` and `betaWeights` gave the callee the concatenated values of `gamma` and `beta` weights for all styles

- calling `gamma` and `beta` only gave the `gamma` and `beta` values for the current selected style.

In conclusion `MPS` was happy and `TuriCreate` exported all the weights it needed for multi style style-transfer.

Issue #2748, opened 2019-11-23
OD output regression for evaluation.


jakesabathia2:
5.8:
```
pred = model.evaluate(data, metric=mean_average_precision", output_type="sframe)
print(pred)
+----------+
|  label   |
+----------+
|   bike   |
| carsgraz |
+----------+
```

6.0:
```
pred = model.evaluate(data, metric=mean_average_precision", output_type="sframe)
print(pred)
++
||
++
++
```

PR #2749, opened 2019-11-23
od output regression evalaution.

jakesabathia2:
fix https://github.com/apple/turicreate/issues/2748

### DIFF
--- a/src/python/turicreate/test/test_object_detector.py
+++ b/src/python/turicreate/test/test_object_detector.py
@@ -289,8 +289,7 @@ class ObjectDetectorTest(unittest.TestCase):
         metrics = ["mean_average_precision_50","mean_average_precision"]
         for metric in metrics:
             pred = self.model.evaluate(self.sf.head(), metric=metric, output_type="sframe")
-            self.assertEqual(len(pred.column_names()), 1)
-            self.assertEqual(pred.column_names()[0], "label")
+            self.assertEqual(pred.column_names(), ["label"])
 
     def test_evaluate_invalid_metric(self):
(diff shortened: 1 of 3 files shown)


TobyRoseman:
Would it be simple to create a unit test which fails without your fix? If it is easy to do, I think it's worth doing.

jakesabathia2:
Yes it should be easy to create a unit test for this. Thanks for pointing it out.

jakesabathia2:
Done @TobyRoseman

src/python/turicreate/test/test_object_detector.py
```diff
@@ -285,6 +285,13 @@ def test_evaluate_invalid_threshold(self):
         with self.assertRaises(_ToolkitError):
             self.model.evaluate(self.sf.head(), iou_threshold  =-1)
 
+    def test_evaluate_sframe_format(self):
+        metrics = ["mean_average_precision_50","mean_average_precision"]
+        for metric in metrics:
+            pred = self.model.evaluate(self.sf.head(), metric=metric, output_type="sframe")
+            self.assertEqual(len(pred.column_names()), 1)
+            self.assertEqual(pred.column_names()[0], "label")
```
TobyRoseman: This looks wrong to me, but I think it's alright. You can't count on the column ordering of SFrames to always be the same, but I think this is ok cause there is only one column. In which case I think it would be better to do:
`self.assertEqual(pred.column_names(), ["label"])

jakesabathia2: good!

PR #2751, opened 2019-11-25
turicreate.load_audio does not fail on invalid path

### LINKED ISSUE #2686: turicreate.load_audio does not fail on invalid path
Turicreate load_audio does not fail if a path does not exist:

```
data = tc.load_audio("Path does not exist")
```

simply produces an empty sframe.

dhivyaaxim:
This PR resolves #2686 
turicreate.load_audio does not fail on invalid path

### DIFF
--- a/src/python/turicreate/toolkits/audio_analysis/audio_analysis.py
+++ b/src/python/turicreate/toolkits/audio_analysis/audio_analysis.py
@@ -65,10 +65,13 @@ def load_audio(path, with_path=True, recursive=True, ignore_failure=True, random
     if _fnmatch(path, '*.wav'):    # single file
         all_wav_files.append(path)
     elif recursive:
-        for (dir_path, _, file_names) in _os.walk(path):
-            for cur_file in file_names:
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/audio_analysis/audio_analysis.py
```diff
@@ -65,10 +65,13 @@ def load_audio(path, with_path=True, recursive=True, ignore_failure=True, random
     if _fnmatch(path, '*.wav'):    # single file
         all_wav_files.append(path)
     elif recursive:
-        for (dir_path, _, file_names) in _os.walk(path):
```
TobyRoseman: This will only fail if the user gives a directory that doesn't exist. It should also fail if the user gives a single file which does not exist.

It's probably best to check early (i.e. outside of the `if`/`elif`) if the path exists.

hoytak:
This was merged incorrectly.  Sorry.  I meant to merge a different.  Please reopen it and sorry for the hassle.

PR #2754, opened 2019-11-27
Add abs SArray method

### LINKED ISSUE #1885: Feature request: abs()
pandas has DataFrame.abs() operation while TC does not. It will be useful to have abs() operation as well.

dhivyaaxim:
This PR resolves #1885 
Feature request: abs()

### DIFF
--- a/src/python/turicreate/data_structures/sarray.py
+++ b/src/python/turicreate/data_structures/sarray.py
@@ -4433,3 +4433,6 @@ class SArray(object):
         immutable, this is identical to __copy__.
         """
         return SArray(_proxy = self.__proxy__)
+
+    def abs(self):
+        return abs(self)
--- a/src/python/turicreate/test/test_sarray.py
+++ b/src/python/turicreate/test/test_sarray.py
@@ -3215,3 +3215,10 @@ class SArrayTest(unittest.TestCase):
         self.assertEqual(sorted(x.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'])), ['cat', 'dog'])
         self.assertEqual(sorted(x.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'], exclude=True)), ['cow', 'horse'])
         self.assertEqual(sorted(x.filter_by('dog')), ['dog'])
+
+    def test_abs(self):
+        sa = SArray([-1.10, 2, -3.33, 4])
+        result = sa.abs()
+        expected = [1.1, 2.0, 3.33, 4.0]
+        self.__test_equal(result, expected, float)
+


src/python/turicreate/data_structures/sarray.py
```diff
@@ -4433,3 +4433,6 @@ def __deepcopy__(self, memo):
         immutable, this is identical to __copy__.
         """
         return SArray(_proxy = self.__proxy__)
+
+    def abs(self):
```
TobyRoseman: Please add a docstring comment here.

src/python/turicreate/data_structures/sarray.py
```diff
@@ -4433,3 +4433,6 @@ def __deepcopy__(self, memo):
...
+        return abs(self)
```
TobyRoseman: `return self.apply(abs)` will be faster since it will be done in parallel.

hoytak: abs(self) is actually hardcoded to simply call the self.__abs__(self) method, so this is fine.

TobyRoseman: I didn't realize we had a `__abs__` method. Perhaps it would be better to just call that directly.

hoytak: It's a weird intersection -- most mathematical stuff is done through methods, but python has a smattering of built-ins that break that pattern.  I'm fine with both existing here.

TobyRoseman:
@dhivyaaxim - Once we get a docstring, this change is ready to go in.

src/python/turicreate/test/test_sarray.py
```diff
@@ -3215,3 +3215,10 @@ def test_filter_by(self):
         self.assertEqual(sorted(x.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'])), ['cat', 'dog'])
         self.assertEqual(sorted(x.filter_by(['cat', 'hamster', 'dog', 'fish', 'bird', 'snake'], exclude=True)), ['cow', 'horse'])
         self.assertEqual(sorted(x.filter_by('dog')), ['dog'])
+
+    def test_abs(self):
```
guihao-liang: well, if you search `__abs__`, you can move those tests to your `test_abs` section.

src/python/turicreate/data_structures/sarray.py
guihao-liang:
```
        """
        Returns the absolute value of each element of the Sarray.

        Examples
        --------
        >>> abs(tc.SArray([1, -1, -2]))
        dtype: int
        Rows: 3
        [1, 1, 2]

        >>> abs(tc.SArray([-1., -2.]))
        dtype: float
        Rows: 2
        [1.0, 2.0]
        """
```
Here's an example to do it. I cannot push to your @dhivyaaxim branch, and could you find some time to put the docstring on? Thanks!

PR #2755, opened 2019-11-27
Issue_1909_Add_a_to_dict_method_on_SFrame

### LINKED ISSUE #1909: Add a to_dict method on SFrame
Pandas has a `to_dict` method, that returns this equivalent representation:
```
{k:df[k] for k in df.columns}
```

For convenience (especially for those porting scripts from Pandas to SFrame), we should provide an equivalent. In the meantime, here's a workaround:
```
sf_dict = {k:sf[k] for k in sf.column_names()}
```

dhivyaaxim:
This PR resolves #1909 
Add a to_dict method on SFrame

### DIFF
--- a/src/python/turicreate/data_structures/sframe.py
+++ b/src/python/turicreate/data_structures/sframe.py
@@ -5881,6 +5881,17 @@ class SFrame(object):
         """
         gsf = GroupedSFrame(self, key_columns)
         return gsf
+    
+    def to_dict(self):
+        dict_sframe = {}
+        for key in self.column_names():
+            sa = SArray(self[key])
+            val_list = []
+            for val in sa:
+                val_list.append(val)
+            dict_sframe.update({key:val_list})
+            del val_list
+        return dict_sframe
 
     @property
     def shape(self):
--- a/src/python/turicreate/test/test_sframe.py
+++ b/src/python/turicreate/test/test_sframe.py
@@ -3787,6 +3787,12 @@ class SFrameTest(unittest.TestCase):
         sf2 = SFrame.read_json(f.name)
         _assert_sframe_equal(sf, sf2)
 
+    def test_to_dict(self): 
+        sf = SFrame({'col1': [1, 2],'col2': [0.5, 0.75]}) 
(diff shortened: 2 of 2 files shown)


src/python/turicreate/data_structures/sframe.py
```diff
@@ -5881,6 +5881,17 @@ def _group(self, key_columns):
         """
         gsf = GroupedSFrame(self, key_columns)
         return gsf
+    
+    def to_dict(self):
+        dict_sframe = {}
+        for key in self.column_names():
+            sa = SArray(self[key])
+            val_list = []
+            for val in sa:
```
znation: Hmm; Pandas actually seems to behave a bit differently. I'm not sure why but it gives dictionaries of dictionaries (where the inner dictionaries have numeric keys ranging from 0 to n for n rows), instead of dictionaries of lists. I suspect this is to better handle sparse dataframes? Example of this behavior:

```python
In [23]: d = df.to_dict()

In [24]: type(d)
Out[24]: dict

In [25]: d.keys()
Out[25]: ['path', 'image']

In [26]: type(d['path'])
Out[26]: dict

In [27]: d['path'].keys()
Out[27]: 
[0,
 1,
 2,
...
```

In any case, we should probably preserve the same behavior here -- let's use a dict instead of a list for the inner structure, using row index as the key.

hoytak: I'm not sure I understand pandas' behavior here.

znation: Pandas always gives a sparse representation (even if the original data was dense) -- i.e. using dictionaries instead of lists, with row index as the dictionary key.

TobyRoseman: @znation - what do you recommend we do here? Should we modify this pull so we match pandas' behavior? Should we close the issue (#1909) as will not fix?

znation: I'm not sure I have an opinion on this one. I am not a fan of Pandas behavior here, but the goal of #1909 was to match Pandas behavior.

PR #2756, opened 2019-11-27
SFrame.read_json_should_accept_a_JSON_string_in_addition_to_files_2564

### LINKED ISSUE #2564: SFrame.read_json should accept a JSON string in addition to files
There are use cases which require a JSON string to accepted to the `SFrame.read_json` initializer instead of a path to a file. This could happen when the SFrame needs to be built with the results of a web-request API, in which case avoiding writing to a file would be more performant.

dhivyaaxim:
This PR resolves #2564 
 SFrame.read_json should accept a JSON string in addition to files

### DIFF
--- a/src/python/turicreate/data_structures/sframe.py
+++ b/src/python/turicreate/data_structures/sframe.py
@@ -1603,13 +1603,18 @@ class SFrame(object):
         [3 rows x 1 columns]
         """
         if orient == "records":
-            g = SArray.read_json(url)
-            if len(g) == 0:
-                return SFrame()
-            if g.dtype != dict:
-                raise RuntimeError("Invalid input JSON format. Expected list of dictionaries")
-            g = SFrame({'X1':g})
-            return g.unpack('X1','')
+            if type(url)==str and url[0] !='{' :
+                g = SArray.read_json(url)
+                if len(g) == 0:
+                    return SFrame()
+                if g.dtype != dict:
+                    raise RuntimeError("Invalid input JSON format. Expected list of dictionaries")
+                g = SFrame({'X1':g})
+                return g.unpack('X1','')
+            elif type(url)==str and url[0] =='{' :
+                url=pandas.read_json(url)
+                url=SFrame(url)
+                return url
         elif orient == "lines":
             g = cls.read_csv(url, header=False,na_values=['null'],true_values=['true'],false_values=['false'],
                     _only_raw_string_substitutions=True)
--- a/src/python/turicreate/test/test_sframe.py
+++ b/src/python/turicreate/test/test_sframe.py
@@ -201,6 +201,13 @@ class SFrameTest(unittest.TestCase):
             sf = SFrame.read_csv(csvfile.name, header=True)
             self.assertEqual(sf.dtype, [float, int, str])
             self.__test_equal(sf, df)
+    def test_read_json(self):
+        x='{"row 1":{"col 1":"a","col 2":"b"},"row 2":{"col 1":"c","col 2":"d"}}'
+        sf=SFrame.read_json(x)
+        df =pd.read_json(x)
+        df=df.reset_index()
+        df=df.drop(['index'],axis=1)
+        self.__test_equal(sf, df)
 
     def test_auto_parse_csv(self):
         with tempfile.NamedTemporaryFile(mode='w', delete=False) as csvfile:


src/python/turicreate/data_structures/sframe.py
```diff
@@ -1603,13 +1603,18 @@ def read_json(cls,
...
+            if type(url)==str and url[0] !='{' :
+                g = SArray.read_json(url)
+                if len(g) == 0:
+                    return SFrame()
+                if g.dtype != dict:
+                    raise RuntimeError("Invalid input JSON format. Expected list of dictionaries")
+                g = SFrame({'X1':g})
+                return g.unpack('X1','')
+            elif type(url)==str and url[0] =='{' :
+                url=pandas.read_json(url)
```
Jarvi-Izana: actually we don't want to introduce pandas if it's not necessary due to the reason that pandas import time is lengthy (~260ms) and duplicated with SFrame in terms of functionality. Could you use json lib to read the json as a dictionary (you can lazily import here)? Then you can pass the dictionary to SFrame constructor.

Jarvi-Izana:
it would be better to rename this pr. :-)

src/python/turicreate/data_structures/sframe.py
TobyRoseman: @Jarvi-Izana - Even without this change `pandas` has already been loaded. So I don't think we should worry about that load time in this pull request.
The reason `pandas` is being used here is to prevent having to write the JSON string to disk when creating a `SFrame`. I don't think anything in the standard `json` lib is going to be able to help with that.

Jarvi-Izana: nope, the lazy module deferred this, if someday we'd like put that back. In general, we shouldn't eagerly load pandas since we encourage people to use sframe.

TobyRoseman: I'm confused. Nothing in this pull request changes when `pandas` is loaded/imported. It's only adding an additional usage of `pandas` when a user calls `SFrame.read_json`.

Jarvi-Izana: That's right. The lazy loading in the future should handle it.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -1603,13 +1603,18 @@ def read_json(cls,
...
+                url=SFrame(url)
+                return url
         elif orient == "lines":
```
TobyRoseman: Can we support JSON strings with lines orient? If we can, I think that should be part of this pull request. If we can't support this we should mention that in the docstring.

TobyRoseman: I think it should be possible. It looks like `pandas.read_json` has a similar `orient` parameter.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -1603,13 +1603,18 @@ def read_json(cls,
         [3 rows x 1 columns]
         """
```
TobyRoseman: I don't think `url` is the right name for the first parameter any more. Maybe we should call it `url_or_string`.

src/python/turicreate/data_structures/sframe.py
```diff
@@ -1603,13 +1603,18 @@ def read_json(cls,
         [3 rows x 1 columns]
```
TobyRoseman: The docstring needs to be updated to talk about this new functionality.

src/python/turicreate/test/test_sframe.py
```diff
@@ -201,6 +201,13 @@ def test_auto_parse_csv_with_bom(self):
             sf = SFrame.read_csv(csvfile.name, header=True)
             self.assertEqual(sf.dtype, [float, int, str])
             self.__test_equal(sf, df)
+    def test_read_json(self):
+        x='{"row 1":{"col 1":"a","col 2":"b"},"row 2":{"col 1":"c","col 2":"d"}}'
```
TobyRoseman: So this is Panda's default JSON format which is different than what our SFrame expects. If you were to write this string to a file and try to read it with `tc.SFrame.read_json(...)`, you would get an error.

The expected JSON format should not be different based on weather a string literal is used or a filename is given.

TobyRoseman:
There is a significant amount of work needed to merge this pull request. The default input JSON format should be the same regardless of whether a JSON file or a JSON string is used. Passing a JSON string should also work for `orient="lines"`.

I'm going to move the issue (#2564) for this pull request to the next milestone.

@dhivyaaxim - please let us know if you would like to continue to work on this pull. I'm happy to answer any questions.

Issue #2763, opened 2019-12-02
6.0 toolkits doesn't have model.name() function


jakesabathia2:
```
model = tc.object_detector.create(data)
print(model.name())
```
results in
```
Traceback (most recent call last):
  File "OD Test Cases.py", line 66, in <module>
    print(model.name())
  File "/Users/henry/Desktop/turicreate/release/src/python/turicreate/toolkits/_model.py", line 397, in __getattribute__
    return object.__getattribute__(self, attr)
AttributeError: 'ObjectDetector' object has no attribute 'name'
```

jakesabathia2:
hmmm turns out all toolkit don't have this function yet.

srikris:
We should not fix this. It's not pythonic to add a function.

PR #2765, opened 2019-12-02
add name() for all toolkits.

jakesabathia2:
https://github.com/apple/turicreate/issues/2763

### DIFF
--- a/src/python/turicreate/test/test_activity_classifier.py
+++ b/src/python/turicreate/test/test_activity_classifier.py
@@ -445,6 +445,9 @@ class ActivityClassifierTest(unittest.TestCase):
         model = self.model
         model.summary()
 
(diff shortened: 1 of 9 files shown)


srikris:
@TobyRoseman thoughts on this? I don' think we should add this to all our classes. In fact, we should consider deprecating it from the places we add it to.

TobyRoseman:
> @TobyRoseman thoughts on this? I don' think we should add this to all our classes. In fact, we should consider deprecating it from the places we add it to.

@srikris - I agree. I don't think we should add this pull request and we should deprecate our current `name` functions.  I think people should get that information from `type(my_model)` rather than `my_model.name()`.

jakesabathia2:
Cool!

Issue #2774, opened 2019-12-03
Version requirement of coremltools should not depend on beta version when non-beta is available. 


hoytak:
coremltools 3.1 was just released; we should bump the requirements to 3.1 from 3.0b3.

hoytak:
@srikris -- Why is this removed from the 6.0 milestone?  It doesn't make sense to force us and users to depend on a beta version of coremltools with a new release is out that isn't in beta.

PR #2775, opened 2019-12-03
Upgraded coremltools to version 3.1.

hoytak:
Fixes https://github.com/apple/turicreate/issues/2774.

### DIFF
--- a/src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
+++ b/src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
@@ -794,7 +794,6 @@ class SoundClassifier(_CustomModel):
         """
         import coremltools
         from coremltools.proto.FeatureTypes_pb2 import ArrayFeatureType
-
         prob_name = self.target + 'Probability'
 
         def get_custom_model_spec():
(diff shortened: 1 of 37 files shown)


hoytak:
@TobyRoseman -- I see what you're saying, but it also makes very little sense to depend on a beta version when the non-beta has been released.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -806,35 +805,39 @@ def get_custom_model_spec():
...
-            builder.set_input([input_name], [(input_length,)])
-            builder.set_output([self.target], [(self.num_classes,)])
+                input_name = output_name
 
+                if cur_layer['act']:
+                    output_name = next_layer_name()
+                    builder.add_activation("activation"+str(i), 'RELU', input_name, output_name)
+                    input_name = output_name
+
+            builder.add_softmax('softmax', input_name, self.target)
```
srikris: The bug was simply a difference in behavior between 3.0b1 and 3.1 where setting class labels before output vs after output overrides the class names.

srikris:
Passes all tests on all platforms.

hoytak:
Looks good @srikris !

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
TobyRoseman: Is this something that worth a coremltools GitHub issue? Or do you think the coremltools behavior is better in 3.1? Overriding the class names sounds bad to me.

srikris: No I mis-understood the behavior. The new API is actually cleaner.

PR #2783, opened 2019-12-03
Fixed coremltools to non-beta 3.0

hoytak:
Fixes our coremltools dependencies to the non-beta version of 3.0 to minimize risk in releasing 6.0.

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -1,4 +1,4 @@
-coremltools==3.0b3
+coremltools==3.0
 scipy==1.2.1
 numpy==1.16.4
 cython==0.29.10
--- a/src/python/setup.py
+++ b/src/python/setup.py
@@ -112,7 +112,7 @@ if __name__ == '__main__':
         long_description = f.read().decode('utf-8')
 
     install_requires = [
-        "coremltools==3.0b3",
+        "coremltools==3.0",
         "decorator >= 4.0.9",
         "numpy==1.16.4",
         "pandas >= 0.23.2",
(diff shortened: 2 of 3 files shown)


hoytak:
This one is the low-risk change for 6.0, the other PR can go in after we release 6.0 to upgrade the version.

hoytak:
Marking this as 6.0 and p1 to make sure we don't let it slip.  @nickjong -- would you sign off on this?  This just pins it to the non-beta 3.0 version. Beta eggs are often not kept up after the actual egg is released.  This may not be the case with the coremltools package, but depending on them when the full version is released is bad practice in general.

TobyRoseman:
> This one is the low-risk change for 6.0, the other PR can go in after we release 6.0 to upgrade the version.

But then we'll be running our unit tests using a different set of dependency than our users get.

hoytak:
> > This one is the low-risk change for 6.0, the other PR can go in after we release 6.0 to upgrade the version.
> 
> But then we'll be running our unit tests using a different set of dependency than our users get.

I'm not sure what you mean here ?

TobyRoseman:
> > > This one is the low-risk change for 6.0, the other PR can go in after we release 6.0 to upgrade the version.
> > 
> > 
> > But then we'll be running our unit tests using a different set of dependency than our users get.
> 
> I'm not sure what you mean here ?

@hoytak - We run our unit tests via the `test_wheel.sh` script. This uses the `install_python_toolchain.sh` [to create its' virtual environment](https://github.com/apple/turicreate/blob/ab8bdf42a62baeb8b66c753fbdfdfd944234869d/scripts/test_wheel.sh#L80). `install_python_toolchain.sh` use [the requirement file that you're editing to install its' dependencies](https://github.com/apple/turicreate/blob/ad9c748aa4b255f3c57f6a08da967d9785073664/scripts/install_python_toolchain.sh#L39).

So if we make this change without also making #2775, we will be running unit tests using a different version of `coremltools` than our user will get when they install `turicreate`.

I think both this change and #2275 should be the same pull request. What is the purpose of separating these change into two separate pull requests?

srikris:
Closing this PR for the one that checks on coremltools-3.1

Issue #2786, opened 2019-12-03
OD predicting unreasonable bounding boxes


jakesabathia2:
```
model = tc.load('5_8.model')
predict = model.predict(img, 0.001)

predict
[{'confidence': 0.0032273794058710337, 'type': 'rectangle', 'coordinates': {'y': 7.100619544269842e+19, 'x': 4.5197563914702523e+24, 'width': 6.55339966711328e+23, 'height': 1.153618045223043e+19}, 'label': 'carsgraz'}, {'confidence': 0.003223825478926301, 'type': 'rectangle', 'coordinates': {'y': 1.0207597725556107e-19, 'x': 1.190615349848963e-11, 'width': 5.658245681890162e-13, 'height': 1.1959743226777501e-20}, 'label': 'carsgraz'}, {'confidence': 0.0031641607638448477, 'type': 'rectangle', 'coordinates': {'y': 2.6505702010983458e+28, 'x': 3.08780606686637e+21, 'width': 1.9845394992203406e+20, 'height': 4.477404642043274e+27}, 'label': 'carsgraz'}, {'confidence': 0.0031594019383192062, 'type': 'rectangle', 'coordinates': {'y': 1.041885749239123e+25, 'x': 9.955128683747101e+21, 'width': 8.024232341072039e+21, 'height': 3.0644791943030545e+24}, 'label': 'carsgraz'}]
```

Issue #2773, opened 2019-12-03
od 6.0 behaves inconsistent among different platforms


Jarvi-Izana:
# repro

1. get a wheel of 6.0rc1 for python3.6
2. export weights on darwin and save it
3. scp the saved model to linux box
4. load the model and predict.

evaluation on darwin:
```
Using 'image' as feature column
Using 'annotations' as annotations column
Using GPU to create model (AMD Radeon Pro 560X)
Setting 'batch_size' to 32
+--------------+--------------+--------------+
| Iteration    | Loss         | Elapsed Time |
+--------------+--------------+--------------+
| 1            | 5.25522      | 1.29s        |
| 2            | 5.40305      | 2.16s        |
| 3            | 5.45075      | 3.10s        |
| 4            | 5.49147      | 3.96s        |
| 5            | 5.57757      | 4.91s        |
| 6            | 5.6226       | 5.77s        |
| 11           | 5.41964      | 10.34s       |
| 16           | 5.17099      | 14.81s       |
| 20           | 5.12047      | 18.99s       |
+--------------+--------------+--------------+
dump evaluation dictionary:
 {'average_precision': {'bike': 0.012373289093375206, 'carsgraz': 0.013162309303879738}, 'average_precision_50': {'bike': 0.07835809886455536, 'carsgraz': 0.06423525512218475}, 'mean_average_precision': 0.012767799198627472, 'mean_average_precision_50': 0.07129667699337006}
```
on ubuntu 16.04
```
{'average_precision': {'bike': 0.007338845171034336, 'carsgraz': 0.013124960474669933}, 'average_precision_50': {'bike': 0.04268103092908859, 'carsgraz': 0.06689130514860153}, 'mean_average_precision': 0.010231902822852135, 'mean_average_precision_50': 0.05478616803884506}
```

scripts to load model and save model is mentioned in #2769

Jarvi-Izana:
as always, 20 iterations. Since models on Darwin and Linux are using the same weights, we should expect a very similar result even though the training hasn't converged due to a small or insufficient iteration number.

PR #2787, opened 2019-12-03
OD difference in predictions between MPS and MXNet

jakesabathia2:
fix https://github.com/apple/turicreate/issues/2786


srikris:
Also fixes #2773

srikris:
Passes gitlab.

Issue #2795, opened 2019-12-04
output regression for OD export_coreml meta data


jakesabathia2:
When run 
```
model.export_coreml(filename,include_non_maximum_suppression=True)
```
6.0:
```
metadata {
  shortDescription: "Object detector (darknet-yolo) with non-maximum suppression created by Turi Create (version 5.8)"
  userDefined {
    key: "annotations"
    value: "annotations"
  }
  userDefined {
    key: "classes"
    value: "bike,carsgraz"
  }
  userDefined {
    key: "confidence_threshold"
    value: "0.25"
  }
  userDefined {
    key: "feature"
    value: "image"
  }
  userDefined {
    key: "include_non_maximum_suppression"
    value: "True"
  }
  userDefined {
    key: "iou_threshold"
    value: "0.45"
  }
  userDefined {
    key: "max_iterations"
    value: "500"
  }
  userDefined {
    key: "model"
    value: "darknet-yolo"
  }
  userDefined {
    key: "non_maximum_suppression_threshold"
    value: "0.45"
  }
  userDefined {
    key: "training_iterations"
    value: "500"
  }
  userDefined {
    key: "turicreate_version"
    value: "5.8"
  }
  userDefined {
    key: "type"
    value: "ObjectDetector"
  }
  userDefined {
    key: "version"
    value: "1"
  }
}

```
5.8:
```
metadata {
  shortDescription: "Object detector created by Turi Create (version 6.0rc1)"
  userDefined {
    key: "annotations"
    value: "annotations"
  }
  userDefined {
    key: "classes"
    value: "bike,carsgraz"
  }
  userDefined {
    key: "com.github.apple.os.platform"
    value: "Darwin-19.0.0-x86_64-i386-64bit"
  }
  userDefined {
    key: "com.github.apple.turicreate.version"
    value: "6.0rc1"
  }
  userDefined {
    key: "confidence_threshold"
    value: ""
  }
  userDefined {
    key: "feature"
    value: "image"
  }
  userDefined {
    key: "include_non_maximum_suppression"
    value: "True"
  }
  userDefined {
    key: "iou_threshold"
    value: ""
  }
  userDefined {
    key: "max_iterations"
    value: "500"
  }
  userDefined {
    key: "model"
    value: "darknet-yolo"
  }
  userDefined {
    key: "training_iterations"
    value: "500"
  }
  userDefined {
    key: "type"
    value: "object_detector"
  }
}
```
We got three issues need to be fixed:
[...]

PR #2796, opened 2019-12-04
Od coreml meta

jakesabathia2:
fix https://github.com/apple/turicreate/issues/2795

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/object_detector.py
+++ b/src/python/turicreate/toolkits/object_detector/object_detector.py
@@ -1663,9 +1663,9 @@ class ObjectDetector(_Model):
         options = {}
         options['include_non_maximum_suppression'] = include_non_maximum_suppression
         options['version'] = self._get_version()
-        if confidence_threshold != None:
+        if confidence_threshold is not None:
             options['confidence_threshold'] = confidence_threshold
-        if iou_threshold != None:
+        if iou_threshold is not None:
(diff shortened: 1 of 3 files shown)


src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1662,8 +1662,11 @@ def export_coreml(self, filename,
         """
         options = {}
         options['include_non_maximum_suppression'] = include_non_maximum_suppression
-        options['confidence_threshold'] = confidence_threshold
-        options['iou_threshold'] = iou_threshold
+        options['version'] = self._get_version()
+        if confidence_threshold != None:
```
shantanuchhabra: Let's follow PEP 8 and replace `!=` with `is not`.

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -1662,8 +1662,11 @@ def export_coreml(self, filename,
...
+            options['confidence_threshold'] = confidence_threshold
+        if iou_threshold != None:
```
shantanuchhabra: Same as above.

src/python/turicreate/toolkits/object_detector/object_detector.py
jakesabathia2: good! ;)

Issue #2785, opened 2019-12-03
needs more informational error msg when image is undefined


Jarvi-Izana:
We can suggest user call `dropna` before training. Or we can check it for them before the training starts.

```
during handling of the above exception, another exception occurred:
turicreate.toolkits._main.ToolkitError: Cannot decode image. Unknown format.
```

traceback:
```
(lldb) up
frame #3: 0x0000000119b2a705 libunity_shared.dylib`turi::image_util::resize_image(turi::flexible_type const&, unsigned long, unsigned long, unsigned long, bool, int) + 197
libunity_shared.dylib`turi::image_util::resize_image:
->  0x119b2a705 <+197>: movq   -0x80(%rbp), %rax
    0x119b2a709 <+201>: movq   -0x78(%rbp), %rcx
    0x119b2a70d <+205>: cmpq   %r13, %rcx
    0x119b2a710 <+208>: jne    0x119b2a721               ; <+225>
(lldb) up
frame #4: 0x0000000119b23ac3 libunity_shared.dylib`void turi::image_util::copy_image_to_memory_impl<float>(turi::image_type const&, float*, std::__1::vector<unsigned long, std::__1::allocator<unsigned long> > const&, std::__1::vector<unsigned long, std::__1::allocator<unsigned long> > const&, bool) + 627
libunity_shared.dylib`turi::image_util::copy_image_to_memory_impl<float>:
->  0x119b23ac3 <+627>: leaq   -0x60(%rbp), %rdi
    0x119b23ac7 <+631>: callq  0x118f78700               ; turi::flexible_type::~flexible_type()
    0x119b23acc <+636>: movq   -0x50(%rbp), %rbx
    0x119b23ad0 <+640>: leaq   0x8(%rbx), %rdi
(lldb) up
frame #5: 0x0000000119b9b4ae libunity_shared.dylib`turi::drawing_classifier::simple_data_iterator::next_batch(unsigned long) + 1998
libunity_shared.dylib`turi::drawing_classifier::simple_data_iterator::next_batch:
->  0x119b9b4ae <+1998>: movq   -0x80(%rbp), %rdi
    0x119b9b4b2 <+2002>: testq  %rdi, %rdi
    0x119b9b4b5 <+2005>: movq   -0x200(%rbp), %r15
    0x119b9b4bc <+2012>: je     0x119b9b4c7               ; <+2023>
(lldb)
```

Issue #2784, opened 2019-12-03
not complain for undefined value until the last stage of training


Jarvi-Izana:
set OMP_NUM_
call trace back:
```
frame #22: 0x00000001131c2a03 libunity_shared.dylib`turi::evaluation::compute_classifier_metrics(turi::gl_sframe, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, turi::gl_sarray, std::__1::map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, turi::flexible_type, std::__1::less<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, turi::flexible_type> > >) + 5347
libunity_shared.dylib`turi::evaluation::compute_classifier_metrics:
```
error reported:
```
~/.pyenv/versions/3.6.8/lib/python3.6/site-packages/turicreate/extensions.py in __run_class_function(self, fnname, args, kwargs)
    275         except RuntimeError as exc:
    276             # Expose C++ exceptions using ToolkitError.
--> 277             raise _ToolkitError(exc)
    278         ret = _wrap_function_return(ret)
    279         return ret

ToolkitError: Invalid type conversion
```
which is quite annoying if you train for a long time and almost hit the last mile of success, an exception is thrown at the evaluation stage and all efforts vanish.

PR #2804, opened 2019-12-05
check na data at init_training

Jarvi-Izana:
#2785 #2784

early termination of a training job if data contains na. 

Also, this pr reorganized `init_training` a little bit.

### DIFF
--- a/src/python/turicreate/test/test_drawing_classifier.py
+++ b/src/python/turicreate/test/test_drawing_classifier.py
@@ -106,7 +106,7 @@ class DrawingClassifierTest(unittest.TestCase):
         sf = self.check_cross_sf
         sf = sf.remove_column(self.target)
         sf = sf.add_column(_tc.SArray(
-            [None for _ in range(len(sf))], dtype=str), self.target)
+            [None] * len(sf), dtype=str), self.target)
         try:
             _tc.drawing_classifier.create(sf, self.target)
         except _ToolkitError as e:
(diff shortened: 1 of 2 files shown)


src/python/turicreate/test/test_drawing_classifier.py
```diff
@@ -96,8 +97,20 @@ def setUpClass(self, warm_start=None):
...
+        try:
+            _tc.drawing_classifier.create(sf, self.target)
+        except _ToolkitError as e:
+            self.assertTrue("dropna" in str(e))
+
+    def test_create_with_missing_value_in_label(self):
+        sf = self.check_cross_sf
+        sf = sf.remove_column(self.target)
+        sf = sf.add_column(_tc.SArray(
+            [None for _ in range(len(sf))], dtype=str), self.target)
```
TobyRoseman: Rather than `[None for _ in range(len(sf))]` you can just do `[None] * len(sf)`.

Jarvi-Izana: Wow, will try.

src/python/turicreate/test/test_drawing_classifier.py
```diff
@@ -96,8 +97,20 @@ def setUpClass(self, warm_start=None):
...
+        try:
             _tc.drawing_classifier.create(sf, self.target)
+        except _ToolkitError as e:
+            self.assertTrue("dropna" in str(e))
```
jakesabathia2: good!

Jarvi-Izana:
passed the internal pipeline. merge it.

PR #2807, opened 2019-12-05
Performance improvements for TF OD

nickjong:
- Parallelize data augmentation across CPU cores
  - Refactor `float_array_image_augmenter` to allow for map-reduce-style parallelization
- Dispatch training to a task queue, to allow augmentation to occur in parallel with NN operations

To minimize risk, the async training is only implemented for OD training, for now

### DIFF
--- a/src/ml/neural_net/image_augmentation.hpp
+++ b/src/ml/neural_net/image_augmentation.hpp
@@ -285,29 +285,26 @@ class float_array_image_augmenter : public image_augmenter {
   result prepare_images(std::vector<labeled_image> source_batch) override;
 
  protected:
-  /** The output sent from TensorFlow after augmenting the images. */
-  struct float_array_result {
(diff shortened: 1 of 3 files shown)


src/ml/neural_net/image_augmentation.hpp
```diff
@@ -288,7 +288,7 @@ class float_array_image_augmenter : public image_augmenter {
   /** The output sent from TensorFlow after augmenting the images. */
   struct float_array_result {
     /** The images after augmenting sent from Tensorflow */
-    shared_float_array images;
+    std::vector<shared_float_array> images;
```
shreyajain17: This might have made the struct (labelled_float_image) that is going into tensorflow and this the same. Can we just remove one of them?

nickjong: Good call!

PR #2812, opened 2019-12-06
Fixes backcompat issue with model.__repr__() in ST.

### LINKED ISSUE #2811: Style Transfer: model.__repr__() on model loaded from 5.8 throws error
Repro steps:

1. Load a model trained in Turi Create 5.8 into 6.0rc1.

2. Run `print(model)` in Python. 

Actual output:
```
KeyError: 'Field "_training_time_as_string" not in model. Available fields are batch_size, max_iterations, model, num_styles.'
```

Expected output:
No error.

shantanuchhabra:
Fixes #2811

### DIFF
--- a/src/python/turicreate/test/test_style_transfer.py
+++ b/src/python/turicreate/test/test_style_transfer.py
@@ -292,6 +292,17 @@ class StyleTransferTest(unittest.TestCase):
             self.test_get_styles_success()
             print("Get styles passed")
 
+    def test_state(self):
+        model = self.model
+        fields = model.__proxy__.list_fields()
+        assert "model" in fields
+        assert "num_styles" in fields
+        assert "_training_time_as_string" in fields
+        assert "training_epochs" in fields
+        assert "training_iterations" in fields
+        assert "num_content_images" in fields
+        assert "training_loss" in fields
+
 
 @unittest.skipIf(tc.util._num_available_cuda_gpus() == 0, 'Requires CUDA GPU')
 @pytest.mark.gpu
--- a/src/python/turicreate/test/test_style_transfer.py
+++ b/src/python/turicreate/test/test_style_transfer.py
@@ -295,13 +295,13 @@ class StyleTransferTest(unittest.TestCase):
     def test_state(self):
         model = self.model
         fields = model.__proxy__.list_fields()
-        assert "model" in fields
-        assert "num_styles" in fields
-        assert "_training_time_as_string" in fields
-        assert "training_epochs" in fields
-        assert "training_iterations" in fields
-        assert "num_content_images" in fields
-        assert "training_loss" in fields
+        self.assertTrue("model" in fields)
+        self.assertTrue("num_styles" in fields)
+        self.assertTrue("_training_time_as_string" in fields)
+        self.assertTrue("training_epochs" in fields)
+        self.assertTrue("training_iterations" in fields)
+        self.assertTrue("num_content_images" in fields)
+        self.assertTrue("training_loss" in fields)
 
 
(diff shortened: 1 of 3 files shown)


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -484,8 +484,9 @@ gl_sarray style_transfer::convert_style_indices_to_filter(
 
 gl_sframe style_transfer::predict(variant_type data,
                                   std::map<std::string, flexible_type> opts) {
-  gl_sframe_writer result({"style_idx", "stylized_image"},
```
srikris: Is this needed? Bad merge?

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -854,11 +855,21 @@ void style_transfer::import_from_custom_model(variant_map_type model_data,
...
+  const flex_int training_epochs = read_opts<flex_int>(model_data, "training_epochs");
+  const flex_int training_iterations = read_opts<flex_int>(model_data, "training_iterations");
+  const flex_int num_content_images = read_opts<flex_int>(model_data, "num_content_images");
+  const flex_float training_loss = read_opts<flex_float>(model_data, "training_loss");  
 
   add_or_update_state({{"model", model_type},
                        {"num_styles", num_styles},
                        {"max_iterations", max_iterations},
-                       {"batch_size", DEFAULT_BATCH_SIZE}});
+                       {"batch_size", DEFAULT_BATCH_SIZE},
```
srikris: Won't this effect a unit test that check what things are in the state? If not, we don't have one so lets' add one

src/toolkits/style_transfer/style_transfer.cpp
shantanuchhabra: We are changing the output format of the SFrame, so this should be needed?

src/toolkits/style_transfer/style_transfer.cpp
shantanuchhabra: Does affect Python unit tests, looking into those right now.

src/toolkits/style_transfer/style_transfer.cpp
shantanuchhabra: My bad, sorry.

src/python/turicreate/test/test_style_transfer.py
```diff
@@ -292,6 +292,17 @@ def test_save_and_load(self):
             self.test_get_styles_success()
             print("Get styles passed")
 
+    def test_state(self):
+        model = self.model
+        fields = model.__proxy__.list_fields()
+        assert "model" in fields
```
TobyRoseman: In unit tests `self.assertTrue` is usually preferable to just `assert`. I believe using `assert` will cause failures to show up as errors which isn't ideal.

src/python/turicreate/test/test_style_transfer.py
```diff
@@ -292,6 +292,17 @@ def test_save_and_load(self):
...
+        assert "num_styles" in fields
+        assert "_training_time_as_string" in fields
+        assert "training_epochs" in fields
+        assert "training_iterations" in fields
+        assert "num_content_images" in fields
+        assert "training_loss" in fields
```
TobyRoseman: Are we expecting other keys might in `fields`? If not, it would be good to check there aren't any other thing in there.

src/python/turicreate/test/test_style_transfer.py
shantanuchhabra: Done, thanks!

src/python/turicreate/test/test_style_transfer.py
shantanuchhabra: We are expecting there to be other keys.

PR #2816, opened 2019-12-06
Fixed typo in image augmentation.

hoytak:
On machines with a high thread count, there would be an out-of-bounds segfault on writing augmented images.


nickjong:
Thank you for looking into this in my absence, I really appreciate it!

nickjong:
FWIW in hindsight the comment on `labeled_float_image` could have been generalized to say just that the concatenation of the image tensors should yield the same batch size as implied by the annotations vector

PR #2819, opened 2019-12-06
Remove tensorflow.compat.v1.reset_default_graph()

### LINKED ISSUE #2731: Never call tensorflow.compat.v1.reset_default_graph()
Calling this function basically clears TensorFlow's state, which will destroy the image feature extractors and any Sound Classifier instances. If the user has TensorFlow objects in their environment, it will also remove them.

Many of our toolkit assume only their objects are in the default graph. We need to fix this assumption before we remove the call to reset the default graph. 

I think there are two basic approaches for this fix:
1 - Use a scope/namespace. #2712 is doing this for the Sound Classifier. 
2 - Use different TensorFlow sessions/graphs for each toolkit instances.
The second one is almost certainly preferable.

fareeha:
Closes #2731 

Investigating the safety of removing reset default graph.


fareeha:
We are using `tensorflow.compat.v1.reset_default_graph()` in: 
* Activity Classifier 
* Drawing Classifier
* Object Detector
* Style Transfer

After removing `reset_default_graph` from all four toolkits, only Activity Classifier and Drawing Classifier tests break. Adding the graph context manager fixes the breaks.

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
```diff
@@ -21,13 +21,13 @@ def __init__(self, input_h, input_w, batch_size, output_size, out_h, out_w, init
         self.gpu_policy = _utils.TensorFlowGPUPolicy()
         self.gpu_policy.start()
 
-        #reset tensorflow graph when a new model is created
-        _tf.reset_default_graph()
 
         # Converting incoming weights from shared_float_array to numpy
         for key in init_weights.keys():
             init_weights[key] = _utils.convert_shared_float_array_to_numpy(init_weights[key])
+        self.init_weights = init_weights
```
nickjong: Could we instead just pass `init_weights` as an argument to `self.init_object_detector_graph`? Doesn't look like we need it afterwards.

src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
```diff
@@ -488,7 +493,7 @@ def __init__(self, config, net_params):
 
         self.__define_graph();
         
-        self.sess = _tf.Session()
+        #self.sess = _tf.Session()
```
nickjong: Let's just remove this line

src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py
fareeha: Thats a good idea! Done

Issue #2810, opened 2019-12-06
Inconsistent output schema for model.stylize between 5.8 and current master.


shantanuchhabra:
As of current master, the output SFrame of `model.stylize` on an SArray with multiple content images does not have a `row_id` column to reference the row in the input SFrame.

Repro:

1. Load a model trained using Turi Create 5.8 in Turi Create 6.0rc1. Model shared internally.
2. Call `model.stylize` on an SArray with 2 content images.

Actual output:
```
Columns:
	style_idx	int
	stylized_image	Image

Rows: 16

Data:
+-----------+------------------------+
| style_idx |     stylized_image     |
+-----------+------------------------+
|     0     | Height: 600 Width: 447 |
|     0     | Height: 600 Width: 447 |
|     1     | Height: 600 Width: 447 |
|     1     | Height: 600 Width: 447 |
|     2     | Height: 600 Width: 447 |
|     2     | Height: 600 Width: 447 |
|     3     | Height: 600 Width: 447 |
|     3     | Height: 600 Width: 447 |
|     4     | Height: 600 Width: 447 |
|     4     | Height: 600 Width: 447 |
+-----------+------------------------+
```

Expected: Another column in the SFrame called `row_id` to match 5.8 which refers to the row number in the input SFrame.

example output from TC 5.8:
```
Columns:
	row_id	int
	style	int
	stylized_image	Image

Rows: 8

Data:
+--------+-------+------------------------+
| row_id | style |     stylized_image     |
+--------+-------+------------------------+
|   0    |   0   | Height: 600 Width: 447 |
|   0    |   1   | Height: 600 Width: 447 |
|   0    |   2   | Height: 600 Width: 447 |
|   0    |   3   | Height: 600 Width: 447 |
|   0    |   4   | Height: 600 Width: 447 |
|   0    |   5   | Height: 600 Width: 447 |
|   0    |   6   | Height: 600 Width: 447 |
|   0    |   7   | Height: 600 Width: 447 |
+--------+-------+------------------------+
[8 rows x 3 columns]


```

shantanuchhabra:
After an initial investigation, seems like fixing this will involve more of a code change than I initially anticipated. We may have to add a `is_prediction` option in the data iterator that would then return the row index from the input SFrame as well. Currently, the data iterator doesn’t store or return the row number in the original dataset. @abhishekpratapa thoughts?

PR #2820, opened 2019-12-06
Style transfer

jakesabathia2:
fix https://github.com/apple/turicreate/issues/2810


Jarvi-Izana:
better to have a unit test as well.

src/python/turicreate/test/test_style_transfer.py
```diff
@@ -168,6 +168,8 @@ def test_stylize_success(self):
 
         for style in styles:
             stylized_out = model.stylize(sf, style=style)
+            self.assertEqual(set(stylized_out.column_names()),
```
Jarvi-Izana: Bravo!

PR #2822, opened 2019-12-08
Fixed breaking unit test in master.

srikris:
- Breaking unit test in master due to my final commit in #2810


srikris:
Passed unit tests on internal build.

PR #2823, opened 2019-12-08
Fixed C++ unit test for export-coreml

srikris:
The PR contains 2 commits
- Fixed the C++ test for export_coreml to include calling with the user defined metadata
- Added an additional test to preserve the original behavior and added default parameters


srikris:
Passes C++ tests on all platforms on the internal builds.

PR #2824, opened 2019-12-09
C API bug fixes

TobyRoseman:
This can be merged post 6.0 release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -440,9 +440,9 @@ elseif (APPLE)
   set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,-rpath,@loader_path -Wl,-rpath,@loader_path/..")
 else()
   # LINUX
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath,$ORIGIN")
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-rpath,$ORIGIN -Wl,-rpath,$ORIGIN/..")
-  set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,-rpath,$ORIGIN -Wl,-rpath,$ORIGIN/..")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath,\\$ORIGIN")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-rpath,\\$ORIGIN -Wl,-rpath,\\$ORIGIN/..")
+  set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,-rpath,\\$ORIGIN -Wl,-rpath,\\$ORIGIN/..")
 endif()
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -440,9 +440,9 @@ elseif (APPLE)
   set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,-rpath,@loader_path -Wl,-rpath,@loader_path/..")
 else()
   # LINUX
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath,$ORIGIN")
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-rpath,$ORIGIN -Wl,-rpath,$ORIGIN/..")
-  set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,-rpath,$ORIGIN -Wl,-rpath,$ORIGIN/..")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath,\\$ORIGIN")
```
Jarvi-Izana: Just curious, is ORIGIN a shell variable or a cmake variable? If it's a cmake variable, you may need the variable interpolation to happen here.

TobyRoseman: It is a cmake variable. I'm not sure I understand what you mean by variable interpolation here. Are you suggesting it should be `\\${ORIGIN}` rather than just `\\$ORIGIN`?

TobyRoseman: I was wrong. The literal sting `$ORIGIN`needs to be passed to the linker. It should not be expanded. It should not be `${ORIGIN}`.

Jarvi-Izana: Sorry, I should make myself clear. variable interpolation means expansion.

Jarvi-Izana: https://stackoverflow.com/questions/29584348/escaping-dollar-sign-in-cmake.

Jarvi-Izana: shell will receive "\\$" as well, where reference variable after `$` will be expanded as well, with a pending `\`.

a single back slach is enough for shell.
```
echo "\$HOST"
```

it should also work for CMake.

src/capi/framework/tc_capi_exports_linux.ver
```diff
@@ -1,6 +1,7 @@
 {
   global:
     _tc_*;
+    tc_*;
```
hoytak: This should not be needed.  All C symbols have a _ prepended to them.

src/capi/impl/capi_operators.cpp
```diff
@@ -129,7 +129,7 @@ static inline flexible_type apply_binary_op_ff(
     case binary_op::OP_GT:
       return (t > u);
     case binary_op::OP_GE:
-      return (t <= u);
+      return (t >= u);
```
hoytak: Nice catch.  I'm surprised the unit tests didn't catch this one...

CMakeLists.txt
```diff
@@ -440,9 +440,9 @@ elseif (APPLE)
...
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-rpath,\\$ORIGIN -Wl,-rpath,\\$ORIGIN/..")
+  set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,-rpath,\\$ORIGIN -Wl,-rpath,\\$ORIGIN/..")
```
hoytak: Out of curiosity, what made you discover this issue?

Issue #2828, opened 2019-12-09
od evaluate segfaulted when missing value is involved


Jarvi-Izana:
#2827 
```
+------------------------+-------------------------------+
|         image          |          annotations          |
+------------------------+-------------------------------+
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
+------------------------+-------------------------------+
[156 rows x 2 columns]
Note: Only the head of the SFrame is printed.
You can use print_rows(num_rows=m, num_columns=n) to print more rows and columns.
[1]    49565 segmentation fault (core dumped)  python3 test_od_bad_input.py
```

## repro
```
import turicreate as tc

data = tc.SFrame('/Users/guihaoliang/Work/tc-testing/'
                 'tc_test_dataset/Object_Detection/ig02.sframe')

data.remove_column("name", inplace=True)
data.remove_column("label", inplace=True)
print(data)

train_data, test_data = data.random_split(0.8, seed=1)
model = tc.object_detector.create(train_data, max_iterations=1)


# bad annotations
def func(x):
    for y in x:
        y["label"] = None
    return x


# alt_train_data = train_data["annotations"].apply(lambda x: func(x))
# bad_train_data = train_data.remove_column("annotations")
# bad_train_data = bad_train_data.add_column(alt_train_data, "annotations")

# print(bad_train_data)
# print(isinstance(bad_train_data, tc.SFrame))
# model = tc.object_detector.create(bad_train_data, max_iterations=1)

alt_test_data = test_data["annotations"].apply(lambda x: func(x))
bad_test_data = test_data.remove_column("annotations")
bad_test_data = bad_test_data.add_column(alt_test_data, "annotations")

print(bad_test_data)
model = model.evaluate(bad_test_data)
```

jakesabathia2:
good.

[...]

Issue #2827, opened 2019-12-09
OD: segfault when annotation has "None" label during training phase


Jarvi-Izana:
```
+------------------------+-------------------------------+
|         image          |          annotations          |
+------------------------+-------------------------------+
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
+------------------------+-------------------------------+
[629 rows x 2 columns]
Note: Only the head of the SFrame is printed.
You can use print_rows(num_rows=m, num_columns=n) to print more rows and columns.
True
Using 'image' as feature column
Using 'annotations' as annotations column
Using GPU to create model (AMD Radeon Pro 560X)
Setting 'batch_size' to 32
+--------------+--------------+--------------+
[1]    47325 segmentation fault (core dumped)  python3 test_od_bad_input.py
```
whereas in 5.8,
```
+------------------------+-------------------------------+
|         image          |          annotations          |
+------------------------+-------------------------------+
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
| Height: 480 Width: 640 | [{'label': None, 'coordina... |
+------------------------+-------------------------------+
[629 rows x 2 columns]
Note: Only the head of the SFrame is printed.
[...]

PR #2829, opened 2019-12-09
OD seg-faults when annotation labels contains "None" in them.

jakesabathia2:
fix  https://github.com/apple/turicreate/issues/2827
fix https://github.com/apple/turicreate/issues/2828

### DIFF
--- a/src/python/turicreate/test/test_object_detector.py
+++ b/src/python/turicreate/test/test_object_detector.py
@@ -185,6 +185,8 @@ class ObjectDetectorTest(unittest.TestCase):
             for y in x:
                 y['label'] = None
             return x
+
+
         try:
             with self.assertRaises(_ToolkitError):
                 sf = self.sf.head()
@@ -335,6 +337,8 @@ class ObjectDetectorTest(unittest.TestCase):
             for y in x:
                 y['label'] = None
             return x
+
+
         try:
             with self.assertRaises(_ToolkitError):
                 sf = self.sf.head()
--- a/src/python/turicreate/test/test_object_detector.py
+++ b/src/python/turicreate/test/test_object_detector.py
@@ -179,6 +179,22 @@ class ObjectDetectorTest(unittest.TestCase):
 
             tc.object_detector.create(sf)
 
+    def test_create_with_missing_annotations_label(self):
+
+        def create_missing_annotations_label(x):
+            for y in x:
+                y['label'] = None
+            return x
+        try:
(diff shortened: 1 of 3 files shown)


src/python/turicreate/test/test_object_detector.py
```diff
@@ -313,6 +329,21 @@ def test_evaluate_missing_annotations(self):
             del sf[self.annotations]
             self.model.evaluate(sf.head())
 
+    def test_evaluate_with_missing_annotations_label(self):
+
+        def create_missing_annotations_label(x):
+            for y in x:
+                y['label'] = None
+            return x
```
Jarvi-Izana: well, 2 newlines are recommended as the interval between function the other codes.

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -103,6 +103,10 @@ std::vector<image_annotation> parse_annotations(
 
       if (key == "label") {
         // If the label is invalid (not in class_to_index_map) then ignore it.
+        if (kv.second.get_type() == flex_type_enum::UNDEFINED) {
+          log_and_throw("Annotation labels must have valid type.");
```
Jarvi-Izana: add more informational text like: please consider 'dropna()' to clean column <blah>. Also, Check #2804 to early error out user the missing value will halt the training.

src/python/turicreate/test/test_object_detector.py
jakesabathia2: good

src/toolkits/object_detection/od_data_iterator.cpp

jakesabathia2: will not do the same thing in #2804 in `init_training` tho,
bc OD allows empty annotation (they ignore them and this is covered in a test case),
and the missing value for image is handled in the Python side.

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -103,6 +103,10 @@ std::vector<image_annotation> parse_annotations(
 
       if (key == "label") {
         // If the label is invalid (not in class_to_index_map) then ignore it.
+        if (kv.second.get_type() == flex_type_enum::UNDEFINED) {
+          log_and_throw("Annotation label is missing.");
```
srikris: "All annotations must contain label information."

srikris:
Passes internal gitlab.

Issue #2830, opened 2019-12-09
OD predict function shouldn't depend on the annotation column


jakesabathia2:
Now the `predict` function for OD is still depending on the `annotation` column.

repro:
```
import turicreate as tc

data = tc.SFrame('ig02.sframe')

data.remove_column("name", inplace=True)
data.remove_column("label", inplace=True)
print(data)

train_data, test_data = data.random_split(0.8, seed=1)
model = tc.object_detector.create(train_data, max_iterations=1)

alt_train_data = train_data["annotations"].apply(lambda x: "invalid")
bad_train_data = train_data.remove_column("annotations")
bad_train_data = bad_train_data.add_column(alt_train_data, "annotations")

print(bad_train_data)
print(isinstance(bad_train_data, tc.SFrame))
pred = model.predict(bad_train_data)
```

5.8:
```
+------------------------+-------------------------------+
|         image          |          annotations          |
+------------------------+-------------------------------+
| Height: 480 Width: 640 | [{'coordinates': {'y': 296... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 242... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 266... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 162... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 240... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 248... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 221... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 144... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 222... |
| Height: 480 Width: 640 | [{'coordinates': {'y': 286... |
+------------------------+-------------------------------+
[785 rows x 2 columns]
Note: Only the head of the SFrame is printed.
You can use print_rows(num_rows=m, num_columns=n) to print more rows and columns.
Using 'image' as feature column
Using 'annotations' as annotations column
Setting 'batch_size' to 32
Using GPU to create model (AMD Radeon Pro 560X)
+--------------+--------------+--------------+
| Iteration    | Loss         | Elapsed Time |
+--------------+--------------+--------------+
| 1            | 6.235        | 9.6          |
+--------------+--------------+--------------+
+------------------------+-------------+
|         image          | annotations |
+------------------------+-------------+
| Height: 480 Width: 640 |   invalid   |
| Height: 480 Width: 640 |   invalid   |
[...]

PR #2832, opened 2019-12-10
OD predict method should ignore the annotation column if it's present in the input data.

jakesabathia2:
close https://github.com/apple/turicreate/issues/2830.

### DIFF
--- a/src/toolkits/object_detection/object_detector.cpp
+++ b/src/toolkits/object_detection/object_detector.cpp
@@ -659,6 +659,12 @@ variant_type object_detector::predict(
   // Convert data to SFrame
   std::string image_column_name = read_state<flex_string>("feature");
   gl_sframe sframe_data = convert_types_to_sframe(data, image_column_name);
+
+  // Predict function should only depends on the feature column
+  // So we extract the image column only.
+  if (!sframe_data.contains_column(image_column_name)) {
+    log_and_throw("Column name '" + image_column_name + "' does not exist.");
+  }
   gl_sframe sframe_image_data(
       {{image_column_name, sframe_data[image_column_name]}});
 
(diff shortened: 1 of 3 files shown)


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -659,8 +659,11 @@ variant_type object_detector::predict(
   // Convert data to SFrame
   std::string image_column_name = read_state<flex_string>("feature");
   gl_sframe sframe_data = convert_types_to_sframe(data, image_column_name);
+  gl_sframe sframe_image_data(
```
Jarvi-Izana: please comment on why you need to extract the sframe. Otherwise, without the context of #2830, other ppl may feel strange.

jakesabathia2: good!

jakesabathia2:
fail on gitlab.
looking into it ;(.

test/unity/toolkits/object_detection/test_object_detector.cxx
```diff
@@ -761,7 +761,8 @@ BOOST_AUTO_TEST_CASE(test_object_detector_predict) {
   // Create an arbitrary SFrame with test_num_examples rows, since
   // object_detector uses the number of rows to compute num_examples, which is
   // used as a normalizer.
-  gl_sframe data({{"ignored", gl_sarray::from_sequence(0, test_num_examples)}});
+  gl_sframe data(
+      {{test_image_name, gl_sarray::from_sequence(0, test_num_examples)}});
```
Jarvi-Izana: could you explain?

jakesabathia2: in the predict function we are extracting the image column (test_image_name).
But the mocked data itself doesn't have such column which causes the error ;).

jakesabathia2:
re-run on gitlab.

jakesabathia2:
pass the internal c++ unit test.
will merge it.

Issue #2833, opened 2019-12-10
OD: invalid coordinate value gets no error and make it through


Jarvi-Izana:
#2830 

same repro, except change the `func` into:
```
 39 def func(x):
 40     for y in x:
 41         # y["coordinates"]['x'] = None # y["coordinates"]['y'] = None
 42         y["coordinates"]['width'] = None
 43         y["coordinates"]['height'] = None
 44     return x
```

feel free to play with it.

jakesabathia2:
hmmmm cool ;)

jakesabathia2:
```
+--------------+--------------+--------------+
| Iteration    | Loss         | Elapsed Time |
+--------------+--------------+--------------+
| 1            | 4.39937      | 2.92s        |
| 2            | 4.5508       | 3.53s        |
| 3            | 4.50975      | 4.63s        |
| 4            | 4.43646      | 5.56s        |
| 5            | 4.30856      | 6.29s        |
| 6            | 4.39513      | 7.02s        |
| 11           | 4.18712      | 10.68s       |
| 16           | 4.43339      | 14.34s       |
| 21           | 4.43974      | 18.50s       |
| 26           | 3.96112      | 22.16s       |
| 31           | 3.98385      | 25.81s       |
| 36           | 3.64385      | 29.48s       |
| 41           | 3.83445      | 34.50s       |
| 46           | 4.22252      | 38.24s       |
| 51           | 3.56531      | 43.04s       |
| 56           | 4.0538       | 46.69s       |
| 61           | 3.85165      | 51.21s       |
| 66           | 3.63041      | 54.87s       |
| 71           | 3.45167      | 58.53s       |
| 76           | 3.37094      | 1m 2s        |
| 81           | 3.50457      | 1m 6s        |
| 86           | 3.10482      | 1m 10s       |
| 91           | 3.43324      | 1m 13s       |
| 96           | 3.18027      | 1m 17s       |
| 100          | 3.40443      | 1m 20s       |
+--------------+--------------+--------------+
```
haha interesting ;)

Jarvi-Izana:
I bet the missing value is skipped. Check the log while you run this code snippet. If there's a log associated with it, it should be better we can surface it to users.

what you need to do is to get the log file location of current running python process
```
tc.config.get_server_log_location()
```

don't close the process. At the same time,
```
tail -f <log>
```
and run the repro code. In this way, you can check whether the missing value is logged or not.

jakesabathia2:
[...]

PR #2836, opened 2019-12-10
[6.1] OD bounding box co-ordinates cannot contain None values

jakesabathia2:
https://github.com/apple/turicreate/issues/2833

### DIFF
--- a/src/toolkits/object_detection/od_data_iterator.cpp
+++ b/src/toolkits/object_detection/od_data_iterator.cpp
@@ -126,7 +126,7 @@ std::vector<image_annotation> parse_annotations(
           if (coord_kv.second.get_type() != flex_type_enum::FLOAT and
               coord_kv.second.get_type() != flex_type_enum::INTEGER) {
             log_and_throw(
-                "Bounding box co-ordinates must be int or float values.");
+                "Bounding box coordinates must be int or float values.");
           }
 
           float coord_val = coord_kv.second;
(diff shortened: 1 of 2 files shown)


src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -121,6 +121,16 @@ std::vector<image_annotation> parse_annotations(
...
 
           const flex_string& coord_key = coord_kv.first.get<flex_string>();
+
+          // Check the type of coordinate
+          if (coord_kv.second.get_type() != flex_type_enum::FLOAT and
+              coord_kv.second.get_type() != flex_type_enum::INTEGER) {
+            log_and_throw(
+                "Invalid object annotations discovered. Annotation coordinate "
+                "'" +
+                coord_key + "' must have type of Int or Float.");
```
srikris: "Bounding box co-ordinates must be int or float values" is more accurate.

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -125,6 +125,14 @@ std::vector<image_annotation> parse_annotations(
         for (const auto& coord_kv : coordinates) {
 
           const flex_string& coord_key = coord_kv.first.get<flex_string>();
+
+          // Check the type of coordinate
```
Jarvi-Izana: good job. One more thing, please update the docstring if the type constraint of coordinates is not mentioned.

Jarvi-Izana: also, don't forget to check the width and height. ;-)

jakesabathia2: It does include width and height ;)

Jarvi-Izana: Ah, it's a map. Good!

PR #2840, opened 2019-12-11
WIP : Changing TF data augmenter to match MPS data augmenter

shreyajain17:


src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,38 +22,267 @@
...
+        
+        # TODO: Replace this condition soon
+    
+        mat_intersection_points = tf.expand_dims(intersection_points, 0)
+        transformed_intersection_points = apply_transformation(mat_intersection_points, transformation)
+        
+        overlap_condition = math_ops.greater_equal(area_fraction, 0.5)
+        new_annotations_arr.append(control_flow_ops.cond(overlap_condition,
+                                                        lambda: tf.squeeze(transformed_intersection_points),
+                                                        lambda: np.array([0.0, 0.0, 0.0, 0.0], dtype=np.float32)))
```
gustavla: For these kinds of conditionals, I usually just cast the bool to a float and multiply with it. I don't know what's faster in TF though. I tend to borrow notions from straight up compiled languages where control flow can screw up pipelining. I don't know if this notion holds in the TF graph.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,38 +22,267 @@
...
+
+def _crop_augmenter_worker(image, annotation, num_attempts_var, max_attempts, perform_cropped_annotations, image_height, image_width, min_height_var, max_height_var, aspect_ratio_var, min_aspect_ratio, max_aspect_ratio, min_area_fraction, max_area_fraction, image_box, keep_annotations):
+    image_box, keep_annotations, annotation, num_attempts_var, max_attempts, perform_cropped_annotations, aspect_ratio_var, min_aspect_ratio, max_aspect_ratio, min_height_var, max_height_var, image_height, image_width, min_area_fraction, max_area_fraction = tf.while_loop(lambda image_box, keep_annotations, annotation, num_attempts_var, max_attempts, perform_cropped_annotations, aspect_ratio_var, min_aspect_ratio, max_aspect_ratio, min_height_var, max_height_var, image_height, image_width, min_area_fraction, max_area_fraction:_crop_augmenter_loop_conditional(num_attempts_var, max_attempts, perform_cropped_annotations),
```
gustavla: Does it need the identity lambda here, why not just pass in `_crop_augmenter_loop_conditional`?

gustavla: This is also almost impossible to read. Maybe some newlines?

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,38 +22,267 @@
...
+    bounding_box2_unstacked = tf.unstack(bounding_box2)
+    is_intersection_cond = math_ops.logical_and(math_ops.logical_and(math_ops.less(bounding_box1_unstacked[0], bounding_box2_unstacked[2]),
+                                                                     math_ops.less(bounding_box1_unstacked[1], bounding_box2_unstacked[3])),
+                                                math_ops.logical_and(math_ops.less(bounding_box2_unstacked[0], bounding_box1_unstacked[2]),
+                                                                     math_ops.less(bounding_box2_unstacked[1], bounding_box1_unstacked[3])))
+    
+    return control_flow_ops.cond(is_intersection_cond,
+                                lambda: True,
+                                lambda: False)
+def _get_middle_points(bounding_box1, bounding_box2):
```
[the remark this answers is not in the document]
gustavla: The names are a bit confusing too, `_get_intersection_point` and `_get_middle_points` - they return bounding boxes though, so it's really getting the intersection between two bounding boxes. `_get_intersection_box`?

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,38 +22,267 @@
...
+    
+    flip_values_3 = math_ops.less(uts[3][1], uts[3][0])
+    uts[3] = control_flow_ops.cond(flip_values_3,
+                        lambda: array_ops.reverse(uts[3], [0]),
+                        lambda: uts[3])
+
+    restack_points = tf.stack(uts)[:, :1]
+    reshaped_points = tf.reshape(restack_points, [ -1, 4])
+    
+    return tf.squeeze(reshaped_points)
```
gustavla: If there is only a single bounding box, couldn't this oversqueezed? I always prefer to have an explicit axis defined when squeezing to prevent corner cases.

gustavla: I guess this function only operates on a single box?

I feel like this operation can be done vectorized for batches, though.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,38 +22,267 @@
...
+    tx = tf.to_float(tf.to_int32(x_offset))
+
+    # Make the transformation matrix
+    transformation = tf.reshape(tf.stack([
+        1.0,     0.0,     -ty,
+        0.0,     1.0,     -tx,
+        0.0,     0.0,    1.0]
+    ), (3, 3))
+
+    for ann in unstacked_ann:
```
gustavla: I'm a little confused what's happening here. So `unstacked_ann` are different boxes in the same image, right? But that axis is not defined, so how can you unstack it? What will this iterate over when building the graph?

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,38 +22,267 @@
...
+    intersection_image_box = tf.unstack(image_box)
+    
+    intersection_image_box[0] = tf.to_float(tf.to_int32(y_offset))
+    intersection_image_box[1] = tf.to_float(tf.to_int32(x_offset))
+    intersection_image_box[2] = tf.to_float(tf.to_int32(cropped_height + y_offset))
+    intersection_image_box[3] = tf.to_float(tf.to_int32(cropped_width + x_offset))
+
+    intersection_bounding_box = tf.stack(intersection_image_box)
+    
+    identifier, box, confidence = decompose_annotations(annotations, tf.to_float(image_height), tf.to_float(image_width))
```
gustavla: Most of the time, annotation (singular) is passed into `decompose_annotations`, but here it's plural. How does it work and is `annotations` here really multiple boxes?

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,38 +22,267 @@
...
+    
+    unstacked_ann = tf.unstack(box)
```
gustavla: I think `box` has shape `[None, 4]` here, so how is it possible to unstack it?

PR #2846, opened 2019-12-12
Changes in TF Data Augmenter to match 5.8 performance and accuracy

shreyajain17:


src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,351 @@
...
+  'max_contrast' : 1.25,
+  'max_saturation' : 1.25,
+  'flip_pr' : True,
+  'skip_probability' : 0.1,
+  'min_aspect_ratio' : 0.8,
+  'max_aspect_ratio' : 1.25,
+  'min_area_fraction' : 1.0,
+  'max_area_fraction' : 2.0,
+  'max_attempts' : 50,
+  'output_shape' : (416, 416),
```
nickjong: Many of these appear to be unused. Can we find some consistent way to reason about / collect these constants?

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,351 @@
...
-
-        image = images[i]
-        image = _utils.convert_shared_float_array_to_numpy(image)
+    for i in range(0, len(annotation)):
+        if np.any(annotation[i][1:5]):
+            annotation[i][1] = 1 - annotation[i][1] - annotation[i][3]
+
+    return flipped_image, annotation
+
+def padding_augmenter(image,
```
nickjong: Could we better document the logic of this function? It's hard to follow

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,351 @@
...
+        y_offset = np.random.uniform(0.0, (image_height - cropped_height))
+        
+        crop_bounds_x1 = x_offset
+        crop_bounds_y1 = y_offset
+        
+        crop_bounds_x2 = x_offset + cropped_width
+        crop_bounds_y2 = y_offset + cropped_height
+        
+        formatted_annotation = []
+        condition_met = True
```
nickjong: The variable name could be more helpful

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,351 @@
...
+            width = bounds[2]
+            height = bounds[3]
+            
+            x1 = bounds[0] * image_width
+            y1 = bounds[1] * image_height
+            
+            x2 = (bounds[0] + width) * image_width
+            y2 = (bounds[1] + height) * image_height
+            
+            if crop_bounds_x1 < x2 and crop_bounds_y1 < y2 and x1 < crop_bounds_x2 and y1 < crop_bounds_y2:
```
nickjong: The logic from here also needs more explanatory comments, I think

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,351 @@
...
+    img_tf, ann_tf = color_augmenter(img_tf, ann_tf)
+    img_tf, ann_tf = hue_augmenter(img_tf, ann_tf)
+    img_tf, ann_tf = resize_augmenter(img_tf, ann_tf, (output_height, output_width))
+    return img_tf, ann_tf
+
+
+
+class DataAugmenter(object):
+    def __init__(self, output_height, output_width, resize_only):
+        self.batch_size = 32
```
nickjong: Should this be hard-coded?

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,352 @@
...
-        image = _utils.convert_shared_float_array_to_numpy(image)
+    for i in range(0, len(annotation)):
+        if np.any(annotation[i][1:5]):
+            annotation[i][1] = 1 - annotation[i][1] - annotation[i][3]
+
+    return flipped_image, annotation
+
+def padding_augmenter(image,
+                      annotation,
+                      skip_probability=0.1,
```
srikris: We should probably pick this up from the defaults dict above.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
shreyajain17: Getting from C++ now.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
shreyajain17: Yeah just saw it.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
shreyajain17: Just put in the right place. Doing it now.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,358 @@
 
 tf.disable_v2_behavior()
 
-def get_augmented_data(images, annotations, output_height, output_width, resize_only):
+_DEFAULT_AUG_PARAMS = {
+  'max_hue_adjust' : 0.05,
+  'max_brightness' : 0.05,
+  'max_contrast' : 1.25,
+  'max_saturation' : 1.25,
+  'flip_pr' : True,
```
TobyRoseman: It doesn't look like this is getting used. Also I don't understand what "pr" means.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,358 @@
...
-    horizontal_flip=True 
-
-    for i in range(len(images)):
-
-        image = images[i]
-        image = _utils.convert_shared_float_array_to_numpy(image)
+    image_height, image_width, _ = image.shape
+    flipped_image = np.flip(image, 1)
+    
+    for i in range(0, len(annotation)):
```
TobyRoseman: You don't need the `0, `.

More importantly - it's really not clear what this for-loop is doing.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -22,351 +22,358 @@
...
+                    feed_dict[self.ann_tf[i]] = _utils.convert_shared_float_array_to_numpy(annotations[i])
+            aug_output = session.run(graph_op, feed_dict=feed_dict)
+            processed_images = []
+            processed_annotations = []
+            for o in aug_output:
+                processed_images.append(o[0])
+                processed_annotations.append(np.ascontiguousarray(o[1], dtype=np.float32))
+            processed_images = np.array(processed_images, dtype=np.float32)
+            processed_images = np.ascontiguousarray(processed_images, dtype=np.float32)
+            return tuple((processed_images, processed_annotations))
```
TobyRoseman: `return (processed_images, processed_annotations)` will do the same thing. Calling `tuple` is unnecessary.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
shreyajain17: I need to remove it

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
shreyajain17: One image can have more than one annotation. Will have to do this for all annotations. Can put up comments for this.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
shreyajain17: Yeah will change it.

srikris:
Passes on all platforms and accuracy tested!

Issue #2844, opened 2019-12-11
AC: Classify still go through for data with missing feature column


jakesabathia2:
In the documentation (https://apple.github.io/turicreate/docs/api/generated/turicreate.activity_classifier.ActivityClassifier.classify.html#turicreate.activity_classifier.ActivityClassifier.classify)
it mentions that the data must include columns with the same names as the features and session id used for model training.
However, when running
```
model = tc.activity_classifier.create(data,session_id='exp_id',target='activity',
	max_iterations=3,prediction_window=prediction_window)

print('\n==classify with missing feature==')
missing_data = data.copy()
del missing_data['acc_x']
pred = model.classify(missing_data)
```
It still go through without error out hmmmm.
And I observe the same issue for 5.8, so I don't know it is meant to behave like this or not.

PR #2847, opened 2019-12-12
[6.1] Ac data with missing feature for predict/classify/predict_topk/evaluate.

jakesabathia2:
close https://github.com/apple/turicreate/issues/2844

### DIFF
--- a/src/toolkits/activity_classification/ac_data_iterator.cpp
+++ b/src/toolkits/activity_classification/ac_data_iterator.cpp
@@ -296,6 +296,12 @@ variant_map_type _activity_classifier_prepare_data_aug_impl(
   }
 #endif
 
+  for (auto &feat : features) {
+    if (!data.contains_column(feat)) {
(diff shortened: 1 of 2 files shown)


src/toolkits/activity_classification/ac_data_iterator.cpp
```diff
@@ -296,6 +296,12 @@ variant_map_type _activity_classifier_prepare_data_aug_impl(
   }
 #endif
 
+  for (auto &feat : features) {
```
TobyRoseman: It's probably best not to use `auto` here.

Jarvi-Izana: yeah, flexible_type is enough.

Jarvi-Izana: sorry, std::string.

jakesabathia2: very good!

Issue #2799, opened 2019-12-05
Style Transfer model loads lazily


TobyRoseman:
[This unit test](https://github.com/apple/turicreate/blob/4e9fe88dc634fa3d694fbfd48427c9eddbe99902/src/python/turicreate/test/test_style_transfer.py#L306) is disabled unless you have a CUDA GPU. Running this unit test with a CUDA GPU fails:

```
_________________________________________________________________ StyleTransferGPUTest.test_gpu_save_load_export _________________________________________________________________

self = <turicreate.test.test_style_transfer.StyleTransferGPUTest testMethod=test_gpu_save_load_export>

    def test_gpu_save_load_export(self):
        old_num_gpus = tc.config.get_num_gpus()
        gpu_options = set([old_num_gpus, 0, 1])
        for in_gpus in gpu_options:
            tc.config.set_num_gpus(in_gpus)
            model = tc.style_transfer.create(self.style_sf, self.content_sf, max_iterations=1)
            for out_gpus in gpu_options:
                with test_util.TempDirectory() as path:
>                   model.save(path)

test_style_transfer.py:314: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../toolkits/_model.py:491: in save
    return glconnect.get_unity().save_model(self, _make_internal_url(location))
cy_unity.pyx:96: in turicreate._cython.cy_unity.UnityGlobalProxy.save_model
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
[...]

PR #2851, opened 2019-12-13
Refactor Style Transfer Unit Test

TobyRoseman:
Fixes: #2799

### DIFF
--- a/src/python/turicreate/test/test_style_transfer.py
+++ b/src/python/turicreate/test/test_style_transfer.py
@@ -327,7 +327,7 @@ class StyleTransferGPUTest(unittest.TestCase):
                     tc.config.set_num_gpus(out_gpus)
                     model = tc.load_model(path)
 
-                with test_util.TempDirectory() as path:
-                    model.export_coreml(os.path.join(path, 'model.mlmodel'))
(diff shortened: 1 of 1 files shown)


src/python/turicreate/test/test_style_transfer.py
```diff
@@ -311,22 +311,23 @@ def test_state(self):
...
-            model = tc.style_transfer.create(self.style_sf, self.content_sf, max_iterations=1)
+            original_model = tc.style_transfer.create(self.style_sf, self.content_sf, max_iterations=1)
             for out_gpus in gpu_options:
                 with test_util.TempDirectory() as path:
-                    model.save(path)
+                    original_model.save(path)
                     tc.config.set_num_gpus(out_gpus)
                     model = tc.load_model(path)
+
+                with test_util.TempDirectory() as path:
```
hoytak: What is the reason to delete the model path out from under the model?  I don't think this is supported.

TobyRoseman: Good catch. Ironically, this is basically the same bug I was fixing.

I've updated the code to export while the model is still in scope.

Issue #2855, opened 2019-12-13
[6.1] AC: Strange behavior when k <= 0 for predict_topk. Should check for the input.


jakesabathia2:
When `k < 0`, in the C++ it produces overflow for `size_t` type,
which cause `k` to be something like `18446744073709551615`.
So the predict_topk function perform like `k=num_class`.
For 5.8 it also performs the same way.

However, when `k = 0`,
5.8 :
```
+--------+---------------------+-----------------+
| row_id |        class        |   probability   |
+--------+---------------------+-----------------+
|   0    |       standing      |  0.641275942326 |
|   0    |       sitting       |  0.230049073696 |
|   0    |       walking       | 0.0515022464097 |
|   0    | climbing_downstairs | 0.0410682857037 |
|   0    |  climbing_upstairs  | 0.0238569062203 |
|   0    |        laying       |  0.012247575447 |
|   1    |       standing      |  0.641275942326 |
|   1    |       sitting       |  0.230049073696 |
|   1    |       walking       | 0.0515022464097 |
|   1    | climbing_downstairs | 0.0410682857037 |
+--------+---------------------+-----------------+
```
6.0:
```
+--------+-------+-------------+
| row_id | class | probability |
+--------+-------+-------------+
|   0    |  None |     None    |
|   1    |  None |     None    |
|   2    |  None |     None    |
|   3    |  None |     None    |
|   4    |  None |     None    |
|   5    |  None |     None    |
|   6    |  None |     None    |
|   7    |  None |     None    |
|   8    |  None |     None    |
|   9    |  None |     None    |
+--------+-------+-------------+
```
which is strange.

PR #2856, opened 2019-12-13
[6.1] AC invalid k for predict_topk.

jakesabathia2:
https://github.com/apple/turicreate/issues/2855

### DIFF
--- a/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
+++ b/src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
@@ -606,6 +606,7 @@ class ActivityClassifier(_Model):
         |      ...      |  ...  |        ...        |
         +---------------+-------+-------------------+
         """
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -606,6 +606,8 @@ def predict_topk(self, dataset, output_type='probability', k=3, output_frequency
         |      ...      |  ...  |        ...        |
         +---------------+-------+-------------------+
         """
+        if not isinstance(k, int):
+            raise _ToolkitError('k must be of type int')
```
TobyRoseman: I think `TypeError` makes more sense than a `ToolkitError`.

jakesabathia2: Great!

Issue #2863, opened 2019-12-16
OSOD summary() error out


jakesabathia2:
```
model = tc.one_shot_object_detector.create(starter_images, 'label', background_images, max_iterations=max_iterations)

print('\n==test model summary==')
print(model.summary('dict'))
```

produce error message:
```
  File "/Users/henry/Desktop/turicreate/release/src/python/turicreate/toolkits/_model.py", line 371, in _get
    raise ValueError("There is no model field called {}".format(field))
ValueError: There is no model field called input_image_shape
```

This his also in `5.8`.

Issue #2862, opened 2019-12-16
OSOD print(model) produces wrong output.


jakesabathia2:
```
max_iterations = 1

model = tc.one_shot_object_detector.create(starter_images, 'label', background_images, max_iterations=max_iterations)
print(model)
```
produces
```
OneShotObjectDetector
```

which is supposed to be something like this:
```
Class                                    : OneShotObjectDetector

Model summary
-------------
Number of classes                        : 1
Input image shape                        : (3, 416, 416)

Synthetic data summary
----------------------
Number of synthetically generated examples : 5
Number of synthetically generated bounding boxes : 5

Training summary
----------------
Training time                            : 9s
Training iterations                      : 1
Training epochs                          : 6
Final loss (specific to model)           : 3.4134
```

Jarvi-Izana:
@nickjong not sure this should be p1 or not.

PR #2870, opened 2019-12-19
one-shot od model summary.

jakesabathia2:
fix https://github.com/apple/turicreate/issues/2863
fix https://github.com/apple/turicreate/issues/2862

### DIFF
--- a/src/python/turicreate/test/test_one_shot_object_detector.py
+++ b/src/python/turicreate/test/test_one_shot_object_detector.py
@@ -247,6 +247,18 @@ class OneObjectDetectorSmokeTest(unittest.TestCase):
         model = self.model
         model.summary()
 
+    def test_summary_str(self):
+        model = self.model
+        self.assertTrue(isinstance(model.summary("str"), str))
+
+    def test_summary_dict(self):
+        model = self.model
+        self.assertTrue(isinstance(model.summary("dict"), dict))
+
+    def test_summary_invalid(self):
+        model = self.model
+        model.summary("invalid")
+
     def test_repr(self):
         # Repr after fit
         model = self.model
--- a/src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
+++ b/src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
@@ -12,7 +12,6 @@ from turicreate.toolkits._model import PythonProxy as _PythonProxy
 from turicreate.toolkits.object_detector.object_detector import ObjectDetector as _ObjectDetector
 from turicreate.toolkits.one_shot_object_detector.util._augmentation import preview_synthetic_training_data as _preview_synthetic_training_data
 import turicreate.toolkits._internal_utils as _tkutl
-from turicreate.toolkits._internal_utils import _toolkit_serialize_summary_struct
 from turicreate.toolkits import _coreml_utils
 
 USE_CPP = _tkutl._read_env_var_cpp('TURI_OD_USE_CPP_PATH')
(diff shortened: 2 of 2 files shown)


src/python/turicreate/test/test_one_shot_object_detector.py
```diff
@@ -247,6 +247,18 @@ def test_summary(self):
...
+        model = self.model
+        self.assertTrue(isinstance(model.summary("str"), str))
+
+    def test_summary_dict(self):
+        model = self.model
+        self.assertTrue(isinstance(model.summary("dict"), dict))
+
+    def test_summary_invalid(self):
+        model = self.model
+        model.summary("invalid")
```
TobyRoseman: It seems like this should throw an exception.

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -12,6 +12,7 @@
 from turicreate.toolkits.object_detector.object_detector import ObjectDetector as _ObjectDetector
 from turicreate.toolkits.one_shot_object_detector.util._augmentation import preview_synthetic_training_data as _preview_synthetic_training_data
 import turicreate.toolkits._internal_utils as _tkutl
+from turicreate.toolkits._internal_utils import _toolkit_serialize_summary_struct
```
TobyRoseman: This is fine, don't worry about changing it. If it's only being used in one function, I would just import it at the start of that function rather than for the whole file.

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -296,6 +295,20 @@ def __repr__(self):
             class_name='OneShotObjectDetector')
         return out
 
+    def summary(self, output=None):
+
```
TobyRoseman: Needs a doctoring.

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
```diff
@@ -296,6 +295,20 @@ def __repr__(self):
...
+        if output is None or output == 'stdout':
```
TobyRoseman: This code appears to be an exact copy of the code in the base class. Am I missing some thing?

If this code is an exact copy of what's in the base class, I don't understand why we're overriding it.

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
jakesabathia2: good!

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
jakesabathia2: For OSOD, the summary is actually the OD model,
so instead of doing 
```
return _toolkit_serialize_summary_struct( self, *self._get_summary_struct() )
```
Here is 
```
return _toolkit_serialize_summary_struct( self.__proxy__['detector'],*self._get_summary_struct() )
```

src/python/turicreate/toolkits/one_shot_object_detector/one_shot_object_detector.py
jakesabathia2: good!

src/python/turicreate/test/test_one_shot_object_detector.py
jakesabathia2: For all the toolkits now for invalid inputs,
they return print(self.__repr__()).
Do I need to change the logic for all toolkits?

TobyRoseman: Let's not address that in this PR. Please file a GitHub issue for this problem with all toolkits.

jakesabathia2: good!

PR #2876, opened 2019-12-23
Utility script for fixing header include paths

hoytak:
This PR adds a script, ./scripts/fix_headers.py, that fixes include paths for all the source files in the project.  It's mainly useful when headers get moved, as it then allows the user to adjust all the include paths throughout the project with a single command. 

Examples:

To fix one header file in all source files.

./scripts/fix_headers.py <header_name>

To fix all references to a collection of header files (note regex match):

./scripts/fix_headers.py --all-match='src/<subdir/.*\.h.*'.


TobyRoseman:
@hoytak - It's very difficult to tell what has changed. Could you make this pull request two commit, where the first commit just uses `git mv`?

scripts/fix_headers.py
```diff
@@ -1,23 +1,47 @@
...
+import re
 import sys
 from os.path import join, split, normpath, exists
 import subprocess
 import argparse
+from collections import defaultdict
+
+parser = argparse.ArgumentParser(description=
+"""
+Script to fix the headers in the repository source files.  Useful when
```
TobyRoseman: Thanks for adding a description. It would be great if you could elaborate a little on what you mean by "fix". It finds the correct path and updates it?

scripts/fix_headers.py
```diff
@@ -98,7 +138,45 @@ def print_usage_and_exit():
     print("  --fix-headers <header_name>        Fixes all import paths of a unique header.")
     sys.exit(1)
 
+def all_headers_matching(match_regex):
+
+    rhl = [h for h in raw_header_list if match_regex.match(h)]
+
+
+    # Split the list into two components -- those with
```
TobyRoseman: Did you want to finish this comment?

scripts/fix_headers.py
hoytak: Yes, when you move headers or source files around, the include statements are wrong.  This script fixes those include statements.

PR #2877, opened 2019-12-24
[WIP] New build system.

hoytak:
Rearchitecture of the the build system.  A vast simplification.

Changes

- ./configure can produce an XCode project of the Turi source code.
- deps/ can be compiled and installed out of the regular build process. Helpful for the XCode project.
- All packages in src/external/ and src/visualization compile into static libraries.
- The rest of the code base compiles into a single shared library.  This means that most everything occurs in one single CMakeLists.txt file using standard cmake commands.
- All headers are installed into targets/include when `make install` is called.  The main library is installed into targets/lib.
- The code definitions that reflect compiler oddities (e.g. is std::hash<int128_t> defined) are dumped into a single header file, src/turi_common.h, at configure time.  All source files and headers now include this file first.  This allows other programs to link against this library / headers reliably.
- The python part of TuriCreate now builds through the standard setup.py method. Cython / pybind11 extensions are compiled and packaged using the standard extensions mechanisms and linking against the installed Turi headers.

WIP TODO: The python installation, deployment building.

To review:  Essentially it's all in the root CMakeLists.txt file, with some changes to ./configure.  Just focus on those two.


build_python_wheel.sh
```diff
@@ -0,0 +1,114 @@
...
+function print_help {
+  echo "Configures and builds a specified target."
+  echo
+  echo "Usage: ./build.sh [build options] [configure options]"
+  echo
+  echo "Common Options:"
+  echo "  --target-dir, -t                 The target directory to install artifact to."
+  echo "                                   default: `pwd`/targets."
+  echo
+  echo "  --release                        Build in release mode."
```
TobyRoseman: It doesn't look like debug or release is actually getting used any where.

deps/src/libxml2-2.9.1/doc/xmllint.html
```diff
@@ -101,7 +101,7 @@
 	   Do not use any catalogs.
 	  </dd><dt><span class="term"><tt class="option">--auto</tt></span></dt><dd>
 	   Generate a small document for testing purposes.
-	  </dd><dt><span class="term"><tt class="option">--xinclude</tt></span></dt><dd>
+	  </dd><dt><span class="term"><tt class="option">--xinclude </tt></span></dt><dd>
```
znation: Is this change intentional?

deps/src/libxml2-2.9.1/doc/html/libxml-xlink.html
```diff
@@ -10,7 +10,7 @@
...
... border="0" cellspacing="0" cellpadding="1" width="100%" bgcolor="#000000"><tr><td><table border="0" cellpadding="3" cellspacing="1" width="100%"><tr><td bgcolor="#fffacd"><table class="navigation" width="100%" summary="Navigation header" cellpadding="2" cellspacing="2"><tr valign="middle"><td><a accesskey="p" href="libxml-xinclude.html"><img src="left.png" width="24" height="24" border="0" alt="Prev" /></a></td><th align="left"><a href="libxml-xinclude.html">xinclude </a></th><td><a accesskey="u" href="index.html"><img src="up.png" width="24" height="24" border="0" alt="Up" /></a></td><th align="left"><a href="index.html">API documentation</a></th><td><a accesskey="h" href="../index.html"><img src="home.png" width="24" height="24" border="0" alt="Home" /></a></td><th align="center"><a href="../index.html">The XML C parser and toolkit of Gnome</a></th><th align="right"><a href="libxml-xmlIO.html">xmlIO</a></th><td><a accesskey="n" href="libxml-xmlIO.html"><img src="right.png" width="24" height="24" border="0" alt="Next" /></a></td></tr></table><p>unfinished XLink detection module </p><h2>Table of Contents</h2><pre class="programlisting">Enum <a href="#xlinkActuate">xlinkActuate</a>
```
znation: Is this change intentional?

deps/src/libxml2-2.9.1/doc/html/libxml-valid.html
```diff
@@ -10,7 +10,7 @@
...
...2" cellspacing="2"><tr valign="middle"><td><a accesskey="p" href="libxml-uri.html"><img src="left.png" width="24" height="24" border="0" alt="Prev" /></a></td><th align="left"><a href="libxml-uri.html">uri</a></th><td><a accesskey="u" href="index.html"><img src="up.png" width="24" height="24" border="0" alt="Up" /></a></td><th align="left"><a href="index.html">API documentation</a></th><td><a accesskey="h" href="../index.html"><img src="home.png" width="24" height="24" border="0" alt="Home" /></a></td><th align="center"><a href="../index.html">The XML C parser and toolkit of Gnome</a></th><th align="right"><a href="libxml-xinclude.html">xinclude </a></th><td><a accesskey="n" href="libxml-xinclude.html"><img src="right.png" width="24" height="24" border="0" alt="Next" /></a></td></tr></table><p>API for the DTD handling and the validity checking </p><h2>Table of Contents</h2><pre class="programlisting">#define <a href="#XML_CTXT_FINISH_DTD_0">XML_CTXT_FINISH_DTD_0</a></pre><pre class="programlisting">#define <a href="#XML_CTXT_FINISH_DTD_1">XML_CTXT_FINISH_DTD_1</a></pre><pre class="programlisting">Structure <a href="#xmlAttributeTable">xmlAttributeTable</a><br />struct _xmlHashTable
```
znation: Is this change intentional?

deps/src/libxml2-2.9.1/doc/xmllint.xml
```diff
@@ -105,7 +105,7 @@
 			<arg choice="plain"><option>--catalogs</option></arg>
 			<arg choice="plain"><option>--nocatalogs</option></arg>
 			<arg choice="plain"><option>--auto</option></arg>
-			<arg choice="plain"><option>--xinclude</option></arg>
+			<arg choice="plain"><option>--xinclude </option></arg>
```
znation: Is this change intentional?

znation: In general it looks like lots of "include<" became "include <" -- this doesn't seem right for non-source-code files.

deps/src/libxml2-2.9.1/doc/xmllint.html
hoytak: Probably not

cmake/CompilerOddities.cmake
```diff
@@ -60,11 +62,11 @@ macro(Set_Compiler_Specific_Flags)
 
     message(FATAL_ERROR "Cannot determine noexcept fladg on std::ios_base::failure.  See log.")
   elseif(COMPILER_HAS_NOEXCEPT_WHAT_ON_EXCEPTIONS_V1) 
-    add_definitions(-DCOMPILER_MODIFIER_ON_EXCEPTION_WHAT=noexcept)
+    add_compiler_source_define(COMPILER_MODIFIER_ON_EXCEPTION_WHAT=noexcept)
   elseif(COMPILER_HAS_NOEXCEPT_WHAT_ON_EXCEPTIONS_V2)
-    add_definitions(-DCOMPILER_MODIFIER_ON_EXCEPTION_WHAT=_NOEXCEPT)
+    add_compiler_source_define(DCOMPILER_MODIFIER_ON_EXCEPTION_WHAT=_NOEXCEPT)
```
guihao-liang: should remove `D`, will modify this in my branch.

src/external/google/protobuf/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
-project(TuriExternalDependencies)
+project(Turi)
```
guihao-liang: what''s the reason to change here?

configure
```diff
@@ -78,6 +78,13 @@ function print_help {
   echo
   echo "  -D var=value                      Specify FLAGS definitions to be passed on to cmake."
   echo
+  echo "  --create-xcode-project            Create an xcode project file as xcode/Turi.xcodeproj."
+  echo
+  echo "  --build-dependencies              Build the external dependencies as part of configure and install them into deps/local."
```
guihao-liang: what's the relationship between xcode-project with the build-deps?

build_python_wheel.sh
guihao-liang: by default is release. Just add `--debug` should be fine.

build_python_wheel.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash -e
```
guihao-liang: is this file intended to be used interactively? like, `configure`?

cmake/CompilerOddities.cmake
hoytak: Yes, that makes sense.

src/external/google/protobuf/CMakeLists.txt
hoytak: This is the name of the generated Xcode project that these files are added to.  Thus all project specifications should be Turi if we generate an xcode project, which we can do here.

guihao-liang: I see.

configure
hoytak: You have to build the dependencies (those not in src/external) separately, as the xcode project does not integrate with their build systems.  So building them means the xcode project then depends on them as lib / include dependencies, but doesn't rebuild them.

guihao-liang:
forked in #3065.

build_python_wheel.sh
hoytak: Eh, this one seem deprecated.  scripts/make_wheel.sh is better I think?  Not sure.

guihao-liang: this one uses `make_wheel.sh`. I don't know if this is just for the manual interactive build.

configure
guihao-liang: I see. If I understand correctly, xcode proejct under the hood uses cmakes? And those cannot be integrated with xcode project don't use cmake as their build pipeline?

Issue #2873, opened 2019-12-20
SC's strange error message after a model is created.


jakesabathia2:
I have a Python script following the user guide for SC:
https://apple.github.io/turicreate/docs/userguide/sound_classifier/

The script runs through but in the end it throws this error
```
Exception TypeError: "'NoneType' object is not callable" in <bound method VGGishFeatureExtractor.__del__ of <turicreate.toolkits.sound_classifier._audio_feature_extractor.VGGishFeatureExtractor object at 0x11b1168d0>> ignored
```
This error message appears every time as long as a model is created.
I.e.:
```
import turicreate as tc
from os.path import basename
import sys
train_set = tc.SFrame('train_set.sframe')
test_set = tc.SFrame('test_set.sframe')

# Create the model.
model = tc.sound_classifier.create(train_set, target='category', feature='audio')
```
Also produced the same error message in the end.

TobyRoseman:
This likely has the same root cause as #2842

PR #2891, opened 2020-01-03
del error for sound classifier

jakesabathia2:
#2873 

hmmmm so the error comes from that `_mac_ver()` method is destroyed before the object is destroyed.
I think this PR is just a workaround tho.
@TobyRoseman @nickjong any thoughts?


TobyRoseman:
@jakesabathia2 - I agree this fix is just workaround. Are you confident this fix prevent the error message from getting displayed?

jakesabathia2:
> @jakesabathia2 - I agree this fix is just workaround. Are you confident this fix prevent the error message from getting displayed?

In my local it fixed the error message.

Issue #2168, opened 2019-07-27
Object Detector Needs Seed Parameter


TobyRoseman:
The `create` method for the object detector should take a `seed` parameter so that a deterministic model can be produced.

PR #2892, opened 2020-01-03
od random seed.

jakesabathia2:
#2168


src/python/turicreate/test/test_object_detector.py
```diff
@@ -230,6 +230,12 @@ def test_create_with_empty_dataset(self):
         with self.assertRaises(_ToolkitError):
             tc.object_detector.create(self.sf[:0])
 
+    def test_create_with_fixed_random_seed(self):
+        model_1 = tc.object_detector.create(self.sf, max_iterations=3, random_seed=86)
+        pred_1 = model_1.evaluate(self.sf)
+        model_2 = tc.object_detector.create(self.sf, max_iterations=3, random_seed=86)
+        pred_2 = model_2.evaluate(self.sf)
+        self.assertEqual(pred_1, pred_2)
```
TobyRoseman: Would comparing the outputs of `predict` rather than `evaluate` be a better test?

jakesabathia2: Good!
I can do it.

shreyajain17:
Only this would not seed tensorflow. It needs a seed at the top of TF Graph as well as the data augmentations using TF.

jakesabathia2:
> Only this would not seed tensorflow. It needs a seed at the top of TF Graph as well as the data augmentations using TF.

Good!

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -99,7 +99,7 @@ def _raise_error_if_not_detection_sframe(dataset, feature, annotations, require_
 
 def create(dataset, annotations=None, feature=None, model='darknet-yolo',
            classes=None, batch_size=0, max_iterations=0,
-           verbose=True, grid_shape=[13, 13], **kwargs):
+           verbose=True, grid_shape=[13, 13], random_seed=0, **kwargs):
```
nickjong: This is not how other public API accepting random seeds work. With this default, won't the toolkit produce the same behavior every time unless the user manually randomizes the seed?

jakesabathia2: No they won't produce the same result.
For OD if random_seed is not provided, it is `FLEX_UNDEFINED` in C++,
and then a random seed is produced.
Don't know if this answer your question...

jakesabathia2:
So I find out that it seems to be impossible to seed the Numpy.random function inside a tensorflow graph.
So I construct a tensorflow random array first instead, which we can feed,
and pass the array into each augmenter.

src/python/turicreate/toolkits/object_detector/object_detector.py
guihao-liang: Well, in DC, if the value is `FLEX_UNDEFINED`, the random seed will be 0, implicitly. I guess you can mention it in docstring about this implicit behavior and add similar logic in C++ implementation, in which way there's no need to change the C-API.

guihao-liang: https://github.com/apple/turicreate/blob/4e2d9159b27e321689c3aa6db43a0400eb98ad68/src/toolkits/drawing_classifier/drawing_classifier.cpp#L334

jakesabathia2: Good!

jakesabathia2:
> great job! After you resolve @nickjong's concern, you should be good to go!

Done.

jakesabathia2:
> https://www.tensorflow.org/api_docs/python/tf/compat/v1/set_random_seed

We are trying to produce the same result every time when we give a fixed seed.
Yes so I did try `tf.set_random_seed(seed)` for the graph level before:
```
tf.set_random_seed(seed=seed)
with self.graph.as_default():
            self.img_tf = [tf.placeholder(tf.float32, [None, None, 3]) for x in range(0, self.batch_size )]
            self.ann_tf = [tf.placeholder(tf.float32, [None, 6]) for x in range(0, self.batch_size )]
```
but it didn't work out haha.
So I feed seed to each opts in order to achieve the same thing.

guihao-liang:
> > https://www.tensorflow.org/api_docs/python/tf/compat/v1/set_random_seed
> 
> We are trying to produce the same result every time when we give a fixed seed.
> Yes so I did try `tf.set_random_seed(seed)` for the graph level before:
> 
> ```
> tf.set_random_seed(seed=seed)
> with self.graph.as_default():
>             self.img_tf = [tf.placeholder(tf.float32, [None, None, 3]) for x in range(0, self.batch_size )]
>             self.ann_tf = [tf.placeholder(tf.float32, [None, 6]) for x in range(0, self.batch_size )]
> ```
> 
> but it didn't work out haha.
> So I feed seed to each opts in order to achieve the same thing.

that's the behavior we need. ;-)

nickjong:
But we don't want the same result on every image, right? If you give the same annotated image to the augmentation pipeline multiple times during the same training session, they must have different results, right?

guihao-liang:
> But we don't want the same result on every image, right? If you give the same annotated image to the augmentation pipeline multiple times during the same training session, they must have different results, right?

Maybe I have a misunderstanding. If every time is different, how do you make sure it can be reproduced?

jakesabathia2:
> > But we don't want the same result on every image, right? If you give the same annotated image to the augmentation pipeline multiple times during the same training session, they must have different results, right?
> 
> Maybe I have a misunderstanding. If every time is different, how do you make sure it can be reproduced?

Nick is saying that he is wondering whether by fixing the seed at the opts level will produce the same result every time it is called.

jakesabathia2:
> But we don't want the same result on every image, right? If you give the same annotated image to the augmentation pipeline multiple times during the same training session, they must have different results, right?

they will produce different results every time even though the seed is fixed at operation level.
```
import tensorflow.compat.v1 as tf
tf.disable_v2_behavior()

random_num = tf.random.uniform([],0,1,seed=0) #seed is fixed 
session = tf.Session()
print(session.run(random_num)) 
print(session.run(random_num))
```
this get two different numbers.

guihao-liang:
> > > But we don't want the same result on every image, right? If you give the same annotated image to the augmentation pipeline multiple times during the same training session, they must have different results, right?
> > 
> > 
> > Maybe I have a misunderstanding. If every time is different, how do you make sure it can be reproduced?
> 
> Nick is saying that he is wondering whether by fixing the seed at the opts level will produce the same result every time it is called.

I see. Thanks for the clarification!

jakesabathia2:
Have made session of the image augmenter to a class member,
so that it produces different results for different mini-batch haha.

nickjong:
Can you please describe clearly the behavior you observe, across examples during one training run and across training runs? When the seed is held fixed and deliberately changed? We need to understand how this both achieves the originally intended random sampling while also achieving reproducibility

jakesabathia2:
Random sampling:
We want to acquire the following two properties for random sampling:
1. Each sample in a mini-batch has different behaviors:
Our image augmenter creates an array with length of batch_size, and uses a session to run it.
```
#construct the graph
self.img_tf = [tf.placeholder(tf.float32, [None, None, 3]) for x in range(0, self.batch_size )]
self.ann_tf = [tf.placeholder(tf.float32, [None, 6]) for x in range(0, self.batch_size )]
self.resize_op_batch = []
for i in range(0, self.batch_size):
       aug_img_tf, aug_ann_tf = complete_augmenter(self.img_tf[i], self.ann_tf[i], (output_height, output_width))
       self.resize_op_batch.append([aug_img_tf, aug_ann_tf])

#get augmenter output
aug_output = self.session.run(self.resize_op_batch)
```
We expect each element in aug_output has different output, and this can be achieved by assigning different seeds for each sample.
For simplicity we can look at the following behavior when each operation is a simple `tf.random.uniform` operation
```
# fixed seed for each sample
opt_batch = [tf.random.uniform([],0,1,seed=0), tf.random.uniform([],0,1,seed=0), tf.random.uniform([],0,1,seed=0)]
with tf.Session() as session:
        print(session.run(opt_batch))
```
The code produces something like
```
[0.232, 0.232, 0.232]
```
This is not what we want since each element in a batch produces the same result.
So I first use the user-given seed to generate an random seed array with length of the batch size in the beginning, and then assign the seed to each element.
```
np.random.seed(random_seed)
self.random_seed = np.random.randint(low=0, high=2**32-1, size=self.batch_size)
for i in range(0, self.batch_size):
        aug_img_tf, aug_ann_tf = complete_augmenter(self.img_tf[i], self.ann_tf[i], output_height, output_width, self.random_seed[i])
        self.resize_op_batch.append([aug_img_tf, aug_ann_tf])
```
This ensure that each element has different behavior.

2. Each mini-batch produce different results.
Again for each iteration, we want to ensure we get different results. This can be achieved by create only one session and run the augmenter every single time using the session.

What the master doing now is that it basically creates a new session for each iteration,
which will produce the same result across all iterations, which is not good.
```
#construct the graph
self.opt_batch = []
batch_size = 3
for i in range(batch_size):
    self.opt_batch.append(tf.random.uniform([],0,1,seed=i))

#for each iteration a new session is created
#say we run 3 iterations
with tf.Session() as session:
   print(session.run(self.opt_batch)
with tf.Session() as session:
   print(session.run(self.opt_batch)
with tf.Session() as session:
   print(session.run(self.opt_batch)
```
This gets results like
```
[0.231, 0.456, 0.771]
[0.231, 0.456, 0.771]
[0.231, 0.456, 0.771]
```
which is not what we want.

So I change the code to have only one session, for each iteration we use the same session to run the graph
```
self.session = tf.Session()
print(self.session.run(self.opt_batch))
print(self.session.run(self.opt_batch))
print(self.session.run(self.opt_batch))
```
This produces
```
[0.231, 0.456, 0.771]
[0.561, 0.032, 0.112]
[0.345, 0.987, 0.321]
```
which is nice haha.

Reproducibility
In the image augmenter, we basically have two kinds of random operation:
(1)The first one is the four tensorflow operations:
`tf.image.random_hue`, `tf.image.random_brightness`, `tf.image.random_saturation`, and `tf.image.random_contrast`,
like what is reported in https://stackoverflow.com/questions/38469632/tensorflow-non-repeatable-results/40247201, `tf.set_random_set` at the graph level sadly is useless for some operations. And these four operations belong to this as I observed.
So I set the seed for each operation when constructing the graph.
```
tf.image.random_hue(image, max_delta=max_hue_adjust, seed=tf_seed)
...
tf.image.random_contrast(image, lower=np.exp(-log_con), upper=np.exp(log_con), seed=tf_seed)
```

(2) The second one is we use `np.random.uniform` throughout the code in `tf.numpy_function`.
Again, `tf.set_random_seed()` doesn't seem to work, and also `np.random.seed()` doesn't work either (I guess it is because the Numpy function is wrapped in tensorflow haha).
So I replaced every `numpy.random.uniform` in the code by a `tf.random.uniform` tensor.
```
#say if crop_augmenter has 5 np.random.uniform operations included
#I generate a random tensor random_nums with length 5 and pass it as an input
random_nums = tf.random.uniform([5], 0., 1., seed=seed_tf)
img_tf, ann_tf = tf.numpy_function(func=crop_augmenter, inp=[img_tf, ann_tf, random_nums], Tout=[tf.float32, tf.float32])
``` 

So by doing these change we could achieve both random sampling and reproducibility haha.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -41,39 +33,36 @@
   'min_eject_coverage': 0.5
 }
 
-def hue_augmenter(image, annotation,
+def hue_augmenter(image, annotation, tf_seed,
                   max_hue_adjust=_DEFAULT_AUG_PARAMS["max_hue_adjust"]):
 
-    # Sample a random rotation around the color wheel.
```
jakesabathia2: I don't know what `hue_adjust` is used there so I delete it haha.

guihao-liang: you'd better ask the team.

guihao-liang:
I approve of the correctness of the code. But I still have some high-level questions for @jakesabathia2.

Let's say I have 9 different images and augment them in a batch of 3,

one uses seed fixed for each batch,

```
[0.231, 0.456, 0.771]
[0.231, 0.456, 0.771]
[0.231, 0.456, 0.771]
```

The other use different seeds for different batch,

```
[0.231, 0.456, 0.771]
[0.561, 0.032, 0.112]
[0.345, 0.987, 0.321]
```

What's the benefit of the latter approach compared with the former approach?

jakesabathia2:
> I approve of the correctness of the code. But I still have some high-level questions for @jakesabathia2.
> 
> Let's say I have 9 different images and augment them in a batch of 3,
> 
> one uses seed fixed for each batch,
> 
> ```
> [0.231, 0.456, 0.771]
> [0.231, 0.456, 0.771]
> [0.231, 0.456, 0.771]
> ```
> 
> The other use different seeds for different batch,
> 
> ```
> [0.231, 0.456, 0.771]
> [0.561, 0.032, 0.112]
> [0.345, 0.987, 0.321]
> ```
> 
> What's the benefit of the latter approach compared with the former approach?

Good!
Because I guess we want each batch has different transformation in order to achieve random behavior?

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -41,39 +33,36 @@
...
     # Sample a random adjustment to brightness.
     if max_brightness is not None and max_brightness > 0:
-        image = tf.image.random_brightness(image, max_delta=max_brightness)
+        image = tf.image.random_brightness(image, max_delta=max_brightness, seed=tf_seed)
 
     # Sample a random adjustment to contrast.
     if max_saturation is not None and max_saturation > 1.0:
         log_sat = np.log(max_saturation)
-        image = tf.image.random_saturation(image, lower=np.exp(-log_sat), upper=np.exp(log_sat))
+        image = tf.image.random_saturation(image, lower=np.exp(-log_sat), upper=np.exp(log_sat), seed=tf_seed)
```
nickjong: Here you are using same seed as for the `random_brightness`. In the TF implementation (https://github.com/tensorflow/tensorflow/blob/cf7fcf164c9846502b21cebb7d3d5ccf6cb626e8/tensorflow/python/ops/image_ops_impl.py#L1582), I see that this is used as
```
delta = random_ops.random_uniform([], -max_delta, max_delta, seed=seed)
```
Will the delta used for brightness be correlated with the offset used for e.g. saturation? The end of the `set_seed` TF docs suggest that in some cases, ops may or may not share the same internal state.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -337,90 +329,114 @@ def crop_augmenter(image,
...
+
+    max_attempts = _DEFAULT_AUG_PARAMS["max_attempts"]
+    random_nums = tf.random.uniform([1+4*max_attempts], 0., 1., seed=seed_tf)
+    img_tf, ann_tf = tf.numpy_function(func=crop_augmenter, inp=[img_tf, ann_tf, random_nums], Tout=[tf.float32, tf.float32])
+    return img_tf, ann_tf
+
+def padding_augmenter_wrapper(img_tf, ann_tf, seed_tf):
+
+    max_attempts = _DEFAULT_AUG_PARAMS["max_attempts"]
+    random_nums = tf.random.uniform([1+max_attempts+3], 0., 1., seed=seed_tf)
```
nickjong: Is there correlation between the `random_nums` values computed in this function versus in e.g. `crop_augmenter_wrapper`? Note that the first value of `random_nums` is often used to evaluate the skip probability. So if these are correlated, then whether or not each augmentation is performed will be improperly correlated.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -1117,13 +1119,11 @@ void object_detector::init_training(gl_sframe data,
...
   // Read options from user.
   init_options(opts);
-
-  // Choose a random seed if not set.
   if (read_state<flexible_type>("random_seed") == FLEX_UNDEFINED) {
-    std::random_device random_device;
-    int random_seed = static_cast<int>(random_device());
-    add_or_update_state({{"random_seed", random_seed}});
-  }
+     std::random_device random_device;
```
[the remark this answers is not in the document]
jakesabathia2: Will fix it

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
jakesabathia2: Oh cool. I'll take a look.

jakesabathia2: Haha actually will.
As long as two random tensors are declared with the same type and the same seed,
if they are ran by the same session,
they produce the same result haha.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
jakesabathia2: Good!

jakesabathia2:
Add random_seed table in order to seed augmenter differently,
solve master conflict,
wrap uniform distribution function.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -36,6 +36,11 @@
   'resize_method': 'turicreate',
 }
 
+def uniform_num(random_num, min_value, max_value):
+    #return a uniform distribution number between min_value and max_value
```
TobyRoseman: major nit: PEP8 says you should have a space between the `#` and the start of the comment.

jakesabathia2: Good!

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -32,42 +32,42 @@
   'skip_probability_crop' : 0.1,
   'min_object_covered': 0.0,
   'min_eject_coverage': 0.5,
-  'resize_method': 'turicreate',
```
nickjong: Why are we switching this back? Is this just a merge issue? (We need to resolve the merge conflicts with master)

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -407,52 +409,87 @@ def crop_augmenter(image,
...
+
+        #Total num of augmenters with randomness
+        num_augmenter = 5
+
+        #Max num of functions with randomness in one augmenter
+        num_function = 3
+
+        #Generate a random table
+        random_table_shape = (self.batch_size, num_augmenter, num_function)
+        np.random.seed(random_seed if random_seed >=0 else -random_seed)
```
nickjong: Setting the global `numpy` random seed can effect user code that uses `numpy`. Please use `RandomState` instead to avoid clobbering user-level state

https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.random.RandomState.html#numpy.random.RandomState

nickjong:
Superseded by #3093

PR #2893, opened 2020-01-04
Delete unused C++ code

TobyRoseman:


guihao-liang:
It might be fine to leave these header-only utils there in case of future usage.

TobyRoseman:
> It might be fine to leave these header-only utils there in case of future usage.

What would the benefit of that be? These files are under source control, so we can recover them whenever we what. Right now having them around is just extra code to worry about and it's confusing.

guihao-liang:
> > It might be fine to leave these header-only utils there in case of future usage.
> 
> What would the benefit of that be? These files are under source control, so we can recover them whenever we what. Right now having them around is just extra code to worry about and it's confusing.

those headers are not even compiled. Besides that, since those files are rarely modified, the git blob store should be very efficient to track it and store it.

nickjong:
Unless we think we are likely to use this stuff, let’s not keep it around just because git is efficient at keeping it. The cost is more conceptual complexity and clarity, not build performance

PR #2896, opened 2020-01-06
Exposing fixing a seed to AC

shreyajain17:


src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
```diff
@@ -21,13 +21,19 @@
...
         self.gpu_policy = _utils.TensorFlowGPUPolicy()
         self.gpu_policy.start()
 
         for key in net_params.keys():
             net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
+        if 'random_seed' in config:
+            seed = int(_utils.convert_shared_float_array_to_numpy(config['random_seed']))
+            _tf.set_random_seed(seed)
+        else:
+            seed=None
```
shantanuchhabra: Let's follow PEP8 here and surround the assignment with spaces. https://www.python.org/dev/peps/pep-0008/

src/python/turicreate/toolkits/activity_classifier/_activity_classifier.py
```diff
@@ -33,7 +33,7 @@
 USE_CPP = _tkutl._read_env_var_cpp('TURI_AC_USE_CPP_PATH')
 
 def create(dataset, session_id, target, features=None, prediction_window=100,
-           validation_set='auto', max_iterations=10, batch_size=32, verbose=True):
+           validation_set='auto', max_iterations=10, batch_size=32, verbose=True, random_seed=None):
     """
```
TobyRoseman: Please add a description for this new parameter in the Parameters section of the docstring.

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -324,7 +324,7 @@ std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
 
     // Make an instance of python object
     pybind11::object activity_classifier = tf_ac_backend.attr(
-        "ActivityTensorFlowModel")(weights, n, c_in, c_out, pw, w_out);
+        "ActivityTensorFlowModel")(weights, n, c_in, c_out, pw, w_out, config);
```
nickjong: This `config` dictionary is a vestige of the original TCMPS interface for binding a C backend to the Python frontend. Could we just add normal Pythonic arguments to `ActivityTensorFlowModel`?

src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
```diff
@@ -21,13 +21,19 @@
...
 
+        _utils.suppress_tensorflow_warnings()
         self.gpu_policy = _utils.TensorFlowGPUPolicy()
         self.gpu_policy.start()
 
         for key in net_params.keys():
             net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
+        if 'random_seed' in config:
+            seed = int(_utils.convert_shared_float_array_to_numpy(config['random_seed']))
+            _tf.set_random_seed(seed)
```
nickjong: What is this seed used for? Aren't we already specifying a seed for the dropout layer?

src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
```diff
@@ -83,9 +89,9 @@ def __init__(self, net_params, batch_size, num_features, num_classes, prediction
             beta_initializer=_tf.initializers.constant(net_params['bn_beta'], verify_shape=True),
             gamma_initializer=_tf.initializers.constant(net_params['bn_gamma'], verify_shape=True),
             moving_mean_initializer=_tf.initializers.constant(net_params['bn_running_mean'], verify_shape=True),
-            moving_variance_initializer=_tf.initializers.constant(net_params['bn_running_var'], verify_shape=True), training=self.is_training )
+            moving_variance_initializer=_tf.initializers.constant(net_params['bn_running_var'], verify_shape=True), training=self.is_training)
         dense = _tf.nn.relu(dense)
-        dense = _tf.layers.dropout(dense, rate=0.5, training=self.is_training)
+        dense = _tf.layers.dropout(dense, rate=0.5, training=self.is_training, seed=seed)
```
jakesabathia2: hmmm just found something strange and doesn't make sense
because we re-run a session every single time for each mini batch,
set a fixed seed either from graph-level or opt-level with produce the same result
every single time.
Say:
```
random_num = tf.random.unifrom([],0,1,seed=0)
with tf.Session() as session:
   print(session.run(random_num)

with tf.Session() as session:
   print(session.run(random_num))
```
will produce two identical random number.
That is to say, the behavior across all mini-batch will be the same.
@nickjong  @shreyajain17  I don't think this is what we expect.

shreyajain17: I don't really understand completely. But you get different results with different seeds and same result with the same seed. Isn't that what is expected?

```In [4]: model = tc.activity_classifier.create(data,'exp_id','activity',random_seed=1)
Pre-processing 672978 samples...
Using sequences of size 2000 for model creation.
Processed a total of 55 sessions.
Using GPU to create model (AMD Radeon Pro 560X)
+--------------+----------------+---------------------+--------------+
| Iteration    | Train Accuracy | Validation Accuracy | Elapsed Time |
+--------------+----------------+---------------------+--------------+
| 1            | 0.53856        | 0.767315            | 12.20s       |
| 2            | 0.745894       | 0.853582            | 13.84s       |
| 3            | 0.814608       | 0.892493            | 15.52s       |
| 4            | 0.85464        | 0.905128            | 17.16s       |
| 5            | 0.877809       | 0.910418            | 18.77s       |
| 6            | 0.885565       | 0.915482            | 20.52s       |
| 10           | 0.90907        | 0.918054            | 27.72s       |
+--------------+----------------+---------------------+--------------+
Training complete
Total Time Spent: 50.2019

In [5]: model = tc.activity_classifier.create(data,'exp_id','activity',random_seed=2)
Pre-processing 671511 samples...
Using sequences of size 2000 for model creation.
Processed a total of 55 sessions.
Using GPU to create model (AMD Radeon Pro 560X)
+--------------+----------------+---------------------+--------------+
| Iteration    | Train Accuracy | Validation Accuracy | Elapsed Time |
+--------------+----------------+---------------------+--------------+
| 1            | 0.546871       | 0.786748            | 12.26s       |
| 2            | 0.757693       | 0.855413            | 13.91s       |
| 3            | 0.829075       | 0.871994            | 15.56s       |
| 4            | 0.859569       | 0.894701            | 17.20s       |
| 5            | 0.881342       | 0.89708             | 18.88s       |
| 6            | 0.886383       | 0.905612            | 20.76s       |
| 10           | 0.906395       | 0.919267            | 27.95s       |
+--------------+----------------+---------------------+--------------+
Training complete
Total Time Spent: 50.7373

In [6]: model = tc.activity_classifier.create(data,'exp_id','activity',random_seed=1)
Pre-processing 672978 samples...
Using sequences of size 2000 for model creation.
Processed a total of 55 sessions.
Using GPU to create model (AMD Radeon Pro 560X)
+--------------+----------------+---------------------+--------------+
| Iteration    | Train Accuracy | Validation Accuracy | Elapsed Time |
+--------------+----------------+---------------------+--------------+
| 1            | 0.53856        | 0.767315            | 12.53s       |
| 2            | 0.745894       | 0.853582            | 14.22s       |
| 3            | 0.814608       | 0.892493            | 15.93s       |
| 4            | 0.85464        | 0.905128            | 17.62s       |
| 5            | 0.877809       | 0.910418            | 19.32s       |
| 6            | 0.885565       | 0.915482            | 21.13s       |
| 10           | 0.90907        | 0.918054            | 28.65s       |
+--------------+----------------+---------------------+--------------+```

Comment (jakesabathia2) [reply src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py]: What I'm saying is,
say the input feature of the dropout layer has 10 dimensions, 
`dropout(seed=0)` with dropout the `dim=0`,
so for every mini batch the first dim will be dropped.
But the model can learn better if it change.

say we have '3' total mini batch
==now==
minibatch_1:drop dim=0
minibatch_2:drop dim=0
minibatch_10:drop dim=0

I'm just wondering we expect thing more like
minibatch_1:drop dim=0
minibatch_2:drop dim=3
minibatch_10:drop dim=4

and the dim [0,3,4] is fixed for every experiment if we provide the same seed.

Hope this clarify.

File: src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py (hunk 4 above)
Comment (shreyajain17) [reply src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py]: `_tf.set_random_seed(seed)` sets the seed at graph level and setting the seed for dropout level is more op-level. And I have realized that it is redundant to call it here and it is sufficient to specify only at dropout layer.

File: src/ml/neural_net/tf_compute_context.cpp (hunk 3 above)
Comment (shreyajain17) [reply src/ml/neural_net/tf_compute_context.cpp]: Yes

File: src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py (hunk 5 above)
Comment (shreyajain17) [reply src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py]: So this is what tensorflow says - 
seeds can be set at graph-level and op-level.

To generate different sequences across sessions, set neither graph-level nor op-level seeds:
```
a = tf.random.uniform([1])
b = tf.random.normal([1])

print("Session 1")
with tf.compat.v1.Session() as sess1:
  print(sess1.run(a))  # generates 'A1'
  print(sess1.run(a))  # generates 'A2'
  print(sess1.run(b))  # generates 'B1'
  print(sess1.run(b))  # generates 'B2'

print("Session 2")
with tf.compat.v1.Session() as sess2:
  print(sess2.run(a))  # generates 'A3'
  print(sess2.run(a))  # generates 'A4'
  print(sess2.run(b))  # generates 'B3'
  print(sess2.run(b))  # generates 'B4'
```
To generate the same repeatable sequence for an op across sessions, set the seed for the op:
```
a = tf.random.uniform([1], seed=1)
b = tf.random.normal([1])

# Repeatedly running this block with the same graph will generate the same
# sequence of values for 'a', but different sequences of values for 'b'.
print("Session 1")
with tf.compat.v1.Session() as sess1:
  print(sess1.run(a))  # generates 'A1'
  print(sess1.run(a))  # generates 'A2'
  print(sess1.run(b))  # generates 'B1'
  print(sess1.run(b))  # generates 'B2'

print("Session 2")
with tf.compat.v1.Session() as sess2:
  print(sess2.run(a))  # generates 'A1'
  print(sess2.run(a))  # generates 'A2'
  print(sess2.run(b))  # generates 'B3'
  print(sess2.run(b))  # generates 'B4'
```

So wouldn't it be generating same sequence of dimensions rather than the same dimension  across sessions?

Comment (jakesabathia2) [reply src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py]: You are right :).
We are using the same session to run multiple times haha.

**Comment by shreyajain17** [comment]
> Good!
> Can you add an unit test for this PR?

Done.

File: src/ml/neural_net/tf_compute_context.cpp [hunk 6]
```diff
@@ -317,14 +317,22 @@ std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
   const float* pred_window = prediction_window.data();
   int pw = static_cast<int>(*pred_window);
 
+  int seed = -1;
+  auto random_seed_it = config.find("random_seed");
+  if (random_seed_it != config.end()) {
+    shared_float_array random_seed = random_seed_it->second;
+    const float* rd = random_seed.data();
+    seed = static_cast<int>(*rd);
```
nickjong: Using static cast here loses precision. See how this field is set here:
https://github.com/apple/turicreate/blob/76618e56d00c7db542a47a301e9c4b6f39ee883e/src/toolkits/activity_classification/activity_classifier.cpp#L65

And how the MPS implementation reads it:
https://github.com/apple/turicreate/blob/76618e56d00c7db542a47a301e9c4b6f39ee883e/src/ml/neural_net/mps_layers.mm#L655

In general, using on the "integer" part of the `float` value won't use the full 32 bits of entropy we allow to be passed in the random seed

src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
```diff
@@ -21,14 +21,15 @@
...
+    def __init__(self, net_params, batch_size, num_features, num_classes, prediction_window, seq_len, random_seed):
 
+        _utils.suppress_tensorflow_warnings()
         self.gpu_policy = _utils.TensorFlowGPUPolicy()
         self.gpu_policy.start()
 
         for key in net_params.keys():
             net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
-
+        seed = random_seed if random_seed!=-1 else None
```
nickjong: Why isn't `-1` a valid value for the seed?

src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
```diff
@@ -67,7 +68,7 @@ def __init__(self, net_params, batch_size, num_features, num_classes, prediction
         conv = _tf.nn.bias_add(conv, self.biases['conv_bias'])
         conv = _tf.nn.relu(conv)
 
-        dropout = _tf.layers.dropout(conv, rate=0.2, training=self.is_training)
+        dropout = _tf.layers.dropout(conv, rate=0.2, training=self.is_training, seed=seed)
```
nickjong: To future-proof this, could we just do this at the graph level?

src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
shreyajain17: I should try to send random seed as FLEX_UNDEFINED at inference time.

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: Okay read the difference makes a bit more sense.

src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
jakesabathia2: Actually I think there we got the same problem as OD.
`tf.set_random_seed` doesn't work here either ;(

shreyajain17: Yeah he is right. Somehow only operation level seed works for dropout!

jakesabathia2: haha seems like this is a known issue?
https://stackoverflow.com/questions/38469632/tensorflow-non-repeatable-results/40247201

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -317,14 +318,22 @@ std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
   const float* pred_window = prediction_window.data();
   int pw = static_cast<int>(*pred_window);
 
+  int seed = FLEX_UNDEFINED;
+  auto random_seed_it = config.find("random_seed");
+  if (random_seed_it != config.end()) {
```
jakesabathia2: what happens if FLEX_UNDEFINED is passed into the tensorflow code?
I guess in the inference time it doesn't matter but just want to make sure.

shreyajain17: FLEX_UNDEFINED is like None in python side. Also it doesn't matter because seed is used in dropouts which is disabled at inference time.

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -11,6 +11,7 @@
 #include <core/util/try_finally.hpp>
 #include <ml/neural_net/image_augmentation.hpp>
 #include <ml/neural_net/model_backend.hpp>
+#include <model_server/lib/variant_deep_serialize.hpp>
```
nickjong: Why do we need to add this import?

src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -317,14 +318,22 @@ std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
   const float* pred_window = prediction_window.data();
   int pw = static_cast<int>(*pred_window);
 
+  int seed = FLEX_UNDEFINED;
```
nickjong: `FLEX_UNDEFINED` doesn't have `int` type. Can we just require that the `config` continue the random seed?

shreyajain17: What do you mean? Inference config has no random seed. I just have to pass something to python side, it won't be used at inference anyway.

src/ml/neural_net/tf_compute_context.cpp
shreyajain17: It was for FLEX_UNDEFINED.

src/python/turicreate/test/test_activity_classifier.py
```diff
@@ -161,6 +161,14 @@ def test_invalid_model(self):
                                                   prediction_window=self.prediction_window,
                                                   validation_set=None)
 
+    def test_create_with_fixed_random_seed(self):
+        model_1 = tc.activity_classifier.create(self.data, target=self.target, session_id=self.session_id, max_iterations=3, random_seed=86)
+        pred_1 = model_1.predict(self.data)
```
jakesabathia2: better to test on `output_type='probability_vector'`

src/python/turicreate/test/test_activity_classifier.py
```diff
@@ -161,6 +161,14 @@ def test_invalid_model(self):
...
+        model_2 = tc.activity_classifier.create(self.data, target=self.target, session_id=self.session_id, max_iterations=3, random_seed=86)
+        pred_2 = model_2.predict(self.data)
+        for i in range(len(pred_1)):
```
jakesabathia2: I think we should make sure `len(pred_1) == len(Fred_2)` here.

PR #2897, opened 2020-01-06
Fix seed for DC

shreyajain17:


src/ml/neural_net/tf_compute_context.hpp
```diff
@@ -46,8 +46,8 @@ EXPORT class tf_compute_context: public compute_context {
 
   std::unique_ptr<model_backend> create_drawing_classifier(
       /* TODO: const float_array_map& config if needed */
```
shantanuchhabra: Let's remove the TODO

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
```diff
@@ -27,6 +27,12 @@ def __init__(self, net_params, batch_size, num_classes):
         for key in net_params.keys():
             net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
 
+        if 'random_seed' in config:
+            seed = int(_utils.convert_shared_float_array_to_numpy(config['random_seed']))
+            _tf.set_random_seed(seed)
+        else:
+            seed=None
```
shantanuchhabra: Let's follow PEP8 and add spaces around the assignment.

"Always surround these binary operators with a single space on either side: assignment (=), augmented assignment (+=, -= etc.), comparisons (==, <, >, !=, <>, <=, >=, in, not in, is, is not), Booleans (and, or, not)." from https://www.python.org/dev/peps/pep-0008/

shreyajain17: I think I can remove the else completely. It is not needed to assign `seed` to None.

src/ml/neural_net/tf_compute_context.hpp
shreyajain17: Okay

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
```diff
@@ -27,6 +27,12 @@ def __init__(self, net_params, batch_size, num_classes):
         for key in net_params.keys():
             net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
 
+        if 'random_seed' in config:
+            seed = int(_utils.convert_shared_float_array_to_numpy(config['random_seed']))
+            _tf.set_random_seed(seed)
```
nickjong: What is the random seed used for in the DC backend?

shreyajain17:
> Can you add an unit test for this?

Done.

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
shreyajain17: Didn't need this. Removed this and config everywhere.

PR #2901, opened 2020-01-07
Better tarfile management for OSOD.

### LINKED ISSUE #2005: One Shot - needs better tar file management for SArray background images

shantanuchhabra:
Fixes #2005 

**Summary**

- We move the logic to unzip the backgrounds tarball to when the tarball is downloaded the first time.
- We save the unzipped tarball in the data cache instead of the current directory. This facilitates looking up in the data cache and avoiding multiple downloads and unzips.

### DIFF
--- a/src/python/turicreate/toolkits/_data_zoo.py
+++ b/src/python/turicreate/toolkits/_data_zoo.py
@@ -10,6 +10,7 @@ import os as _os
 import sys as _sys
 import requests as _requests
 import turicreate as _tc
+import shutil as _shutil
 import tarfile as _tarfile
 import hashlib as _hashlib
 from six.moves.urllib import parse as _urlparse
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/_data_zoo.py
```diff
@@ -21,14 +22,21 @@ class OneShotObjectDetectorBackgroundData(object):
...
         self.sarray_url = _urlparse.urljoin(
             DATA_URL_ROOT, self.source_tar_filename)
         self.sarray_url_md5_pairs = [
             (self.sarray_url, "08830e90771897c1cd187a07cdcb52b4")
             ]
 
     def get_backgrounds_path(self):
-        backgrounds_path = _download_and_checksum_files(
+        if _os.path.exists(self.destination_sarray_path):
+            return self.destination_sarray_path
```
nickjong: How do we know the contents of this path are valid?

shantanuchhabra: Added a check!

Issue #2551, opened 2019-11-01
Warming message if the provide class_labels doesn't exist in the training dataset.


jakesabathia2:
The OD data iterator ignore the bounding box with invalid label.
It might be helpful to proving warning if the "class_labels" contains labels which 
do not exist in the training data.
Say user provide [catt,dog], while the training data has [cat,dog],
or user just provide [chair, horse] with data only contains [cat,dog].

PR #2906, opened 2020-01-09
warning when user-defined 'classes' contains labels not in the training data.

jakesabathia2:
#2551

### DIFF
--- a/src/toolkits/object_detection/od_data_iterator.cpp
+++ b/src/toolkits/object_detection/od_data_iterator.cpp
@@ -309,14 +309,11 @@ simple_data_iterator::compute_properties(
     result.classes = std::move(expected_class_labels);
     int i = 0;
     for (const std::string& label : result.classes) {
-      if (classes_inferred.find(label) == classes_inferred.end()) {
-
-        std::stringstream ss;
-        ss << "Class label " << label << " is not presented in the dataset.";
-        log_and_warn(ss.str());
-        //logstream("q86!");
-        //std::cout << "Warning: class label '" + label +
-        //                 "' is not presented in the dataset.\n";
+      if (classes_inferred.find(label) == classes_inferred.end() &&
+          is_training_) {
+        std::cout << "Warning: User provided 'classes' includes label '" +
+                         label +
+                         "', which is not presented in the training dataset.\n";
       }
       result.class_to_index_map[label] = i++;
     }
@@ -350,6 +347,9 @@ simple_data_iterator::simple_data_iterator(const parameters& params)
       // Whether to traverse the SFrame more than once, and whether to shuffle.
       repeat_(params.repeat),
       shuffle_(params.shuffle),
+
+      // Whether it is in the training stage
+      is_training_(params.is_training),
 
       // Identify/verify the class labels and other annotation properties.
       annotation_properties_(
--- a/src/toolkits/object_detection/od_data_iterator.cpp
+++ b/src/toolkits/object_detection/od_data_iterator.cpp
@@ -311,7 +311,7 @@ simple_data_iterator::compute_properties(
     for (const std::string& label : result.classes) {
       if (classes_inferred.find(label) == classes_inferred.end() &&
           is_training_) {
-        logstream(5)
+        logprogress_stream
             << "Warning: User provided 'classes' includes label '" + label +
                    "', which is not presented in the training dataset.\n"
             << std::endl;
--- a/src/toolkits/object_detection/od_data_iterator.cpp
+++ b/src/toolkits/object_detection/od_data_iterator.cpp
@@ -310,8 +310,13 @@ simple_data_iterator::compute_properties(
     int i = 0;
     for (const std::string& label : result.classes) {
       if (classes_inferred.find(label) == classes_inferred.end()) {
-        std::cout << "Warning: class label '" + label +
-                         "' is not presented in the dataset.\n";
+
+        std::stringstream ss;
+        ss << "Class label " << label << " is not presented in the dataset.";
+        log_and_warn(ss.str());
+        //logstream("q86!");
+        //std::cout << "Warning: class label '" + label +
+        //                 "' is not presented in the dataset.\n";
       }
       result.class_to_index_map[label] = i++;
     }
--- a/src/toolkits/object_detection/od_data_iterator.cpp
+++ b/src/toolkits/object_detection/od_data_iterator.cpp
@@ -309,12 +309,13 @@ simple_data_iterator::compute_properties(
     result.classes = std::move(expected_class_labels);
     int i = 0;
     for (const std::string& label : result.classes) {
-      if (classes_inferred.find(label) == classes_inferred.end() &&
-          is_training_) {
-        logprogress_stream
-            << "Warning: User provided 'classes' includes label '" + label +
-                   "', which is not presented in the training dataset.\n"
-            << std::endl;
+      if (is_training_) {
+        if (classes_inferred.find(label) == classes_inferred.end()) {
+          logprogress_stream
+              << "Warning: User provided 'classes' includes label '" + label +
+                     "', which is not presented in the training dataset.\n"
+              << std::endl;
+        }
       }
       result.class_to_index_map[label] = i++;
     }
--- a/src/toolkits/object_detection/od_data_iterator.cpp
+++ b/src/toolkits/object_detection/od_data_iterator.cpp
@@ -289,12 +289,17 @@ simple_data_iterator::compute_properties(
 
   std::tie(classes, result.num_instances) = get_annotation_info(annotations);
 
+  // Infer the class names from the observed labels.
+  std::set<std::string> classes_inferred;
+  for (const flexible_type& label : classes.range_iterator()) {
+    classes_inferred.insert(label);
+  }
+
   if (expected_class_labels.empty()) {
 
-    // Infer the class-to-index map from the observed labels.
     result.classes.reserve(classes.size());
     int i = 0;
-    for (const flexible_type& label : classes.range_iterator()) {
+    for (const std::string& label : classes_inferred) {
       result.classes.push_back(label);
       result.class_to_index_map[label] = i++;
     }
@@ -304,6 +309,10 @@ simple_data_iterator::compute_properties(
     result.classes = std::move(expected_class_labels);
     int i = 0;
     for (const std::string& label : result.classes) {
+      if (classes_inferred.find(label) == classes_inferred.end()) {
+        std::cout << "Warning: class label '" + label +
+                         "' is not presented in the dataset.\n";
+      }
       result.class_to_index_map[label] = i++;
     }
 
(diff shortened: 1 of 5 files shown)


src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -304,6 +309,10 @@ simple_data_iterator::compute_properties(
     result.classes = std::move(expected_class_labels);
     int i = 0;
     for (const std::string& label : result.classes) {
+      if (classes_inferred.find(label) == classes_inferred.end()) {
+        std::cout << "Warning: class label '" + label +
```
guihao-liang: you can use a log stream with a warning flag.

jakesabathia2:
> It's not obvious to me that the current behavior is incorrect. What is the precise use case you're worried about?
> 
> If I have an object detector trained on classes A, B, and C, and I find some test data that might have annotations for A, B, C, and D, why do I need to strip out the D annotations to use this test data without warning?

@nickjong @guihao-liang
I agree the case you mention. I think the case is most in the training:
Say the training data has ["dogs","cats"] labels, but an user gives ["dog","cats"] due to typo.

src/toolkits/object_detection/od_data_iterator.cpp
srikris: +1 to that. Please don't use count. log progress_stream is better.

srikris:
I agree with @nickjong, I'm not sure we should do anything about this. If the test data has different labels, then it's really on the user to fix them. It might be okay to just warn if there exists labels in the test set that are missing in the training dataset. So the overall experience would be simply a warning that says the following

```
Warning: There are 3 labels present in the evaluation set that are missing from the training set. 
```

and then you continue to do the evaluation (as you do in this PR)

srikris:
> It's not obvious to me that the current behavior is incorrect. What is the precise use case you're worried about?
> 
> If I have an object detector trained on classes A, B, and C, and I find some test data that might have annotations for A, B, C, and D, why do I need to strip out the D annotations to use this test data without warning?

@nickjong I don't think this PR is proposing to strip it. @jakesabathia2 is simply proposing to warn the user that they might not have intended for "D" to exist which is a good thing to do IMO.

jakesabathia2:
@srikris @nickjong 
https://github.com/apple/turicreate/pull/2542#issuecomment-548629729
Here is the origin why there is this open issue.

TobyRoseman:
I agree we should give a warning rather than error here.

The more interesting question for: what should we do with examples in the test set that have previously unseen labels? Do we drop them? Does the model automatically get those wrong?

jakesabathia2:
> I agree we should give a warning rather than error here.
> 
> The more interesting question for: what should we do with examples in the test set that have previously unseen labels? Do we drop them? Does the model automatically get those wrong?



> I agree we should give a warning rather than error here.
> 
> The more interesting question for: what should we do with examples in the test set that have previously unseen labels? Do we drop them? Does the model automatically get those wrong?

@TobyRoseman  Great question. We drop those unseen labels, so it doesn't effect the evaluation.

nickjong:
> > I agree we should give a warning rather than error here.
> > The more interesting question for: what should we do with examples in the test set that have previously unseen labels? Do we drop them? Does the model automatically get those wrong?
> 
> > I agree we should give a warning rather than error here.
> > The more interesting question for: what should we do with examples in the test set that have previously unseen labels? Do we drop them? Does the model automatically get those wrong?
> 
> @TobyRoseman Great question. We drop those unseen labels, so it doesn't effect the evaluation.

This is part of my point. OD is not a classifier. Asking a detector to detect things in an image that happens to have objects that the detector wasn't trained on isn't inherently problematic, in the same way that (absent multi-label classification), it's inherently wrong to ask a classifier to classify an image whose one true label is one the classifier doesn't know about.

At present, OD evaluation works by computing average precision scores for each class on which the detector was trained. Suppose that you have a dataset of images annotated with cat and dog bounding boxes. It feels weird to me if you later go back, add additional annotations for "hamster" to a _test set_ and now get a warning.

I would argue that the right solution here is closer to including a note in the evaluation results. Perhaps the output of evaluation should simply note that the test data has classes that the detector cannot detect, and therefore there is no average precision score for those classes.

guihao-liang:
@jakesabathia2 Yeah, just warn them is good enough. You don't want to help people if they don't request you to help, which means explicit is better than implicit. 

Correct me if I'm wrong. Since our toolkit lacks the functionality of [strafity](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html), we cannot guarantee that train and test has the same data distribution (by labels). Just warn them. Even though we warn, they still don't have any tool to correct it. It should be a feature request if we hear complaints from user.

good work.

jakesabathia2:
@nickjong @guihao-liang 
Okay.
So I think I should clarify the purpose of this PR since I think I didn't make a clear title and everyone is misunderstanding it.
So I think the OD is doing the right thing now whatever training/test dataset has same labels or not (just ignore the labels not in the training data is totally fine.)

The purpose here is to warn user only when they create model with explicit 'classes' input argument provided.

Say the user use
```
model = tc.object_detector.create(dataset, classes=['catt', 'dog']),
```
when the training dataset has labels 'cat' and 'dog'.
We should warn the user that they might provide invalid 'classes' argument here (typo in this case),
since the model treat the user-defined 'classes' as ground truth labels instead of the labels inferred from the training dataset.

So it only warns the user during model creation with 'classes' argument provided.
(But now it also has warning message printed for evaluation haha which I should get rid of.)

guihao-liang:
> > > I agree we should give a warning rather than error here.
> > > The more interesting question for: what should we do with examples in the test set that have previously unseen labels? Do we drop them? Does the model automatically get those wrong?
> > 
> > 
> > > I agree we should give a warning rather than error here.
> > > The more interesting question for: what should we do with examples in the test set that have previously unseen labels? Do we drop them? Does the model automatically get those wrong?
> > 
> > 
> > @TobyRoseman Great question. We drop those unseen labels, so it doesn't effect the evaluation.
> 
> This is part of my point. OD is not a classifier. Asking a detector to detect things in an image that happens to have objects that the detector wasn't trained on isn't inherently problematic, in the same way that (absent multi-label classification), it's inherently wrong to ask a classifier to classify an image whose one true label is one the classifier doesn't know about.
> 
> At present, OD evaluation works by computing average precision scores for each class on which the detector was trained. Suppose that you have a dataset of images annotated with cat and dog bounding boxes. It feels weird to me if you later go back, add additional annotations for "hamster" to a _test set_ and now get a warning.
> 
> I would argue that the right solution here is closer to including a note in the evaluation results. Perhaps the output of evaluation should simply note that the test data has classes that the detector cannot detect, and therefore there is no average precision score for those classes.

For my understanding, it would be nice to tell me that 'hamster' is not in training set so I can add 'hamster' to my training set in order to get better knowledge (under the assumption that cat images probably don't share deep features with hamsters). 

In practice, the test set serves a simulation as a production environment. True that production may have unseen data and the cycle to figure this out might be very long. That's why the training set could have unseen data over the test set.

Usually, people would promote the model to production if it works well in test set. The test set data distribution is actually very important. I think rarely people want to simulate the production with unseen data in the test set during the model training phase.

If the test set has unseen data over the training set, probably nothing I can do but add unseen data to the training set to start it over, which might be the handiest way to improve overall accuracy for the test set before promoting the model to production. Otherwise, the test performance might be poor and user doesn't have a clear clue to fix it.

guihao-liang:
> @nickjong @guihao-liang
> Okay.
> So I think I should clarify the purpose of this PR since I think I didn't make a clear title and everyone is misunderstanding it.
> So I think the OD is doing the right thing now whatever training/test dataset has same labels or not (just ignore the labels not in the training data is totally fine.)
> 
> The purpose here is to warn user only when they create model with explicit 'classes' input argument provided.
> 
> Say the user use
> 
> ```
> model = tc.object_detector.create(dataset, classes=['catt', 'dog']),
> ```
> 
> when the training dataset has labels 'cat' and 'dog'.
> We should warn the user that they might provide invalid 'classes' argument here (typo in this case),
> since the model treat the user-defined 'classes' as ground truth labels instead of the labels inferred from the training dataset.
> 
> So it only warns the user during model creation with 'classes' argument provided.
> (But now it also has warning message printed for evaluation haha which I should get rid of.)

Ah, I would admit I misunderstand it even though I was the guy who reported the issue. Do you check the user input label on the whole data set or just training?

guihao-liang:
One more follow up, what's the purpose of user input `classes`?

jakesabathia2:
> One more follow up, what's the purpose of user input `classes`?

Haha the user can define the class labels for the model,
and once provided it override the inferred labels from the training dataset haha.

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -304,6 +309,12 @@ simple_data_iterator::compute_properties(
     result.classes = std::move(expected_class_labels);
     int i = 0;
     for (const std::string& label : result.classes) {
+      if (classes_inferred.find(label) == classes_inferred.end() &&
+          is_training_) {
+        std::cout << "Warning: User provided 'classes' includes label '" +
```
guihao-liang: let me verify whether you can use log_stream to do the warning to stderr.

jakesabathia2:
```
# training data has label ['cat', 'dog']
model = tc.object_detector(data, classes=['catt' , 'dog']) # warning that 'catt' is not in the training set.
predict = model.evaluate(data) # No warning.
```
This behavior is what I think is expected.

src/toolkits/object_detection/od_data_iterator.cpp
jakesabathia2: Good!

src/toolkits/object_detection/od_data_iterator.cpp
guihao-liang: use `logprogress_stream << "Warning: class label '" ...`

logpgress_stream has a call back at the python layer!

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -304,6 +309,13 @@ simple_data_iterator::compute_properties(
     result.classes = std::move(expected_class_labels);
     int i = 0;
     for (const std::string& label : result.classes) {
+      if (classes_inferred.find(label) == classes_inferred.end() &&
+          is_training_) {
+        logstream(5)
```
guihao-liang: logpregress_stream is better

jakesabathia2: Done!

src/toolkits/object_detection/od_data_iterator.cpp
```diff
@@ -304,6 +309,13 @@ simple_data_iterator::compute_properties(
     result.classes = std::move(expected_class_labels);
     int i = 0;
     for (const std::string& label : result.classes) {
+      if (classes_inferred.find(label) == classes_inferred.end() &&
```
hoytak: Move the if(is_training_) around the entire for loop for clarity.

srikris:
@jakesabathia2 Can we merge this?

jakesabathia2:
> @jakesabathia2 Can we merge this?

need @nickjong 's approval :) .

Issue #2859, opened 2019-12-16
All models' name() should be removed for documentation.


jakesabathia2:
```
print(model.name())

AttributeError: 'ActivityClassifier' object has no attribute 'name'
```
`name()` functions should be removed from the TC API documentation.

PR #2907, opened 2020-01-09
clean up models name function.

jakesabathia2:
#2859 

change the `name()` function to `_name()` for `custom_model` class,
in order to remove the `name` function from all documentations.

### DIFF
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -653,7 +653,7 @@ class CustomModel(ExposeAttributesFromProxy):
         --------
         >>> model_name = m.name()
         """
-        warnings.warn("CustomModel.name() is deprecated. It will be removed in the next major release. Use type(CustomModel) instead.")
+        warnings.warn("This function is deprecated. It will be removed in the next release.")
         return self.__class__.__name__
 
     def summary(self, output=None):
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -642,7 +642,7 @@ class CustomModel(ExposeAttributesFromProxy):
         Returns the name of the model.
 
         ..WARNING:: This function is deprecated, It will be removed in the next
-        major release. Use type(CustomModel) instead.
+        release. Use type(CustomModel) instead.
 
         Returns
         -------
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -24,6 +24,7 @@ import turicreate.util._file_util as file_util
 import os
 from copy import copy as _copy
 import six as _six
+import warnings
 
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/_model.py
```diff
@@ -636,7 +636,7 @@ def _load_version(cls, state, version):
     def __init__(self):
         pass
 
-    def name(self):
+    def _name(self):
         """
```
srikris: Doesn't this effect public API. I think the right way to fix this would be to deprecate this for 1 release and then remove it for the next release. https://docs.python.org/3/library/exceptions.html#DeprecationWarning.

@TobyRoseman Thoughts?

TobyRoseman: That sounds good to me. Also please include a warning in the docstring.

Taking a look at #2321 will probably be helpful.

jakesabathia2: Very good!

jakesabathia2: Oh but somehow the models which rely on C++ model (AC,OD,SC...) the `name()` function are already gone in `6.0`.
@TobyRoseman @srikris

srikris: That's okay, But this is public API for all other models so let's follow the deprecation procedure and then file a bug to fix it in 6.2

jakesabathia2: Good!

src/python/turicreate/toolkits/_model.py
```diff
@@ -649,6 +653,7 @@ def _name(self):
         --------
         >>> model_name = m.name()
         """
+        warnings.warn("CustomModel.name() is deprecated. It will be removed in the next major release. Use type(CustomModel) instead.")
```
srikris: Change the message to. 'This function is deprecated. It will be removed in the next release.'

src/python/turicreate/toolkits/_model.py
jakesabathia2: Good!

src/python/turicreate/toolkits/_model.py
jakesabathia2: Good!

src/python/turicreate/toolkits/_model.py
```diff
@@ -640,6 +641,9 @@ def name(self):
         """
         Returns the name of the model.
 
+        ..WARNING:: This function is deprecated, It will be removed in the next
+        major release. Use type(CustomModel) instead.
```
TobyRoseman: Sounds like we're going to remove this in 6.2. So that's not "the next major release". It's just the next release.

jakesabathia2: Good!

PR #2909, opened 2020-01-10
Fix coreml import issues.

hoytak:
This fixes the coreml custom model issues on 10.14, and disable the tests (temporarily) on 10.13.  The 10.13 libraries are failing due to a missing symbol in libSystem.B.dylib, which appears to be a compilation sdk issue.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -387,23 +387,23 @@ test_python_linux_python37:
     paths:
       - pytest.*
 
-test_python_mac_python27_10_13:
-  tags:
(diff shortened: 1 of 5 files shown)


hoytak:
Passes all internal tests.

hoytak:
This fixes the coreml custom model issues on 10.14, and disable the tests (temporarily) on 10.13. The 10.13 libraries are failing due to a missing symbol in libSystem.B.dylib, which appears to be a compilation sdk issue.  

This brings all tests back to passing on OSX.

PR #2914, opened 2020-01-11
API for allowing calls to extension functions to run on a background thread

hoytak:
Many of the current speed regressions are due to a lack of proper pipelining support in python.  This PR adds the ability to run extension functions on a background thread while the main python thread does other things.  This is done by using a futures interface that works on top of extension functions and can be called using a run_background method on the extension function call. 

Previously, we could run the following code:

```
for i in range(n):
    prepared_data = tc.extensions.prepared_data(data_info, i)
    run_tf_model(prepared_data)
```

The problem is that this is run in serial, meaning the CPU and GPU are not both busy.  With this API, you can replace this with: 

```
# Init for the first iteration here
data_future = tc.extensions.prepared_data.run_background(data_info, 0)

for i in range(n):
    # block until background function finishes; get result.
    prepared_data = data_future.wait()

    # start data prep for next index in background
    if i + 1 != n:
        data_future = tc.extensions.prepared_data.run_background(data_info, i + 1)
   
    run_tf_model(prepared_data)
```

This should make pipelining support much easier.

### DIFF
--- a/src/model_server/lib/CMakeLists.txt
+++ b/src/model_server/lib/CMakeLists.txt
@@ -12,6 +12,7 @@ make_library(unity_core OBJECT
     unity_global.cpp
     unity_global_singleton.cpp
     get_toolkit_function_from_closure.cpp
+    toolkit_function_response.cpp
     toolkit_class_registry.cpp
     toolkit_function_registry.cpp
     simple_model.cpp
--- a/src/model_server/lib/toolkit_function_response.cpp
+++ b/src/model_server/lib/toolkit_function_response.cpp
@@ -1,0 +1,33 @@
+#include <model_server/lib/toolkit_function_response.hpp>
+#include <iostream>
+
+namespace turi {
+
+toolkit_function_response_future::toolkit_function_response_future(std::function<toolkit_function_response_type()> exec_function) :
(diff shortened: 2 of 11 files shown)


srikris:
@hoytak have you verified that this PR fixes any of the specific performance issues for 6.1?

hoytak:
@srikris -- That PR is in https://github.com/apple/turicreate/pull/2918.  I'm running benchmarks on it now.

guihao-liang:
A very high level question before I dig into it:
If I understand correctly, the perf gain only shows when people use it in an async way. If they use it sequentially, no gain should be there. If that correct?

hoytak:
@guihao-liang Yes that is correct.

guihao-liang:
FYI, @hoytak. This pr duplicates with #2918. Will leave all my comments on #2918.

hoytak:
> FYI, @hoytak. This pr duplicates with #2918. Will leave all my comments on #2918.

Yes, that PR depends on this one.  Github doesn't de-deplicate diffs in dependent PRs until the first one is merged.

hoytak:
Closing this PR as this functionality is folded into the image pipelining PR and tested more fully there.

PR #2916, opened 2020-01-13
DC loss reintroduced and accuracy calculated in C++

shreyajain17:


src/toolkits/drawing_classifier/drawing_classifier.hpp
```diff
@@ -68,7 +68,7 @@ class EXPORT drawing_classifier : public ml_model_base {
                              variant_type validation_data,
                              std::map<std::string, flexible_type> opts);
 
-  virtual void iterate_training();
+  virtual void iterate_training(bool show_loss);
```
nickjong: We'll have to resolve how to deal with this argument when we expose this method for checkpointing, but this is okay for now.

shreyajain17: Okay!

Issue #2813, opened 2019-12-06
Regression in image classifier due to pipelining


srikris:
Image classifier training on Linux GPUs has a regression in training time between 6.0 and 5.8 due to missing pipelining support.

TobyRoseman:
Related: #2849
This is also affecting Image Similarity.

nickjong:
We ended up reverting #2918 due to suspected involvement in segmentation faults observed while running our Image Classification scenario tests

PR #2918, opened 2020-01-13
Pipelining support for image feature extraction

hoytak:
This PR adds background data preparation for the image feature extraction stage. 

Closes https://github.com/apple/turicreate/issues/2813


src/model_server/lib/toolkit_function_response.cpp
```diff
@@ -0,0 +1,33 @@
+#include <model_server/lib/toolkit_function_response.hpp>
+#include <iostream>
+
+namespace turi {
+
+toolkit_function_response_future::toolkit_function_response_future(std::function<toolkit_function_response_type()> exec_function) :
+  m_info(new response_info) {
+
+    auto captured_info = this->m_info;
```
TobyRoseman: Does the use of `auto` here increase readability?

src/python/turicreate/extensions.py
```diff
@@ -115,24 +115,11 @@ def _setattr_wrapper(mod, key, value):
...
         setattr(_sys.modules[__name__], key, value)
 
-def _run_toolkit_function(fnname, arguments, args, kwargs):
+def _toolkit_function_pack_args(arguments, args, kwargs):
     """
-    Dispatches arguments to a toolkit function.
-
-    Parameters
-    ----------
-    fnname : string
```
TobyRoseman: I think you probably just want to remove the `fnname` parameter from the docstring and leave the rest.

TobyRoseman: Or perhaps I'm getting confused by how the diff is getting rendered and you haven't actually removed it. At any rate, you'll want this method to have a docstring.

src/model_server/extensions/internal_demo.cpp
```diff
@@ -5,9 +5,11 @@
  */
 #include <string>
 #include <vector>
+#include <thread>
 #include <model_server/lib/toolkit_function_macros.hpp>
 #include <model_server/lib/toolkit_class_macros.hpp>
 #include <model_server/lib/extensions/model_base.hpp>
+#include <core/data/sframe/gl_sframe.hpp>
```
guihao-liang: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/model_server/extensions/internal_demo.cpp
```diff
@@ -173,6 +189,7 @@ REGISTER_FUNCTION(_demo_addone, "param");
 REGISTER_FUNCTION(_demo_identity, "param");
 REGISTER_FUNCTION(_demo_add, "param", "param2");
 REGISTER_FUNCTION(_demo_to_string, "param");
+REGISTER_FUNCTION(_demo_extract_row, "X", "row");
```
guihao-liang: better parameter name?

src/model_server/lib/toolkit_function_response.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include <iostream>
+
+namespace turi {
+
+toolkit_function_response_future::toolkit_function_response_future(std::function<toolkit_function_response_type()> exec_function) :
+  m_info(new response_info) {
+
+    auto captured_info = this->m_info; 
+
+m_info->response_future = std::async(std::launch::async,
```
guihao-liang: the indentation level is not right.

src/model_server/lib/toolkit_function_response.cpp
```diff
@@ -0,0 +1,35 @@
+#include <model_server/lib/toolkit_function_response.hpp>
+#include <iostream>
+
+namespace turi {
+
+toolkit_function_response_future::toolkit_function_response_future(std::function<toolkit_function_response_type()> exec_function) :
```
guihao-liang: toolkit_function_response_future::toolkit_function_response_future(
    std::function<toolkit_function_response_type()> exec_function) {

}

https://google.github.io/styleguide/cppguide.html#Functions

src/model_server/lib/toolkit_function_response.cpp
guihao-liang: I think if you get used to it, the answer is yes.

src/model_server/lib/toolkit_function_response.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+ ASSERT_TRUE(m_info->response_future.valid());
+}
+
+const toolkit_function_response_type& toolkit_function_response_future::wait() const {
+  ASSERT_TRUE(m_info != nullptr); 
+
+  if(!m_info->is_completed) { 
+     ASSERT_TRUE(m_info->response_future.valid()); 
+     m_info->response_future.wait();
```
guihao-liang: I think you can only use `future` without `is_completed` to tell whether it's completed or not.

src/model_server/lib/unity_global.cpp
```diff
@@ -477,6 +477,19 @@ namespace turi {
     }
   }
 
+  toolkit_function_response_future unity_global::run_toolkit_background(
+      std::string toolkit_name,
+      variant_map_type& _arguments) {
+
+    auto capture_args = std::make_shared<variant_map_type>(_arguments);
```
guihao-liang: why this copy is needed?

guihao-liang:
I doubt this pipelining will gain perf improv. When I profile tf, it spins a thread pool internally. If image feature extraction is dispatched to a certain thread in tf, what we need to do is to call tf again to let another thread run the new task.

guihao-liang:
But this feature is quite useful in TC internally. Since all functions being delegated have quite similar signature 

```cpp
toolkit_function_response_type()
```

I wonder if we will go further have a thread pool to delegate async tasks? In which way, we should probably have less system thread construct and destruct footprints.

src/model_server/lib/unity_global.cpp
hoytak: Because the arguments are used in another thread after the call below returns, and it's passed by reference into this function, it may be invalid when used later.  Thus the function needs to capture all arguments.

hoytak:
> I doubt this pipelining will gain perf improv. When I profile tf, it spins a thread pool internally. If image feature extraction is dispatched to a certain thread in tf, what we need to do is to call tf again to let another thread run the new task.

I'm not sure I follow.  The issue is that the GPU is waiting around while our code prepares the next batch.  This is designed to keep both the GPU and CPU busy while all this is happening.

src/python/turicreate/extensions.py
hoytak: Must be the rendering.  It's got a docstring and that parameter.

src/model_server/lib/unity_global.cpp
guihao-liang: Yeah, totally understand that. I mean, since you are using `[=]` to capture by value, can't `_arguments` be copied implicitly?

```cpp
return toolkit_function_response_future(
         [=]() {
            return this->run_toolkit(toolkit_name, _arguments);
         });
   }
```

src/model_server/lib/toolkit_function_response.cpp
hoytak: Added a comment.  But yeah, auto could be something else.

guihao-liang:
> > I doubt this pipelining will gain perf improv. When I profile tf, it spins a thread pool internally. If image feature extraction is dispatched to a certain thread in tf, what we need to do is to call tf again to let another thread run the new task.
> 
> I'm not sure I follow. The issue is that the GPU is waiting around while our code prepares the next batch. This is designed to keep both the GPU and CPU busy while all this is happening.

Ah! Thanks for the clarification! I see your point. 

(each line represents a thread)
I thought we can do, date parp in one step in parallel, and trigger G/Cpu process in parallel too,

|-prep-|       | -G/C- |
|-prep-| ->  | -G/C- |
...
|-prep-|       | -G/C- |

But in reality, we do data preparation batch by batch.

without pipeline, 
|-prep-|-G/C-|-prep-|-G/C-|-prep-|-G/C-|-prep-|-G/C-| ...

with the pipeline, (with assumption prep time is longer than G/Cpu processing time)
                | -G/C- |  |-G/C- |   |-G/C- |
|-prep---|-prep---|-prep---|-prep---| ...

if G/Cpu processing time is longer than data prep time,
                                        | -G/C--------|  
                               | -G/C--------|  
                | -G/C--------|  
|-prep---|-prep---|-prep---|-prep---| ...

In this way, I think this pipelining is very valuable and can improve the overall throughput.

It's better to keep data preparation time less than G/Cpu processing time, then we can keep G/Cpu busy all the time.

guihao-liang:
If my above understanding is right,

G := G/Cpu process time,
P := data prep time,
n := number of batches, 

if G > P, through put improvement ratio is,



when n -> , it's equal to fomula blow,

if G <= P,



In order to get large gain, we should have G >> P.

src/model_server/lib/toolkit_function_response.cpp
```diff
@@ -0,0 +1,47 @@
...
+  if(!m_info->is_completed) { 
+     ASSERT_TRUE(m_info->response_future.valid()); 
+     m_info->response_future.wait(); 
+     ASSERT_EQ(m_info->response_future.get(), true); 
+     m_info->future_finished = true;
+  } else if(!m_info->future_finished) {
+     // Call get() to close out the state of the future.
+     ASSERT_TRUE(m_info->response_future.valid()); 
+     ASSERT_EQ(m_info->response_future.get(), true); 
+     m_info->future_finished = true;
```
guihao-liang: why extra state `future_finished` is needed? To be consistent with the [future design](https://en.cppreference.com/w/cpp/thread/future/future), the `toolkit_function_response_future` should be only movable, so no shared state between different `toolkit_function_response_future`.

Which means, if `m_info->response_future.valid() == false`, just do nothing.

guihao-liang:
> > If my above understanding is right,
> > G := G/Cpu process time,
> > P := data prep time,
> > n := number of batches,
> > if G > P, through put improvement ratio is,
> > G%2BP}{(n-1)P%2BG})
> > when n -> , it's equal to fomula blow,
> > if G <= P,
> > 
> > In order to get large gain, we should have G >> P.
> 
> Eh, with pipelining, the total time is `n * max(G, P).`. So as long as P < G, we're fine.

I understand, the `n * max(G, P).` is right under the condition that data prep can start only when G/Cpu finishes processing. 
________________________________________| -G/C--------|
_________________________| -G/C--------|
__________| -G/C--------|
|-prep---|-prep---|____ |-prep---|____ |-prep---| ...

Since the design is not blocking, we don't have to wait until a G/Cpu job to be finished to start the data preparation, as the graph below shows (also assuming we are using `async`, we can start as many threads as we request)

_____________________________| -G/C--------|
____________________| -G/C--------|
__________| -G/C--------|
|-prep---|-prep---|-prep---|-prep---| ...

guihao-liang:
I see your point. Instead of

```python
prep()
for i in range(0, n-1):
  baz = foo.run_backgrond()
  prep()
  baz.wait()

foo.run()
```

we can do

```python
futures = []
for i in range(n):
  prep()
  baz = foo.run_backgrond()
  futures.append(baz)

// wait all at last
for f in futures:
  f.wait()
```

guihao-liang:
you are right, our situation is `n * max(G, P)`.

G/Cpu is already fully busy (if underlying code is using multiprocessing), there's no need to start another G/Cpu job.

If we let P ~= G, we gain most.

hoytak:
The picture is more like: 
```
    |--G--|--G--|--G--|--G--|--G--|--G--|
|-P-|-P-| |-P-| |-P-| |-P-| |-P-| 
```
We don't want to start all the data prep at once due to memory use, and it wouldn't actually gain us anything.

src/model_server/lib/unity_global.cpp
hoytak: So I learned a minor detail of the C++ standard just now governing this.  I assumed the reference was the type used in the value capture, but it turns out when that type is a reference it's the referenced type that's used.  So I don't need that statement: 
```
5.1.2 Lambda expressions

(...) 14. An entity is captured by copy if it is implicitly captured and the capture-default is = or if it is explicitly captured with a capture that does not include an &. For each entity captured by copy, an unnamed nonstatic data member is declared in the closure type. The declaration order of these members is unspecified. The type of such a data member is the type of the corresponding captured entity if the entity is not a reference to an object, or the referenced type otherwise. 
```

src/model_server/lib/toolkit_function_response.cpp
hoytak: I'm not sure I get what you you're saying.  is_completed is a side effect of the function the future is managing, and may be valid before the future returns.

src/model_server/lib/unity_global.cpp
guihao-liang: glad to know that! But this is still not quite safe because the value is referred by 

```cpp
auto capture_args = std::make_shared<variant_map_type>(_arguments);
```

which will expire soon after the `run_toolkit_background` ends. So the reference value refers to an invalid memory section (the value referred by the shared pointer might be deleted or overridden).

```cpp
    return toolkit_function_response_future(
         [toolkit_name, _arguments]() {
            return this->run_toolkit(toolkit_name, _arguments);
         });
```
explicitly capture by value, even though it looks not elegant.

src/model_server/lib/toolkit_function_response.cpp
guihao-liang: let me clarify, the bool value `is_completed` can be stored into `future` to indicate the job is done successfully or not. If you are unlocked on future.wait(), them it means the job is done and you can retrieve its job status from future.get().

hoytak: The future isn't valid anymore after get() is called; once called, it can't be called again.  We want to be able to call this function multiple times if necessary, so storing it in our own state is better.

src/model_server/lib/unity_global.cpp
hoytak: How is it not safe?  The shared_ptr creates a copy of the variant_map_type, which lasts independent of the reference.

src/model_server/lib/toolkit_function_response.cpp
hoytak: It could be, but cython/python doesn't support move parameters, so we need to make toolkit_function_response_future default initializable and copyable, which is why we use a shared pointer to all the information.  We also need to call the future<...> exactly once to release its internal state.

guihao-liang: `is_completed` is not atomic, may have a potential issue that multiple threads see `is_complete` as `false` and call future.wait() multiple times.

guihao-liang: in short, not thread-safe.

hoytak:
This PR gets it within 10% of 5.8 according to benchmarks.  There could be more improvements, but merging as this PR passed all internal tests and all feedback has been addressed.

PR #2920, opened 2020-01-14
rewrite s3 stack with AWS SDK

guihao-liang:
Even though I really doubt the union_fstream design.

I'd say, the guy who wrote the original `curl` and `XML` path is an excellent programmer. If I were assigned with this task at that time without AWS SDK, probably I would already go crazy on testing it and reading the RESTful spec. Respect for that guy!


guihao-liang:
Decoupled dmlcio s3 curl stack completely with aws SDK. Through many days of segfaults, I feel the end is about to come.

guihao-liang:
```
Finished parsing file s3://tc_qa/integration/manual/csvs/mushroom.csv
Parsing completed. Parsed 100 lines in 0.1899 secs.
------------------------------------------------------
Inferred types from first 100 line(s) of file as
column_type_hints=[str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str,str]
If parsing fails due to incorrect types, you can correct
the inferred type list above and pass it to read_csv in
the column_type_hints argument
------------------------------------------------------
Finished parsing file s3://tc_qa/integration/manual/csvs/mushroom.csv
Parsing completed. Parsed 8124 lines in 0.077515 secs.
```
It now can use s3 sdk to read files. But it's intermittent. I need to do further investigation.

guihao-liang:
Fixed a bunch of bugs and add some good features.

* in s3_fstream.hpp, where seed(0, end) is completely wrong.
* enhanced s3_url to hold more information from env var settings
* move AwsInitAPI to a global level since only one initialization is enough.

guihao-liang:
```
In [11]: gui
Out[11]:
Columns:
        path    str
        image   Image
        label   str

Rows: 24766

Data:
+-------------------------------+------------------------+-------+
|              path             |         image          | label |
+-------------------------------+------------------------+-------+
| ./kagglecatsanddogs_3367a/... | Height: 375 Width: 500 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 281 Width: 300 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 500 Width: 489 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 500 Width: 403 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 150 Width: 150 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 360 Width: 431 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 375 Width: 500 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 471 Width: 500 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 375 Width: 500 |  cat  |
| ./kagglecatsanddogs_3367a/... | Height: 240 Width: 320 |  cat  |
+-------------------------------+------------------------+-------+
[24766 rows x 3 columns]

In [12]: gui.groupby("label", operations={'count': tc.aggregate.COUNT()})
Out[12]:
Columns:
        label   str
        count   int

Rows: 2

Data:
+-------+-------+
| label | count |
+-------+-------+
|  dog  | 12361 |
|  cat  | 12405 |
+-------+-------+
[2 rows x 2 columns]
In [13]: gui.apply(lambda x: len(x["label"]))
Out[13]:
dtype: int
Rows: 24766
[3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, ... ]
```

Successfully applied with a large sframe (777MB) and operated groupby and lambda on it.

src/core/storage/fileio/fs_utils.cpp
```diff
@@ -286,17 +294,16 @@ EXPORT bool create_directory_or_throw(const std::string& path) {
...
       case file_status::MISSING:
         log_and_throw_io_failure(
-          "Unable to create directory structure at " +
-          sanitize_url(path) +
-          ". Ensure that you have write permission to this location, or try again with a different path."
-        );
+            "Unable to create directory structure at " + sanitize_url(path) +
+            ". Ensure that you have write permission to this location, or try "
+            "again with a different path." +
+            "error messge: " + status.second);
```
hoytak: messge -> message.

hoytak: Also add space after period.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -111,20 +114,57 @@ bool bucket_name_valid(const std::string& bucket_name) {
 }
 
 
-std::string string_from_s3url(const s3url& parsed_url) {
-  std::string ret = "s3://" + parsed_url.access_key_id + ":"
-      + parsed_url.secret_key + ":";
+} // anonymous namespace
+
+S3Client init_aws_sdk_with_turi_env(s3url& parsed_url) {
```
hoytak: Is it expected to change the parsed_url struct and pass those changes back?  If it's not expected to pass the changes back, then make it pass-by-value; otherwise add a comment here about how it will change.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -111,20 +114,57 @@ bool bucket_name_valid(const std::string& bucket_name) {
...
+  // s3 client config
+  // DefaultCredentialProviderChain
+  Aws::Client::ClientConfiguration clientConfiguration;
+
+  // a little bit too long, anyway
+  clientConfiguration.requestTimeoutMs = 5 * 60000;
```
hoytak: What is the reasoning behind these values?  If there are situations where they could be adjusted, we should make them global variables (search for REGISTER_GLOBAL).

src/core/storage/fileio/s3_api.cpp
```diff
@@ -111,20 +114,57 @@ bool bucket_name_valid(const std::string& bucket_name) {
...
+
+  std::string region = fileio::get_region_name_from_endpoint(
+      clientConfiguration.endpointOverride.c_str());
+
+  if (!region.empty()) {
+    clientConfiguration.region = region.c_str();
+  } else {
+    auto env_var = get_auth_region_from_env();
+    if (env_var) {
+      clientConfiguration.region = env_var->c_str();
```
hoytak: The pointers returned from c_str() become invalid after env_var goes out of scope and may cause a segfault unless the .region and .sdk_region perform a copy on assignment.  I think we should put these strings (or optionsals) in the parsed_url struct to make sure the data is always valid.  I.e. add an std::string env_auth_region (may be empty) and a std::string env_endpoint; field to the parsed_url and then set these clientConfigurations from that.

hoytak: Otherwise, just make sure they're at function scope.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -136,8 +176,9 @@ const std::vector<std::string> S3Operation::_enum_to_str = {
  *
  * Returns true on success, false on failure.
  */
-bool parse_s3url(std::string url, s3url& ret, std::string& err_msg) {
+bool parse_s3url(std::string s3_url, s3url& ret, std::string& err_msg) {
```
hoytak: Let's do const std::string& out of habit here.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -346,249 +397,268 @@ list_objects_response list_objects_impl(s3url parsed_url,
     request.WithDelimiter("/"); // seperate objects from directories
 
     bool moreResults = false;
-    do
-    {
+    bool success = false;
+    int backoff = 50;  // ms
```
hoytak: Could you add a comment about your reasoning behind this number here?

src/core/storage/fileio/s3_api.cpp
hoytak: Update:  Ah, I see, they are there before.  Ideally I think they should be global variables still.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -346,249 +397,268 @@ list_objects_response list_objects_impl(s3url parsed_url,
...
                                std::string endpoint) {
-    std::string ret;
+  std::string ret;
 
-    // initialization
-    Aws::SDKOptions options;
-    options.loggingOptions.logLevel = Aws::Utils::Logging::LogLevel::Info;
-    Aws::InitAPI(options);
+  // credentials
+  Aws::Auth::AWSCredentials credentials(parsed_url.access_key_id.c_str(),
```
hoytak: Not a comment on you, but it seems like there could be a lot of logic in some of these places consolidated into common logic functions.  Lots of duplicated config code.

src/core/storage/fileio/s3_api.cpp
guihao-liang: I thought about it. might be a good-first-issue for the new comer.

src/core/storage/fileio/s3_api.cpp
guihao-liang: yes, it will change the optional sections of s3url, such as region or endpoint.

src/core/storage/fileio/s3_api.cpp
guihao-liang: clientConfiguration.region  is a Aws::string, which cannot copy the std::string. Using `c_str` is to force it copy the contents from std::string.

src/core/storage/fileio/s3_api.cpp
guihao-liang: inherited from previous impl. will do.

src/core/storage/fileio/s3_api.cpp
guihao-liang: no reason. Get this number by running the actual program.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -346,249 +397,268 @@ list_objects_response list_objects_impl(s3url parsed_url,
...
+          // jump out the retry loop
+          break;
+
+        } else {
+          auto error = outcome.GetError().GetResponseCode();
+
+          if (error == Aws::Http::HttpResponseCode::TOO_MANY_REQUESTS) {
+            n_retry++;
+            std::this_thread::sleep_for(std::chrono::milliseconds(backoff));
+            backoff *= 2;
```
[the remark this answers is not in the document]
guihao-liang: exponential backoff is the most adopted one. 1 sec is too long for io.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -346,249 +397,268 @@ list_objects_response list_objects_impl(s3url parsed_url,
...
+            request.WithContinuationToken(result.GetContinuationToken());
+          }
+
+          // jump out the retry loop
+          break;
+
+        } else {
+          auto error = outcome.GetError().GetResponseCode();
+
+          if (error == Aws::Http::HttpResponseCode::TOO_MANY_REQUESTS) {
```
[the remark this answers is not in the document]
guihao-liang: .Time out will be handled internally by sdk and sdk will retry 3 times.

I have thought about other failure reasons but I decide not to handle them prematurely. When we have concrete business need, it's not too late to add more error handling logic.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -346,249 +397,268 @@ list_objects_response list_objects_impl(s3url parsed_url,
...
                                   clientConfiguration, outcome)
                 << std::endl;
             ret.error = ss.str();
+            logstream(LOG_DEBUG)
+                << "list_objects_impl failed:" << ret.error << std::endl;
+            // no need to retry, just jump out all while loop
+            break;
+          }
         }
+      } while (n_retry < 3);  // finished retry
```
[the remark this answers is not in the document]
guihao-liang: when we want to apply retry to other operations, we should add this. I only observe restrictions on list operations so far. Hardcoding 3 is okay, IMO.

src/core/storage/fileio/s3_api.cpp
guihao-liang: Yeah, it's one of my complaints. Just leave it here for people to fix it in the future.

src/core/storage/fileio/s3_api.cpp
guihao-liang: use `c_str` is also a very common when using reference-count-based std::string for the multithreaded environment.

hoytak: Oh, then it does copy it.  So we're okay.  

Yes, it is common, but the original std::string cannot be deleted while the .c_str() result is being used.

src/core/storage/fileio/s3_api.cpp
hoytak: Fair.  It wasn't the best code to begin with.

src/core/storage/fileio/s3_api.cpp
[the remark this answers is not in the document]
hoytak: Okay, makes sense.

src/core/storage/fileio/s3_api.cpp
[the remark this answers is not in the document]
hoytak: Exponential backoff is good, yes.

src/core/storage/fileio/s3_api.cpp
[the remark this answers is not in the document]
hoytak: If the timeouts and data issues are handled internally, than that's fine.

src/core/storage/fileio/s3_api.cpp
hoytak: Why later and not now?   It's hard to review copy-pasted code.

src/core/storage/fileio/s3_filesys.hpp
```diff
@@ -0,0 +1,384 @@
...
+      // 64 MB
+      const size_t kDefaultBufferSize = 64 << 20UL;
+      max_buffer_size_ = kDefaultBufferSize;
+    }
+
+    InitRequest(url_);
+  }
+
+  virtual size_t Read(void *ptr, size_t size) {
+    if (no_exception_) {
```
hoytak: I don't understand this no_exception logic.

src/core/storage/fileio/s3_api.cpp
guihao-liang: It will be deleted in recent std::string implementation since it returns a const char*. There's no way to track a pointer usage. You can easily get segfault from dangling c_str() pointer in current impl.

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: this is the logic inherits from dlmcio code. Some data stream impl requires no exception. if the size is returned as 0, it means no data from stream. And the user should check the io bits such as eof, bad, to determine their error handling logic.

src/core/storage/fileio/s3_api.cpp
guihao-liang: due to the time constraint, I don't want to break things that are working now. The logic was there for a long time and we can push to the next release to clean up code.

src/core/storage/fileio/s3_filesys.hpp
hoytak: I don't understand why the logic is if(no_exception) { // throw exception }

src/core/storage/fileio/s3_filesys.hpp
```diff
@@ -0,0 +1,384 @@
...
+   *  create an stream, the stream will close the underlying files upon deletion
+   *
+   * \param uri the uri of the input currently we support
+   *            hdfs://, s3://, and file:// by default file:// will be used
+   * \param flag can be "w", "r", "a"
+   * \param allow_null whether NULL can be returned, or directly report error
+   * \return the created stream, can be NULL when allow_null == true and file do
+   * not exist
+   */
+  static Stream *Create(const char *uri, const char *const flag,
```
hoytak: This should return an std::shared_ptr<Stream> instead of a raw pointer.   Also, any reason the url is const char* instead of const std::string& ?

src/core/storage/fileio/s3_filesys.hpp
```diff
@@ -0,0 +1,384 @@
...
+   *  create an SeekStream for read only,
+   *  the stream will close the underlying files upon deletion
+   *  error will be reported and the system will exit when create failed
+   * \param uri the uri of the input currently we support
+   *            hdfs://, s3://, and file:// by default file:// will be used
+   * \param allow_null whether NULL can be returned, or directly report error
+   * \return the created stream, can be NULL when allow_null == true and file do
+   * not exist
+   */
+  static SeekStream *CreateForRead(const char *uri, bool allow_null = false);
```
hoytak: Avoid using raw pointers in any situation where they are not isolated to a single scope.

src/core/storage/fileio/s3_filesys.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+  /*!
+   * \brief try to fill the buffer with at least wanted bytes
+   * \param want_bytes number of bytes we want to fill
+   * \return number of remainning running curl handles
+   */
+
+  int FillBuffer(size_t want_bytes);
+  // data buffer
+  std::string buffer_;
```
hoytak: The buffer here makes it such that these classes cannot be used by more than one thread at a time -- i.e. each thread will have to create its own instance of this class.  Is that the intent?

src/core/storage/fileio/s3_filesys.hpp
```diff
@@ -0,0 +1,384 @@
...
+  /*!
+   * \brief open a stream, will report error and exit if bad thing happens
+   * NOTE: the Stream can continue to work even when filesystem was destructed
+   * \param path path to file
+   * \param uri the uri of the input
+   * \param flag can be "w", "r", "a"
+   * \return the created stream, can be NULL when allow_null == true and file
+   * do not exist
+   */
+  virtual Stream *Open(const s3url &path, const char *const flag);
```
hoytak: Let's work with shared pointers here.

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: Yeah, that's the intent. But our file stream is used through shared_ptr, so threads will share the same instance.

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: Sure! That's a good plan!

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: Yeah, inherits from the previous implementation. Will change it to smart pointer.

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: The interface is inherited. Will change.

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: I see! Let me check. might be my sloppiness.

guihao-liang: yeah, forget one exclamation sign! Nice catch!

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: Yeah, the original implementation from dmlcio is C style.

src/core/storage/fileio/s3_filesys.hpp
hoytak: So there's a problem then?  Are you saying that different threads access the same instance?  In that case we need to put in locking or something to ensure that different threads don't compete here.

hoytak: Actually, why do we need a buffer here?

src/core/storage/fileio/s3_filesys.cpp
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  return nwant;
+}
+
+void WriteStream::Write(const void *ptr, size_t size) {
+  size_t rlen = buffer_.length();
+  buffer_.resize(rlen + size);
+  std::memcpy(const_cast<char *>(buffer_.data()) + rlen, ptr, size);
+  if (buffer_.length() >= max_buffer_size_) Upload();
```
hoytak: I don't understand what's going on here with the buffer_.  Why only upload then?

hoytak: Ah, you're chunking it so that writing is done with bigger blocks.

guihao-liang: Yeah, I'm using AWS multipart upload API.

src/core/storage/fileio/s3_filesys.hpp
guihao-liang: because for the normal stream, you should be able to seek. So this one is going to save download from seeking within the same chunk.

guihao-liang:
passed all local tests.

PR #2924, opened 2020-01-14
Fixes bug in save/load in DC model

### LINKED ISSUE #2923: DC crash when load 5.8 model.
train a model in 5.8 and save it into "square_triangle.model"
```
model = tc.load_model("square_triangle.model")
predictions = model.predict(test_data)
```
this cause crash.

shreyajain17:
fixes #2923

### DIFF
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -312,7 +312,6 @@ std::unique_ptr<data_iterator> drawing_classifier::create_iterator(
   }
 
   data_params.feature_column_name = read_state<flex_string>("feature");
-  data_params.random_seed = read_state<int>("random_seed");
 
   return create_iterator(data_params);
 }


jakesabathia2:
> https://github.com/apple/turicreate/blob/d102f2b26573c283d967f239ccdcae06bb002194/src/toolkits/drawing_classifier/dc_data_iterator.cpp#L106
> 
> A better way to do it is to catch exception and ignore it. Let it use the default value,
> 
> ```c++
>     /** Determines results of shuffle operations if enabled. */
>     int random_seed = 0;
> ```

I think deleting this line is more suitable,
since other toolkit are always using the default random_seed for data iterator as well.

guihao-liang:
> > https://github.com/apple/turicreate/blob/d102f2b26573c283d967f239ccdcae06bb002194/src/toolkits/drawing_classifier/dc_data_iterator.cpp#L106
> > 
> > A better way to do it is to catch exception and ignore it. Let it use the default value,
> > ```c++
> >     /** Determines results of shuffle operations if enabled. */
> >     int random_seed = 0;
> > ```
> 
> I think deleting this line is more suitable,
> since other toolkit are always using the default random_seed for data iterator as well.

Then why you need to expose that option from `data_iterator`? You want to control the input order, otherwise, where is the randomness coming from?

jakesabathia2:
> > > https://github.com/apple/turicreate/blob/d102f2b26573c283d967f239ccdcae06bb002194/src/toolkits/drawing_classifier/dc_data_iterator.cpp#L106
> > > 
> > > A better way to do it is to catch exception and ignore it. Let it use the default value,
> > > ```c++
> > >     /** Determines results of shuffle operations if enabled. */
> > >     int random_seed = 0;
> > > ```
> > 
> > 
> > I think deleting this line is more suitable,
> > since other toolkit are always using the default random_seed for data iterator as well.
> 
> Then why you need to expose that option from `data_iterator`? You want to control the input order, otherwise, where is the randomness coming from?

I agree but what I'm saying is other toolkits now are all using the default value `0` for data iterator shuffle. So maybe it is better to put a separate PR which expose `param.random_seed` for all toolkits' data iterator?

shreyajain17:
I agree. I don't think this line should have been added since all the other toolkits use the default always. I think we need to introduce the parameter but must be done across all toolkits for consistency.

guihao-liang:
> > > > https://github.com/apple/turicreate/blob/d102f2b26573c283d967f239ccdcae06bb002194/src/toolkits/drawing_classifier/dc_data_iterator.cpp#L106
> > > > 
> > > > A better way to do it is to catch exception and ignore it. Let it use the default value,
> > > > ```c++
> > > >     /** Determines results of shuffle operations if enabled. */
> > > >     int random_seed = 0;
> > > > ```
> > > 
> > > 
> > > I think deleting this line is more suitable,
> > > since other toolkit are always using the default random_seed for data iterator as well.
> > 
> > 
> > Then why you need to expose that option from `data_iterator`? You want to control the input order, otherwise, where is the randomness coming from?
> 
> I agree but what I'm saying is other toolkits now are all using the default value `0` for data iterator shuffle. So maybe it is better to put a separate PR which expose `param.random_seed` for all toolkits' data iterator?

If you really think so, open one pr and mention it. I don't know the net effect so you guys decide.

guihao-liang:
> I agree. I don't think this line should have been added since all the other toolkits use the default always. I think we need to introduce the parameter but must be done across all toolkits for consistency.

Sure. I keep my disagreement on this. But I will follow the team's decision.

Issue #2798, opened 2019-12-05
Style Transfer inference is 4x slower in master versus in 5.8.


shantanuchhabra:
Repro steps:
1. Follow the user guide and train a model.
2. Stylize a content image.

Inference time was about 250 milliseconds per image in 5.8, and is 1 second per image in master.

guihao-liang:
Reproduced. Working on it. Good catch.

guihao-liang:
![image](https://user-images.githubusercontent.com/14238915/72392102-07d40400-36e4-11ea-8112-afdd76aeec1a.png)
from my inspection, the most time spent on `stylize` function is on
1. create_compute_context::create_style_transfer 400ms
2. tf_model_backend::predict 414 ms

The bottleneck is on the python call layer.
![image](https://user-images.githubusercontent.com/14238915/72392769-b3ca1f00-36e5-11ea-9069-ee0954aa50bf.png)

Besides that, revealed another bug the table printer will duplicate the result for the last entry,
![image](https://user-images.githubusercontent.com/14238915/72392833-e1af6380-36e5-11ea-986d-0cd1575aaf3e.png)
Where I only have one image to stylize. This is an easy fix.

guihao-liang:
Per discussion with @shreyajain17, the Tensorflow graph is re-initialized every time when `predict` is called, which is to accommodate different image sizes. @abhishekpratapa can give more explanation on this issue and make a decision to mark this as a known issue or not. My investigation stops here. I will wait for final decisions from @abhishekpratapa.

abhishekpratapa:
If you pass in the shape as a placeholder then we can prevent re-initializing the graph every-time. This I suspect is the correct solution.

guihao-liang:
talked with @shantanuchhabra and he will provide more information about platform, CPU or GPU path, for me to reproduce.

guihao-liang:
As it's tested locally on 10.15 with **GPU** path by @shreyajain17, the performance of stylizing a single image is around 110ms. #2937 resolves regressions on the **CPU** path.

PR #2926, opened 2020-01-15
duplicated printer for the last entry of each batch

guihao-liang:
related to #2798

### DIFF
--- a/src/toolkits/style_transfer/style_transfer.cpp
+++ b/src/toolkits/style_transfer/style_transfer.cpp
@@ -484,19 +484,22 @@ gl_sarray style_transfer::convert_style_indices_to_filter(
 
 gl_sframe style_transfer::predict(variant_type data,
                                   std::map<std::string, flexible_type> opts) {
-  gl_sframe_writer result({"row_id","style", "stylized_image"},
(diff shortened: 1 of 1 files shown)


src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -624,7 +629,6 @@ void style_transfer::perform_predict(gl_sarray data, gl_sframe_writer& result,
     ++row_idx;
   }
 
-  table.print_row(idx, progress_time(), "100%");
```
guihao-liang: only this line matters.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -565,18 +568,20 @@ void style_transfer::perform_predict(gl_sarray data, gl_sframe_writer& result,
...
-  table.print_header();
-
   // looping through all of the data
   data_iter->reset();
   ASSERT_EQ(batch_size, 1);
   std::vector<st_example> batch = data_iter->next_batch(batch_size);
   int row_idx = 0;
 
+  // Style Printer
+  // record time of process
```
guihao-liang: move the recording start point to be closer to the critical code blocks. might be only a few milliseconds more accurate.

Issue #2312, opened 2019-09-15
pep8 compatible code style check before build


guihao-liang:
I found this in a project of my personal interest and it might help us:
```
# .travis
before_script:
    # stop the build if there are Python syntax errors or undefined names
    - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
    # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
    - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
A universal python coding style consistent with pep8 is super helpful, isn't it?

TobyRoseman:
@guihao-liang - generally you shouldn't be assigning issues to other people. Also generally more than one person shouldn't be assigned to the same issues.

guihao-liang:
@TobyRoseman Thanks for letting me know!

PR #2929, opened 2020-01-15
Adding weights to DC

shreyajain17:
To make it consistent with all other toolkits, DC should also have consistent batch_size in the entire epoch. This can be done by using weights. Weights are arrays of 0/1 of the size of targets. For example, the last batch in the epoch is smaller than the specified batch_size then weights are zeros for the rest of the batch. While calculating loss and accuracy, the data corresponding the zeros is not taken into consideration.

### DIFF
--- a/src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
@@ -151,10 +151,7 @@ class DrawingClassifierTensorFlowModel(TensorFlowModel):
             self.cost
         )
 
-<<<<<<< HEAD
-=======
         self.sess = _tf.Session()
->>>>>>> f65e2cd45c... Adding weights to DC
         self.sess.run(_tf.global_variables_initializer())
 
         # Assign the initialised weights from C++ to tensorflow
--- a/src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
@@ -11,10 +11,11 @@ import numpy as _np
 from .._tf_model import TensorFlowModel
 import turicreate.toolkits._tf_utils as _utils
 import tensorflow.compat.v1 as _tf
+
 _tf.disable_v2_behavior()
 
-class DrawingClassifierTensorFlowModel(TensorFlowModel):
 
+class DrawingClassifierTensorFlowModel(TensorFlowModel):
     def __init__(self, net_params, batch_size, num_classes):
         """
         Defines the TensorFlow model, loss, optimisation and accuracy. Then
@@ -25,7 +26,9 @@ class DrawingClassifierTensorFlowModel(TensorFlowModel):
         self.gpu_policy.start()
 
         for key in net_params.keys():
-            net_params[key] = _utils.convert_shared_float_array_to_numpy(net_params[key])
+            net_params[key] = _utils.convert_shared_float_array_to_numpy(
+                net_params[key]
+            )
 
 
         self.dc_graph = _tf.Graph()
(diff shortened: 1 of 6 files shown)


jakesabathia2:
Seems like it crashes when evaluation?

shreyajain17:
> Seems like it crashes when evaluation?

After your change evaluation change or before? Will look into it.

jakesabathia2:
> > Seems like it crashes when evaluation?
> 
> After your evaluation change or before? Will look into it.

In Travis?
One unit test crash out.

shreyajain17:
> > > Seems like it crashes when evaluation?
> > 
> > 
> > After your evaluation change or before? Will look into it.
> 
> In Travis?
> One unit test crash out.

Cool will look into it.

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
```diff
@@ -33,6 +33,15 @@ def __init__(self, net_params, batch_size, num_classes):
         self.batch_size = batch_size
 
         self.input = _tf.placeholder(_tf.float32, [None, 28, 28, 1])
```
jakesabathia2: If the batch size is fixed to a constant, why is there still has the shape None in the first dimension?
Like in OD the shape is fixed:
https://github.com/apple/turicreate/blob/dc07e46d4e5562f0989e8532feb735391061e170/src/python/turicreate/toolkits/object_detector/_tf_model_architecture.py#L47

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
```diff
@@ -91,16 +100,12 @@ def __init__(self, net_params, batch_size, num_classes):
         self.predictions = softmax_out
 
         # Loss
-        self.cost = _tf.losses.softmax_cross_entropy(logits=out, onehot_labels=self.one_hot_labels, reduction=_tf.losses.Reduction.NONE)
+        self.cost = _tf.losses.softmax_cross_entropy(logits=out, onehot_labels=one_hot_labels, weights=reshaped_weights, reduction=_tf.losses.Reduction.NONE)
```
guihao-liang: if you can use pep8 format, that will be better for reviewing purposes.

guihao-liang:
`While calculating loss and accuracy, the data corresponding to the zeros is not taken into consideration.`

Just curious why?

src/toolkits/drawing_classifier/dc_data_iterator.cpp
```diff
@@ -192,41 +195,33 @@ data_iterator::batch simple_data_iterator::next_batch(size_t batch_size) {
   }
 
   if (target_index_ >= 0) {
-    DASSERT_EQ(real_batch_size, batch_targets.size());
+    batch_targets.resize(batch_size, 0.f);
```
guihao-liang: not sure why you don't want to use `real_batch_size`, which is quite critical.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -746,7 +749,7 @@ gl_sframe drawing_classifier::perform_inference(data_iterator* data) const {
       pending_batches.pop();
 
       size_t num_images = batch.data_info.num_samples;
-      ASSERT_EQ(num_images * num_classes, batch.data_info.predictions.size());
+      ASSERT_EQ(batch_size * num_classes, batch.data_info.predictions.size());
```
guihao-liang: batch_size may not be real_batch_size, say you have 65 images, the batch size is 64, then you have 2 batches of 64 and 1.

shreyajain17: Well the point of going this is that there will be two batch of 64. The batch size would never change. The weights are to weigh the loss. The weights for the first batch would all ones and for the second batch will be [1, 0, 0 ...]. The predictions come from the softmax would also be equal to the batch_size=64 but if you see the code after this, it will compute accuracy only till the num samples in the batch (which would be the real batch size)

src/toolkits/drawing_classifier/dc_data_iterator.cpp
shreyajain17: I am making the batch_targets to be of batch_size so assertion makes no sense to me.

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
shreyajain17: Will do

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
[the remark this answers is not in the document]
shreyajain17: Makes sense.

src/toolkits/drawing_classifier/drawing_classifier.cpp
guihao-liang: I see. What's the benefit of doing this? I'm curious, if the batch size is 64, and for the current batch, I only have 1 image, what you will do? Padding the rest 63 images?

guihao-liang: Will this padding affect the batch norm?

shreyajain17: There is no batch norm in drawing classifier.

jakesabathia2: Very good observe!
I also wondering why in our code we fix the batch size,
for instance, in AC we are doing the same thing and there are batch norm layers in the model.
And we might also change the model structure to something else for DC in the future (don't know this would happen tho haha), which might also include some cross-samples operation :).
But for now in the interest of consistent it makes sense to fix the batch size I guess haha.

guihao-liang: @shreyajain17 I mean for other toolkits with batch norms. I think this skewed batch norm will affect the final results.

guihao-liang:
> es the friction/overhead of reviewing!

That's right! To alleviate the burden, I think we can prioritize #2312 and #2943.

I'm glad that _black_ helps @shreyajain17 on formatting!

PR #2932, opened 2020-01-15
Dc evaluate perf

### LINKED ISSUE #2803: Drawing Classifier Evaluation is 20% slower.
Datasets shared internally.

Times recorded below in seconds.

5.8: 62.4904
6.0: 66.92402

5.8: 58.59062 
6.0: 69.92545

5.8: 56.76585
6.0: 70.56921

5.8: 57.45789
6.0: 71.88781

jakesabathia2:
fix #2803

### DIFF
--- a/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -1214,8 +1214,12 @@ class DrawingClassifier(_Model):
         evaluation_result = self.__proxy__.evaluate(dataset, metric)
 
         # TODO: fix the three passes through the data.
-        class_label = self.__proxy__.predict(dataset, "class")
-        probability_vector = self.__proxy__.predict(dataset, "probability_vector")
+
+        class_label = evaluation_result["prediction_class"]
+        probability_vector = evaluation_result["prediction_prob"]
+
+        del evaluation_result["prediction_class"]
+        del evaluation_result["prediction_prob"]
 
         predicted  = _tc.SFrame({"label": class_label, "probability": probability_vector})
         labels = self.classes
--- a/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
+++ b/src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
@@ -1213,8 +1213,6 @@ class DrawingClassifier(_Model):
 
         evaluation_result = self.__proxy__.evaluate(dataset, metric)
 
-        # TODO: fix the three passes through the data.
-
         class_label = evaluation_result["prediction_class"]
         probability_vector = evaluation_result["prediction_prob"]
 
--- a/src/toolkits/drawing_classifier/drawing_classifier.cpp
+++ b/src/toolkits/drawing_classifier/drawing_classifier.cpp
@@ -933,11 +933,27 @@ gl_sframe drawing_classifier::predict_topk(gl_sframe data,
 
 variant_map_type drawing_classifier::evaluate(gl_sframe data,
                                               std::string metric) {
-  gl_sarray predictions = predict(data, "probability_vector");
 
-  return evaluation::compute_classifier_metrics(
-      data, read_state<flex_string>("target"), metric, predictions,
+  gl_sarray predictions_prob = predict(data, "probability_vector");
+
+  flex_list class_labels = read_state<flex_list>("classes");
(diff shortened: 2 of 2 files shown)


src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -933,11 +933,27 @@ gl_sframe drawing_classifier::predict_topk(gl_sframe data,
...
-  gl_sarray predictions = predict(data, "probability_vector");
 
-  return evaluation::compute_classifier_metrics(
-      data, read_state<flex_string>("target"), metric, predictions,
+  gl_sarray predictions_prob = predict(data, "probability_vector");
+
+  flex_list class_labels = read_state<flex_list>("classes");
+    auto max_prob_label = [=](const flexible_type& ft) {
+      const flex_vec& prob_vec = ft.get<flex_vec>();
+      auto max_it = std::max_element(prob_vec.begin(), prob_vec.end());
```
TobyRoseman: Is the use of `auto` here increasing readability?

jakesabathia2: Yes. And also follow other code's style.

jakesabathia2:
pass internal test.

src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -933,11 +933,27 @@ gl_sframe drawing_classifier::predict_topk(gl_sframe data,
...
 variant_map_type drawing_classifier::evaluate(gl_sframe data,
                                               std::string metric) {
-  gl_sarray predictions = predict(data, "probability_vector");
 
-  return evaluation::compute_classifier_metrics(
-      data, read_state<flex_string>("target"), metric, predictions,
+  gl_sarray predictions_prob = predict(data, "probability_vector");
+
+  flex_list class_labels = read_state<flex_list>("classes");
+    auto max_prob_label = [=](const flexible_type& ft) {
```
guihao-liang: The indentation level is not correct. Run clang-format if possible.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -1213,9 +1213,11 @@ def evaluate(self, dataset, metric='auto'):
 
         evaluation_result = self.__proxy__.evaluate(dataset, metric)
 
-        # TODO: fix the three passes through the data.
-        class_label = self.__proxy__.predict(dataset, "class")
-        probability_vector = self.__proxy__.predict(dataset, "probability_vector")
+        class_label = evaluation_result["prediction_class"]
+        probability_vector = evaluation_result["prediction_prob"]
+
+        del evaluation_result["prediction_class"]
```
guihao-liang: you don't have to delete those. They will be collected by GC.

guihao-liang:
Sorry for this late review. @jakesabathia2, I think it's better to discover why predict class is slower than before.

jakesabathia2:
> Sorry for this late review. @jakesabathia2, I think it's better to discover why predict class is slower than before.

Predict is faster tam 5.8 tbh.
But we run here three times,
hays why it is slower (but fixed now hah.)

guihao-liang:
Duplicate the logic inside is not an optimal design. Besides, be careful about the code style.

jakesabathia2:
> Duplicate the logic inside is not an optimal design. Besides, be careful about the code style.

I am not duplicating that logic. The duplication in previous code is what I fixed in this PR.

guihao-liang:
> > Duplicate the logic inside is not an optimal design. Besides, be careful about the code style.
> 
> I am not duplicating that logic. The duplication in previous code is what I fixed in this PR.

That's what I mean. A better way to do it is to have a subroutine (inline function, for example) to wrap the same logic so that you will have only one place to edit in the future if there's a need to modify that logic.

jakesabathia2:
> > > Duplicate the logic inside is not an optimal design. Besides, be careful about the code style.
> > 
> > 
> > I am not duplicating that logic. The duplication in previous code is what I fixed in this PR.
> 
> That's what I mean. A better way to do it is to have a subroutine (inline function, for example) to wrap the same logic so that you will have only one place to edit in the future if there's a need to modify that logic.

good!

guihao-liang:
To put it in a simple way, make it more maintainable is also a key point for good code.

jakesabathia2:
> To put it in a simple way, make it more maintainable is also a key point for good code.

good! will put an other pr.

PR #2937, opened 2020-01-16
Resolves TF perf regression

### LINKED ISSUE #2798: Style Transfer inference is 4x slower in master versus in 5.8.

guihao-liang:
## TL;DR
The perf regression is caused by that the TF graph is initialized (adapted) for each image input during stylization (predict). The change will make the TF graph only initialized once during a prediction.

The way to do it is to define the image H and W as the dynamic dimension value instead of static.

resolves #2798 on the CPU path.

---

## review suggestion
When you review the code, please ignore src/python/turicreate/test/test_style_transfer.py because it mainly has formatting changes.

---

## report

load the same weights and predict on the same image.

```
In [9]: print(content_sf)
+------------------------+
|  content_feature_name  |
+------------------------+
| Height: 403 Width: 311 |
+------------------------+
```

### conclusion

* compared with 6.0: from 800ms -> 500ms.
* compared with 5.8:
  * on single image predict, 800ms -> 1290ms, expected slow down since initializing the graph will take 400ms.
  * batch stylizing images: 750ms -> 540 + (**880** / `number_of_images * number_of_styles`)ms, the initialization cost is amortized by subsequent calls to the `stylize`. **880** is the estimated data preparation (440ms) and graph initialization (440ms) overhead, which is pretty accurate, IMO, for my machine. We can do better to see how much we can reduce the data preparation overhead at the CPP implementation level.

---

### 5.8 result

since everything runs in the python layer, the model state is reserved and no extra initialization is needed,

```
In [6]: %timeit img = model.stylize(content_sf["content_feature_name"], style=0)
751 ms ± 8.32 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This profile is pretty accurate for `stylize` since there is **no** overhead of neural net architecture initialization.

---

### 6.0 result

if I run predict on singe image with a single style, the overhead of graph construction (880 ms) is not amortized,

```
+------------------+--------------+------------------+
| Images Processed | Elapsed Time | Percent Complete |
+------------------+--------------+------------------+
| 1                | 539.738ms    | 100.00%          |
+------------------+--------------+------------------+
1.29 s ± 20.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

But if we run in a batch manner with data replicated 127 times,

```
In [9]: print(content_sf)
+------------------------+
|  content_feature_name  |
+------------------------+
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
| Height: 403 Width: 311 |
+------------------------+
[128 rows x 1 columns]
```

and then feed it into predict (stylization) compuation flow,

```
In [8]: %timeit img = model.stylize(content_sf["content_feature_name"], style=0)
+------------------+--------------+------------------+
| Images Processed | Elapsed Time | Percent Complete |
+------------------+--------------+------------------+
| 1                | 577.957ms    | 0.78%            |
| 2                | 1.04s        | 1.56%            |
| 3                | 1.48s        | 2.34%            |
| 4                | 1.93s        | 3.12%            |
| 5                | 2.38s        | 3.91%            |
| 10               | 4.69s        | 7.81%            |
| 20               | 9.97s        | 15.62%           |
| 30               | 14.91s       | 23.44%           |
| 40               | 20.05s       | 31.25%           |
| 50               | 24.93s       | 39.06%           |
| 60               | 29.94s       | 46.88%           |
| 70               | 35.06s       | 54.69%           |
| 80               | 40.54s       | 62.50%           |
| 90               | 45.77s       | 70.31%           |
| 100              | 50.89s       | 78.12%           |
| 110              | 55.98s       | 85.94%           |
| 120              | 1m 1s        | 93.75%           |
+------------------+--------------+------------------+
1min 10s ± 7.07 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

The performance is averaged to ` 546.875` ms for each image input, which is approximate 540 + 880 / 128.


src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
```diff
@@ -63,30 +76,30 @@ def define_instance_norm(tf_input, tf_index, weights, prefix):
...
     batch_norm_array = []
     for g, b, i in zip(gamma_split, beta_split, input_split):
-        inputs_shape = i.shape
+        # inputs_shape = i.shape
         inputs_rank = i.shape.ndims
-        
+
         reduction_axis = inputs_rank - 1
-        params_shape = inputs_shape[reduction_axis:reduction_axis + 1]
+        # params_shape = inputs_shape[reduction_axis:reduction_axis + 1]
```
guihao-liang: unused parameter, comment it.

src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
```diff
@@ -467,27 +562,29 @@ def __init__(self, config, net_params):
...
         self._finetune_all_params = True
-        self._define_training_graph = bool(config['st_training'])
+        self._define_training_graph = bool(config["st_training"])
         self._tf_variables = define_tensorflow_variables(net_params)
 
         # TODO: take care of batch size
-        self.tf_input = _tf.placeholder(dtype = _tf.float32, shape = [None, 256, 256, 3])
-        self.tf_style = _tf.placeholder(dtype = _tf.float32, shape = [None, 256, 256, 3])
-        self.tf_index = _tf.placeholder(dtype = _tf.int64, shape = [self.batch_size])
+        self.tf_input = _tf.placeholder(dtype=_tf.float32, shape=[None, None, None, 3])
```
guihao-liang: make the input H, W dynamic

src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
```diff
@@ -168,73 +184,130 @@ def define_resnet(tf_input,
...
+
+    decode_1_inst_1 = define_instance_norm(
+        decode_1_conv_1, tf_index, weights, prefix + "decoding_1_inst_"
+    )
+
     decode_1_relu_1 = _tf.nn.relu(decode_1_inst_1)
 
     # decode 2
-    decode_2_image_shape = decode_1_relu_1.get_shape()
+    decode_2_image_shape = _tf.shape(decode_1_relu_1)
```
guihao-liang: accommodate the dynamic shapes.

guihao-liang:
When you review the code, please ignore `src/python/turicreate/test/test_style_transfer.py` because it mainly has formatting changes.

src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
TobyRoseman: Why not just delete it?

guihao-liang: I want to, but I don't know if it's needed here.

TobyRoseman:
At a high level, what was the change to improve perf?

guihao-liang:
> At a high level, what was the change to improve perf?

I updated my PR description. Feel free to check it!

src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
jakesabathia2: good!

guihao-liang:
Per @nickjong's request, provide the profile comparison between 6.0 and 5.8 on the CPU path.

guihao-liang:
Passed internal pipeline at hash 15e5865. The following 2 commits are trivial.

PR #2938, opened 2020-01-16
Adds a distance field in the Nearest Neighbors summary struct.

### LINKED ISSUE #2460: Nearest Neighbors Log
`model = tc.nearest_neighbors.create(references)` with an SFrame of all strings outputs `Starting brute force nearest neighbors model training.` with no more logs to follow it. Although I think the model still gets created properly, or at least I hope. But maybe just the log does not reflect the progress and completion.

shantanuchhabra:
Fixes the second bullet point of https://github.com/apple/turicreate/issues/2460#issuecomment-543303328

Fixes #2460 .


src/toolkits/nearest_neighbors/nearest_neighbors.cpp
```diff
@@ -318,6 +318,25 @@ void nearest_neighbors_model::train(const sframe& X,
...
+      {"distance_for_summary_struct", "composite"}
+    });
+  } else if (composite_distance_params.size() == 1 
+      && std::get<0>(composite_distance_params[0]).size() > 1) {
+    add_or_update_state({
+      {"distance_for_summary_struct", "composite"}
+    });
+  } else {
+    function_closure_info distance_fn = std::get<1>(composite_params[0]);
+    auto dist_name = extract_distance_function_name(distance_fn);
```
TobyRoseman: I think should be `std::string` rather than `auto`.

src/toolkits/nearest_neighbors/nearest_neighbors.cpp
```diff
@@ -318,6 +318,25 @@ void nearest_neighbors_model::train(const sframe& X,
   train(X, ref_labels, composite_distance_params, opts);
 }
 
+void nearest_neighbors_model::populate_distance_for_summary_struct(
+    const std::vector<dist_component_type>& composite_distance_params) {
+  if (composite_distance_params.size() > 1) {
```
TobyRoseman: I'm struggling to understand what each of these three cases are for. I think a few short comments would really help.

src/toolkits/nearest_neighbors/nearest_neighbors.cpp
shantanuchhabra: Done.

src/toolkits/nearest_neighbors/nearest_neighbors.cpp
shantanuchhabra: Done.

PR #2939, opened 2020-01-16
Auto generate all documentations

### LINKED ISSUE #284: Build Python User Guide and C++ API docs from make_wheel.sh
Currently, [make_wheel.sh](https://github.com/apple/turicreate/blob/master/scripts/make_wheel.sh) only builds Python API docs and not the Python User Guide or C++ API docs. (The User Guide and C++ API docs are built manually by running `npm run build` and `doxygen` respectively). We should add both to this script so we can collect them as build artifacts at the same time as the Python API docs.

TobyRoseman:
- Remove documentation building from make_wheel.sh
- Create standalone script to make documentation
- Creaet standalone GitLab job

Fixes #284

### DIFF
--- a/scripts/make_documentation.sh
+++ b/scripts/make_documentation.sh
@@ -8,6 +8,10 @@ WORKSPACE=${SCRIPT_DIR}/..
 TARGET_DIR=${WORKSPACE}/target
 
 cd ${WORKSPACE}
+
+${VIRTUALENV} ./deps/env
+source deps/env/bin/activate
+
 pip install target/turicreate-*.whl
 
 OUTPUT_DIR=${TARGET_DIR}/docs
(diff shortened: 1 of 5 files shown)


scripts/make_documentation.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+set -e
+set -x
+
+SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
guihao-liang: space around parenthesis is not necessary. User `pwd -P` to get the physical location to avoid chained symlink, i.e., a -> b -> c and `c` is where source scripts are stored.

TobyRoseman:
Internal tests passed.

PR #2947, opened 2020-01-22
Dc eval revise

jakesabathia2:
Code format fixed and clean up.
Follow-up from https://github.com/apple/turicreate/pull/2932


src/toolkits/drawing_classifier/drawing_classifier.cpp
```diff
@@ -51,6 +51,18 @@ struct result {
   data_iterator::batch data_info;
 };
 
+gl_sarray get_predictions_class(gl_sarray predictions_prob,
```
guihao-liang: do you want to inline it? or make it a private/protected member function? Default static linkage is not recommended.

jakesabathia2: Oh okay I'll make it private.

guihao-liang: I think protected is better :-)

guihao-liang: let's do protected inline. Just put the whole definition to the header file!

jakesabathia2: good!

jakesabathia2:
pass internal test.

PR #2953, opened 2020-01-23
Added type field in CoreML model metadata

### LINKED ISSUE #2800: "Machine Learning Essentials" toolkits need model metadata
[These toolkits](https://apple.github.io/turicreate/docs/userguide/modeling-data/) need metadata toolkit type associated with the exported Core ML model (Boosted Trees, Logistic Classifier, etc.). Related: https://github.com/apple/turicreate/issues/2794.

For example, the Style Transfer toolkit exposes the `type` key as part of the Creator Defined Core ML keys:

```
creatorDefinedKey: {
    "com.github.apple.os.platform" = "Darwin-19.0.0-x86_64-i386-64bit";
    "com.github.apple.turicreate.version" = "6.0rc1";
    "content_feature" = image;
    "max_iterations" = 2;
    model = "resnet-16";
    "num_styles" = 1;
    "style_feature" = image;
    "training_iterations" = 2;
    type = "style_transfer";
    version = 1;
}
```

fareeha:
Exposed the `type` field in exported CoreML models from [toolkits](https://apple.github.io/turicreate/docs/userguide/modeling-data/) it was missing in.

Closes #2800


src/python/turicreate/toolkits/_tree_model_mixin.py
```diff
@@ -317,6 +317,7 @@ def _get_summary_struct(self):
 
     def _export_coreml_impl(self, filename, context):
         tc_version_info = _coreml_utils._get_tc_version_info()
+        tc_version_info['type'] = context['class']
         if 'user_defined' not in context:
```
srikris: How are we testing this?

fareeha: I'm thinking of how to do that, specifically how to obtain the class name for asserting the `type` in https://github.com/apple/turicreate/blob/46d5a4de4388dd72d445dfe69fa6064a8e5cba1d/src/python/turicreate/test/test_coreml_export.py#L71 since `_test_coreml_export()` is run for all classifiers and regressors.

fareeha: It works on manual testing obviously but we'd still want to update the testing suite.

fareeha: Added `type` field check in test.

nickjong: It looks like `_coreml_utils.py` already provides a function that adds the toolkit type to this dictionary:
https://github.com/apple/turicreate/blob/b12f67fb523b26cbbff263f6260d0cf2674d8647/src/python/turicreate/toolkits/_coreml_utils.py#L27
Could we be using that instead of the direct calls to `_get_tc_version_info`?
This is in fact what the style transfer toolkit (the motivating example from the bug) uses:

fareeha: The reason I didn't use `_get_model_metadata` is because we don't have `metadata` given in `_export_coreml_impl`. 

I'm wondering if there is a way to get `metadata` from `self.__proxy__` in python as `_xgboost_export_as_model_asset(...)` does in:

https://github.com/apple/turicreate/blob/9815427e8dba58e3ead0807552238644a73b68c4/src/toolkits/coreml_export/coreml_extension.cpp#L33

nickjong: It looks like it's okay for `metadata` to be `None`

fareeha: Oh that's right! Would it be ok if I also changed:

`def _get_model_metadata(model_class, metadata, version=None):
`

to

`def _get_model_metadata(model_class, metadata=None, version=None):
`

for better readability?

nickjong: Seems reasonable to me!

nickjong:
Could we rebase your changes on top of Guihao's formatting pass? (Note the merge conflicts that GitHub identified...)

fareeha:
> Could we rebase your changes on top of Guihao's formatting pass? (Note the merge conflicts that GitHub identified...)

Yes on it

PR #2943, opened 2020-01-18
make python code pep8 compliant

guihao-liang:
## Upstream dependencies

#2826, which removes redundant MXnet **Python** code after major 6.0 release. 

---

## Goal, goal, goal, ale, ale, ale

partially resolve #2312 

1. make our current python codebase fully pep8 compatible.
2. provide GitHub hook to check pep8 compatibility for future commits.
3. build the team culture for the pep8 coding style for our python codebase.

In order to fully resolve #2312, we need to run a `linter`, which can be only done incrementally. I will cover this later.

---

## Methodology

No magic, but [black][0] magic.

I'm pretty confident about the tool [black][0] and its reformatting result because of,

> Also, as a temporary safety measure, Black will check that the reformatted code still produces a valid AST that is equivalent to the original. This slows it down. If you're feeling confident, use --fast.

---

## Step by step

- [x] make source code pep8 compliant, **excluding docstring**.

[black][0] the `python` dir first to make sure the `test` directory is left intact. If all tests are passed, I'm confident to move to the next step.

current `tox.ini` setting,

```bash
[flake8]
max-line-length = 88
```

Just let me know what errors you think should be ignored by the linter and black.

```
All done! ✨ 🍰 ✨
281 files reformatted, 17 files left unchanged.
```

---

- [x] make test code pep8

[0]: https://github.com/psf/black
[1]: https://www.python.org/dev/peps/pep-0008/

PR #2956, opened 2020-01-24
add pre-commit and black support

guihao-liang:
downstream of #2943 

```bash
(tc58) ➜  gui-1-36 git:(01-24-pre-commit) ✗ git commit -m "add pre-commit and black support"
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...............................................................Passed
Check for added large files..............................................Passed
black................................................(no files to check)Skipped
[01-24-pre-commit c94fdd374] add pre-commit and black support
 3 files changed, 50 insertions(+)
 create mode 100644 .pre-commit-config.yaml
 create mode 100644 pyproject.toml
```
When I do this commit, it performs a bunch of correctness checks for me. The setup is one-time thing.


guihao-liang:
```
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...........................................(no files to check)Skipped
Check for added large files..............................................Passed
black................................................(no files to check)Skipped
[01-24-pre-commit da42b5367] install pre-commit hooks when ./configure
 2 files changed, 4 insertions(+), 7 deletions(-)
```
You will love this tool.

znation:
Does it block committing if a check fails? Is it easy to opt out of this behavior (either in general, or for a specific commit)?

guihao-liang:
> Does it block committing if a check fails? Is it easy to opt out of this behavior (either in general, or for a specific commit)?

If the check fails, it won't let you commit. You can skip the check by using 

```
# SKIP=<check_id> git commit -m "blah"
SKIP=check_yaml git commit -m "blah" # will force to skip yaml check
```
Yes! It's easy to opt-out. I can add script to opt-out. But for us, I think it's better everybody opts in.

guihao-liang:
@znation @TobyRoseman, I provide an option for the user to opt-out. Just let me know if you have more questions.

guihao-liang:
Adding `per-commit` introduces a failure in `pytest`, #2973, which is resolved by #2973. Ready to mege.

guihao-liang:
Passed all relevant tests. Ready to merge hand in hand with #2943.

pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+# https://github.com/psf/black/blob/master/pyproject.toml
+
+# NOTE: you have to use single-quoted strings in TOML for regular expressions.
+# It's the equivalent of r-strings in Python.  Multiline strings are treated as
+# verbose regular expressions by Black.  Use [ ] to denote a significant space
+# character.
+
+[tool.black]
+line-length = 88
+target-version = ['py27', 'py35', 'py36', 'py37', 'py38']
```
TobyRoseman: We don't support Python 3.8. So you probably don't want that here.

scripts/install_python_toolchain.sh
```diff
@@ -35,24 +34,32 @@ function linux_patch_sigfpe_handler {
   fi
 }
 
-$PIP install --upgrade "pip>=8.1"
+$PIP install --upgrade "pip"
```
TobyRoseman: Is there a reason you removed the `>=8.1`?

guihao-liang: I think pip 8.1 is really old now. We don't have to hardcode it. The current pip is already 20.0.2. `>=8.1` is redundant.

pyproject.toml
guihao-liang: That's right! I will delete it.

scripts/install_python_toolchain.sh
TobyRoseman: ok, you have the `--upgrade` so that should work.

PR #2957, opened 2020-01-24
Fixes break on loading OSOD 

### LINKED ISSUE #2954: Unable to do prediction on OneshotDetection
When i save the model and load it back i get 
turicreate.toolkits._main.ToolkitError: No neural network compute context provided

furthermore when i tried to print back the value before saving the model i am getting [] 

i am trying to run the example code of stop sign , still fails . hope someone can help me

shreyajain17:
fixes #2954

### DIFF
--- a/src/python/turicreate/toolkits/_model.py
+++ b/src/python/turicreate/toolkits/_model.py
@@ -132,6 +132,7 @@ def load_model(location):
                     return cls(model)
 
                 if name=='one_shot_object_detector' and OD_USE_CPP:
+                    import turicreate.toolkits.libtctensorflow
                     od_cls = MODEL_NAME_MAP['object_detector']
                     if 'detector_model' in model_data['detector']:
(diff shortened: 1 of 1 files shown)


src/python/turicreate/toolkits/_model.py
```diff
@@ -132,6 +132,7 @@ def load_model(location):
                     return cls(model)
 
                 if name=='one_shot_object_detector' and OD_USE_CPP:
+                    import turicreate.toolkits.libtctensorflow
```
jakesabathia2: Just curious why does this solve the issue reported?

shreyajain17: It solves the first part actually, which the break that reports no compute context. The TensorFlow compute context which was designed using PyBind needs to be linked at runtime. Without this import, the shared library is not linked and the compute context will not be found.

PR #2955, opened 2020-01-24
[OD] 6.0 inference regression between CPU and GPU.

jakesabathia2:
close #2865 

6.0's CPU and GPU has prediction regression.
For the same image,
GPU:
```
#predictions
{'confidence': 0.6841502785682678, 'type': 'rectangle', 'coordinates': {'y': 167.98015236854553, 'x': 289.0607535839081, 'height': 238.21115112304688, 'width': 301.1848449707031}, 'label': 'Croissant'}
{'confidence': 0.5636424422264099, 'type': 'rectangle', 'coordinates': {'y': 168.73490810394287, 'x': 351.99418142437935, 'height': 183.61944580078125, 'width': 161.6607208251953}, 'label': 'Croissant'}
{'confidence': 0.5182375311851501, 'type': 'rectangle', 'coordinates': {'y': 101.34550631046295, 'x': 123.91261160373688, 'height': 176.6996612548828, 'width': 241.56607055664062}, 'label': 'Croissant'}
{'confidence': 0.4028134346008301, 'type': 'rectangle', 'coordinates': {'y': 210.60317158699036, 'x': 199.7049316763878, 'height': 175.12136840820312, 'width': 237.15415954589844}, 'label': 'Croissant'}
{'confidence': 0.1619817614555359, 'type': 'rectangle', 'coordinates': {'y': 106.30784332752228, 'x': 233.0048382282257, 'height': 219.98683166503906, 'width': 288.35015869140625}, 'label': 'Croissant'}
{'confidence': 0.10118640214204788, 'type': 'rectangle', 'coordinates': {'y': 213.20754289627075, 'x': 361.471713334322, 'height': 101.99461364746094, 'width': 127.2068862915039}, 'label': 'Croissant'}
{'confidence': 0.027622569352388382, 'type': 'rectangle', 'coordinates': {'y': 65.32773971557617, 'x': 268.4330843389034, 'height': 119.54254913330078, 'width': 164.8693084716797}, 'label': 'Croissant'}
{'confidence': 0.010357137769460678, 'type': 'rectangle', 'coordinates': {'y': 84.39711928367615, 'x': 118.78975331783295, 'height': 107.74749755859375, 'width': 172.99261474609375}, 'label': 'Croissant'}
{'confidence': 0.00467892037704587, 'type': 'rectangle', 'coordinates': {'y': 90.58223068714142, 'x': 315.67183434963226, 'height': 175.4208984375, 'width': 256.6126708984375}, 'label': 'Croissant'}

[...]

PR #2958, opened 2020-01-24
[WIP] Pre-commit hooks for cpp files

guihao-liang:
Downstream of #2955, #2965

To achieve Cpp pre-commit, I  need to install external dependencies, which is kind of troublesome since those dependencies I have are distributed as binaries. I can rely on the local system to install them. So I need further discussion here.

Otherwise, I should make this pre-commit check optional so that developers can opt-in and opt-out.


guihao-liang:
Feel free to grab it [here](https://github.com/guihao-liang/pre-commit-cpp). Close it due to inactivity.

Issue #2934, opened 2020-01-16
Flakey Unit Test: test_json.py - test_arbitrary_json


TobyRoseman:
This [unit test](https://github.com/apple/turicreate/blob/d8aa46a5e79966029796b22ab52616dccfbc2e64/src/python/turicreate/test/test_json.py#L360) should be rewritten or removed. 

Failure example 1:
```
_________________________ JSONTest.test_arbitrary_json _________________________

self = <turicreate.test.test_json.JSONTest testMethod=test_arbitrary_json>

    @hypothesis.settings(derandomize=True, suppress_health_check=[hypothesis.HealthCheck.too_slow])
>   @hypothesis.given(hypothesis_json)
    def test_arbitrary_json(self, json_obj):

test_json.py:359: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../hypothesis/core.py:599: in execute
    % (test.__name__, text_repr[0])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <hypothesis.core.StateForActualGivenExecution object at 0x13401e710>
message = "Hypothesis test_arbitrary_json(self=<turicreate.test.test_json.JSONTest testMethod=test_arbitrary_json>, json_obj=[{'t': 0}]) produces unreliable results: Falsified on the first call but did not on a subsequent one"

    def __flaky(self, message):
        if len(self.falsifying_examples) <= 1:
>           raise Flaky(message)
E           hypothesis.errors.Flaky: Hypothesis test_arbitrary_json(self=<turicreate.test.test_json.JSONTest testMethod=test_arbitrary_json>, json_obj=[{'t': 0}]) produces unreliable results: Falsified on the first call but did not on a subsequent one

../../hypothesis/core.py:769: Flaky
```

Example 2:
```
Falsifying example: test_arbitrary_json(self=<turicreate.test.test_json.JSONTest testMethod=test_arbitrary_json>, json_obj=[{'Y': [{'2AB': -0.0,
    '7h': 13,
    'A': '',
    'CkEm': 0.5,
    'EUUY': -2.220446049250313e-16,
    'JBbEZ': '',
    'LgZcU': '',
    'Y': '',
    'Zvgdo': '',
[...]

PR #2963, opened 2020-01-28
Disable flaky unit test for JSON parsing

nickjong:
The long-term change needed here is tracked by https://github.com/apple/turicreate/issues/2934


nickjong:
I'm honestly still in the process of verifying that putting the @pytest annotation above the @hypothesis annotation works as intended

PR #2978, opened 2020-02-04
make bulding python_source fast again

### LINKED ISSUE #2477: "make python_source" build is slow
Steps to reproduce:
```bash
cd debug/src/python/
make -j
touch ../../../src/python/turicreate/__init__.py 
make python_source
```
Despite there being only python changes, the `make python_source` seems to be doing just as much work as `make -j`.

`make python_source` should be very quick, just a couple of seconds. It should just copy the python files. On my machine the `make python_source` takes almost 30 seconds.

This only started to be an issue recently. It's likely related to the recent remotefs build changes.

TobyRoseman:
Fixes #2477 
Removes a dependency on a target which no longer exists (`numpy_integration`).
Also removes some very outdated comments.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -9,10 +9,6 @@ add_subdirectory(turicreate)
 # source is processed directly.  To handle this, we reproduce the availability
 # of the source files in the build tree.
 
-# Copy all the open source python stuff over first
-
-# Overwrite with the proprietary stuff
-
 make_copy_target(python_source
   FILES
     ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE.txt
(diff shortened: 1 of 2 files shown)


src/python/CMakeLists.txt
```diff
@@ -24,9 +20,8 @@ make_copy_target(python_source
   DIRECTORIES
     ${CMAKE_CURRENT_SOURCE_DIR}/turicreate
 )
-add_dependencies(python_source numpy_integration tctensorflow)
```
guihao-liang: is this `numpy_integration` making building python_source slow?

TobyRoseman: No. See my description. That target doesn’t exist any more.

tctensorflow is what’s making it slow.

guihao-liang: Ah, I see. I was wondering why `make` would block on a non-existent object.

PR #2982, opened 2020-02-04
Adding option to model type in OD

shreyajain17:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -281,6 +281,13 @@ void object_detector::init_options(
       "If True, print progress updates and model details.",
       true,
       true);
+  options.create_categorical_option(
+     /* name              */ "model",
+      /* description       */
+      "Defines the model type",
```
TobyRoseman: FYI - Looks like there are some indentation issues here. I also don't think you want a new line here.

shreyajain17: Yeah. I just did it.

shreyajain17: I just did it.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -281,6 +281,14 @@ void object_detector::init_options(
       "If True, print progress updates and model details.",
       true,
       true);
+  options.create_categorical_option(
+      /* name              */ "model",
+      /* description       */
+      "Defines the model type",
+      /* default_value     */ "darknet-yolo",
+      /* allowed_values    */
+      {flexible_type("darknet-yolo"), flexible_type("other")},
```
nickjong: If we don't know yet what the other categorical values will be, for now we can just make this an arbitrary string value

shreyajain17: Done

PR #2983, opened 2020-02-04
Changed test to be far less intensive.

hoytak:
Still tests the functionality, but uses a much smaller dataset.


TobyRoseman:
@hoytak - looks like most of the changes here are stylistic. Are these done by [black](https://github.com/psf/black), our new code formatter?

hoytak:
PR updated.

nickjong:
Do we still want/need this PR?

hoytak:
I assume we do -- it's just making the test more of a unit test.  I wasn't following this work to be honest, though.

TobyRoseman:
I agree, we should still merge this change. It should make running our unit tests a few minutes quicker.

PR #2984, opened 2020-02-05
Adding ac_parameters struct

shreyajain17:


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -311,20 +311,22 @@ std::unique_ptr<model_backend> tf_compute_context::create_object_detector(
 }
 
 std::unique_ptr<model_backend> tf_compute_context::create_activity_classifier(
-    int n, int c_in, int h_in, int w_in, int c_out, int h_out, int w_out,
-    const float_array_map& config, const float_array_map& weights) {
-  shared_float_array prediction_window = config.at("ac_pred_window");
-  const float* pred_window = prediction_window.data();
-  int pw = static_cast<int>(*pred_window);
+    ac_parameters ac_params) {
+  // shared_float_array prediction_window = config.at("ac_pred_window");
```
TobyRoseman: Should we just delete this code rather than comment it out?

shreyajain17: Yeah I should do that. I also should add comments I think. I will change it.

src/ml/neural_net/compute_context.hpp
```diff
@@ -16,6 +16,19 @@
...
+/** A struct to define all the parameters used to create the activity classifier
+ * model backend
+ */
+struct ac_parameters {
+  int batch_size;
+  int num_features;
+  int prediction_window;
+  int num_classes;
+  int num_predictions_per_chunk;
+  int random_seed;
```
nickjong: For these let's specify reasonable default values, as in `int batch_size = 32;` or whatever the actual default is.

(And add documentation)

src/ml/neural_net/compute_context.hpp
```diff
@@ -16,6 +16,19 @@
...
+  float mode;
```
nickjong: As discussed offline, would be nice to make this a `bool` instead, maybe `is_training`?

src/ml/neural_net/compute_context.hpp
```diff
@@ -16,6 +16,19 @@
...
+  float_array_map weights;
```
nickjong: Someday, maybe not today, we should actually define what keys and shapes we expect. But maybe this is something we need to give more thought about, as we look ahead to scaling to support other architectures.

src/ml/neural_net/compute_context.hpp
```diff
@@ -102,9 +115,9 @@ EXPORT class compute_context {
    *       weights as a float_array_map.
    */
   virtual std::unique_ptr<model_backend> create_activity_classifier(
-      int n, int c_in, int h_in, int w_in, int c_out, int h_out, int w_out,
-      const float_array_map& config, const float_array_map& weights) { ASSERT_TRUE(false); }
-
+      ac_parameters ac_params) {
```
nickjong: Let's pass by `const` reference, to avoid an unnecessary copy

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -51,32 +52,21 @@ constexpr size_t LSTM_HIDDEN_SIZE = 200;
...
+  ac_parameters ac_params;
+  ac_params.batch_size = batch_size;
+  ac_params.num_features = num_features;
+  ac_params.prediction_window = prediction_window;
+  ac_params.num_classes = num_classes;
+  ac_params.num_predictions_per_chunk = num_predictions_per_chunk;
+  ac_params.random_seed = random_seed;
+  ac_params.mode = mode;
+  ac_params.weights = weights;
+  return ac_params;
```
nickjong: I'm not sure this function adds a lot of value. Might actually be clearer at the call site just to instantiate a `ac_parameters` and set all the parameters explicitly.

src/ml/neural_net/compute_context.hpp
shreyajain17: The default values come from `init_options`. For example, `batch_size` is already 32 in it.

src/ml/neural_net/compute_context.hpp
shreyajain17: Okay

Issue #2936, opened 2020-01-16
OSOD non-deterministic unit test failure


shantanuchhabra:
There is a unit test failure (`test_create_empty_dataset`) in OSOD that fails non-deterministically with two different errors:

On macOS 10.14 with Python 2.7:
```
__________ OneObjectDetectorSmokeTest.test_create_with_empty_dataset ___________

self = <turicreate.test.test_one_shot_object_detector.OneObjectDetectorSmokeTest testMethod=test_create_with_empty_dataset>

    def test_create_with_empty_dataset(self):
        with self.assertRaises(_ToolkitError):
>           tc.one_shot_object_detector.create(self.train[:0], target=self.target)

test_one_shot_object_detector.py:138: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../toolkits/one_shot_object_detector/one_shot_object_detector.py:64: in create
    data, target, backgrounds)
../toolkits/one_shot_object_detector/util/_augmentation.py:54: in preview_synthetic_training_data
    backgrounds = backgrounds.apply(lambda im: _tc.image_analysis.resize(
../data_structures/sarray.py:1841: in apply
    dryrun = [fn(i) for i in self.head(100) if i is not None]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

im = '{'

    backgrounds = backgrounds.apply(lambda im: _tc.image_analysis.resize(
        im,
>       int(im.width/2),
        int(im.height/2),
        im.channels
        ))
E   AttributeError: 'str' object has no attribute 'width'

../toolkits/one_shot_object_detector/util/_augmentation.py:56: AttributeError
```

On macOS 10.14 with Python 3.7:
```
__________ OneObjectDetectorSmokeTest.test_create_with_empty_dataset ___________

self = <turicreate.test.test_one_shot_object_detector.OneObjectDetectorSmokeTest testMethod=test_create_with_empty_dataset>

    def test_create_with_empty_dataset(self):
        with self.assertRaises(_ToolkitError):
>           tc.one_shot_object_detector.create(self.train[:0], target=self.target)

test_one_shot_object_detector.py:138: 
[...]

PR #2985, opened 2020-02-05
Fix flakey 1SOD test

TobyRoseman:
Fixes: #2936

This is in many ways just an extension of #2901. There is [evidence](https://github.com/apple/turicreate/issues/2936#issuecomment-582182773) to suggest it's possible for an SArray to be incorrect extracted from a tar file and still be loadable.

This change checks the md5 of each extracted file.

This change also remove the xfail from the unit test and makes other minor changes.


TobyRoseman:
Internal tests passed

Issue #2874, opened 2019-12-21
TC 6.0 style transfer buggy and broken on Ubuntu cuda


romanoss:
Hi,
I tested 6.0 on my ubuntu machine with an internal RTX 2080Ti GPU and style transfer seems broken and useless.

Issues:
1) after installing in a new conda env, a tf version was installed, which disables GPU usage. Additionally installing tensorflow-gpu solved this

2) GPU memory is not freed after training

3) using batch_size param in training crashes the python kernel

4) advanced params dont work:
a) non def input_shape leads to tf error
b) print_loss_breakdown does not work
c) probably the other options dont work either

5) styling output produces similar results for different styles

6) stylized output looks bad and unusable and is absolutely not comparable to old version

7) wavy artefacts are worse than in last mx version

regards

rainhut:
Same issues here

romanoss:
Did you try it on a mac?

rainhut:
Yes, the new MacBook Pro 16 Inch with an AMD Radeon Pro 5500M.

Rob

abhishekpratapa:
@rainhut @romanoss We recently switched to another backend for style transfer and some of these got introduced. Currently investigating the bugs. Thanks for a detailed GitHub issue! Will help us track the issue down faster.

longleb777:
you can do nothing cause the server is a beautiful hacker her name is ruby

rainhut:
Any progress?

abhishekpratapa:
I’m actively working on it. I’ve reproduced the problem but haven’t identified the issue yet.

rainhut:
Thanks for the update

abhishekpratapa:
@rainhut @romanoss I've Identified the similar results issue to be in the conditional instance normalization layer. The gradient wasn't properly being calculated through the backwards pass even though the forward pass was correct. I've fixed it in this PR: #2986

The waviness was due to loading pre-trained weights (not present in the previous model) rather than using the proper initializer. I've fixed that in this PR: #2988

More updates to come!

romanoss:
thx for the info

abhishekpratapa:
@rainhut @romanoss Another issue that was identified was the updatable parameters, only the transformer weights are supposed to be updated, consequently it turns out that all of the parameters of the VGG network were set to be trainable as well causing additional issues. I've fixed that in this PR: #2990

abhishekpratapa:
[...]

PR #2988, opened 2020-02-05
Updated Initialization for Style Transfer

abhishekpratapa:
# Overview
Initialization for style transfer is a uniform between -0.7 and 0.7. Previously only pertained weights were used for this in TensorFlow.

# Fixes
- Weight initialization in style transfer to use uniform between -0.7 and 0.7
- Random Seed values for initialization

### DIFF
--- a/src/ml/neural_net/weight_init.hpp
+++ b/src/ml/neural_net/weight_init.hpp
@@ -55,8 +55,8 @@ class uniform_weight_initializer {
   /**
    * Creates a weight initializer that performs Uniform initialization
    *
-   * \param lower_bound The number of inputs that affect each output from the layer
-   * \param upper_bound The number of outputs affected by each input to the layer
+   * \param lower_bound The lower bound of the uniform distribution to be sampled
+   * \param upper_bound The upper bound of the uniform distribution to be sampled
    * \param random_engine The random number generator to use, which must remain
    *     valid for the lifetime of this instance.
    */
--- a/src/ml/neural_net/weight_init.hpp
+++ b/src/ml/neural_net/weight_init.hpp
@@ -48,6 +48,32 @@ private:
   std::mt19937& random_engine_;
 };
 
+
+class uniform_weight_initializer {
+  public:
+
+  /**
+   * Creates a weight initializer that performs Uniform initialization
+   *
+   * \param lower_bound The number of inputs that affect each output from the layer
+   * \param upper_bound The number of outputs affected by each input to the layer
+   * \param random_engine The random number generator to use, which must remain
+   *     valid for the lifetime of this instance.
+   */
+  uniform_weight_initializer(float lower_bound, float upper_bound,
+                            std::mt19937* random_engine);
+
+  /**
+   * Initializes each value in uniformly at random in the range [-lower_bound, upper_bound]
+   */
+  void operator()(float* first_weight, float* last_weight);
+
+private:
+
+  std::uniform_real_distribution<float> dist_;
+  std::mt19937& random_engine_;
+};
+
 struct scalar_weight_initializer {
   /**
    * Creates a weight initializer that initializes all of the weights to a
--- a/src/python/turicreate/toolkits/style_transfer/style_transfer.py
+++ b/src/python/turicreate/toolkits/style_transfer/style_transfer.py
@@ -247,6 +247,7 @@ def create(
     options["num_styles"] = len(style_dataset)
     options["resnet_mlmodel_path"] = pretrained_resnet_model.get_model_path("coreml")
     options["vgg_mlmodel_path"] = pretrained_vgg16_model.get_model_path("coreml")
+    options["pretrained_weights"] = params["pretrained_weights"]
 
     model.train(style_dataset[style_feature], content_dataset[content_feature], options)
     return StyleTransfer(model_proxy=model, name=name)
--- a/src/toolkits/style_transfer/style_transfer.cpp
+++ b/src/toolkits/style_transfer/style_transfer.cpp
@@ -686,6 +686,12 @@ void style_transfer::init_train(gl_sarray style, gl_sarray content,
   }
   size_t num_styles = num_styles_iter->second;
 
+  auto pretrained_weights_iter = opts.find("pretrained_weights");
+  bool pretrained_weights = false;
+  if (pretrained_weights_iter != opts.end()) {
+    pretrained_weights = pretrained_weights_iter->second;
+  }
+  
   init_options(opts);
 
   if (read_state<flexible_type>("random_seed") == FLEX_UNDEFINED) {
@@ -694,9 +700,11 @@ void style_transfer::init_train(gl_sarray style, gl_sarray content,
     add_or_update_state({{"random_seed", random_seed}});
   }
 
+  int random_seed = read_state<int>("random_seed");
+
   m_training_data_iterator =
       create_iterator(content, style, /* repeat */ true,
-                      /* training */ true, static_cast<int>(num_styles));
+                      /* training */ true, random_seed);
 
   m_training_compute_context = create_compute_context();
   if (m_training_compute_context == nullptr) {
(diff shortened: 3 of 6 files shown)


src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -20,14 +22,24 @@ using CoreML::Specification::NeuralNetworkLayer;
...
+constexpr float UPPER_BOUND = 0.7; 
+
 // TODO: refactor code to be more readable with loops
-void define_resnet(model_spec& nn_spec, size_t num_styles) {
+void define_resnet(model_spec& nn_spec, size_t num_styles, int random_seed=0) {
+  std::mt19937 random_engine;
+  std::seed_seq seed_seq{random_seed};
+  random_engine = std::mt19937(seed_seq);
+
+  auto initializer = xavier_weight_initializer(LOWER_BOUND, UPPER_BOUND, &random_engine);
```
TobyRoseman: I'm really confused by this line. First off, it would probably be best not to use `auto` here.

This is what you're calling, right? https://github.com/apple/turicreate/blob/fb1725cffa5cbd188e281626771cb8b6a53b5ae8/src/ml/neural_net/weight_init.cpp#L25

If so, that doesn't look right to me. The first two parameters seem wrong. fan in/out seems different than lower/upper bound. At the very least these are getting covered from float to size_t which does seem right since they aren't whole numbers.

src/toolkits/style_transfer/style_transfer.cpp
```diff
@@ -686,6 +686,12 @@ void style_transfer::init_train(gl_sarray style, gl_sarray content,
   }
   size_t num_styles = num_styles_iter->second;
 
+  auto pretrained_weights_iter = opts.find("pretrained_weights");
```
TobyRoseman: Is `auto` helping or hurting readability here?

src/toolkits/style_transfer/style_transfer_model_definition.cpp
abhishekpratapa: Good catch! I was testing it and the stylization didn't work, properly.

nickjong:
What is the relationship between this PR and #2874 ?

src/ml/neural_net/weight_init.hpp
```diff
@@ -48,6 +48,32 @@ class xavier_weight_initializer {
...
 };
 
+
+class uniform_weight_initializer {
+  public:
+
+  /**
+   * Creates a weight initializer that performs Uniform initialization
+   *
+   * \param lower_bound The number of inputs that affect each output from the layer
```
nickjong: This documentation does not appear to be correct

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -247,6 +247,7 @@ def create(
     options["num_styles"] = len(style_dataset)
     options["resnet_mlmodel_path"] = pretrained_resnet_model.get_model_path("coreml")
     options["vgg_mlmodel_path"] = pretrained_vgg16_model.get_model_path("coreml")
+    options["pretrained_weights"] = params["pretrained_weights"]
```
nickjong: Why is this needed?

src/toolkits/style_transfer/style_transfer.cpp
nickjong: I believe the alternative is
```
std::map<std::string, flexible_type>::iterator pretrained_weights_iter = opts.find("pretrained_weights");
```
So I would argue yes.

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -20,14 +22,24 @@ using CoreML::Specification::NeuralNetworkLayer;
...
+constexpr float UPPER_BOUND = 0.07; 
+
 // TODO: refactor code to be more readable with loops
-void define_resnet(model_spec& nn_spec, size_t num_styles) {
+void define_resnet(model_spec& nn_spec, size_t num_styles, int random_seed=0) {
+  std::mt19937 random_engine;
+  std::seed_seq seed_seq{random_seed};
+  random_engine = std::mt19937(seed_seq);
+
+  auto initializer = uniform_weight_initializer(LOWER_BOUND, UPPER_BOUND, &random_engine);
```
nickjong: This one I think is more readable, well idiomatic, as
```
uniform_weight_initializer initializer(LOWER_BOUND, UPPER_BOUND, &random_engine);
```

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -46,7 +58,7 @@ void define_resnet(model_spec& nn_spec, size_t num_styles) {
       /* stride_height */ 1,
       /* stride_width */ 1,
       /* padding */ padding_type::VALID,
-      /* weight_init_fn */ zero_weight_initializer());
+      /* weight_init_fn */ initializer);
```
nickjong: Doesn't this mean we're performing a bunch of random initialization even when we're about to overwrite with the pre-trained model?

abhishekpratapa:
@nickjong 
> What is the relationship between this PR and #2874 ?

This fixes the waviness users were experiencing. The pre-trained model let's the users get faster models but this is more likely to introduce artifacts into the stylization. The uniform initialization let's the user train a higher quality model but with a higher number of iterations. It resolves (7.) of the issue

src/python/turicreate/toolkits/style_transfer/style_transfer.py
abhishekpratapa: This is to allow the user to specify through the advanced parameters whether they want uniform initialization or initialization from the pre-trained model.

src/toolkits/style_transfer/style_transfer_model_definition.cpp
abhishekpratapa: Yeah, I noticed that as well. I wasn't sure if it was worth it to write in a flag to keep the model uninitialized. It seemed like a small one time startup cost for the model, but I'll add a flag [Boolean] (if you think that's the right way to do it). Or should it be an enum type of initialization for future proofing? What's your suggestion?

abhishekpratapa: I think I got it.

src/toolkits/style_transfer/style_transfer_model_definition.cpp
```diff
@@ -33,12 +35,20 @@ constexpr float LOWER_BOUND = -0.07;
...
+void define_resnet(model_spec& nn_spec, size_t num_styles, bool initialize=false, int random_seed=0) {
+  weight_initializer initializer;
+
+  // This is to make sure that when the uniform initialization is not needed extra work is avoided
+  if (initialize) {
+    std::mt19937 random_engine;
+    std::seed_seq seed_seq{random_seed};
+    random_engine = std::mt19937(seed_seq);
+
+    initializer = uniform_weight_initializer(LOWER_BOUND, UPPER_BOUND, &random_engine);
```
nickjong: I believe `&random_engine` will become a dangling pointer once `random_engine` goes out scope after this line. I suspect you will need to move `random_engine` outside the scope of this `if` statement to obtain well-defined behavior

shreyajain17: > @nickjong
> 
> > What is the relationship between this PR and #2874 ?
> 
> This fixes the waviness users were experiencing. The pre-trained model let's the users get faster models but this is more likely to introduce artifacts into the stylization. The uniform initialization let's the user train a higher quality model but with a higher number of iterations. It resolves (7.) of the issue

Are we exposing both pre trained weights as well as just the uniform initialization. I think if we are exposing it would be good for the users to actually have this literature somewhere. I don't think it is intuitive and I think we should put it in our userguide. 

I also think maybe in another PR you should expose the random_seed parameter. I think it would be great if users could generate reproducible models.

nickjong: I propose trying to have all the deep-learning toolkits controlled by random seeds for 6.2...

PR #2990, opened 2020-02-07
Disable weight update for the vgg parameters

abhishekpratapa:
# Overview
All the parameters were being trained in the TF style transfer implementation. Only the Transformer network needed to be trained.

# Previous Results

latest_400
latest_300
latest_200

# New Results
latest_5000
latest_6400
latest_6300


src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
```diff
@@ -24,32 +24,37 @@ def define_tensorflow_variables(net_params, trainable=True):
...
-            if "conv" in key:
+        if 'weight' in key:
+            # only set the parameter to train if in the transformer network
+            train_param = trainable and 'transformer_' in key
+            if 'conv' in key:
                 tensorflow_variables[key] = _tf.Variable(
                     initial_value=_utils.convert_conv2d_coreml_to_tf(net_params[key]),
                     name=key,
-                    trainable=trainable,
+                    trainable=train_param,
```
shreyajain17: What is the transformer layer? Why is trainable switching between True and False?

abhishekpratapa: So trainable specifies if the convolutional layers of the `transformer` network are trainable or not. The instance normalization layers are always trainable whereas the convolutional layers of the `vgg` network should never be trainable. Since the `vgg` network only contains convolutional layers. and the only other network that contains convolutional layers is the `transformer` network, this split works.

src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
```diff
@@ -24,32 +24,37 @@ def define_tensorflow_variables(net_params, trainable=True):
     Parameters
     ----------
     trainable: boolean
-        If `true` the network updates the convolutional layers as well as the
-        instance norm layers of the network. If `false` only the instance norm
-        layers of the network are updated.
+        If `true` the transformer network updates the convolutional layers as
```
TobyRoseman: In Python it's `True` and `False`, not `true` and `false`.

abhishekpratapa: Updated!

src/python/turicreate/toolkits/style_transfer/_tf_model_architecture.py
```diff
@@ -24,32 +24,37 @@ def define_tensorflow_variables(net_params, trainable=True):
...
     Returns
     -------
     out: dict
         The TF Variable dictionary.
     """
     tensorflow_variables = dict()
     for key in net_params.keys():
-        if "weight" in key:
-            if "conv" in key:
+        if 'weight' in key:
```
nickjong: Why are we going back to single quotes here? Are not using the new formatter/linter?

PR #2991, opened 2020-02-07
Added support for AC to use single input feature column

### LINKED ISSUE #2945: Can't export activity classifier to CoreML.
Can't export activity classifier to CoreML. It works well with TuriCreate 5, broken with 6.0.

`
turicreate.toolkits._main.ToolkitError: Could not export model: validator error: Layer 'features' of type 320 has 1 inputs but expects at least 2.
`

fareeha:
Closes #2945 

Currently AC expects input data to include >=2 input feature columns. For this a concat layer is used to concatenate all input feature columns. When a single input column is provided, `export_coreml` breaks due to NN spec `validateConcatLayer` check failing.

This change checks number of input feature columns and **does not** add concat layer if feature columns are of size 1.

### DIFF
--- a/src/python/turicreate/test/test_activity_classifier.py
+++ b/src/python/turicreate/test/test_activity_classifier.py
@@ -150,6 +150,26 @@ class ActivityClassifierCreateStressTests(unittest.TestCase):
                 validation_set=sf,
             )
 
+    def test_create_single_input_column(self):
+        sf_label = random.randint(0, self.num_labels - 1)
+        sf_session_id = max(self.data[self.session_id])
+        input_data = tc.SFrame(
+                {
+                    self.features[0]: [3.14],
+                    self.target: [sf_label],
+                    self.session_id: [sf_session_id],
+                }
+        )
+        model = tc.activity_classifier.create(
+                input_data,
+                features=["X1-r"],
+                target=self.target,
+                session_id=self.session_id,
+                prediction_window=self.prediction_window,
+        )
+        filename = tempfile.mkstemp("ActivityClassifier.mlmodel")[1]
+        model.export_coreml(filename)
+
     def test_create_invalid_batch_size(self):
         with self.assertRaises(_ToolkitError):
             tc.activity_classifier.create(
--- a/src/python/turicreate/test/test_activity_classifier.py
+++ b/src/python/turicreate/test/test_activity_classifier.py
@@ -170,6 +170,48 @@ class ActivityClassifierCreateStressTests(unittest.TestCase):
         filename = tempfile.mkstemp("ActivityClassifier.mlmodel")[1]
         model.export_coreml(filename)
 
+        # Load the model back from the CoreML model file
+        import coremltools
+        coreml_model = coremltools.models.MLModel(filename)
+
+        # Create a small dataset, and compare the models' predict() output
+        rs = np.random.RandomState(1234)
+        dataset = tc.util.generate_random_sframe(column_codes="r" * 3, num_rows=10)
+        dataset["session_id"] = 0
+        dataset[self.target] = random_labels = [
+            rs.randint(0, self.num_labels - 1,) for i in range(10)
+        ]
+
+        if _mac_ver() >= (10, 13):
+            w = self.prediction_window
+            labels = list(map(str, sorted(model.classes)))
+
+            input_features = {}
+            for f in self.features:
+                input_features[f] = dataset[f].to_numpy()
+            first_input_dict = {}
+            second_input_dict = {}
+            for key, value in input_features.items():
+                first_input_dict[key] = value[:w].copy()
+                second_input_dict[key] = value[w : 2 * w].copy()
+            first_input_dict["stateIn"] = np.zeros((400))
+            ret0 = coreml_model.predict(first_input_dict)
+
+            second_input_dict["stateIn"] = ret0["stateOut"]
+            ret1 = coreml_model.predict(second_input_dict)
+
+            pred = model.predict(dataset, output_type="probability_vector")
+            model_time0_values = pred[0]
+            model_time1_values = pred[w]
+            model_predictions = np.array([model_time0_values, model_time1_values])
+            coreml_time0_values = [ret0[self.target + "Probability"][l] for l in labels]
+            coreml_time1_values = [ret1[self.target + "Probability"][l] for l in labels]
+            coreml_predictions = np.array([coreml_time0_values, coreml_time1_values])
+
+            np.testing.assert_array_almost_equal(
+                model_predictions, coreml_predictions, decimal=3
+            )
+
     def test_create_invalid_batch_size(self):
         with self.assertRaises(_ToolkitError):
             tc.activity_classifier.create(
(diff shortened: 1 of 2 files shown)


fareeha:
We should add a test for this case. Will push a commit

src/python/turicreate/test/test_activity_classifier.py
```diff
@@ -150,6 +150,26 @@ def test_create_missing_validation_set(self):
...
+                }
+        )
+        model = tc.activity_classifier.create(
+                input_data,
+                features=["X1-r"],
+                target=self.target,
+                session_id=self.session_id,
+                prediction_window=self.prediction_window,
+        )
+        filename = tempfile.mkstemp("ActivityClassifier.mlmodel")[1]
```
guihao-liang: I remember that you are responsible to clear the temp file? how about using context like `with TempFile... as f`

src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -919,11 +919,21 @@ std::unique_ptr<model_spec> activity_classifier::init_model(
...
+    const flex_string& feature_column_name = features_list.front();
+    std::string single_input_feature{feature_column_name};
+    result->add_reshape("reshape", single_input_feature,
+                        {{1, num_features, 1, prediction_window}});
+  } else {
+    // Multiple feature columns. Add concatenate layer to concat all columns.
+    result->add_channel_concat(
+        "features",
+        std::vector<std::string>(features_list.begin(), features_list.end()));
+    result->add_reshape("reshape", "features",
```
guihao-liang: you probably want to add similar logic for tf?

src/python/turicreate/test/test_activity_classifier.py
fareeha: Oh good point!

> Unlike TemporaryFile(), the user of mkstemp() is responsible for deleting the temporary file when done with it.

 https://docs.python.org/3/library/tempfile.html

We do use  `tempfile.mkstemp` in a few places in the `test_activity_classifier.py` without clearing. Maybe I should do this for all places using it.

src/toolkits/activity_classification/activity_classifier.cpp
fareeha: We don't use tempfile here. Can you clarify which logic change you meant?

src/python/turicreate/test/test_activity_classifier.py
```diff
@@ -150,6 +150,26 @@ def test_create_missing_validation_set(self):
...
+        model.export_coreml(filename)
```
TobyRoseman: In addition to making sure we can export the model without error, it would be good to make sure that we can call `predict` (on the exported model) without error and that the results are very close to those from `model.predict` (on the TuriCreate model).

src/python/turicreate/test/test_activity_classifier.py
TobyRoseman: I would just create a GitHub issues for this. I don't think we should worry about this problem at this point in the release cycle. I think it would be a good thing to fix early into the next release cycle.

fareeha: That sounds good. It makes sense to open another issue and track the usage across the repo and make proper file clearing logic where needed. Since we have this usage in a lot of places:

```
scenario-tests/tests/toolkits/test_image_classifier.py:        filename = tempfile.mkstemp('bingo.mlmodel')[1]
scenario-tests/tests/toolkits/test_image_classifier.py:        filename = tempfile.mkstemp('bingo.mlmodel')[1]
scenario-tests/tests/toolkits/test_image_similarity.py:        filename = tempfile.mkstemp('ImageSimilarity.mlmodel')[1]
src/python/turicreate/test/test_activity_classifier.py:        filename = tempfile.mkstemp("ActivityClassifier.mlmodel")[1]
src/python/turicreate/test/test_activity_classifier.py:        filename = tempfile.mkstemp("ActivityClassifier.mlmodel")[1]
src/python/turicreate/test/test_activity_classifier.py:                filename = tempfile.mkstemp("ActivityClassifier.mlmodel")[1]
src/python/turicreate/test/test_image_classifier.py:        filename = tempfile.mkstemp("bingo.mlmodel")[1]
src/python/turicreate/test/test_image_classifier.py:        filename = tempfile.mkstemp("bingo.mlmodel")[1]
src/python/turicreate/test/test_image_similarity.py:        filename = tempfile.mkstemp("ImageSimilarity.mlmodel")[1]
src/python/turicreate/test/test_object_detector.py:        filename = tempfile.mkstemp("bingo.mlmodel")[1]
src/python/turicreate/test/test_object_detector.py:        filename2 = tempfile.mkstemp("bingo2.mlmodel")[1]
src/python/turicreate/test/test_object_detector.py:        filename = tempfile.mkstemp("bingo.mlmodel")[1]
src/python/turicreate/test/test_object_detector.py:        filename2 = tempfile.mkstemp("bingo2.mlmodel")[1]
src/python/turicreate/test/test_one_shot_object_detector.py:        filename = tempfile.mkstemp("bingo.mlmodel")[1]
src/python/turicreate/test/test_one_shot_object_detector.py:        filename2 = tempfile.mkstemp("bingo2.mlmodel")[1]
src/python/turicreate/test/test_style_transfer.py:            filename = tempfile.mkstemp("my_style_transfer.mlmodel")[1]
src/python/turicreate/test/test_text_classifier.py:        filename = tempfile.mkstemp("bingo.mlmodel")[1]
src/python/turicreate/test/test_text_classifier.py:        filename = tempfile.mkstemp("bingo.mlmodel")[1]
```

src/python/turicreate/test/test_activity_classifier.py
fareeha: Great idea! On it

src/toolkits/activity_classification/activity_classifier.cpp
shreyajain17: @guihao-liang This has nothing to do with tf Graph or MPS Graph.

src/python/turicreate/test/test_activity_classifier.py
fareeha: Opened an issue to track it https://github.com/apple/turicreate/issues/2993

guihao-liang: That's why we jam the test machines.

src/toolkits/activity_classification/activity_classifier.cpp
guihao-liang: @shreyajain17 you mean the extra concat layer won't affect the tf architecture?

fareeha: @guihao-liang  The concat layer was in the default AC graph. I added logic to not use concat layer where there isn't more than one feature columns i.e. nothing to concat.

guihao-liang: I understand that. this is for coreml. What I want to say is that where the `concat` logic is implemented. Is it in C++ backend before feeding the data into any compute context?

fareeha: Yes this happens in `init_model()` before instantiating the NN backend from the compute context.

shreyajain17: Yeah that's what I meant.

nickjong: Yes, at present the only interdependence between the CoreML spec here and what we do in the backends is the weights/parameters we use to initialize the backends. There are no NN weights or parameters involved with this change.

PR #3000, opened 2020-02-11
Introduce a minimal reactive streams framework, inspired somewhat by Swift's Combine

nickjong:
The main goal is make it easier to reason about the various portions of our
deep-learning training pipelines. For example, in Object Detection, we have a
data iterator that produces raw training examples, an image augmenter that
performs data augmentation, and a model backend that consumes a stream of
augmented training examples and produces progress updates and models.

As we integrate this framework, it should be easier to maintain unit tests for
these evolving pipelines and to manipulate threading and concurrency. For
example, if done right, this can replace our implementations of "double
buffering" currently copied across each model toolkit.

This PR turned out to be quite large, but FWIW it introduces no additions or modifications to the Turi Create release binaries, just headers (not imported into the main codebase yet) and unit tests. I'm explicitly going to use a separate PR for anything that might affect the shipping codebase.

### DIFF
--- a/src/ml/neural_net/combine.hpp
+++ b/src/ml/neural_net/combine.hpp
@@ -19,8 +19,6 @@
 #include <ml/neural_net/combine_base.hpp>
 #include <ml/neural_net/combine_futures_subscriber.hpp>
 #include <ml/neural_net/combine_iterator.hpp>
-
-}  // namespace neural_net
-}  // namespace turi
+#include <ml/neural_net/combine_map.hpp>
 
 #endif  // ML_NEURAL_NET_COMBINE_HPP_
--- a/src/ml/neural_net/combine.hpp
+++ b/src/ml/neural_net/combine.hpp
@@ -1,0 +1,26 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#ifndef ML_NEURAL_NET_COMBINE_HPP_
+#define ML_NEURAL_NET_COMBINE_HPP_
+
+/**
+ * \file combine.hpp
+ *
+ * Defines a reactive-streams library inspired by the Swift Combine
+ * framework. Its intent is to simplify reasoning about and testing of NN
+ * model-training pipelines.
+ */
+
+#include <ml/neural_net/combine_base.hpp>
+#include <ml/neural_net/combine_futures_subscriber.hpp>
+#include <ml/neural_net/combine_iterator.hpp>
+
+}  // namespace neural_net
+}  // namespace turi
+
+#endif  // ML_NEURAL_NET_COMBINE_HPP_
--- a/src/ml/neural_net/combine_base.hpp
+++ b/src/ml/neural_net/combine_base.hpp
@@ -24,6 +24,9 @@ namespace neural_net {
 // Forward declarations for types defined by other headers included by
 // combine.hpp.
 
+template <typename T, typename Callable>
+class CallableTransform;
+
 template <typename T>
 class FuturesStream;
 
@@ -221,19 +224,26 @@ class Publisher : public std::enable_shared_from_this<Publisher<T>> {
   }
 
   std::shared_ptr<FuturesStream<Output>> AsFutures() {
-    auto subscriber =
-        std::make_shared<FuturesSubscriber<Output>>(this->shared_from_this());
+    auto subscriber = std::make_shared<FuturesSubscriber<Output>>();
     Subscribe(subscriber);
     return std::make_shared<FuturesStream<Output>>(std::move(subscriber));
   }
 
-  template <typename Transform>
-  std::shared_ptr<Publisher<typename Transform::Output>> Map(
-      std::shared_ptr<Transform> transform) {
-    using TransformInput = typename Transform::Input;
-    using TransformOutput = typename Transform::Output;
+  template <typename TransformType>
+  std::shared_ptr<Publisher<typename TransformType::Output>> Map(
+      std::shared_ptr<TransformType> transform) {
+    using TransformInput = typename TransformType::Input;
+    using TransformOutput = typename TransformType::Output;
     return std::make_shared<MapPublisher<TransformInput, TransformOutput>>(
         this->shared_from_this(), std::move(transform));
+  }
+
+  template <typename Callable>
+  std::shared_ptr<Publisher<typename std::result_of<Callable(Output)>::type>>
+  Map(Callable fn) {
+    using TransformType = CallableTransform<Output, Callable>;
+    auto transform = std::make_shared<TransformType>(std::move(fn));
+    return Map(std::move(transform));
   }
 };
 
--- a/src/ml/neural_net/combine_base.hpp
+++ b/src/ml/neural_net/combine_base.hpp
@@ -1,0 +1,243 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#ifndef ML_NEURAL_NET_COMBINE_BASE_HPP_
+#define ML_NEURAL_NET_COMBINE_BASE_HPP_
+
+/**
+ * \file combine_base.hpp
+ *
+ * Defines the core data types for a reactive-streams library inspired by the
+ * Swift Combine framework. Client code should generally import combine.hpp.
+ */
+
+#include <exception>
+#include <memory>
+
+namespace turi {
+namespace neural_net {
+
+// Forward declarations for types defined by other headers included by
+// combine.hpp.
+
+template <typename T>
+class FuturesStream;
+
+template <typename T>
+class FuturesSubscriber;
+
+template <typename T, typename U>
+class MapPublisher;
+
+template <typename T, typename U>
+class Transform;
+
+/**
+ * Simple type expressing how many values a Subscriber is ready to receive
+ * from its Publisher.
+ */
+class Demand {
+ public:
+  static Demand Unlimited() { return Demand(-1); }
+  static Demand None() { return Demand(0); }
+
+  /** Any negative value is interpreted as "unlimited". */
+  explicit Demand(int max) : max_(max) {}
+
+  bool IsUnlimited() const { return max_ < 0; }
+  bool IsNone() const { return max_ == 0; }
+
+  /** Returns a negative number to indicate "unlimited." */
+  int max() const { return max_; }
+
+  /** Additively combines another Demand value into this one. */
+  Demand& Add(Demand other) {
+    if (IsUnlimited() || other.IsUnlimited()) {
+      max_ = -1;
+    } else {
+      max_ += other.max_;
+    }
+    return *this;
+  }
+
+  /** Decrease this demand by one if the current max is positive and finite. */
+  Demand& Decrement() {
+    if (max_ > 0) {
+      --max_;
+    }
+    return *this;
+  }
+
+ private:
+  int max_ = 0;
+};
+
+/**
+ * Interface for objects that Publishers send to Subscribers to allow the
+ * Subscribers to (potentially asynchronously) control the flow of values that
+ * the Subscriber receives from the Publisher.
+ */
+class Subscription {
+ public:
+  virtual ~Subscription() = default;
+
+  /**
+   * Requests the Publisher to stop sending anything to the Subscriber.
+   *
+   * After receiving Cancel() from a Subscriber, a Publisher should thereafter
+   * ignore all future messages from that Subscriber, including future calls to
+   * Cancel.
+   *
+   * Publishers must support Subscribers calling Cancel() from inside
+   * Subscriber::Receive(...).
+   */
+  virtual void Cancel() = 0;
+
+  /**
+   * Requests the Publisher to send the indicated number of values to the
+   * Subscriber.
+   *
+   * Publishers must support Subscribers calling Request(Demand) from inside
+   * Subscriber::Receive(Subscription), but Subscribers should avoid calling
+   * Request(Demand) inside Subscriber::Receive(Input). Instead, they should
+   * send additional Demand via the return value of Subscriber::Receive(Input)
+   * (to help prevent infinite recursion).
+   */
+  virtual void Request(Demand demand) = 0;
+};
+
+/**
+ * Type representing a message from a Publisher to a Subscriber indicating that
+ * the Subscriber will no longer receive any further messages.
+ */
+class Completion {
+ public:
+  /** Returns an instance that signals successful completion. */
+  static Completion Finished() { return Completion(); }
+
+  /**
+   * Returns an instance that signals failure, described by the given
+   * exception.
+   */
+  static Completion Failure(std::exception_ptr e) { return Completion(e); }
+
+  bool IsFinished() const { return failure_ == nullptr; }
+
+  /** Returns the exception if a failure and a null pointer otherwise. */
+  std::exception_ptr failure() const { return failure_; }
+
+ private:
+  explicit Completion(std::exception_ptr e = nullptr) : failure_(e) {}
+
+  std::exception_ptr failure_;
+};
+
+/**
+ * Interface for objects that consume values from a Publisher.
+ *
+ * Unless otherwise specified by the concrete implementation, external
+ * synchronization must be used to avoid concurrent calls the Subscriber
+ * interface from different threads.
+ */
+template <typename T>
+class Subscriber {
+ public:
+  /** The type of the values that this Subscriber consumes. */
+  using Input = T;
+
+  virtual ~Subscriber() = default;
+
+  /**
+   * The first signal that a Subscriber receives from a Publisher, passing the
+   * Subscription that the Subscriber can use to control the flow of values.
+   *
+   * A Subscriber may only have one Publisher. If it somehow receives more than
+   * one Subscription, it should call Subscription::Cancel() on any instances
+   * received after the first.
+   *
+   * A Subscriber is explictly allowed to demand values synchronously from
+   * within its implementation of this method.
+   */
+  virtual void Receive(std::shared_ptr<Subscription> subscription) = 0;
+
+  /**
+   * Transmits a value from the Publisher to this Subscriber.
+   *
+   * A Subcriber should never receive more calls to this method than the total
+   * Demand it has requested from its publisher. Subscribers should only demand
+   * more elements from within this method via its return value.
+   */
+  virtual Demand Receive(Input element) = 0;
+
+  /**
+   * Signals completion of the stream of values from the Publisher.
+   *
+   * A Subscriber should not receive any further signals of any kind after
+   * receiving a Completion.
+   */
+  virtual void Receive(Completion completion) = 0;
+};
+
+/**
+ * Interface for objects that produce values on demand from its Subscribers.
+ *
+ * Unless otherwise specified by the concrete implementation, external
+ * synchronization must be used to avoid concurrent calls on multiple threads to
+ * a Publisher, including via the Subscriptions that it passes to its
+ * Subscribers.
+ *
+ * Each concrete implementation defines whether it is unicast or multicast:
+ * whether multiple Subscribers observe the same values or not. (An
+ * implementation might only support one Subscriber, by passing an immediate
+ * Completion to each Subscriber after the first.)
+ */
+template <typename T>
+class Publisher : public std::enable_shared_from_this<Publisher<T>> {
+ public:
+  /** The type of values that this Publisher produces. */
+  using Output = T;
+
+  virtual ~Publisher() = default;
+
+  /**
+   * Establishes a connection between this Publisher and the given Subcriber.
+   *
+   * The Publisher must eventually call Subscriber::Receive(Subscription) on the
+   * given Subscriber (and may do so synchronously). The Publisher must then
+   * conform to the protocol established by the Subscription.
+   */
+  virtual void Receive(std::shared_ptr<Subscriber<Output>> subscriber) = 0;
+
+  // Convenienience methods, supporting the chaining together of operations.
+  // Many of these rely on the forward declarations above. Client code should
+  // include combine.hpp to ensure these are defined before they are used.
+
+  void Subscribe(std::shared_ptr<Subscriber<Output>> subscriber) {
+    Receive(std::move(subscriber));
+  }
+
+  std::shared_ptr<FuturesStream<Output>> AsFutures() {
+    auto subscriber =
+        std::make_shared<FuturesSubscriber<Output>>(this->shared_from_this());
+    Subscribe(subscriber);
+    return std::make_shared<FuturesStream<Output>>(std::move(subscriber));
+  }
+
+  template <typename Transform>
+  std::shared_ptr<Publisher<typename Transform::Output>> Map(
+      std::shared_ptr<Transform> transform) {
+    using TransformInput = typename Transform::Input;
+    using TransformOutput = typename Transform::Output;
+    return std::make_shared<MapPublisher<TransformInput, TransformOutput>>(
+        this->shared_from_this(), std::move(transform));
+  }
+};
+
+}  // namespace neural_net
+}  // namespace turi
+
+#endif  // ML_NEURAL_NET_COMBINE_BASE_HPP_
--- a/src/ml/neural_net/combine_base.hpp
+++ b/src/ml/neural_net/combine_base.hpp
@@ -197,6 +197,10 @@ class Subscriber {
  * whether multiple Subscribers observe the same values or not. (An
  * implementation might only support one Subscriber, by passing an immediate
  * Completion to each Subscriber after the first.)
+ *
+ * Note: instances of this class are intended to be stored using shared_ptr.
+ * Many of the operators rely on generating strong references to the instance
+ * being augmented.
  */
 template <typename T>
 class Publisher : public std::enable_shared_from_this<Publisher<T>> {
--- a/src/ml/neural_net/combine_futures_subscriber.hpp
+++ b/src/ml/neural_net/combine_futures_subscriber.hpp
@@ -8,6 +8,7 @@
 #ifndef ML_NEURAL_NET_COMBINE_FUTURES_SUBSCRIBER_HPP_
 #define ML_NEURAL_NET_COMBINE_FUTURES_SUBSCRIBER_HPP_
 
+#include <cassert>
 #include <future>
 #include <memory>
 #include <queue>
@@ -82,6 +83,10 @@ class FuturesSubscriber : public Subscriber<T> {
   }
 
   void Receive(std::shared_ptr<Subscription> subscription) override {
+    // It is a programmer error to attach the same Subscriber to more than one
+    // Publisher.
+    assert(subscription_ == nullptr);
+
     // Reject any subscriptions after the first. Reject the first subscription
     // if we cancelled before it could start.
     if (subscription_ || completed_) {
--- a/src/ml/neural_net/combine_futures_subscriber.hpp
+++ b/src/ml/neural_net/combine_futures_subscriber.hpp
@@ -1,0 +1,160 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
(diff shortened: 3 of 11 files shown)


src/ml/neural_net/combine_mock.hpp
```diff
@@ -0,0 +1,84 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
```
TobyRoseman: Would it be better for mocks to go somewhere under the top level `test` directory rather than the top level `src`?

src/ml/neural_net/combine.hpp
```diff
@@ -0,0 +1,26 @@
...
+ * Defines a reactive-streams library inspired by the Swift Combine
+ * framework. Its intent is to simplify reasoning about and testing of NN
+ * model-training pipelines.
+ */
+
+#include <ml/neural_net/combine_base.hpp>
+#include <ml/neural_net/combine_futures_subscriber.hpp>
+#include <ml/neural_net/combine_iterator.hpp>
+
+}  // namespace neural_net
```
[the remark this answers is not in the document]
nickjong: Sigh, those were an error. I didn't notice that none of my tests actually exercised this header!

src/ml/neural_net/combine_mock.hpp
nickjong: Good question. I'd like to encourage the mocks to live near the interfaces they are mocking, so that it's easier to remember that they need to change in concert. Ideally the responsibility for maintaining a mock is for the maintainer of that interface, not the clients who want to consume that interface. Since these are header-only, it's harmless to have in that directory: it doesn't get compiled into the shipping binary

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Returns an instance that signals failure, described by the given
+   * exception.
+   */
+  static Completion Failure(std::exception_ptr e) { return Completion(e); }
+
+  bool IsFinished() const { return failure_ == nullptr; }
+
+  /** Returns the exception if a failure and a null pointer otherwise. */
+  std::exception_ptr failure() const { return failure_; }
```
guihao-liang: it's better to use boost::optional

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+  virtual void Receive(std::shared_ptr<Subscription> subscription) = 0;
+
+  /**
+   * Transmits a value from the Publisher to this Subscriber.
+   *
+   * A Subcriber should never receive more calls to this method than the total
+   * Demand it has requested from its publisher. Subscribers should only demand
+   * more elements from within this method via its return value.
+   */
+  virtual Demand Receive(Input element) = 0;
```
guihao-liang: this may cause an infinite subscription. If need more data, it's better to have a new subscription. In this way, the subscriber is not responsible for accounting but only for receiving published values.

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+  virtual ~Publisher() = default;
+
+  /**
+   * Establishes a connection between this Publisher and the given Subcriber.
+   *
+   * The Publisher must eventually call Subscriber::Receive(Subscription) on the
+   * given Subscriber (and may do so synchronously). The Publisher must then
+   * conform to the protocol established by the Subscription.
+   */
+  virtual void Receive(std::shared_ptr<Subscriber<Output>> subscriber) = 0;
```
guihao-liang: the name `Receive` is confusing. It means to receive new subscribers but in most contexts, `Receive` means receiving data. I find this is not intuitive.

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+ * a Publisher, including via the Subscriptions that it passes to its
+ * Subscribers.
+ *
+ * Each concrete implementation defines whether it is unicast or multicast:
+ * whether multiple Subscribers observe the same values or not. (An
+ * implementation might only support one Subscriber, by passing an immediate
+ * Completion to each Subscriber after the first.)
+ */
+template <typename T>
+class Publisher : public std::enable_shared_from_this<Publisher<T>> {
```
guihao-liang: I think it's better to design that a publisher can also be a subscriber, in which way you can chain publishers.

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Establishes a connection between this Publisher and the given Subcriber.
+   *
+   * The Publisher must eventually call Subscriber::Receive(Subscription) on the
+   * given Subscriber (and may do so synchronously). The Publisher must then
+   * conform to the protocol established by the Subscription.
+   */
+  virtual void Receive(std::shared_ptr<Subscriber<Output>> subscriber) = 0;
+
+  // Convenienience methods, supporting the chaining together of operations.
```
guihao-liang: inconsistent comment blocks

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+  // Many of these rely on the forward declarations above. Client code should
+  // include combine.hpp to ensure these are defined before they are used.
+
+  void Subscribe(std::shared_ptr<Subscriber<Output>> subscriber) {
```
guihao-liang: do you mean publisher can subscribe subscriber?

src/ml/neural_net/combine_base.hpp
nickjong: This portion of the codebase is not taking a dependency on boost for now. Might adopt std::optional if we go to C++17, or write a simple implementation ourselves.

src/ml/neural_net/combine_base.hpp
nickjong: We’re hewing to the design of the Combine API published in Swift.

This will help with familiarity for contributors more familiar with Swift, and in case we migrate this codebase to Swift in future

src/ml/neural_net/combine_base.hpp
nickjong: This also comes from Swift Combine

https://developer.apple.com/documentation/combine/publisher/3229093-receive

src/ml/neural_net/combine_base.hpp
nickjong: Nothing stops an implementation from deriving from both interfaces

src/ml/neural_net/combine_base.hpp
nickjong: This comment is not meant to be exported to Doxygen, since it doesn’t describe a single symbol or file

nickjong:
See #3006 for how this framework can integrate into our current object-detection implementation

src/ml/neural_net/combine_base.hpp
guihao-liang: but our codebase already used boost 1.68. Exception is better than segfault.

src/ml/neural_net/combine_base.hpp
guihao-liang: Thanks for letting me know what's the motivation behind.

src/ml/neural_net/combine_base.hpp
guihao-liang: I guess I know where rust takes `where` clause from. Anyway, it's not an intuitive design.

src/ml/neural_net/combine_base.hpp
guihao-liang: But the interface is deceiving, all called `Receive`. The only differences are their parameter lists.

Besides that, Published is being `Subscribed` by subscribers? I feel kind of strange to call `publisher.subcribte(subscriber)`. Shouldn't it be `publisher.distribute(subscriber)` or `subscriber.subcribte(publisher)`?

src/ml/neural_net/combine_base.hpp
guihao-liang: glad to know. Normally use `///` to denote this purpose.

src/ml/neural_net/combine_base.hpp
nickjong: What segfault do you mean? We are using `std::exception_ptr` explicitly to capture exceptions and support those in client code

src/ml/neural_net/combine_base.hpp
nickjong: FWIW, I think it's to help emphasize all the places where different elements in a Combine pipeline can receive signals from one another. The four methods named `Receive`, along with a `Request` from a `Subscription`, are the messages that get routed to different threads or dispatch queues in the full Combine framework.

src/ml/neural_net/combine_base.hpp
nickjong: In practice, most users don't call `Receive` explicitly. It's an implementation detail of the class authors. The API is designed to facilitate chaining together operations between a Publisher and a Subscriber. For example, one draft of my client integration had this line of code:
```
data_iterator->AsPublisher()->Map(augmenter)->Map(encoder)->Map(trainer)->Subscribe(sink);
```
The alternative, following your suggestion, would be:
```
sink->Subscribe(data_iterator->AsPublisher()->Map(augmenter)->Map(encoder)->Map(trainer)
```
or maybe something like
```
sink->Subscribe(encoder->Map(augmenter->Map(ToPublisher(data_iterator))))
```

src/ml/neural_net/combine_base.hpp
nickjong: Yes, this follows the Swift Combine design from here: https://developer.apple.com/documentation/combine/publisher/3204756-subscribe

src/ml/neural_net/combine_base.hpp
guihao-liang: Yeah, it's more like a functional programming style. Personally I like it. Or,

```
data_iterator->AsPublisher()->Map(augmenter)->Map(encoder)->Map(trainer)->PublishTo(sink);
```

Anyway, I understand this now.

src/ml/neural_net/combine_base.hpp
guihao-liang: I mean one can use std::exception_ptr with null value.

src/ml/neural_net/combine_futures_subscriber.hpp
```diff
@@ -0,0 +1,160 @@
...
+      subscription->Cancel();
+      return;
+    }
+
+    subscription_ = std::move(subscription);
+
+    // If we already have promises queued, request their values now.
+    if (!promises_.empty()) {
+      Demand demand(static_cast<int>(promises_.size()));
+      subscription_->Request(demand);
```
guihao-liang: shouldn't we use the last subscription to do it? I think you want to exhaust the last subscription here.

guihao-liang: and you also want to clean up `failure_` as well.

src/ml/neural_net/combine_futures_subscriber.hpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  Demand Receive(Input element) override {
```
guihao-liang: I think this won't work well if used with multithreading and async jobs.

If I have some requests sent in an order of <1,2,3,4>, the publisher may finish jobs (assume job is async) in order <4,3,2,1> and dispatch then back to me not in order of <1,2,3,4>.

I think the promise should be moved to publisher and publisher should fulfill the output into it. And the subscriber awaits on futures.

guihao-liang: assume the call is like `subscriber.Receive(publisher's output)`. Maybe I'm wrong.

src/ml/neural_net/combine_iterator.hpp
```diff
@@ -0,0 +1,147 @@
...
+          // On any exception, *value was not initialized, since it was the last
+          // statement in the try block.
+          value = nullptr;
+          failure = std::current_exception();
+        }
+
+        // Send the appropriate signal.
+        if (failure) {
+          // Signal failure and ensure we don't send any more signals.
+          subscriber_->Receive(Completion::Failure(failure));
```
guihao-liang: I see! Subscriber is deeply attached to Publisher. That restricts one to one relationships between subscribers and publishers, which means only allows singe streaming. Oh, well, I thought it's one-to-many at the very beginning. It's worth mentioning at the beginning.

In short,
1. publisher holds only one subscriber
2. The only one subscriber holds a subscription from the publisher.

If the publisher and the subscriber only run at most one thread for each of them, this will be good.

Again, I think it's better not to attach a subscriber closely attached to the publisher. You can use the `subscription` object to send` promises` to the publisher's task queue (maybe multiple queues for each subscriber so that one copy of data can be sent to many). That can allow one publisher and multiple subscribers running in their own threads and do the test independently.

src/ml/neural_net/combine_futures_subscriber.hpp
guihao-liang: After reading the code from the publisher, I know that the publisher has a one-to-one relationship with the subscriber. The above problem won't show up.

src/ml/neural_net/combine_map.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+ private:
+  Callable impl_;
+};
+
+/**
+ * Concrete operator Publisher that wraps a Transform.
+ *
+ * The resulting Publisher inherits the semantics of the upstream Publisher that
+ * it subscribes to, with regard to the semantics of multiple downstream
```
guihao-liang: only one downstream.

guihao-liang:
Thanks for introducing this framework. Awesome work and solid coding skills. This can free us from wiring up the mock objects.

I think it twice, single threading design model is fine since the deep learning toolkits mostly have single-threaded data flow.

guihao-liang:
I will 

> See #3006 for how this framework can integrate into our current object-detection implementation

Just figure out that you already pointed out an easy path to understanding all. I missed it. T_T

src/ml/neural_net/combine_iterator.hpp
```diff
@@ -0,0 +1,147 @@
...
+          subscriber_->Receive(Completion::Finished());
+          Cancel();
+        } else {
+          // Pass the value to the Subscriber, adding any new demand.
+          demand.Decrement();
+          Demand new_demand = subscriber_->Receive(std::move(*value));
+          demand.Add(new_demand);
+
+          // We must manually destroy the (moved-from) value.
+          value->~Output();
```
guihao-liang: not because of the `move` but because we are using placement new.

src/ml/neural_net/combine_map.hpp
guihao-liang: Just think twice. My bad, you can have multiple different subscribers subscribe to a publisher.

guihao-liang:
this is using **pull** model, instead of **push** model.

Since there's no **mailbox** and **watermark** concept, it's hard for a publisher to send one data to many subscribers (unless each buffered task queue for each subscriber). All subscribers consume the publisher.

I kind of understand why `publisher` calls `subscribe` on `subscriber`.  `subscriber` is the initiator of this dataflow, instead of `publisher`. It's like `publisher` subscribes **requests** from `subscriber`.

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+    Subscribe(subscriber);
+    return std::make_shared<FuturesStream<Output>>(std::move(subscriber));
+  }
+
+  template <typename TransformType>
+  std::shared_ptr<Publisher<typename TransformType::Output>> Map(
+      std::shared_ptr<TransformType> transform) {
+    using TransformInput = typename TransformType::Input;
+    using TransformOutput = typename TransformType::Output;
+    return std::make_shared<MapPublisher<TransformInput, TransformOutput>>(
```
guihao-liang: I kept thinking about this weak reference usage. I think there's some trouble with this design or you should state it clearly in the documentation. The Map is only useful when you use it immediately along with the original publisher or the owner is kept living longer than that publisher generated from Map.

```
shared_ptr<Publisher> foo() {
    auto A = make_shared<Publisher>(args);
    ....
    return A->Map(some_transformer_fu);
}

Publisher B = foo();
B->asFutures(); // should throw bad weak_ref because the real owner A expires.
```

src/ml/neural_net/combine_base.hpp
nickjong: Yes, in this case, a null exception pointer corresponds to the no-error case, when a stream finished normally

src/ml/neural_net/combine_futures_subscriber.hpp
nickjong: What do you mean by "last subscription" and "exhaust the subscription"? Typically there's only one subscription between a publisher and subscriber. For example, see rule 12 under "Subscriber" in this open-source specification of the API protocol, which appears to roughly match Apple's Combine: https://github.com/reactive-streams/reactive-streams-swift

nickjong: Also, what is there to clean up for `failure_`? That variable cannot be non-null at this point, since we must receive the subscription before any completion.

src/ml/neural_net/combine_base.hpp
guihao-liang: Talked with @hoytak and I realize I have some misunderstanding of the underlying implementation of `shared_from_this`. The use here is totally legitimate. My mistake.

src/ml/neural_net/combine_iterator.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+    void Request(Demand demand) override {
+      // Keep sending signals to the Subscriber until we're cancelled or we
+      // exhaust the demand.
+      while (IsActive() && !demand.IsNone()) {
+        // Invoke the iterator to determine what signal we'll send.
+
+        // Don't assume that Output has a (cheap) default constructor.
+        // TODO: Use an Optional type instead.
+        using OutputStorage =
```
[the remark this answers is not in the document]
nickjong: I believe `std::aligned_storage` is the safest way to get raw memory compatible with the desired (unknown) type `Output`. In my mind, this is effectively the same as doing
```
Output value;
```
without actually calling `Output`'s constructor

Or what do you mean by manually align the storage type? How would that be done? (Yes, this is another spot where we could use `std::optional` or `boost::optional`)

src/ml/neural_net/combine_futures_subscriber.hpp
guihao-liang: I mean, the already queued promises correspond to get data from the previous subscription. From here,

```
    if (subscription_ || completed_) {
       subscription->Cancel();
       return;
     }
```

Also `failure_` might be the exception of that canceled subscription, we should reset it for the new subscription passed in.

src/ml/neural_net/combine_iterator.hpp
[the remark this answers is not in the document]
guihao-liang:
```
            typename std::aligned_storage<sizeof(Output),
                                           alignof(Output)>::type;
```
I mean `alignof`.

src/ml/neural_net/combine_iterator.hpp
nickjong: Yes! It turns out that our use case doesn't really need "multicast" publishing, so I haven't worried too much about this.

We do want to introduce multiple threads at some point. Indeed, that's one of the main values of Combine and similar frameworks. These tend to reason about dispatch to other task queues as "scheduling", but for us, it will probably suffice just to worry about hopping threads (not time delays and such).

So in the example that you specify, you would typically set each subscriber to receive elements on a different queue. Something like:
```
iterator->ReceiveOn(thread1)->Subscribe(subscriber1);
iterator->ReceiveOn(thread2)->Subscribe(subscriber2);
```
Effectively the middle "pipe" in this pipeline just switches to the desired thread before sending the value along

nickjong:
> this is using **pull** model, instead of **push** model.
> 
> Since there's no **mailbox** and **watermark** concept, it's hard for a publisher to send one data to many subscribers (unless each buffered task queue for each subscriber). All subscribers consume the publisher.
> 
> I kind of understand why `publisher` calls `subscribe` on `subscriber`. `subscriber` is the initiator of this dataflow, instead of `publisher`. It's like `publisher` subscribes **requests** from `subscriber`.

Yes, good point. Honestly, I suspect most clients of Apple's Combine framework do use a push model, by requesting unlimited demand from a root publisher that is ultimately event-based. (To respond to e.g. user interaction on a UI widget or a notification).

In our case, we want to use the back-pressure mechanism built into Combine to instead use a pull model, where the iterator only emits another batch on demand.

Thanks for taking a close look at all of this!

src/ml/neural_net/combine_base.hpp
nickjong: To be fair, I find `shared_from_this` somewhat confusing too. It basically adds an intrusive ref count, instead of relying on the external refcount that `shared_ptr` normally uses. There must be some special logic in all of the `shared_ptr` constructors to detect when it wraps a class that has an intrusive ref count, and then it uses that instead.

src/ml/neural_net/combine_futures_subscriber.hpp
nickjong: Hmm, let me try to explain my intent for how this class works. As a Subscriber, this class receives from its Publisher a Subscription, zero or more values, then (maybe) a Completion, in that order. Each value corresponds to one value that this class’s user demands via `Request`.

At present, all of this is synchronous, but the class is written to guard against a future case when all of this is asynchronous. In that scenario, the user can call `Request` (and generate a future) _before_ the class has even received a Subscription from its Subscriber. That’s the situation the code here (line 94) is handling. In this case, when we finally receive the Subscription, then we submit the demands to the Publisher to correspond to the requests our user has already made

nickjong: Another way of explaining it: in general, each time `FuturesSubscriber::Request` is called by the user, the implementation submits a Request to its underlying Subscription. But it’s possible that it hasn’t received the Subscription yet.

src/ml/neural_net/combine_iterator.hpp
guihao-liang: Yeah, it depends on the original publisher to decide how to distribute (multicast) the data. It can keep track of different queues for the different subscribers. I think this protocol is very flexible. Learned a ton! Thanks for this great work!

src/ml/neural_net/combine_futures_subscriber.hpp
guihao-liang: I understand the intent to design for `Request`.

I'm concerned with if I added another subscription mistakenly. As a user, I would expect the subscription is consistent and data I will get is from same subscription. No surprises coming out from a mistakenly switching to other subscriptions.

Also, if I use a new subscription, that means the stream restarts from this publisher and the `failure_` value is from the previous subscription (publisher).

src/ml/neural_net/combine_iterator.hpp
nickjong: Yeah, that's something that wasn't obvious to me: the behavior with multiple subscribers is not all standardized. It's up to the publisher to define. This is one reason that Combine and similar frameworks introduce an explicit `multicast` operator, which supports adding multiple subscribers and then having a trigger/barrier that begins pushing values only after all subscribers are ready.

src/ml/neural_net/combine_futures_subscriber.hpp
nickjong: Line 85 is intended to handle this case, by rejecting any subscriptions after the first.

(This aligns with Subscriber Rule 5 here: https://github.com/reactive-streams/reactive-streams-swift.)

The intent is to never accept any subscription after the first. It's generally up to the user not to attach one Subscriber to multiple Publishers, but the API design helps here by encouraging the use of operations (like `Map`) that instantiate a Subscriber and then attach to it all at once. In short, by having factory functions on Publisher that both generate the next pipe in the pipeline and connect it immediately, yielding a new Publisher.

src/ml/neural_net/combine_base.hpp
guihao-liang: I finally reasoning it through. Again, if we use this `shared_from_this`,

https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

> it is permitted to call shared_from_this only on a previously shared object, i.e. on an object managed by std::shared_ptr<T>. Otherwise the behavior is undefined (until C++17)std::bad_weak_ptr is thrown (by the shared_ptr constructor from a default-constructed weak_this) (since C++17).

We should let the user know they should first have a shared_ptr to call this class!

> The constructors of std::shared_ptr detect the presence of an unambiguous and accessible (ie. public inheritance is mandatory) (since C++17) enable_shared_from_this base and assign the newly created std::shared_ptr to the internally stored weak reference if not already owned by a live std::shared_ptr (since C++17). 

image

I finally understand `owned by a live std::shared_ptr `. @hoytak

src/ml/neural_net/combine_iterator.hpp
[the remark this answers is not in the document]
guihao-liang: That's why I feel weird. I used to this way

```
char value[sizeof(Output));
```

src/ml/neural_net/combine_futures_subscriber.hpp
guihao-liang: Yeah, I know the subscription is canceled. I mean the receiver doesn't know this.

The `future` that the subscriber gets corresponds to the `promise` from the canceled subscription. I would expect this way. Otherwise, it's like I subscribe to _New York Times_, and I paid for the annual fee and suddenly I received Washington Posts someday.

I think you could send a failure saying subscription change, renew your subscription to the new publisher.

src/ml/neural_net/combine_iterator.hpp
```diff
@@ -0,0 +1,147 @@
...
+   public:
+    IteratorSubscription(std::shared_ptr<Subscriber<Output>> subscriber,
+                         std::shared_ptr<Iterator<Output>> iterator)
+        : subscriber_(std::move(subscriber)), iterator_(std::move(iterator)) {}
+
+    bool IsActive() const { return subscriber_ != nullptr; }
+
+    void Cancel() override { subscriber_.reset(); }
+
+    void Request(Demand demand) override {
```
guihao-liang: https://github.com/apple/turicreate/pull/3000/files#r382174687

I think the problem also comes from here. If it's canceled, the subscriber kept queueing its `promise` queue, while subscription doesn't reply but ignores its request.

If one canceled the subscription and used its subscriber to get a future inadvertently, the future of that will hang forever and it's hard to debug. I think it's better to explicitly reply something like `SubscriptionCancelled`.

src/ml/neural_net/combine_iterator.hpp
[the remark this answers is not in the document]
nickjong: Hmm, I believe that may not satisfy the alignment requirements of `Output`. Suppose we're on a platform where `double` is 8 bytes and should be allocated at an address that is a multiple of 8. Your code above wouldn't guarantee the proper alignment

src/ml/neural_net/combine_base.hpp
nickjong: Makes sense, I'll add a comment!

src/ml/neural_net/combine_futures_subscriber.hpp
nickjong: Well, this case is documented in the comments here:
https://github.com/apple/turicreate/blob/8805840d3679fbdc9ab9f59b7df08ecfe2a07826/src/ml/neural_net/combine_futures_subscriber.hpp#L43

This test case covers it: https://github.com/apple/turicreate/blob/15bd9f8ff8b56d511038854a9571028c64fa89ee/test/unity/toolkits/neural_net/test_combine_futures_subscriber.cxx#L87

Since the implementation treats cancellation and receiving a Completion as having the same effect here, setting `completed_` to true

guihao-liang: Yeah, but this is the case where subscription is completed not due to a subscription change. 

I mean the promises are generated based on the **previous** subscription. We should set those promises with nullptr as well since the **previous** subscription is forced to be finished, which is also consistent with the behavior that a subscription is completed.

nickjong: What previous subscription? By design, we only allow one subscription per subscriber. Can you give a specific code path or test case illustrating your concern?

guihao-liang: let's discuss that in person later.

src/ml/neural_net/combine_futures_subscriber.hpp
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  void Receive(std::shared_ptr<Subscription> subscription) override {
+    // It is a programmer error to attach the same Subscriber to more than one
+    // Publisher.
+    assert(subscription_ == nullptr);
+
+    // Reject any subscriptions after the first. Reject the first subscription
+    // if we cancelled before it could start.
+    if (subscription_ || completed_) {
```
guihao-liang: if we already assert `subscription_` to be null, we don't have to check it here.

nickjong: This is a failsafe for when `NDEBUG` is defined

PR #3006, opened 2020-02-18
Integrate new Combine-like framework into OD training

nickjong:
Also begins encapsulating the darknet-yolo-specific bits. However, the inference
path and CoreML export still need to be refactored

This PR builds upon #3000. For this PR, please focus feedback on the commits after the first two (outside of `neural_net`)

(Although I do hope to move much of the code here into `neural_net` once we untangle the dependencies)


nickjong:
High-level overview of changes: defined a new interface in `od_model.hpp` meant to abstract over various model architectures. The code in that file deals with what should be shared among architectures: how we process (and augment) the data.

Then `od_darknet_yolo_model.hpp` declares code that is specific to the current `darknet-yolo` model. Much of this code is moved from `object_detector.cpp`, which way too strongly assumed one specific model.

Finally, `object_detector.cpp` adopts the new Combine-like framework. The class itself no longer maintains direct references to all the dependencies, just to the "outlets" of the training pipeline. This also moves us away from using `nn_spec_` as the source of truth for a model (since this was out of date anyway the moment we train anything). Instead, there's an explicit `Checkpoint` type that represents what we need to serialize, which also handily defines the minimum required to instantiate a backend.

nickjong:
Also note that in this new code I'm adopting the Google style guide a bit more closely: in particular using CamelCase instead of under_score for type names and function names. This is especially helpful for telling at a glance whether a name is a variable or a type.

`Foo bar()` is a variable `bar` with type `Foo` that is being default constructed.
`Foo Bar()` is a function `Bar` that takes no arguments and returns a `Foo`

And to me `int fn(Foo foo)` is more readable than `int fn(foo_type foo)` or `int fn(foo f)`

nickjong:
One more one more thing: I've tried to do minimal changes to the OD unit tests, to help have confidence that these changes largely preserve the old behaviors. So note that this integration does not yet demonstrate the potential for simpler unit tests. For example, the OD tests can someday be written in a way that they just mock out the end of the backend pipeline, and we can focus our testing efforts on each individual pipe in the pipeline, without having to bend over backwards to manipulate mocks for all the important dependencies.

src/ml/neural_net/combine_base.hpp
```diff
@@ -0,0 +1,253 @@
...
+  static Demand None() { return Demand(0); }
+
+  /** Any negative value is interpreted as "unlimited". */
+  explicit Demand(int max) : max_(max) {}
+
+  bool IsUnlimited() const { return max_ < 0; }
+  bool IsNone() const { return max_ == 0; }
+
+  /** Returns a negative number to indicate "unlimited." */
+  int max() const { return max_; }
```
abhishekpratapa: Curiosity, is this method lower case since it's pretty much referencing an private member variable.

nickjong: Good question. The Google style guide makes an exception for getters/setters that are conceptually used like variables: https://google.github.io/styleguide/cppguide.html#Function_Names

src/ml/neural_net/combine_futures_subscriber.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+    // Cancel the subscription if active.
+    if (subscription_) {
+      subscription_->Cancel();
+      subscription_.reset();
+    }
+
+    // Fulfill any outstanding promises.
+    while (!promises_.empty()) {
+      promises_.front().set_value(nullptr);
```
abhishekpratapa: Are we just setting the value to `nullptr` and popping since the subscriber doesn't care about these promises anymore after calling `Cancel`?

nickjong: Short answer: yes. We don't just want to leave the promises unfulfilled, in case someone is waiting for them.

nickjong: Also, note that it would be ideal to keep feedback/discussion on the `combine*.hpp` files in the base PR: #3000

shreyajain17:
This comment is only regarding the OD files (object_detector, od_model and od_darket_yolo_model)
The `object_detector` is the most high level file which takes out all of the model specific pieces. 
The `od_model` makes the isolated pieces (mid level). 
The `od_darket_yolo_model` derives from the pieces from od_model and implements exactly how darknet-yolo needs the implementation to be. So if we want to ever add another model backend for OD, we would need to only add a file like `od_new_model`.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -335,16 +279,42 @@ size_t object_detector::get_version() const {
 }
 
 void object_detector::save_impl(oarchive& oarc) const {
-  if (training_model_) {
-    // If checkpointing during training, first copy weights from the backend.
-    synchronize_model(nn_spec_.get());
-  }
-  _save_impl(oarc, *nn_spec_, state);
+  Checkpoint* checkpoint = read_checkpoint();
```
abhishekpratapa: So to truly checkpoint would we also have to store the learning rate and current iterations to say resume training or does that work differently?

abhishekpratapa: Ohhh nvmd you also store the state. Got it.

nickjong:
> This comment is only regarding the OD files (object_detector, od_model and od_darket_yolo_model)
> The `object_detector` is the most high level file which takes out all of the model specific pieces.
> The `od_model` makes the isolated pieces/interfaces for specific model backends. (mid level).
> The `od_darket_yolo_model` (lowest level file) derives from the pieces from `od_model` and implements exactly how darknet-yolo needs the implementation to be. So if we want to ever add another model backend for OD, we would need to only add a file like `od_darknet_yolo_model`. Is this right?

I think there are different ways of looking at it, but I would construe `od_model` and `od_darknet_yolo_model` as being at the same level, just one is model-specific and one is abstract across specific models.

I see three levels here:

1. The `model_backend` implementation, which is platform-specific and only knows about one batch at a time: how to update model parameters given one batch of data
2. The new `Model` and `DarknetYOLOModel` classes, which know about training pipelines that combine data, augmentation, and a model backend
3. The `object_detector` class, which talks to Python and knows about save/load and SFrame, etc.

When we add new model architectures, we'll probably want a new `model_backend` and `Model` subclass for each, but ideally we should need few changes elsewhere

src/toolkits/object_detection/object_detector.cpp
nickjong: Hmm, but I don't pass the state to the backend. So I think you found a bug: resuming from a checkpoint shouldn't restart the learning-rate schedule.

I think the way to think about it is that the checkpoint should contain everything that the `DarknetYOLOModel` needs to operate correctly. The state is a higher-level concept, which is used to communicate with Python and to save to disk. That does imply that the state should somehow contain everything in the checkpoint (we explicitly handle the model weights separately)

src/toolkits/object_detection/od_darknet_yolo_model.cpp
```diff
@@ -0,0 +1,399 @@
...
+        static_cast<int>(GetAnchorBoxes().size()), config.num_classes);
+  };
+
+  // Wrap the model_backend.
+  auto trainer = std::make_shared<DarknetYOLOTrainer>(
+      backend_, BASE_LEARNING_RATE, config_.max_iterations);
+
+  // Append the encoding function and the model backend to the pipeline.
+  // TODO: Dispatch augmentation to a separate thread/queue.
+  return augmented_data->Map(encoder)->Map(trainer);
```
guihao-liang: Map example here.

src/toolkits/object_detection/object_detector.cpp
```diff
@@ -1197,49 +1080,43 @@ void object_detector::resume_training(gl_sframe data,
...
-      /* h_out   */ grid_height,
-      /* w_out   */ grid_width,
-      /* config  */ train_config,
-      /* weights */ get_model_params());
+void object_detector::connect_training_backend(
+    std::unique_ptr<Model> backend, std::unique_ptr<data_iterator> iterator,
+    int batch_size) {
+  // Subscribe to the backend model using futures, for compatibility with our
+  // current synchronous API surface.
+  training_futures_ =
```
guihao-liang: reference, create publisher.

src/toolkits/object_detection/od_darknet_yolo_model.cpp
```diff
@@ -0,0 +1,399 @@
...
+// TODO: These should be exposed in a way that facilitates experimentation.
+// TODO: A struct instead of a map would be nice, too.
+
+float_array_map GetBaseBackendConfig() {
+  float_array_map config;
+  config["learning_rate"] = shared_float_array::wrap(BASE_LEARNING_RATE);
+  config["gradient_clipping"] = shared_float_array::wrap(0.025f);
+  // TODO: Have MPS path use these parameters, instead
+  // of the values hardcoded in the MPS code.
+  config["od_rescore"] = shared_float_array::wrap(1.0f);
```
shreyajain17: Should we change config to a struct specific to OD?

nickjong:
> ~I don't see the new test framework being used. Is it in test_object_detector.cxx?~
> 
> Never mind, thought combine will also be used in the test.

I'd like to use it for any new tests, but I thought it might actually be valuable to show that this preserves the behavior tested by the old tests (with some minimal modifications).

You can also take that as my rationalization for not making time to straight up rewrite the existing tests!

src/toolkits/object_detection/od_darknet_yolo_model.cpp
nickjong: Yeah, that's something I'd like to do. Should probably be a separate PR though. Ultimately we should decide whether we actually will ever expose these as user-controllable parameters. If not, they can be implementation details of individual model backends

PR #3012, opened 2020-02-20
clean up some unnecessary files

guihao-liang:
* .flake8.log we don't need to track it in our repo.
* black doesn't need to check py38.
* ignore .pre-commit-config.yaml so that one can have their own customized pre-commit config living together with the repo.


guihao-liang:
intentionally ignore,

* E501, line too long
* E266 too many leading '#' for block comment
* temporarily ignore all __init__.py files

```bash
flake8 src/python/turicreate --config=tox.ini | wc -l
   864
```

tox.ini
```diff
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 88
-extend-ignore = E501
+extend-ignore = E501, E203, E501, E266
```
TobyRoseman: I don't think you need to list `E501` twice.

.gitignore
```diff
@@ -105,3 +105,6 @@ deps/build
 # CCache within Docker
 /.docker_ccache/
 /.docker_images/
+
+# Pre-commit stuff
```
TobyRoseman: I don't really think this comment is really adding anything. I'd probably prefer just not having a comment here than having this comment. A better comment might be `# git hooks`.

guihao-liang: That's better! Will do.

tox.ini
guihao-liang: Yeah, my oversight.

guihao-liang:
> Do we want to add instructions for running flake8 to our [build document](https://github.com/apple/turicreate/blob/master/BUILD.md)?
> 
> Do we want to add `flake8` to our [development requirements.txt](https://github.com/apple/turicreate/blob/master/scripts/requirements.txt)?

No, this flake8 is only used by pre-commit hooks and also personal use. Currently, there are too many things that we need to fix in order to use flake8.

PR #3024, opened 2020-02-27
Mark unused variable

TobyRoseman:


src/ml/neural_net/tf_compute_context.cpp
```diff
@@ -259,7 +259,8 @@ std::unique_ptr<compute_context> create_tf_compute_context() {
 
 // At static-init time, register create_tf_compute_context().
 // TODO: Codify priority levels?
-static auto* tf_registration = new compute_context::registration(
+static auto* tf_registration __attribute__((unused)) =
+  new compute_context::registration(
```
shreyajain17: What is this for?

TobyRoseman: This tells the complier not to issues a warning about this variable being unused. This change was needed to compile on Ubuntu 16.04.

shreyajain17: Got it.

PR #3026, opened 2020-02-28
Use make URLs internal for load_audio and load_images

### LINKED ISSUE #2685: turicreate.load_audio broken with home directory present.
The following code generates a dataset that has no entries: 

```
In [13]: import turicreate as tc
    ...:
    ...: data = tc.load_audio("~/data/ESC-50/audio/")
```

However, the same code where the ~ is replaced by the absolute path name works fine.

TobyRoseman:
Fixes #2685

### DIFF
--- a/src/python/turicreate/toolkits/image_analysis/image_analysis.py
+++ b/src/python/turicreate/toolkits/image_analysis/image_analysis.py
@@ -62,7 +62,9 @@ def load_images(
     ...                                                       recursive=True)
     """
     from ... import extensions as _extensions
+    from ...util import _make_internal_url
 
+    url = _make_internal_url(url)
     return _extensions.load_images(
         url, format, with_path, recursive, ignore_failure, random_order
     )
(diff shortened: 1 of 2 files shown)


src/python/turicreate/toolkits/image_analysis/image_analysis.py
```diff
@@ -62,7 +62,9 @@ def load_images(
     ...                                                       recursive=True)
     """
     from ... import extensions as _extensions
+    from ...util import _make_internal_url
```
guihao-liang: mixing relative and absolute is not a good idea in our project. We'd better choose one style and be coherent with it.

TobyRoseman: I haven't heard of that being a problem. What's your concern?

guihao-liang: > Relative imports for intra-package imports are highly discouraged. Always use the absolute package path for all imports. Even now that PEP 328 [7] is fully implemented in Python 2.5, its style of explicit relative imports is actively discouraged; absolute imports are more portable and usually more readable.

guihao-liang: or you can find it SO,
https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module

PR #3030, opened 2020-03-03
Begin adding checkpointing support for Activity Classification

nickjong:


src/toolkits/activity_classification/activity_classifier.cpp
```diff
@@ -392,13 +409,7 @@ void activity_classifier::train(
   }
 
   state_update["verbose"] = read_state<bool>("verbose");
-  state_update["num_examples"] = data.size();
```
nickjong: Note that before we were giving the number of examples in the `data` argument. If the validation argument was `"auto"` then this included the validation set too. I fixed this along the way, so now we already write the actual training set size.

PR #3034, opened 2020-03-06
Refactor Object Detection inference to use new Model Trainer type

nickjong:


src/toolkits/object_detection/od_darknet_yolo_model_trainer.cpp
```diff
@@ -252,6 +260,7 @@ EncodedInputBatch EncodeDarknetYOLO(InputBatch input_batch,
   result.iteration_id = input_batch.iteration_id;
   result.images = std::move(input_batch.images);
   result.annotations = std::move(input_batch.annotations);
+  result.image_sizes = std::move(input_batch.image_sizes);
```
shreyajain17: What is this for?

src/toolkits/object_detection/od_darknet_yolo_model_trainer.cpp
```diff
@@ -388,7 +430,82 @@ DarknetYOLOModelTrainer::DarknetYOLOModelTrainer(
...
+    std::unique_ptr<data_iterator> test_data, size_t batch_size,
+    float confidence_threshold, float iou_threshold) {
+  // Wrap the data_iterator to incorporate into a Combine pipeline.
+  auto iterator = std::make_shared<DataIterator>(std::move(test_data),
+                                                 batch_size, /* offset */ 0);
+
+  // No labels to encode. Just pass the annotations through for potential
+  // evaluation.
+  auto trivial_encoder = [](InputBatch input_batch) {
+    EncodedInputBatch result;
```
shreyajain17: Why are we doing this?

src/toolkits/object_detection/od_darknet_yolo_model_trainer.cpp
nickjong: Good question. For prediction, we want to yield predicted bounding boxes in image coordinates. Without this, we don't have a convenient way to scale our (normalized) bounding box predictions.

shreyajain17: Makes sense

src/toolkits/object_detection/od_darknet_yolo_model_trainer.cpp
nickjong: The augmenter currently outputs an `InputBatch` but the model backend wants an `EncodedInputBatch`, which is basically `InputBatch` plus some stuff for loss labels that don't apply to the inference case. So we just need to copy all the contents of the `InputBatch` into the `EncodedInputBatch`, since there are no loss labels to encode.

Issue #3003, opened 2020-02-12
One Shot Object Detection: Segmentation Fault (core dumped)


ElectricCarbon:
Hello. When I try to run the one shot object detection example, I receive this error:
I am using Tensorflow GPU with CUDA 10.1 and CUDnn 7
```
Augmenting input images using 951 background images.
+------------------+--------------+------------------+
| Images Augmented | Elapsed Time | Percent Complete |
+------------------+--------------+------------------+
| 100              | 19.46s       | 10.5%            |
| 200              | 25.24s       | 21%              |
| 300              | 34.38s       | 31.5%            |
| 400              | 42.32s       | 42%              |
| 500              | 49.74s       | 52.5%            |
| 600              | 56.26s       | 63%              |
| 700              | 1m 3s        | 73.5%            |
| 800              | 1m 9s        | 84%              |
| 900              | 1m 17s       | 94.5%            |
+------------------+--------------+------------------+
Using 'image' as feature column
Using 'annotation' as annotations column
Segmentation fault (core dumped)

```

ElectricCarbon:
I'm also having the same issue with turicreate running on the CPU.

TobyRoseman:
@ElectricCarbon - Looking the the [TensorFlow Documentation](https://www.tensorflow.org/install/source#gpu), it looks like CUDA 10.0 and cuDNN 7.4 should be used with TensorFlow 2.0.

ElectricCarbon:
> @ElectricCarbon - Looking the the [TensorFlow Documentation](https://www.tensorflow.org/install/source#gpu), it looks like CUDA 10.0 and cuDNN 7.4 should be used with TensorFlow 2.0.

Sorry, I meant tensorflow 2.1.

TobyRoseman:
What version of cuDNN are you using? Also what Operating System?

ElectricCarbon:
I'm 

> What version of cuDNN are you using? Also what Operating System?

cuDNN 7.6.5, Operating System is Ubuntu 18.04.4 LTS

ElectricCarbon:
I am also having the same issue running it on the CPU when the GPUs are set to 0 using `tc.config.set_num_gpus(0)`

LordHansolo:
Hello. I have same problem running one shot object detection example.

OS: Ubuntu 16.04.6 LTS
GPU: GTX1050
CUDA: 10.1
CUDnn: 7.6.4.38-1
Python: 3.5.2
Turicreate: 6.1
[...]

PR #3037, opened 2020-03-09
Better GPU/CPU message

TobyRoseman:
I implemented these changes because I thought it would allow us to sidestep #3003. However that issue turned out to be a much more serious. I'm putting these changes up now because I believe this is improves things.

* Call the correct TensorFlow API given the TensorFlow version.
* Stop implying that multiple GPUs will be used on Linux.
* Simplify redundant code by making the `compute_context` print information about the compute context.


shreyajain17:
This looks good.

PR #3042, opened 2020-03-12
dependency injection on block_cache

guihao-liang:
`block_cache` is the front end of `fixed_size_cache_manager`, which means the former one relies on the later.

If `fiex_size_cache_manager` is destroyed before `block_cache`, bad access will show up.

The motivation here is that to enable the client to use `general_fstream` directly without any global setup and tear down, where the destruction order is explicitly handled.

### DIFF
--- a/src/core/storage/fileio/block_cache.hpp
+++ b/src/core/storage/fileio/block_cache.hpp
@@ -8,6 +8,7 @@
 #include <cstdint>
 #include <core/util/lru.hpp>
 #include <core/parallel/mutex.hpp>
+#include <core/storage/fileio/fixed_size_cache_manager.hpp>
 
 namespace turi {
 class general_ifstream;
@@ -54,6 +55,8 @@ class block_cache {
   /**
    * Constructs the block cache. init must be called before the block_cache
    * can be used.
+   *
+   * make sure the cache manager is detroyed after all cache files are deleted.
    */
   block_cache() = default;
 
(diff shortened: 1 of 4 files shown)


src/core/storage/fileio/fixed_size_cache_manager.cpp
```diff
@@ -127,10 +127,17 @@ namespace fileio {
 /*                                                                       */
 /*************************************************************************/
 
- EXPORT fixed_size_cache_manager& fixed_size_cache_manager::get_instance() {
-   static std::unique_ptr<fixed_size_cache_manager> instance(new fixed_size_cache_manager);
-   return *instance;
- }
+std::shared_ptr<const fixed_size_cache_manager> fixed_size_cache_manager::hold_instance() {
```
guihao-liang: mark

src/core/storage/fileio/block_cache.hpp
```diff
@@ -186,6 +197,11 @@ class block_cache {
   /// whether the block_cache is initialized
   bool m_initialized = false;
 
+  // hold the reference of the underlying cache storage provider
+  // put it at the top of the member variables so it will be destructed at last
+  std::shared_ptr<const fileio::fixed_size_cache_manager> m_cache_provider;
```
guihao-liang: mark

guihao-liang:
the basic idea is to hold a reference of fixed_cache_manager inside of block_cache instance. The same underlying concept with dependency injection.

PR #3045, opened 2020-03-14
enable global aws api initialization

guihao-liang:
subtask of #2920. 

Put all `InitAPI` and `ShutdownAPI` at the global setup teardown stage.

PR #3046, opened 2020-03-14
retry on S3 error code 429

guihao-liang:
client SDK handles retry for network level.

> Each AWS SDK implements automatic retry logic. The AWS SDK for Java automatically retries requests, and you can configure the retry settings using the ClientConfiguration class. For example, you might want to turn off the retry logic for a web page that makes a request with minimal latency and no retries. Use the ClientConfiguration class and provide a maxErrorRetry value of 0 to turn off the retries.

But we need to handle our internal logic explicitly after network error is handled by SDK. 

429 is retried here explicitly, which stands for TOO MANY REQUESTS due to some rate limits.

### DIFF
--- a/src/core/storage/fileio/s3_api.cpp
+++ b/src/core/storage/fileio/s3_api.cpp
@@ -215,8 +215,6 @@ bool parse_s3url(std::string s3_url, s3url& ret, std::string& err_msg) {
 
   boost::trim(url);
 
-  // next char is '/' is credential string is in url
-  // it must be this form: s3://[access_key_id]:[secret_key]:[endpoint][/bucket]/[object_name]
   if (url.empty()) {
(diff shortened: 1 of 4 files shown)


guihao-liang:
sub-commit of #2920, sibling to #3045

src/core/storage/fileio/s3_api.cpp
```diff
@@ -346,249 +405,265 @@ list_objects_response list_objects_impl(s3url parsed_url,
...
+            request.WithContinuationToken(result.GetContinuationToken());
+          }
+
+          // jump out the retry loop
+          break;
+
+        } else {
+          auto error = outcome.GetError().GetResponseCode();
+
+          if (error == Aws::Http::HttpResponseCode::TOO_MANY_REQUESTS) {
```
guihao-liang: retried here.

PR #3047, opened 2020-03-14
unify the format for the encoded s3 url through our codebase

guihao-liang:
```
def _try_inject_s3_credentials(url):
    """
    Inject aws credentials into s3 url as s3://[aws_id]:[aws_key]:[bucket/][objectkey]

    If s3 url already contains secret key/id pairs, just return as is.
    """
```

Bucket format should be consistent. Some places use `[/bucket]` and some places use `[bucket]`

Related to #2920.

### DIFF
--- a/src/core/storage/fileio/general_fstream.hpp
+++ b/src/core/storage/fileio/general_fstream.hpp
@@ -41,7 +41,7 @@ typedef boost::iostreams::stream<fileio_impl::general_fstream_source>
  * but random seeks will be disabled.
  *
  * S3 access keys are mediated by having the filename be of the form
- * s3://[access_key_id]:[secret_key]:[endpoint/][bucket/][object_name]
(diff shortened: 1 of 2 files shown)


src/core/storage/fileio/general_fstream.hpp
```diff
@@ -41,7 +41,7 @@ typedef boost::iostreams::stream<fileio_impl::general_fstream_source>
  * but random seeks will be disabled.
  *
  * S3 access keys are mediated by having the filename be of the form
- * s3://[access_key_id]:[secret_key]:[endpoint][/bucket]/[object_name]
+ * s3://[access_key_id]:[secret_key]:[endpoint/][bucket]/[object_name]
```
guihao-liang: this is actually the format used in our codebase by `parse_s3_url`

PR #3051, opened 2020-03-15
remove redundant ignore

guihao-liang:
I realize that there's no way to ignore this `.pre-commit-config.yaml` while everybody can pull it from GitHub, which requires file to be tracked (not ignored) by git.

In relation to #2920.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,3 @@ deps/build
 # CCache within Docker
 /.docker_ccache/
 /.docker_images/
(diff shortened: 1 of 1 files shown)


guihao-liang:
The intention is to enable people to use their own customized pre-commit hooks so that they can modify this file without adding the modifications to git.

What people can do is to just `pre-commit install -c <path-to-their-own-configurations>`.

PR #3054, opened 2020-03-17
test sframe s3 download and upload

guihao-liang:
add test for frame s3 download and upload.

define env var `DTURI_ENDBALE_SF_S3` to enable the test.

add listed env vars in CI,

```
TURI_S3_REGION
TURI_S3_ENDPOINT
AWS_SECRET_ACCESS_KEY
AWS_ACCESS_KEY_ID
```

to run the test.

### DIFF
--- a/src/python/turicreate/test/test_sframe_s3.py
+++ b/src/python/turicreate/test/test_sframe_s3.py
@@ -8,13 +8,15 @@ from __future__ import print_function as _  # noqa
 from __future__ import division as _  # noqa
 from __future__ import absolute_import as _  # noqa
 from ..data_structures.sframe import SFrame
+from turicreate.util import _assert_sframe_equal
 
 import tempfile
 import os
-import math
 import shutil
 import pytest
 import boto3
+
+remote_sframe_folders = ["small_sframe_dc", "medium_sframe_ac", "big_sframe_od"]
 
 
 @pytest.mark.skipif(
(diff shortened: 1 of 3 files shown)


src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,127 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2020 Apple Inc. All rights reserved.
+#
+# Use of this source code is governed by a BSD-3-clause license that can
+# be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+#
+from __future__ import print_function as _  # noqa
```
TobyRoseman: I don't understand the `noqa` comment on this and the next two lines.

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,127 @@
...
+                            if math.isnan(t1):
+                                self.assertTrue(math.isnan(t2))
+                            else:
+                                self.assertEqual(t1, t2)
+                        else:
+                            self.assertEqual(t1, t2)
+                else:
+                    self.assertEqual(v1, v2)
+
+    def _test_sframe_equal(self, sf1, sf2):
```
TobyRoseman: Can you just use the [method we already have for doing this](https://github.com/apple/turicreate/blob/f57ca7c6af95c7497b01ca3326e945f4d5057a1c/src/python/turicreate/util/__init__.py#L276)?

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,127 @@
...
+
+import tempfile
+import os
+import math
+import shutil
+import pytest
+import boto3
+
+
+@pytest.mark.skipif(
```
TobyRoseman: How slow are these test?

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,127 @@
...
+    @classmethod
+    def setup_class(self):
+        self.my_tempdir = tempfile.mkdtemp()
+        self.s3_client = boto3.client(
+            "s3",
+            endpoint_url=os.environ["TURI_S3_ENDPOINT"],
+            region_name=os.environ["TURI_S3_REGION"],
+        )
+        self.bucket = "tc_qa"
+        self.s3_prefix = "integration/manual/"
```
TobyRoseman: I'm not sure "manual" makes sense in the prefix.

Much more importantly - if two different instances of these tests are running at the same time, they will step on each other. I suggest you generate a UUID and append that to the prefix.

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,127 @@
...
+        s3_url = os.path.join("s3://", self.bucket, obj_key)
+        sf_from_s3 = SFrame(s3_url)
+        self._test_sframe_equal(sf_from_disk, sf_from_s3)
+
+    @pytest.mark.parametrize(
+        "folder", ["big_sframe_od", "medium_sframe_ac", "small_sframe_dc"]
+    )
+    def test_s3_sframe_download_and_upload(self, folder):
+        tmp_folder = os.path.join(self.my_tempdir, folder)
+        try:
```
TobyRoseman: I don't think you should need this. I think it should all be handled by the `teardown_class` method.

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,127 @@
...
+            shutil.rmtree(tmp_folder)
+        except FileNotFoundError:
+            pass
+        read_prefix = os.path.join(self.s3_prefix, "sframes/")
```
TobyRoseman: I'm not really understanding the complexity in the below code. Comments would probably help.

If you want to test S3 downloads and uploads, why can't you just do something like the code below?
```python
s3_path = self.s3_prefix + '/sf'
local_sf.save(s3_path)
downloaded_sf = tc.load_sframe(s3_path)
_assert_sframe_equal(local_sf, downloaded_sf)
```

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: that's my s3 prefix, not the prefix on local fs.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: Just make sure I do download the content if someone forgets to delete temp file again.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: 1. I download from a read-only source to verify my download. 
2. I upload to a scratch space where I can mess up at my will. Since I verified my download, it the round trip succeeds, I'm confident to say the upload is right.

In this way, it indeed wastes 1 download. I can save one download as you suggested.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: and the file is stored in S3, without uuid.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: depends on the network bandwidth. For my network, 2mb/s, it will take around 10 mins.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: Yeah, why not. I hope I know it earlier.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: pass flake8 and pylint.

guihao-liang:
depends on https://github.com/boto/boto3/issues/2347

src/python/turicreate/test/test_sframe_s3.py
TobyRoseman: Right, it looks like it always writes to the same place in S3. So if two instances of this script are running at the same time they will be interfering with each other because they're both writing to the same place.

src/python/turicreate/test/test_sframe_s3.py
TobyRoseman: I'm don't really care about saving a download. The reason I make this suggest was because it make the code simpler and easier to understand. I also think it's more like how our users are likely to use it.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: I believe s3 will attach version id to it. And our internal S3 won't allow us to delete folders.

guihao-liang: that's why I'm not going to have a UUID as a grouping folder.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: you mean, the user cares about downloading more? let's consider this scenario, you download the sframe and do a bunch of data cleaning and then you save it back. Do you think that's a reasonable use case or not?

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,115 @@
...
+import os
+import shutil
+import pytest
+import boto3
+
+remote_sframe_folders = ["small_sframe_dc", "medium_sframe_ac", "big_sframe_od"]
+
+
+@pytest.mark.skipif(
+    os.environ.get("DTURI_ENABLE_SF_S3") is None,
```
hoytak: What's the reason for the D here?  TURI_ENABLE_SF_S3 would be more consistent.

src/python/turicreate/test/test_sframe_s3.py
TobyRoseman: Yes, that's very reasonable.

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -0,0 +1,115 @@
...
+                key = s3_object["Key"]
+                fname = os.path.join(tmp_folder, os.path.basename(key))
+                self.s3_client.download_file(self.bucket, key, fname)
+
+            self.downloaded_files[folder + "_path"] = tmp_folder
+            self.downloaded_files[folder + "_sframe"] = SFrame(tmp_folder)
+
+    @classmethod
+    def teardown_class(self):
+        shutil.rmtree(self.my_tempdir)
```
hoytak: Thank you for this.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: removed `D`

PR #3060, opened 2020-03-20
SFrame shuffle

TobyRoseman:


src/core/storage/sframe_interface/unity_sarray.cpp
```diff
@@ -1764,6 +1764,19 @@ std::shared_ptr<unity_sarray_base> unity_sarray::tail(size_t nrows) {
   return copy_range(start, 1, end);
 }
 
+std::shared_ptr<unity_sarray_base> unity_sarray::make_uniform_int_array(size_t size, size_t max_int) {
+  int seed_hash = 12; // XXX: how to set this properly?
```
hoytak: uint64_t seed = random::pure_random_seed(); 

This is best here because group-by doesn't guarantee order because of the multithreaded aspect of this.  If we want deterministic order, we'll need to do a sort.

src/core/storage/sframe_interface/unity_sarray.cpp
```diff
@@ -1764,6 +1764,19 @@ std::shared_ptr<unity_sarray_base> unity_sarray::tail(size_t nrows) {
...
+
+  auto seq = std::static_pointer_cast<unity_sarray>(
+    unity_sarray::create_sequential_sarray(size, 0, false));
+
+  auto func = [max_int, seed_hash](const flexible_type& val)->flexible_type {
+    uint64_t d = hash64(val.get<flex_int>() ^ seed_hash);
```
hoytak: Use `hash64_combine(hash64(val.get<flex_int>(), seed_hash);`

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1201,73 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+  
+  log_func_entry();
+
+  std::shared_ptr<unity_sframe> temp(new unity_sframe());
+
+  size_t length = this->size();
+
+  std::vector<std::string> column_names = this->column_names();
+
+  const size_t num_buckets = 2;   // XXX: how to determine the number of buckets?
```
hoytak: This is hard to do perfect.  I'd create a global variable, call it TURI_SFRAME_SHUFFLE_BUCKET_SIZE and then just do  I'd just do num_rows() / TURI_SFRAME_SHUFFLE_BUCKET_SIZE.  Default to like 512.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1201,73 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+
+  // Group by concatenate on the random int column. This randomly bucketizes.
+  const std::string buckets_column_name("Buckets");
+  std::shared_ptr<unity_sframe_base> bucketized_sframe = temp->groupby_aggregate({rand_int_column_name},
+                                                                                 {{packed_data_column_name}},
+                                                                                 {buckets_column_name},
+                                                                                 {"__builtin__concat__list__"});
+  std::shared_ptr<unity_sarray_base> bucketized_sarray = bucketized_sframe->select_column(buckets_column_name);
+
+  // XXX: do this in parallel
```
hoytak: Get it working first, then let's parallelize it.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1233,11 +1232,41 @@ std::shared_ptr<unity_sframe_base> unity_sframe::shuffle() {
...
   std::shared_ptr<unity_sarray_base> bucketized_sarray = bucketized_sframe->select_column(buckets_column_name);
 
-  // XXX: Iterate over in parallel and shuffle each group
+  // XXX: do this in parallel
+  // Shuffle each bucket
+  gl_sarray_writer writer(flex_type_enum::LIST, 1);
+  gl_sarray gl_bucketized_sarray = gl_sarray(bucketized_sarray);
+  gl_sarray_range ra = gl_bucketized_sarray.range_iterator();
+  auto cur_bucket = ra.begin();
+  unsigned int seed = std::chrono::system_clock::now().time_since_epoch().count();
```
hoytak: Can use random::pure_random_seed() here.

src/core/storage/sframe_interface/unity_sframe.cpp
hoytak: To do this in parallel, use this code template: 
```
size_t num_threads = thread::cpu_count();

gl_sarray_writer writer(flex_type_enum::LIST, num_threads);

in_parallel([&](size_t thread_id, size_t n_threads) { 
...
size_t idx_start = (gl_bucketized_sarray.num_rows() * thread_idx) / num_threads;
size_t idx_end = (gl_bucketized_sarray.num_rows() * (thread_idx + 1) ) / num_threads;
 
gl_sarray_range ra = gl_bucketized_sarray.range_iterator(idx_start, idx_end);

// Same code below, just write to the correct segment.

});

// Now close the writer
```

src/core/storage/sframe_interface/unity_sarray.cpp
TobyRoseman: Done. 

Yeah, I was hoping we could expose a `seed` parameter for `shuffle` in the Python layer but that's not going to be possible.

src/core/storage/sframe_interface/unity_sarray.cpp
TobyRoseman: Done.

FYI: I didn't call `hash64` on the seed. Doesn't really seem like there's much point to doing that.

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: Good idea. Done.

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: Nice. Done.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1202,72 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+
+  // Create a column of random ints between 0 and (num_buckets - 1).
+  auto temp_groupby_column = std::static_pointer_cast<unity_sarray>(
+    unity_sarray::make_uniform_int_array(this->size(), num_buckets));
+  const std::string rand_int_column_name("Random Ints");
+  std::shared_ptr<unity_sframe> temp(new unity_sframe());
+  temp->add_column(temp_groupby_column, rand_int_column_name);
+
+  // Pack columns so we can group by concatenate
+  std::shared_ptr<unity_sarray_base> packed_columns = this->pack_columns(column_names,
```
guihao-liang: great! Reduces overall random read times from the original shuffle index then read from index approach!

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: Thanks @hoytak - I was able to use this code almost verbatim.

TobyRoseman:
@hoytak - I believe this is now ready for final review.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1202,80 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+  gl_sarray_writer writer(flex_type_enum::LIST, num_threads);
+  gl_sarray gl_bucketized_sarray = gl_sarray(bucketized_sarray);
+
+  in_parallel([&](size_t thread_id, size_t n_threads) {
+      size_t idx_start = (gl_bucketized_sarray.size() * thread_id) / n_threads;
+      size_t idx_end = (gl_bucketized_sarray.size() * (thread_id + 1) ) / n_threads;
+
+      gl_sarray_range ra = gl_bucketized_sarray.range_iterator(idx_start, idx_end);
+      auto cur_bucket = ra.begin();
+      unsigned int seed = random::pure_random_seed();
```
guihao-liang: narrow from 64 bits to 32 bits might not be a good idea.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1202,80 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+      while (cur_bucket != ra.end()) {
+        // shuffle the indexes for the current bucket
+        std::vector<int> indexes = std::vector<int>(cur_bucket->size());
```
guihao-liang: I think auto is more readable here.

guihao-liang: you may consider using std::itoa to initialize the indices.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1202,80 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+        for (size_t i = 0; i < cur_bucket->size(); i++) {
+          indexes[i] = i;
+        }
+        std::shuffle(indexes.begin(), indexes.end(), std::default_random_engine(seed));
+
+        // output in random order
+        for (size_t i = 0; i < cur_bucket->size(); i++) {
```
guihao-liang: you may consider a sequential read of all into a vector of rows and then shuffle in memory. Random read from sarray is slow, Correct me if I'm wrong.

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: Good point. This is actually results in a compiler error on our iOS dylib build.

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: The random reads aren't from an SArray. `cur_bucket` is a flexible type (flex_list).

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: Ok, changed to `auto`.

I'm not sure how `std::itoa` makes sense here. I'm not trying to convert ints to strings.

guihao-liang: Sorry, I mean this iota https://en.cppreference.com/w/cpp/algorithm/iota

src/core/storage/sframe_interface/unity_sframe.cpp
guihao-liang: ah, got it.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1202,80 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+  gl_sarray_writer writer(flex_type_enum::LIST, num_threads);
+  gl_sarray gl_bucketized_sarray = gl_sarray(bucketized_sarray);
+
+  in_parallel([&](size_t thread_id, size_t n_threads) {
+      size_t idx_start = (gl_bucketized_sarray.size() * thread_id) / n_threads;
+      size_t idx_end = (gl_bucketized_sarray.size() * (thread_id + 1) ) / n_threads;
+
+      gl_sarray_range ra = gl_bucketized_sarray.range_iterator(idx_start, idx_end);
+      auto cur_bucket = ra.begin();
+      unsigned int seed = std::chrono::system_clock::now().time_since_epoch().count();
```
guihao-liang: why don't we use the pure_random function? Just curious.

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: OK. Done.

src/core/storage/sframe_interface/unity_sframe.cpp
TobyRoseman: I agree exposing a seed parameter to the end user would be nice. However it's not possible. We have to use groupby in order to get a linear time solution. Groupby is non-deterministic. So it doesn't take a seed parameter. So we can't either.

guihao-liang: I think groupby is deterministic if the hash key is generated as the same.

guihao-liang: okay, I see, elements in the bucket are not. Well, I agree. It's hard.

guihao-liang: The only way I can think of is to do the indirect sort on each bucket (sort their index). And then shuffle the indirectly sorted index vector.

TobyRoseman: I don't know what you mean by "indirect sort on each bucket". However if you're doing any sorting, then I don't think the solution can have a linear running time.

guihao-liang: for each hash key, you have a bucket, which is your list (vector<flex>) here. When you perform indirect sort, you have a separate `vector<size_t> indices`, which represents the index in the bucket.

```
indices.sort(ii, jj) { bucket[ii] < bucket[jj]; }
```

This will save extra copies on elements in bucket.

Nlog(N/B) time is not bad. You already save tons of random reads that involve io.

guihao-liang: The worst case is NlogN is when all elements go into the same bucket. It depends on whether it's worthy of sacrificing extra NlogN - N **memory read** for a reproducible model. 

For most of our inputs, N is less than 10k, log(10k) is 9. Assume each memory random read is 1us, and io read is 1ms.

Assume a sframe can be fully fit into memory, there's no io read from disk. The read will be N * 1us = 100ms.

The worst penalty for memory read is 80k more reads if we sort the bucket. you pay extra 800 ms. Not that bad.

If sframe is very large, more than 2G, which is the cache manager capacity. we may have some read that needs to be load from disk. Let's say only 10% of them will miss the cache and force the load from disk, you pay N*10%*1ms. If N is 10k, then it's about 1000ms.

src/core/storage/sframe_interface/unity_sframe.cpp
```diff
@@ -1200,6 +1202,78 @@ unity_sframe::random_split(float percent, uint64_t random_seed, bool exact) {
...
+      size_t idx_end = (gl_bucketized_sarray.size() * (thread_id + 1) ) / n_threads;
+
+      gl_sarray_range ra = gl_bucketized_sarray.range_iterator(idx_start, idx_end);
+      auto cur_bucket = ra.begin();
+      unsigned int seed = static_cast<unsigned int>(random::pure_random_seed());
+      while (cur_bucket != ra.end()) {
+        // shuffle the indexes for the current bucket
+        auto indexes = std::vector<int>(cur_bucket->size());
+        std::iota(indexes.begin(), indexes.end(), 0);
+        std::shuffle(indexes.begin(), indexes.end(), std::default_random_engine(seed));
```
hoytak, line 1250: nit: we can just do `auto rng = std::default_random_engine(seed);` once in each thread, then pass that into the std::shuffle here. Subsequent draws will still be random.

PR #3070, opened 2020-03-26
put back the s3 url sanitize test

guihao-liang:
sub-task of #2920.

modify the parse_s3url to reject the invalid domain name so sanitze_aggressive can do further work.

### DIFF
--- a/src/core/storage/fileio/s3_api.cpp
+++ b/src/core/storage/fileio/s3_api.cpp
@@ -222,6 +222,7 @@ bool parse_s3url(std::string s3_url, s3url& ret, std::string& err_msg) {
     return false;
   }
 
+  // this is a bad design
   auto original_url = sanitize_url(url);
 
   // The rest is parsed using boost::tokenizer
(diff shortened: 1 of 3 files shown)


src/core/storage/fileio/s3_api.cpp
```diff
@@ -240,7 +249,23 @@ bool parse_s3url(std::string s3_url, s3url& ret, std::string& err_msg) {
...
+
+    std::vector<std::string> subs;
+    boost::algorithm::split(subs, endpoint, [](char c) { return c == '.'; });
+    bool is_valid = std::all_of(
+        std::begin(subs), std::end(subs), [](const std::string& name) {
+          return std::regex_match(name, std::regex("[:w:]+"));
+        });
+
+    if (!is_valid) {
+      ss << "endpoint name: " << endpoint << "contains invalid chars" << url
```
hoytak: nit -- spaces here.   `<< " contains`

src/core/storage/fileio/s3_api.cpp
```diff
@@ -240,7 +249,23 @@ bool parse_s3url(std::string s3_url, s3url& ret, std::string& err_msg) {
...
+         << __FILE__ << "at" << __LINE__;
```
hoytak: spaces here too

PR #3072, opened 2020-03-27
enable s3 tests in internal gitlab runners

guihao-liang:


src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -23,7 +23,7 @@
 
 
 @pytest.mark.skipif(
-    os.environ.get("TURI_ENABLE_SF_S3") is None,
+    os.environ.get("TC_ENABLE_S3_TESTS", "0") == "1",
```
hoytak: Doesn't this skip it if it's 1?

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -47,7 +47,7 @@ def setup_class(self):
             # force clean in case same tempdir is reused without cleaning
             try:
                 shutil.rmtree(tmp_folder)
-            except FileNotFoundError:
+            except OSError:
```
hoytak: Just change this to :
```
except Exception, e:
     warning.warn("Error raised while cleaning up %s: %s" % (tmp_folder, str(e)))
```

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: do you set that to 1 to skip? I though 1 means to turn it on.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: that would be good.

src/python/turicreate/test/test_sframe_s3.py
guihao-liang: ah, may bad.

PR #3074, opened 2020-03-28
Fix non-deterministic sound classifier unit test

### LINKED ISSUE #2949: Non-deterministic unit test failures in test_audio_functionality
There are two non-deterministic unit test failures that we have been able to repro on macOS on Python 3.5 and 3.7 and on Linux on Python 3.5.

```
=================================== FAILURES ===================================
_______________ ClassifierTestTwoClassesIntLabels.test_classify ________________

self = <turicreate.test.test_audio_functionality.ClassifierTestTwoClassesIntLabels testMethod=test_classify>

    def test_classify(self):
        classification = self.model.classify(self.data)
        for a, b in zip(classification['class'], self.data['labels']):
>           self.assertEqual(a, b)
E           AssertionError: 0 != 1

test_audio_functionality.py:318: AssertionError
---------------------------- Captured stdout setup -----------------------------

Training a custom neural network -
+-------------------+-------------------+-------------------+
| Iteration         | Training Accuracy | Elapsed Time      |
+-------------------+-------------------+-------------------+
| 1                 | 0.500             | 1.189             |
+-------------------+-------------------+-------------------+
| 2                 | 0.500             | 1.196             |
+-------------------+-------------------+-------------------+
| 3                 | 0.500             | 1.202             |
+-------------------+-------------------+-------------------+
| 4                 | 0.500             | 1.208             |
+-------------------+-------------------+-------------------+
| 5                 | 0.500             | 1.215             |
+-------------------+-------------------+-------------------+
| 6                 | 0.500             | 1.221             |
+-------------------+-------------------+-------------------+
| 7                 | 0.500             | 1.227             |
+-------------------+-------------------+-------------------+
| 8                 | 0.500             | 1.234             |
+-------------------+-------------------+-------------------+
| 9                 | 0.500             | 1.240             |
+-------------------+-------------------+-------------------+
| 10                | 0.833             | 1.247             |
+-------------------+-------------------+-------------------+
________________ ClassifierTestTwoClassesIntLabels.test_predict ________________

self = <turicreate.test.test_audio_functionality.ClassifierTestTwoClassesIntLabels testMethod=test_predict>

    def test_predict(self):
        # default ('class') output_type
        predictions = self.model.predict(self.data['audio'])
        _raise_error_if_not_sarray(predictions)
        self.assertEqual(len(predictions), len(self.data))
        for a, b in zip(predictions, self.data['labels']):
>           self.assertEqual(a, b)
E           AssertionError: 0 != 1

test_audio_functionality.py:197: AssertionError
```

TobyRoseman:
Fixes #2949 

All the other Sound Classifier models where correctness of predictions are checked, have their max iterations set to 100. It just seems like an oversight that this model wasn't having that set too.

Without this change, I was only able to run the test 16 times before I got a failure. With this change, I've been able to run it 100 times without failure.

The "training data" is only six examples that seems to be why we need the large number of iterations. Also because the data is so small, increasing the number of iteration doesn't seem to have an appreciable affect on running time of the unit tests.

### DIFF
--- a/src/python/turicreate/test/test_audio_functionality.py
+++ b/src/python/turicreate/test/test_audio_functionality.py
@@ -455,22 +455,9 @@ class ClassifierTestTwoClassesIntLabels(ClassifierTestTwoClassesStringLabels):
             feature="audio",
             custom_layer_sizes=layer_sizes,
             validation_set=None,
+            max_iterations=100,
         )
         assert self.model.custom_layer_sizes == layer_sizes
-
-    # Remove the following two tests after #2949 is fixed!
-
-    @pytest.mark.xfail(
-        reason="Non-deterministic test failure tracked in https://github.com/apple/turicreate/issues/2949"
-    )
-    def test_classify(self):
-        pass
-
-    @pytest.mark.xfail(
-        reason="Non-deterministic test failure tracked in https://github.com/apple/turicreate/issues/2949"
-    )
-    def test_predict(self):
-        pass
 
 
 class ClassifierTestThreeClassesStringLabels(ClassifierTestTwoClassesStringLabels):


fareeha:
Good catch!

PR #3087, opened 2020-04-02
Revert "Pin TensorFlow version on Linux"

TobyRoseman:
Reverts apple/turicreate#3085


TobyRoseman:
I should not have merged this. This breaks things for Python 2.7 on Linux. Will put up an updated PR soon. Reverting this change for now.

PR #3089, opened 2020-04-03
Restrict TensorFlow Version on Linux

TobyRoseman:
Temporary workaround for #3003

#3085 was a first attempt at doing this. That approach didn't work because TensorFlow 2.0.0 supports Python 2.7 but 2.0.1 does not. This solves that problem by allowing Python 2.7 users to use TensorFlow 2.0.0 but Python 3 users can use TensorFlow 2.0.1.


TobyRoseman:
Passes internal tests.

PR #3090, opened 2020-04-03
fix test error

guihao-liang:
suppress error if the file is deleted before manual clean up.

### DIFF
--- a/src/python/turicreate/test/test_prototype_models.py
+++ b/src/python/turicreate/test/test_prototype_models.py
@@ -2,6 +2,7 @@ import turicreate as tc
 import unittest
 import tempfile
 import shutil
+import errno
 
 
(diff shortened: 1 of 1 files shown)


guihao-liang:
> `e` if the file exists)?

My understanding of the pythonic way is to use exceptions since exception handling is quite efficient.

TobyRoseman:
> > `e` if the file exists)?
> 
> My understanding of the pythonic way is to use exceptions since exception handling is quite efficient.

Yeah, that's fair. It is considered Pythonic to use `try`/`except` with the idea that It's [easier to ask for forgiveness than permission](https://docs.python.org/3.5/glossary.html#term-eafp). I just think the code would be simpler if you checked first, rather than handling an exception. 

I'm good with it either way; go ahead and merge.

PR #3093, opened 2020-04-04
Add random_seed parameter to object_detector

### LINKED ISSUE #2168: Object Detector Needs Seed Parameter

nickjong:
- Plumb random seed parameter from Python to data augmentation
- Plumb random seed from C++ into TensorFlow image augmenter

Incorporates some changes originally authored by Henry Tao (jakesabathia2)

Fixes #2168

### DIFF
--- a/src/python/turicreate/test/test_object_detector.py
+++ b/src/python/turicreate/test/test_object_detector.py
@@ -150,7 +150,7 @@ class ObjectDetectorTest(unittest.TestCase):
         self.get_ans["annotation_origin"] = lambda x: isinstance(x, str)
         self.get_ans["grid_height"] = lambda x: x > 0
         self.get_ans["grid_width"] = lambda x: x > 0
-        self.get_ans["random_seed"] = lambda x: True
+        self.get_ans["random_seed"] = lambda x: isinstance(x, int)
         self.get_ans["verbose"] = lambda x: True
         del self.get_ans["_model"]
         del self.get_ans["_class_to_index"]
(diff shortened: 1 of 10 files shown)


nickjong:
It turns out that the most robust thing to do is not to use TensorFlow's random number generation at all. By generating a batch-specific seed in C++, we can use numpy for all the Python-side RNG, seeded by the C++ seed.

This has some advantages over the approach from #2892:
1. Sampling a numpy random number table during image augmentation initialization implies that we're using the same results for each iteration. For example, if the first image of the first batch skips horizontal flip, then the first image of every batch will skip horizontal flip.
2. Relying on TF RNG, which is scoped to sessions, will make it difficult to support resuming from a checkpoint with the correct semantics.

src/python/turicreate/test/test_object_detector.py
```diff
@@ -288,6 +288,21 @@ def test_create_with_verbose_False(self):
...
+         pred_1 = model_1.predict(self.sf)
+         model_2 = tc.object_detector.create(self.sf, max_iterations=max_iterations, random_seed=random_seed)
+         pred_2 = model_2.predict(self.sf)
+
+         self.assertEqual(len(pred_1), len(pred_2))
+         for i in range(len(pred_1)):
+             self.assertEqual(len(pred_1[i]),len(pred_2[i]))
+             for j in range(len(pred_1[i])):
+                 self.assertEqual(pred_1[i][j], pred_2[i][j])
+
```
TobyRoseman, line 305: I think this test is good, but it's really only testing that the same seed produces the same model with your local changes. It's not testing that both master and your local changes produce the same model with the same seed.

I suggest we hard code the probabilities for one row of the data here and check that those match.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -490,20 +501,29 @@ def crop_augmenter(
...
+    output_height,
+    output_width
+):
     img_tf, ann_tf = tf.numpy_function(
-        func=horizontal_flip_augmenter,
-        inp=[img_tf, ann_tf],
-        Tout=[tf.float32, tf.float32],
+        func=numpy_augmenter,
+        inp=[img_tf, ann_tf, seed_tf],
+        Tout=[tf.float32, tf.float32]
```
TobyRoseman, line 523: Major nit: I actually generally prefer keeping trailing comma. From [PEP8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas):
```
When trailing commas are redundant, they are often helpful when a version control system is used, when a list of values, arguments or imported items is expected to be extended over time. The pattern is to put each value (etc.) on a line by itself, always adding a trailing comma, and add the close parenthesis/bracket/brace on the next line. 
```

src/python/turicreate/test/test_object_detector.py
nickjong: I don't think we're in a position to guarantee that the same golden output will remain valid. We already have to accept some numerical differences across platforms, and we can't control minor behavior changes in our dependencies across their versions. For example, we inherit whatever the semantics are for how numpy uses random seeding.

I do think we should revisit this test in general, since it's quite heavyweight for what it does. I filed #3094

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
nickjong: I honestly did not consciously intend to remove that comma! I actually to punctuate in this fashion too (for languages/syntaxes that support it), for the same reason

PR #3097, opened 2020-04-07
propagate exception

guihao-liang:
The root cause for Linux segfault might be caused by `fin` continues to read and deserialize garbage data when underlying file IO fails.

```
void block_manager::init_segment(std::shared_ptr<block_manager::segment>& seg) {
  ...
  // read 8 bytes
  fin->read(reinterpret_cast<char*>(&footer_size), sizeof(footer_size));

  // deserialize the block information
  fin->clear();
  fin->seekg(filesize - footer_size - sizeof(footer_size), std::ios_base::beg);
  iarchive iarc(*fin);
  iarc >> seg->blocks;

  seg->inited = true;
  seg->file_size = filesize;
}
```

One should always check `fin.good()` after using it. As @ylow mentioned, most designs are based on the assumption that the retry logic (exception handling may involve) happens at very low-level IO operations and those IO will succeed in the end, people tend to not to check `fin` state bits.

This change will enable IO to throw since by default, IO sets its state bits instead of throwing errors.

After discussing with @ylow, I think it's reasonable to propagate exception on `badbit` because those types of errors usually mean irrecoverable.

This is a behavioral change on IO. And it's a good change.

### DIFF
--- a/src/core/storage/fileio/general_fstream.cpp
+++ b/src/core/storage/fileio/general_fstream.cpp
@@ -23,6 +23,7 @@ general_ifstream::general_ifstream(std::string filename)
     // every conceieved.
     try :general_ifstream_base(filename), opened_filename(filename)
     {
+      exceptions(std::ios_base::badbit);
     }  catch (const std::exception& e) {
       log_and_throw_io_failure("Cannot open " + sanitize_url(filename) + " for read. " + e.what());
     } catch (std::string e) {
@@ -35,6 +36,7 @@ general_ifstream::general_ifstream(std::string filename)
 general_ifstream::general_ifstream(std::string filename, bool gzip_compressed)
     try :general_ifstream_base(filename, gzip_compressed),
      opened_filename(filename) {
+       exceptions(std::ios_base::badbit);
      } catch (const std::exception& e) {
       log_and_throw_io_failure("Cannot open " + sanitize_url(filename) + " for read. " + e.what());
     } catch (std::string e) {
@@ -67,6 +69,7 @@ std::shared_ptr<std::istream> general_ifstream::get_underlying_stream() {
 general_ofstream::general_ofstream(std::string filename)
     try :general_ofstream_base(filename), opened_filename(filename) {
       // this space intentionally left blank
+       exceptions(std::ios_base::failbit);
     } catch (const std::ios_base::failure& e) {
       // an I/O error is already the exception type - just re-throw
       throw;
@@ -81,7 +84,9 @@ general_ofstream::general_ofstream(std::string filename)
 
 general_ofstream::general_ofstream(std::string filename, bool gzip_compress)
     try :general_ofstream_base(filename, gzip_compress),
-     opened_filename(filename) { }  catch (std::exception e) {
+     opened_filename(filename) {
+       exceptions(std::ios_base::failbit);
+     }  catch (std::exception e) {
       log_and_throw_io_failure("Cannot open " + sanitize_url(filename) + " for write. " + e.what());
     } catch (std::string e) {
       log_and_throw_io_failure("Cannot open " + sanitize_url(filename) + " for write. " + e);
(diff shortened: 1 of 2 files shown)


guihao-liang:
Unexpectedly included some format changes. I hope it won't bother.

The general idea behind is that `general_fstream` wraps `union_fstream`. In order to let exception to be poped, we need to turn the gate open at each wrapping layer.

src/core/storage/fileio/general_fstream.cpp
```diff
@@ -21,27 +22,33 @@ general_ifstream::general_ifstream(std::string filename)
...
-     } catch (const std::exception& e) {
-      log_and_throw_io_failure("Cannot open " + sanitize_url(filename) + " for read. " + e.what());
-    } catch (std::string e) {
-      log_and_throw_io_failure("Cannot open " + sanitize_url(filename) + " for read. " + e);
-    } catch (...) {
-      log_and_throw_io_failure("Cannot open " + sanitize_url(filename));
-    }
+    try : general_ifstream_base(filename),
+          opened_filename(filename) {
+  exceptions(std::ios_base::badbit);
```
guihao-liang, line 27: That's why you may need, http://www.cplusplus.com/reference/ios/ios/exceptions/

badbit includes failbit, anyway.

src/core/storage/fileio/union_fstream.cpp
```diff
@@ -148,34 +153,29 @@ union_fstream::union_fstream(std::string url,
...
 /// Returns the current stream type. Whether it is a HDFS stream or a STD stream
-union_fstream::stream_type union_fstream::get_type() const {
-  return type;
-}
+union_fstream::stream_type union_fstream::get_type() const { return type; }
 
-std::shared_ptr<std::istream> union_fstream::get_istream () {
+std::shared_ptr<std::istream> union_fstream::get_istream() {
   ASSERT_TRUE(input_stream != NULL);
+  input_stream->exceptions(std::ios_base::badbit);
```
guihao-liang, line 166: this line and line 172 is what you need to review.

guihao-liang:
passed all internal tests.

src/core/storage/fileio/general_fstream.cpp
```diff
@@ -1,13 +1,14 @@
 /* Copyright © 2017 Apple Inc. All rights reserved.
  *
  * Use of this source code is governed by a BSD-3-clause license that can
- * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ * be found in the LICENSE.txt file or at
```
TobyRoseman, line 4: I'd really prefer not add a new line here. Let's keep all of our copyright headers the same. At some point we might want to right a script to change our headers and I could see something like this breaking that.

src/core/storage/fileio/general_fstream.cpp
```diff
@@ -51,9 +58,7 @@ size_t general_ifstream::get_bytes_read() {
   return this->general_ifstream_base::operator->()->get_bytes_read();
 }
 
-std::string general_ifstream::filename() const {
-  return opened_filename;
-}
+std::string general_ifstream::filename() const { return opened_filename; }
```
TobyRoseman, line 61: Should the return value be `const`?

src/core/storage/fileio/union_fstream.cpp
```diff
@@ -1,9 +1,11 @@
 /* Copyright © 2017 Apple Inc. All rights reserved.
  *
  * Use of this source code is governed by a BSD-3-clause license that can
- * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ * be found in the LICENSE.txt file or at
```
TobyRoseman, line 4: Same as my first comment.

src/core/storage/fileio/general_fstream.cpp
guihao-liang: Or we can run clang-format on all of our files to make this exception to be normal. Hoyt had some scripts adding headers and those scripts work on this kind of scenario. Again, if adding headers, a program should be robust for block and line comments.

src/core/storage/fileio/general_fstream.cpp
guihao-liang: it's a copy.

PR #3098, opened 2020-04-07
better retry approach

guihao-liang:
`ShouldRetry` is using 

```
        inline bool IsRetryableHttpResponseCode(HttpResponseCode responseCode)
        {
            switch (responseCode)
            {
                case HttpResponseCode::INTERNAL_SERVER_ERROR:
                case HttpResponseCode::SERVICE_UNAVAILABLE:
                case HttpResponseCode::TOO_MANY_REQUESTS:
                case HttpResponseCode::BANDWIDTH_LIMIT_EXCEEDED:
                case HttpResponseCode::REQUEST_TIMEOUT:
                case HttpResponseCode::AUTHENTICATION_TIMEOUT:
                case HttpResponseCode::LOGIN_TIMEOUT:
                case HttpResponseCode::GATEWAY_TIMEOUT:
                case HttpResponseCode::NETWORK_READ_TIMEOUT:
                case HttpResponseCode::NETWORK_CONNECT_TIMEOUT:
                    return true;
                default:
                    return false;
            }
        }
```

which is better for replacing my hardcoded approach.

Also, I reduce the connection time as well.

### DIFF
--- a/src/core/storage/fileio/s3_api.hpp
+++ b/src/core/storage/fileio/s3_api.hpp
@@ -286,11 +286,14 @@ std::ostream& reportS3Error(std::ostream& ss, const s3url& parsed_url,
                             S3Operation::ops_enum operation,
                             const Aws::Client::ClientConfiguration& config,
                             const Response& outcome) {
+  auto error = outcome.GetError();
   ss << "('" << parsed_url << ", proxy: '" << config.proxyHost << "', region: '"
      << config.region << "')"
      << " Error while performing " << S3Operation::toString(operation)
-     << ". Error Name: " << outcome.GetError().GetExceptionName()
-     << ". Error Message: " << outcome.GetError().GetMessage();
+     << ". Error Name: " << error.GetExceptionName()
+     << ". Error Message: " << error.GetMessage()
+     << ". HTTP Error Code: " << static_cast<int>(error.GetResponseCode());
+
   return ss;
 }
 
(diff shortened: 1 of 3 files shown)


test/fileio/s3_filesys_test.cpp
```diff
@@ -91,15 +91,15 @@ int main(int argc, char** argv) {
...
-    turi::file_download_cache::get_instance().clear();
-    turi::block_cache::release_instance();
-
   } catch (std::string& e) {
     std::cerr << "Exception: " << e << std::endl;
   } catch (std::exception& e) {
     std::cerr << "Exception: " << e.what() << std::endl;
   }
 
+  /* teardown manually */
```
guihao-liang, line 100: teardown should not be skipped. This might cause a core dump during manual tests.

src/core/storage/fileio/s3_api.hpp
```diff
@@ -286,11 +286,14 @@ std::ostream& reportS3Error(std::ostream& ss, const s3url& parsed_url,
                             S3Operation::ops_enum operation,
                             const Aws::Client::ClientConfiguration& config,
                             const Response& outcome) {
+  auto error = outcome.GetError();
   ss << "('" << parsed_url << ", proxy: '" << config.proxyHost << "', region: '"
      << config.region << "')"
      << " Error while performing " << S3Operation::toString(operation)
-     << ". Error Name: " << outcome.GetError().GetExceptionName()
-     << ". Error Message: " << outcome.GetError().GetMessage();
+     << ". Error Name: " << error.GetExceptionName()
```
guihao-liang, line 293: Aws didn't provide a good enum to string method to better render the error message. Often the time, exception name and message is empty. So we need to add HTTP code to better tace error on the client-side.

guihao-liang:
Passed all internal tests. Good to go.

PR #3100, opened 2020-04-08
only use path style be default

guihao-liang:
path style: host/bucket/key
virtual address: bucket.host/key

The new SDK uses a virtual address by default which causes backward compatible issues (because virtual address DNS is not available for normal dev environment).

Well, at least it works for all production environments.

related to https://github.com/aws/aws-sdk-cpp/issues/1357

### DIFF
--- a/src/core/storage/fileio/s3_api.cpp
+++ b/src/core/storage/fileio/s3_api.cpp
@@ -162,12 +162,18 @@ S3Client init_aws_sdk_with_turi_env(s3url& parsed_url) {
   }
 
   if (parsed_url.secret_key.empty()) {
-    return S3Client(clientConfiguration);
+    return S3Client(clientConfiguration,
+                    /* default value */
+                    Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy::Never,
+                    /* use virtual address */ false);
   } else {
     // credentials
     Aws::Auth::AWSCredentials credentials(parsed_url.access_key_id.c_str(),
                                           parsed_url.secret_key.c_str());
-    return S3Client(credentials, clientConfiguration);
+    return S3Client(credentials, clientConfiguration,
+                    /* default value */
+                    Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy::Never,
+                    /* use virtual address */ false);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/core/storage/fileio/s3_api.cpp
```diff
@@ -410,7 +417,7 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
   }
 
   clientConfiguration.proxyHost = proxy.c_str();
-  clientConfiguration.requestTimeoutMs = 5 * 60000;
+  clientConfiguration.requestTimeoutMs = 10000;
```
guihao-liang, line 420: 10s is enough for the list. There should be a global place to set all these configuration properties. Leave it for future people.

src/python/turicreate/test/test_sframe_s3.py
```diff
@@ -21,7 +21,8 @@
 # size from small to big: 76K, 21MB, 77MB.
 # 64MB is the cache block size. The big sframe with 77MB is used to
 # ensure there's no issues when crossing different cache blocks.
-remote_sframe_folders = ["small_sframe_dc", "medium_sframe_ac", "big_sframe_od"]
+# remote_sframe_folders = ["small_sframe_dc", "medium_sframe_ac", "big_sframe_od"]
```
TobyRoseman, line 24: You probably don't want to keep this commented out?

src/core/storage/fileio/s3_api.cpp
guihao-liang: well, it's recommended by blobby to set it to 5 mins.

guihao-liang: since we download by 64 MB, 5 mins are too many. let's reduce it to 2 mins.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -419,7 +426,9 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
     parsed_url.sdk_region = region;
   }
 
-  S3Client client(credentials, clientConfiguration);
+  S3Client client(credentials, clientConfiguration,
```
TobyRoseman, line 429: This one doesn't seem to have your helpful inline comment. Those would be nice to add.

guihao-liang: Yeah, this duplicate part will be removed anyway.

Issue #3101, opened 2020-04-09
New S3 Code Path Ignores TuriCreate SSL Runtime Config Values


TobyRoseman:
The [runtime config values](https://apple.github.io/turicreate/docs/api/generated/turicreate.config.set_runtime_config.html#turicreate.config.set_runtime_config) for `TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE` and `TURI_FILEIO_ALTERNATIVE_SSL_CERT_DIR` are not used by the new S3 code path.

TobyRoseman:
Fixed by #3102

PR #3102, opened 2020-04-09
support  TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE;

guihao-liang:
I have to admit I overlooked this one.

### DIFF
--- a/src/core/storage/fileio/s3_api.cpp
+++ b/src/core/storage/fileio/s3_api.cpp
@@ -147,18 +147,25 @@ S3Client init_aws_sdk_with_turi_env(s3url& parsed_url) {
 
   // TODO: add proxy support
   // clientConfiguration.proxyHost = proxy.c_str();
+
+  // set path or file for ssl certs
   if (!get_alternative_ssl_cert_file().empty()) {
     auto fstaus = get_file_status(get_alternative_ssl_cert_file());
     ASSERT_TRUE(fstaus.second.empty());
-    switch (fstaus.first) {
-      case file_status::REGULAR_FILE:
-        clientConfiguration.caFile = get_alternative_ssl_cert_file().c_str();
-        break;
-      case file_status::DIRECTORY:
-        clientConfiguration.caPath = get_alternative_ssl_cert_file().c_str();
-        break;
-      default:
-        log_and_throw("Invalid path or file for alternative SSL certificate");
+    if (fstaus.first == file_status::REGULAR_FILE) {
+      clientConfiguration.caFile = get_alternative_ssl_cert_file().c_str();
+    } else {
+      log_and_throw("Invalid file for alternative SSL certificate");
+    }
+  }
+
+  if (!get_alternative_ssl_cert_dir().empty()) {
+    auto fstaus = get_file_status(get_alternative_ssl_cert_dir());
+    ASSERT_TRUE(fstaus.second.empty());
+    if (fstaus.first == file_status::DIRECTORY) {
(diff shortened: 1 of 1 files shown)


TobyRoseman:
fixes #3101

guihao-liang:
> It looks like this only works for `TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE`. We also want it to work for `TURI_FILEIO_ALTERNATIVE_SSL_CERT_DIR`.

Just realized that. It should be good now.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -147,6 +147,20 @@ S3Client init_aws_sdk_with_turi_env(s3url& parsed_url) {
 
   // TODO: add proxy support
   // clientConfiguration.proxyHost = proxy.c_str();
+  if (!get_alternative_ssl_cert_file().empty()) {
+    auto fstaus = get_file_status(get_alternative_ssl_cert_file());
```
TobyRoseman, line 153: I don't think this is the right way. I think you should be getting the `clientConfiguration.caPath` value from `get_alternative_ssl_cert_dir()` not from `get_alternative_ssl_cert_file()`.

Also I'm not sure it makes sense but I think the should be possible for both `clientConfiguration.caPath` and `clientConfiguration.caFile` to be set.

guihao-liang:
passed the internal test.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -148,6 +148,27 @@ S3Client init_aws_sdk_with_turi_env(s3url& parsed_url) {
...
   // clientConfiguration.proxyHost = proxy.c_str();
 
+  // set path or file for ssl certs
+  if (!get_alternative_ssl_cert_file().empty()) {
+    auto fstaus = get_file_status(get_alternative_ssl_cert_file());
+    ASSERT_TRUE(fstaus.second.empty());
+    if (fstaus.first == file_status::REGULAR_FILE) {
+      clientConfiguration.caFile = get_alternative_ssl_cert_file().c_str();
+    } else {
+      log_and_throw("Invalid file for alternative SSL certificate");
```
TobyRoseman, line 158: Let's give a little more detail in the error message here. Something like:
`Invalid file for alternative SSL certificate. Value of TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE must be a regular file.`

Including the value of `get_alternative_ssl_cert_file()` in the error message would also be nice.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -148,6 +148,27 @@ S3Client init_aws_sdk_with_turi_env(s3url& parsed_url) {
...
+    }
+  }
+
+  if (!get_alternative_ssl_cert_dir().empty()) {
+    auto fstaus = get_file_status(get_alternative_ssl_cert_dir());
+    ASSERT_TRUE(fstaus.second.empty());
+    if (fstaus.first == file_status::DIRECTORY) {
+      clientConfiguration.caPath = get_alternative_ssl_cert_dir().c_str();
+    } else {
+      log_and_throw("Invalid path for alternative SSL certificate");
```
TobyRoseman, line 168: Let give a better error message here too.

src/core/storage/fileio/s3_api.cpp
guihao-liang: right, I should give some care to customers.

PR #3105, opened 2020-04-09
check on parent dir instead

guihao-liang:
Address #3108


src/core/storage/fileio/fs_utils.cpp
```diff
@@ -290,26 +290,31 @@ EXPORT bool create_directory_or_throw(const std::string& path) {
   // this function throws if the directory still doesn't exist
   // at that location
   if (get_protocol(path) == "s3") {
-    auto status = get_file_status(path);
+    // only check on parent direcotory or prefix
+    auto ppath = get_dirname(path);
```
TobyRoseman, line 294: `dirname` would be much more descriptive variable name than `ppath`.

PR #3110, opened 2020-04-10
add sarray test

guihao-liang:


guihao-liang:
passed all internal tests.

PR #3114, opened 2020-04-11
get the retry logic back

guihao-liang:
I should blame myself. I should read the source files instead of their [documentation](http://sdk.amazonaws.com/cpp/api/LATEST/class_aws_1_1_client_1_1_a_w_s_error.html#a6262c30a4967119d1730cb23dbe30d6f).

It's using `S3Errors` to wrap `CoreErrors`. We should only rely on Http code to do retry on our own.


guihao-liang:
https://guihao-liang.github.io/2020-04-12-aws-s3-retry/

PR #3116, opened 2020-04-13
Customized s3 error handlings when encountering non-standard errors

guihao-liang:
To avoid collisions (HTTP code 429) when 2 - 3 runners running tests at the same time.

### DIFF
--- a/src/core/storage/fileio/s3_api.cpp
+++ b/src/core/storage/fileio/s3_api.cpp
@@ -20,6 +20,7 @@
 #include <chrono>
 #include <core/logging/assertions.hpp>
 #include <core/logging/logger.hpp>
+#include <core/random/random.hpp>
 #include <core/storage/fileio/fs_utils.hpp>
 #include <core/storage/fileio/general_fstream.hpp>
 #include <core/storage/fileio/get_s3_endpoint.hpp>
@@ -513,13 +514,27 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
 
       } else {
         auto error = outcome.GetError();
-        // Unlike CoreErrors, S3Error Never retries. Use Http code instead.
-        // check aws-cpp-sdk-s3/source/S3Error.cpp
-        if (error.GetResponseCode() ==
-            Aws::Http::HttpResponseCode::TOO_MANY_REQUESTS) {
-          n_retry++;
+        /*
+         * Unlike CoreErrors, S3Error Never retries on S3 errors.
+         * Retry can be based on HTTP code or HTTP body.
+         *
+         * 1. if SDK uses HTTP code, e.g., ShouldRetry return true on 429.
+         * We don't need to retry on our own since SDK already retried.
+         * 2. if SDK doesn't use HTTP code, we check HTTP on our own.
+         * check: https://guihao-liang.github.io/2020-04-12-aws-s3-retry/
+         * */
+        if (!error.ShouldRetry()) {
+          // SDK didn't retry for us, check retry on our own decisions
+          if (error.GetResponseCode() ==
+              Aws::Http::HttpResponseCode::TOO_MANY_REQUESTS) {
+            n_retry++;
+          } else {
+            // stop retry immediately
+            n_retry = 3;
+          }
 
-          if (n_retry == 3) {
+          // should be ==, use >= to suppress any surprises
+          if (n_retry >= 3) {
             // amend the error msg on the last retry failure
             std::stringstream ss;
             reportS3ErrorDetailed(ss, parsed_url, S3Operation::List,
@@ -531,10 +546,14 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
           } else {
             // continue retry
             std::this_thread::sleep_for(std::chrono::milliseconds(backoff));
-            backoff *= 2;
+            // [2, 4), float random
+            double rdm = (turi::random::pure_random_seed() % 1000 / 500.) + 2;
+            backoff = static_cast<int>(backoff * rdm);
           }
 
         } else {
+          // error.ShouldRetry() == 3
+          // AWS SDK already retried 3 times
           std::stringstream ss;
           reportS3ErrorDetailed(ss, parsed_url, S3Operation::List,
                                 clientConfiguration, outcome)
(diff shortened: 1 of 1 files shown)


src/core/storage/fileio/s3_api.cpp
```diff
@@ -531,10 +550,14 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
           } else {
             // continue retry
             std::this_thread::sleep_for(std::chrono::milliseconds(backoff));
-            backoff *= 2;
+            // [2, 4), float random
+            double rdm = (turi::random::pure_random_seed() % 1000 / 500.) + 2;
```
TobyRoseman, line 554: I have no idea what `rdm` means. A more descriptive name would be good.

Is this the backoff time in seconds?

src/core/storage/fileio/s3_api.cpp
```diff
@@ -543,7 +566,7 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
           logstream(LOG_DEBUG)
               << "list_objects_impl failed:" << ret.error << std::endl;
           // no need to retry, just jump out all while loop
-          break;
+          break;  // or n_retry = 3;
```
TobyRoseman, line 569: I don't think this comment is helpful.

guihao-liang: I think it's good for people to understand the last n_retry = 3

src/core/storage/fileio/s3_api.cpp
guihao-liang: It means random. The comment above that line doesn't make sense to you?

src/core/storage/fileio/s3_api.cpp
```diff
@@ -513,13 +514,31 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
...
+         * 2. if SDK doesn't use HTTP code but HTTP body, we check HTTP on our
+         * own if SDK doesn't think it should retry based on messages in HTTP
+         * body. Check https://guihao-liang.github.io/2020-04-12-aws-s3-retry/
+         *
+         * */
+        if (!error.ShouldRetry()) {
+          // SDK didn't retry for us, check retry on our own decisions
+          if (error.GetResponseCode() ==
+              Aws::Http::HttpResponseCode::TOO_MANY_REQUESTS) {
+            n_retry++;
```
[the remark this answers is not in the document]
guihao-liang: The internal service states client should explicitly handle 429. That's by design.

src/core/storage/fileio/s3_api.cpp
```diff
@@ -513,13 +514,31 @@ list_objects_response list_objects_impl(s3url parsed_url, std::string proxy,
...
+          } else {
+            // stop retry immediately
+            n_retry = 3;
+          }
 
-          if (n_retry == 3) {
+          // should be ==, use >= to suppress any surprises
```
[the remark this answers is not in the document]
guihao-liang: just want to let people know the why I use >=. Theoriactially, its max value should be 3.

src/core/storage/fileio/s3_api.cpp
TobyRoseman: I never would have guessed that `rdm` meant random. So much more time is spent reading code than writing code. So let make variable names descriptive.

You didn't answer my question: is this in seconds?

src/core/storage/fileio/s3_api.cpp
TobyRoseman: It going to be very easy for this comment to get out of date. If someone changes the while-loop conditional but doesn't update this inline comment, this comment will go from confusing to incomprehensible.

src/core/storage/fileio/s3_api.cpp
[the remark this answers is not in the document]
TobyRoseman: `to suppress any surprises` is kind of a strange phase. I don't think this require a comment but if you really want to add one I would say `// Defensively use >= rather than ==`. It probably make more sense as an inline comment.

TobyRoseman:
@guihao-liang - let hold off on merging this until we know how much more reliable the s3 unit tests become with the internal configuration changes.

src/core/storage/fileio/s3_api.cpp
guihao-liang: The scale for `backoff` is ms. This is just a scalar factor.

src/core/storage/fileio/s3_api.cpp
guihao-liang: The scale for backoff is ms. This is just a scalar factor.

src/core/storage/fileio/s3_api.cpp
guihao-liang: Then they should read my code and change the condition. I use `n_retry==3` in other places. I just want to remind people not to think the previous guy working here only use `break`.

guihao-liang:
related to https://github.com/aws/aws-sdk-cpp/issues/1359

src/core/storage/fileio/s3_api.cpp
hoytak: Why do we need a random backoff?  That's only needed when two identical services are competing for the exact same resource, which isn't the rate-limiting service we're talking about here.

src/core/storage/fileio/s3_api.cpp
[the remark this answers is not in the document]
hoytak: I think Toby's comment is a good one here, or just drop it -- it's obvious in this what's going on.

src/core/storage/fileio/s3_api.cpp
hoytak: I don't understand the comment either.  I don't think people would get that context.

guihao-liang:
reopen when needed.

src/core/storage/fileio/s3_api.cpp
guihao-liang: not exactly the same time. internal aws service blocks subsequent requests within a window time after it decides to throttle the rate. I observed 2 running jobs that start very closely will trigger the rate limiter. Well, let me use regular backoff.

PR #3127, opened 2020-04-15
Allow float_array.cpp to build without other dependencies

nickjong:
This PR is the first step to allowing the `neural_net` directory to build as a standalone library, which some of us have been doing in a hacky way for debugging/development purposes. The main change is to move the Turi serialization code outside, to the toolkits library.

### DIFF
--- a/src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
+++ b/src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
@@ -5,8 +5,7 @@
  * https://opensource.org/licenses/BSD-3-Clause
  */
 
-#ifndef TURI_STYLE_TRANSFER_MPS_STYLE_TRANSFER_BACKEND_H_
-#define TURI_STYLE_TRANSFER_MPS_STYLE_TRANSFER_BACKEND_H_
+#pragma once
 
 #include <functional>
 #include <map>
@@ -50,5 +49,3 @@ private:
 } // namespace turi
 
 #endif // #ifdef HAS_MACOS_10_15
-
-#endif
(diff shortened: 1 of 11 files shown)


src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
```diff
@@ -50,4 +51,4 @@ class EXPORT mps_style_transfer : public turi::neural_net::model_backend {
 
 #endif // #ifdef HAS_MACOS_10_15
 
-#endif
\ No newline at end of file
+#endif
```
brtal, line 54: i think all of our supported compilers should support `#pragma once`? should we incrementally move to that as we touch headers?

src/toolkits/util/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ make_library(unity_util OBJECT
     sframe_utils.cpp
     class_registrations.cpp
     random_sframe_generation.cpp
+    float_array_serialization.cpp
```
brtal, line 11: is there any coding standard around keeping cmakelists sorted alphabetically?

src/ml/neural_net/style_transfer/mps_style_transfer_backend.hpp
nickjong: I wasn't previously familiar with `#pragma once`, but it seems reasonable. I am not aware that any of our build processes copy headers in a way that would be problematic here. I'll kick off a build to verify that the compiler we use for our releases does the right thing

src/toolkits/util/CMakeLists.txt
nickjong: We have been using https://google.github.io/styleguide/cppguide.html in general, but it doesn't explicitly address CMakeLists. Extrapolating from its treatment of headers, I would agree that we should sort these....

PR #3137, opened 2020-04-21
Pinned llvm version to one that compiles correctly with python 2.7

hoytak:
This pins the llvmlite version to one that compiles correctly with python 2.7.

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -8,6 +8,7 @@ pillow==6.2.0
 prettytable==0.7.2
 pytz==2019.1
 resampy==0.2.1
+llvmlite==0.31.0
 scikit-learn==0.20.3
 statsmodels==0.9.0
 tensorflow==2.0.0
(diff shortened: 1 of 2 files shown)


scripts/requirements.txt
```diff
@@ -8,6 +8,7 @@ pillow==6.2.0
 prettytable==0.7.2
 pytz==2019.1
 resampy==0.2.1
+llvmlite==0.31.0
```
TobyRoseman, line 11: Remove this line. It's already here. See line 22/23.

src/python/setup.py
```diff
@@ -132,6 +132,7 @@ def run(self):
         "pillow >= 5.2.0",
         "prettytable == 0.7.2",
         "resampy == 0.2.1",
+        "llvmlite == 0.31.0",
```
TobyRoseman, line 135: Do you think it would be better to only pin this for Python 2.7?

hoytak: Good call.  Fixed.

PR #3138, opened 2020-04-21
Clean up OD CoreML export some more

nickjong:
Also add some experimental support for the NMS NN layer added in Catalina. This is mostly for development/experimental purposes, since it's not clear that the Vision API we tell developers to use to deploy OD models is compatible with the new NMS layer.

### DIFF
--- a/src/toolkits/coreml_export/neural_net_models_exporter.cpp
+++ b/src/toolkits/coreml_export/neural_net_models_exporter.cpp
@@ -246,7 +246,7 @@ void add_non_maximum_suppression_layer(Model* model_nn,
                                        float confidence_threshold,
                                        float iou_threshold) {
   // The model we're modifying must be a NeuralNetwork.
-  assert(model_nn->has_neuralnetwork());
+  ASSERT_TRUE(model_nn->has_neuralnetwork());
 
   // Append the actual NMS layer.
   NeuralNetworkLayer* nms_layer =
@@ -355,7 +355,7 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
 
   if (use_nms_layer) {
     int num_models = model_pipeline.pipeline().models_size();
-    assert(num_models > 0);
+    ASSERT_GT(num_models, 0);
     Model* model_nn =
         model_pipeline.mutable_pipeline()->mutable_models(num_models - 1);
     add_non_maximum_suppression_layer(model_nn, pipeline_desc, num_classes,
--- a/src/toolkits/coreml_export/neural_net_models_exporter.cpp
+++ b/src/toolkits/coreml_export/neural_net_models_exporter.cpp
@@ -17,9 +17,11 @@ using CoreML::Specification::ArrayFeatureType;
 using CoreML::Specification::FeatureDescription;
 using CoreML::Specification::ImageFeatureType;
 using CoreML::Specification::ImageFeatureType_ImageSizeRange;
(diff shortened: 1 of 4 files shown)


src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -262,7 +238,136 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
...
+  // Set up Pipeline
+  CoreML::Specification::Model model_pipeline;
+  model_pipeline.set_specificationversion(
+      CoreML::MLMODEL_SPECIFICATION_VERSION);
+  ModelDescription* pipeline_desc = model_pipeline.mutable_description();
+
+  // Adopt the model pipeline passed to us as input.
+  std::unique_ptr<CoreML::Specification::Pipeline> raw_pipeline_spec =
+      std::move(raw_pipeline).move_coreml_spec();
+  model_pipeline.mutable_pipeline()->Swap(raw_pipeline_spec.get());
```
brtal, line 339: this block of code took me down an interesting code path. it's interesting that it takes 3 lines of code to handoff a pointer.

src/toolkits/coreml_export/neural_net_models_exporter.cpp
```diff
@@ -262,7 +238,136 @@ std::shared_ptr<MLModelWrapper> export_object_detector_model(
...
+
+  if (!include_non_maximum_suppression) {
+    // Only support this case for models supporting spec version 1, which means
+    // no pipeline models.
+    ASSERT_EQ(model_pipeline.pipeline().models_size(), 1);
```
brtal, line 344: if my reading is correct, places where we use `assert` (from `cassert`) don't throw in release builds but `ASSERT_EQ` and friends always do. i've noticed in other PRs folks have switched to `cassert`. maybe better for an offline conversation, but curious what the use conventions are here and why you chose this here and `assert` on line 358.

src/toolkits/coreml_export/neural_net_models_exporter.cpp
nickjong: Yes, much of this is motivated by the desire to minimize the number of files (and especially headers) that much import the Google auto-generated protocol buffer headers. For example, attempting to import both those headers and Foundation.h will generate a compiler error (at least in one order of includes?). So we ended up creating a PIMPL-like design to encapsulate protocol buffer values.

src/toolkits/coreml_export/neural_net_models_exporter.cpp
nickjong: Good question. These should be `ASSERT_EQ` and `ASSERT_GT`. I've been using the standard `assert` in the `neural_net` directory since I am pushing that code to be portable / standalone. But this code is in `toolkits`, which I think is more firmly dependent on the broader Turi Create code base.

brtal: what about debug vs. release?

nickjong: What I have gathered is that the ASSERT macros are intended to catch programming errors, while we use `log_and_throw` to surface user errors. I honestly don't have context on why we have both debug-only and debug-and-release ASSERT macros. Most of this codebase seems to err on the side of compiling out only the assertions in performance-sensitive areas of the code.

PR #3139, opened 2020-04-21
defer the module import

guihao-liang:
in order to bypass some missing dependencies for minimal packages.

For example,

```python
visualization = _DeferredModuleLoader(
    "turicreate.visualization"
)
plot = _DeferredCallableLoader(visualization, "plot")
show = _DeferredCallableLoader(visualization, "show")
```

won't include vis module when calling `import turicreate`.

People can continue using `turicreate.SFrame` even the visualization binary is missing. As long as they don't call "SFrame.show", they should be fine.


src/python/turicreate/_deps/__init__.py
```diff
@@ -9,6 +9,291 @@
 from distutils.version import StrictVersion as _StrictVersion
 import logging as _logging
 import re as _re
+import sys
```
guihao-liang, line 12: this is the most critical part you need to review.

src/ml/neural_net/mps_graph_trainer.mm
```diff
@@ -3,6 +3,7 @@
 #include <tuple>
 
 #include <ml/neural_net/float_array.hpp>
+#include <core/logging/logger.hpp>
```
guihao-liang, line 6: cherry-pick from my other pr. Please ignore this.

guihao-liang:
```
pip uninstall -y tensorflow
pytest -s test/test_sframe.py
```
with this import trick, sframe can run without `tensorflow`.

guihao-liang:
Plan changed.

PR #3141, opened 2020-04-22
typo which results in bad formatting.

bglnelissen:
removed a '`' for the markdown to parse correctlyt


TobyRoseman:
Thanks for the fix @bglnelissen.

PR #3150, opened 2020-04-28
Add build support for clang-tidy and clang build metrics

brtal:
clang-tidy support enables the llvm extras linter to catch a wide array of problems and improve the build. It is off by default. Currently all rules are enabled, but this will fail the build. We will need to turn these rules on one by one.

Usage:
install llvm from homebrew, and make sure clang-tidy is in your path
./configure --with-clang-tidy

Clang build metrics enables a new features in Clang 9.0 which generates chrome://tracing (and similar tools) compatible json files for every compiled object that details where time was spent for that particular target. There are additional tools (such as ClangBuildAnalyzer) that will combine these files to give an overall high-level view of where time is spent in the build. I will share the detailed output from this internally.

Usage:
./configure --with-clang-metrics

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,10 +26,15 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED 1)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+# ##############################################################################
+# Set options
+
+option(TC_BUILD_METRICS "Produce clang build metrics" OFF)
+
 # **************************************************************************/
-# * */
-# * Global Link, Include and Define Flags              */
-# * */
+# *                                                                        */
+# * Global Link, Include and Define Flags                                  */
+# *                                                                        */
 # **************************************************************************/
 
 set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH};${CMAKE_SOURCE_DIR}/cmake")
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -341,6 +349,23 @@ CXXCMD=`./scripts/find_compiler.sh cxx --ccache=$with_ccache --script-dir=${PWD}
 echo "Setting C compiler to $CCCMD."
 echo "Setting C++ compiler to $CXXCMD."
 
+if [[ $with_clangtidy == 1 ]]; then
+  clangtidy_path=`which clang-tidy || echo ''`
```
guihao-liang, line 353: "``" is actually not preferred. You can install and run `shellcheck` to learn more. `$()` is preferred for readability.

configure
```diff
@@ -341,6 +349,23 @@ CXXCMD=`./scripts/find_compiler.sh cxx --ccache=$with_ccache --script-dir=${PWD}
...
+  
+  if ! [ -f $clangtidy_path ]; then
```
guihao-liang, line 355: bash style should be `[[ ! ]]`

guihao-liang: better to wrap `$clangtidy_path` with quotes. Otherwise, the shell is free to expand what's referenced.

CMakeLists.txt
```diff
@@ -148,6 +153,13 @@ check_and_set_compiler_flag(-fpeel-loops RELEASE)
 check_and_set_compiler_flag(-funswitch-loops RELEASE)
 check_and_set_compiler_flag(-ftracer RELEASE)
 
+if(TC_BUILD_METRICS)
+  # TC_BUILD_METRICS enables -ftime-trace in supported compilers, which will
+  # produce a timings json file for each compiled object.
+  
+  check_and_set_compiler_flag(-ftime-trace)
```
guihao-liang, line 160: in order to improve compile? good to know.

PR #3151, opened 2020-04-28
minimal build install

guihao-liang:
we can customize the install when we call `bdist_wheel` by

```
 python setup.py bdist_wheel install --minimal

running bdist_wheel
running build
running build_py
running build_ext
installing to build/bdist.macosx-10.14-x86_64/wheel
running install
```

It will customize the install information `install_requires`, which will be then used by `wheel` to write to `turicreate-6.2.dist-info/METADATA`.

### DIFF
--- a/src/python/setup.py
+++ b/src/python/setup.py
@@ -14,6 +14,8 @@ from setuptools.command.install import install
 
 PACKAGE_NAME = "turicreate"
 VERSION = "6.2"  # {{VERSION_STRING}}
+NON_MINIMAL_LIST = ["tensorflow", "resampy"]
+
 
 # Prevent distutils from thinking we are a pure python package
 class BinaryDistribution(Distribution):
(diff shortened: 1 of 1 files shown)


src/python/setup.py
```diff
@@ -22,9 +24,38 @@ def is_pure(self):
...
+        if self.minimal is not None:
+
+            def do_not_install(require):
+                require = require.strip()
+                for name in NON_MINIMAL_LIST:
+                    if require.startswith(name):
+                        return False
+                return True
+
+            install_requires_minimal = list(
```
hoytak, line 49: `install_requires_minimal = [r for r in self.distribution.install_requires
                                                 if not any(r.startswith(nm) for nm in NON_MINIMAL_LIST)]`

src/python/setup.py
```diff
@@ -14,6 +14,8 @@
 
 PACKAGE_NAME = "turicreate"
 VERSION = "6.2"  # {{VERSION_STRING}}
+NON_MINIMAL_LIST = ["tensorflow", "resampy"]
```
hoytak, line 17: Let's get pandas in here too?  Is that possible?

guihao-liang: let's do that in the next phase (another PR on its way)

src/python/setup.py
guihao-liang: I think my version is more readable, even though it's kind of verbose.

PR #3152, opened 2020-04-29
minimal pkg pipeline

guihao-liang:
The successor of #3151. Sub-pr: #3171.

The minimal build wheel won't have dependencies listed below.

```python
19 NON_MINIMAL_LIST = [
 20     "coremltools",
 21     "pandas",
 22     "tensorflow",
 23     "resampy",
 24     "scipy",
 25 ]
```

And the minimal pkg will run relevant tests, which are not considered as part of the `toolkits` components.

```bash
minimal_test=($(grep -L "toolkits" *.py))
```
The internal runner will run tests on minimal pkg with 2.7 (Linux) and 3.7 (OSX). The reason to do so is that minimal pkg is the subset of full pkg. As long as full pkg succeeds in all platforms we care about, we should be confident minimal pkg will succeed too.

---
update 07/05/2020

The feature shown below is from #3175. The package produced by this PR doesn't have any prompt for the user to install the full package. That is to say, this PR is now not related to #3175 now.

Update: 01/05/2020

in the minimal package, if you use lazy-import wrapped `tensorflow`,
```
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    tc._deps.tensorflow.get_module()
  File "/Users/guihaoliang/Work/guicreate-2/target/env36/lib/python3.6/site-packages/turicreate/_deps/minimal_import.py", line 307, in get_module
    self._load_module()
  File "/Users/guihaoliang/Work/guicreate-2/target/env36/lib/python3.6/site-packages/turicreate/_deps/minimal_import.py", line 326, in _load_module
    self._module = self._init_func(self._name)
  File "/Users/guihaoliang/Work/guicreate-2/target/env36/lib/python3.6/site-packages/turicreate/_deps/minimal_import.py", line 95, in default_init_func
    raise e
  File "/Users/guihaoliang/Work/guicreate-2/target/env36/lib/python3.6/site-packages/turicreate/_deps/minimal_import.py", line 72, in default_init_func
    return importlib.import_module(name)
  File "/Users/guihaoliang/Work/guicreate-2/target/env36/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
[...]

PR #3156, opened 2020-04-30
prompt for minimal pkg when module is missing

guihao-liang:
Branched from #3152.

Lazy import tensorflow, resampy and coremltools and wrap import failure with installation instructions.

Numpy and Pandas are taken cared for by flags `HAS_PANDAS` and `HAS_NUMPY`.


scripts/make_wheel.sh
```diff
@@ -291,69 +291,99 @@ package_wheel() {
...
+    local version_modifier=$2
 
-  VERSION_NUMBER=`${PYTHON_EXECUTABLE} -c "from turicreate.version_info import version; print(version)"`
-  ${PYTHON_EXECUTABLE} setup.py ${dist_type} # This produced an wheel starting with turicreate-${VERSION_NUMBER} under dist/
+    cd ${WORKSPACE}/${build_type}/src/python
 
-  cd ${WORKSPACE}
+    if [[ "$minimal" -eq 1 ]] ; then
+      if [[ "$(uname -s)" == "Darwin" ]] ; then
+        sed -i "" 's/^USE_MINIAL = False$/USE_MINIAL = True/g' turicreate/_deps/minimal_import.py
```
hoytak, line 303: Should this be USE_MINIMAL ?

guihao-liang: Yeah, that's my typo. I use `USE_MINIAL` everywhere. Good Catch!

TobyRoseman:
@guihao-liang - branching from another merge request make it very difficult to review things. Let's get the first merge request in. Then rebase. Then review this one.

guihao-liang:
> @guihao-liang - branching from another merge request make it very difficult to review things. Let's get the first merge request in. Then rebase. Then review this one.

Will do. That's the plan. This pr depends on  #3152.

scripts/install_python_toolchain.sh
```diff
@@ -35,7 +35,11 @@ function linux_patch_sigfpe_handler {
 }
 
 $PIP install --upgrade "pip"
-$PIP install -r scripts/requirements.txt
+if [[ "$USE_MINIMAL" -eq 1  ]]; then
+  $PIP install -r scripts/requirements-minimal.txt
```
TobyRoseman, line 39: Is `requirements-minimal.txt` a subset of `requirements.txt`? If so, why not just install everything? I'm not really sure this is worth adding additional complexity.

src/python/turicreate/_deps/minimal_import.py
```diff
@@ -0,0 +1,367 @@
...
+    # pickle
+    def __reduce__(self):
+        def _init(*args, **kwargs):
+            return DeferredModuleLoader(*args, **kwargs)
+
+        # be aware, singleton must be set false
+        return (_init, (self._name, self._init_func, False))
+
+
+class DeferredCallableLoader(object):
```
TobyRoseman, line 342: How is this getting used?

src/python/turicreate/_deps/minimal_import.py
```diff
@@ -0,0 +1,367 @@
+# -*- coding: utf-8 -*-
```
TobyRoseman, line 1: What is the purpose of this file?

src/python/turicreate/_deps/minimal_import.py
guihao-liang: not used. Just leave it here for future use.

src/python/turicreate/_deps/minimal_import.py
guihao-liang: copied from __init__.py under `_deps`. I don't know either.

scripts/install_python_toolchain.sh
guihao-liang: for example, if we want to test this pkg with no tensorflow installed so that we can make sure this minimal pkg can work without those stripped dependencies.

src/python/turicreate/_deps/minimal_import.py
TobyRoseman: If it's not being used, let's delete it.

scripts/install_python_toolchain.sh
TobyRoseman: ok, that's reasonable.

src/python/turicreate/_deps/minimal_import.py
guihao-liang: removed.

src/python/turicreate/_deps/minimal_import.py
hoytak: If you don't know why it's here, then let's remove it.

hoytak: Or what's going on with this file?  Are you saying there isn't a reason it's in this spot specifically?  This is a lot of code to introduce here, so we should be clear we need it.

guihao-liang:
```
# -*- coding: utf-8 -*-
```
I think Toby concerns about this line, not the rest of the code.

guihao-liang:
> Can we merge the previous commits in and then rebase this commit? It's hard to separate out the changes of the previous commits with what's new in this one.

Adding some new things in. Please review #3172 first and then I will rebase later.

scripts/requirements-minimal.txt
```diff
@@ -0,0 +1,22 @@
+numpy==1.16.4
+cython==0.29.10
+decorator==4.4.0
+pandas==0.24.2
```
hoytak, line 4: Is this meant to still be here?

scripts/test_wheel.sh
```diff
@@ -39,6 +42,7 @@ while [ $# -gt 0 ]
     --docker-python3.5)     USE_DOCKER=1;DOCKER_PYTHON=3.5;;
     --docker-python3.6)     USE_DOCKER=1;DOCKER_PYTHON=3.6;;
     --docker-python3.7)     USE_DOCKER=1;DOCKER_PYTHON=3.7;;
+    --minimal-build)        USE_MINIMAL=1;;
```
hoytak, line 45: Let's change this to --minimal to keep it consistent.

guihao-liang:
Let's get #3152 and #3172 reviewed and merged first. Otherwise, it's hard for you to review.

src/python/turicreate/_deps/minimal_import.py
hoytak: The header is 100% standard -- I think we are both wondering what the purpose of this code is and if it's necessary.  

I honestly think the full lazy-loading approach here is a bit too heavyweight for what we're hoping to do, since we're only worried about delaying the load of third party packages.   It adds more complexity than just wrapping the individual imports in a function to make it clear they are lazily loaded.

Let's just do this.  Here

```
def import_tensorflow(): 
  try: 
    import tensorflow
  except ImportError:
    # Raise helpful error message
  return tensorflow
```
Then instead of `import tensorflow`, we can just do `tensorflow = import_tensorflow()`  This would be a much more maintainable way of doing it.  

When we look at lazily import the third party modules, then we could use this machinery.

guihao-liang:
close for leaving a record for the lazy import method. Will create a new branch to use more maintainable approach suggested by @hoytak

src/python/turicreate/_deps/minimal_import.py
TobyRoseman: My concerns are about the whole file, not this line.

I agree with Hoyt. The lazy loading approach seems too heavyweight. I think we should do Hoyt's suggestion instead.

guihao-liang: This is not doable. Since we use `pytest -m minimal`, it will run all imports first. And if we follow this path, all module depends on module level `import tensorflow` will throw. One work around is to move all `import tensorflow` to function level to do the "lazy" import. I'm not sure that if we want to go this path together with other packages.

nickjong: What you describe as the workaround sounds to me like the correct way to do lazy importing. "To avoid importing heavyweight dependencies that are only used sometimes" seems to be a primary reason to do imports within functions at all

guihao-liang: How do you define heavyweight? The perf overhead is very small. What people's concern is about maintenance costs. It's been tested for #3139 and also this PR. If we in the future want to roll out lazy import to all of our packages, why not we first fully test on these 3rd party dependencies and release it? Then we can have more confidence in using this method for other modules.

is this complicated? no.

This is the second time for me to tweak lazy import for minimal packages (first time #3139 ). We want to hold it back again. Let's plan it first before I do it again!

what's the deal for module-level import such as

```
import tensorflow.compat.v1 as _tf

_tf.disable_v2_behavior()
```

should I move those 2 lines to each place there `_tf` is used? Is this more maintainable than have one place using this lazy import?

nickjong: To me, this problem seems similar to some of the problems surrounding static initialization in C++ (e.g. https://isocpp.org/wiki/faq/ctors#static-init-order). Both to avoid the static initialization order fiasco, and to make initialization lazy, it's often best to defer initialization until the thing being initialized is used.

The best practice for this, in both C++ and Python, seems to be simply to perform the initialization in some function, so that the initialization only occurs when the function that needs it calls it. This is simple and easy to understand, and is self-documenting.

Building custom machinery around this to me sounds like trying to create a C++ wrapper around arbitrary types that avoids calling the constructor until it's actually used. Isn't it simpler just to do what the language supports directly?

guihao-liang: Anyway, I will go ahead and do the labor work.

guihao-liang: > To me, this problem seems similar to some of the problems surrounding static initialization in C++ (e.g. https://isocpp.org/wiki/faq/ctors#static-init-order). Both to avoid the static initialization order fiasco, and to make initialization lazy, it's often best to defer initialization until the thing being initialized is used.
> 
> The best practice for this, in both C++ and Python, seems to be simply to perform the initialization in some function, so that the initialization only occurs when the function that needs it calls it. This is simple and easy to understand, and is self-documenting.
> 
> Building custom machinery around this to me sounds like trying to create a C++ wrapper around arbitrary types that avoids calling the constructor until it's actually used. Isn't it simpler just to do what the language supports directly?

Thanks for sharing this. I don't know there's actually called Construct On First Use Idiom. I use it all the time but don't know its name. Expanding this topic, there's also destruct order issues on static objects. I've seen some in our codebase too. I manually fix with dependency injection. So I think these 2 are twin problems. 

Back to the point, the initialization is done by `import`. So the initialization dependency topology is constructed by different imports. And `import` implementation is using this `Construct On First Use Idiom` by checking the singleton in `sys.modules` (similar to static Foo* bar = new Foo(); return &*bar). Besides that, you can think `import` as a **factory function** instead of the **constructor**. There's no risk to cause `static initialization fiasco`. 

The main concern here is still the maintenance and I fully understand the concern. Thanks for your knowledge sharing, that's really helpful for me.

guihao-liang: For more information, `import` function uses reentrent lock to protect the read and write to the singleton map `sys.modules`.

PR #3157, opened 2020-04-30
Fix number of predictions in object_detector.

brtal:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -93,19 +93,6 @@ constexpr float DEFAULT_CONFIDENCE_THRESHOLD_EVALUATE = 0.001f;
 
 // before generation precision-recall curves.
 
-// Each bounding box is evaluated relative to a list of pre-defined sizes.
-const std::vector<std::pair<float, float>>& anchor_boxes() {
```
nickjong, line 97: Thank you for cleaning this up!

brtal: apparently we don't have the same compiler warnings turned on internally...

Issue #3148, opened 2020-04-27
Object detector doesn't find matches


tagy:
Hi, I have setup the current version of turicreate in linux and tried training a model for object detection. The training completes but the model doesn't find objects in test images, either trying predict() or exporting the model to the sample iOS app.

I have checked the training images and annotations are correct using util.draw_bounding_boxes() and it looks good.

I also tried the same data in the MakeML app, that seems to use turicreate 5.2.1 the model from that works correctly with their test tool and also in iOS.

Does anyone have any clue how to debug this? When training the loss is decreasing by a similar amount as seen in the MakeML app. I would rather be able to run it on my own hardware. Thanks.

tagy:
I have re-installed linux and stared again and now I can create the model and it does find matches, so must have been something I was doing wrong. You can close this. 

The model I am getting does not work as well as the one from MakeML though, I guess I could try installing turicreate 5.2.1 but that is a different question.

TobyRoseman:
We have a serious issue with our TensorFlow implementation of object detection, this is what gets used on Linux. It's completely broken with our latest (6.2) release. If you have more than two classes in your dataset, then things have been broken since 6.0 (see #3031).

So if you have two or less classes/objects in your dataset, please use TuriCreate 6.1. If you have more than two classes, please use 5.8.

I am actively looking into these object detection issues. I'm planning to do a new release of TuriCreate once these issues are resolved.

tagy:
I have installed 6.1 and it seems to be working as expected now (I have one object). Thanks.

TobyRoseman:
We just released a point release (TuriCreate 6.2.2) with this bug fix. Upgrade your version of TuriCreate and you should be good to go.

Issue #3031, opened 2020-03-04
Multiclass Object detection prediction always returns the same class


YassineElbouchaibi:
Multiclass Object detection prediction always returns the same class but different after each training.. Let me clarify.

So I have a dataset and here is its head : 
```
+--------------------+------------------------+-------------------------------+
|        path        |         image          |           annotation          |
+--------------------+------------------------+-------------------------------+
| new_data/00009.jpg | Height: 512 Width: 384 | [{'label': 'Clear plastic ... |
| new_data/00011.jpg | Height: 512 Width: 384 | [{'label': 'Glass bottle',... |
| new_data/00028.jpg | Height: 512 Width: 384 | [{'label': 'Clear plastic ... |
| new_data/00034.jpg | Height: 384 Width: 512 | [{'label': 'Food Can', 'co... |
| new_data/00090.jpg | Height: 384 Width: 512 | [{'label': 'Meal carton', ... |
| new_data/00095.jpg | Height: 512 Width: 384 | [{'label': 'Plastic bottle... |
| new_data/00097.jpg | Height: 310 Width: 512 | [{'label': 'Drink can', 'c... |
| new_data/00098.jpg | Height: 512 Width: 384 | [{'label': 'Styrofoam piec... |
| new_data/00104.jpg | Height: 233 Width: 512 | [{'label': 'Unlabeled litt... |
| new_data/00110.jpg | Height: 512 Width: 233 | [{'label': 'Plastic film',... |
+--------------------+------------------------+-------------------------------+
```

Here is one exemple of annotations :
```
[{
    'coordinates': {
        'height': 85,
        'width': 88,
        'x': 283,
        'y': 208
    },
    'label': 'Meal carton'
},
{
    'coordinates': { 'height': 52, 'width': 109, 'x': 111, 'y': 108 },
    'label': 'Clear plastic bottle'
},
{
    'coordinates': { 'height': 15, 'width': 10, 'x': 161, 'y': 89 },
    'label': 'Plastic bottle cap'
},
{
    'coordinates': { 'height': 33, 'width': 72, 'x': 393, 'y': 136 },
    'label': 'Drink can'
},
{
    'coordinates': { 'height': 46, 'width': 44, 'x': 490, 'y': 47 },
    'label': 'Disposable food container'
},
{
    'coordinates': { 'height': 41, 'width': 65, 'x': 413, 'y': 153 },
    'label': 'Normal paper'
},
{
[...]

PR #3160, opened 2020-05-01
Two Critical TensorFlow Object Detector Fixes

TobyRoseman:
* Fix TensorFlow object detector evaluate always returning zero accuracy #3148
* Fix TensorFlow object detector not working for more than two objects/classes #3031

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
+++ b/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
@@ -566,17 +566,14 @@ class DataAugmenter(object):
             feed_dict = dict()
 
             # Populate feed_dict with images and annotations
-            graph_op = self.resize_op_batch[0 : len(images)]
-            for i in range(0, len(images)):
+            graph_op = self.resize_op_batch[0:len(images)]
+            for i in range(len(images)):
                 feed_dict[self.img_tf[i]] = _utils.convert_shared_float_array_to_numpy(
                     images[i]
                 )
-                if self.resize_only:
-                    feed_dict[self.ann_tf[i]] = self.batch_size * [np.zeros(6)]
-                else:
-                    feed_dict[
-                        self.ann_tf[i]
-                    ] = _utils.convert_shared_float_array_to_numpy(annotations[i])
+                feed_dict[
+                    self.ann_tf[i]
+                ] = _utils.convert_shared_float_array_to_numpy(annotations[i])
 
             # Populate feed_dict with random seed and random alpha values, used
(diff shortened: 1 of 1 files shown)


TobyRoseman:
Internal test pass and benchmarking has verified both fixes.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -151,7 +151,6 @@ def resize_turicreate_image(image, output_shape):
         raise Exception("Non-supported resize method.")
 
     image_clipped = tf.clip_by_value(image_scaled, 0.0, 1.0)
-    annotation = tf.clip_by_value(annotation, 0.0, 1.0)
```
shreyajain17, line 154: The coordinates change if not in normalized format? Is it already in normalized format?

nickjong: I believe the bounding-box coordinates are normalized to [0,1] at this point, but the class labels are just indices from 0 .. n-1

TobyRoseman: I'm not sure. I don't have a deep understanding of this code. I believe at this point the coordinates should all be relative and between zero and one, and the confidences should all be ones.

Certainly, clipping is the wrong behavior for the last dimension of `annotation` (i.e. the dimension which hold the label id). Clipping *might* make sense for the other dimensions, since they should all be between zero and one. However it just seems wrong to me. If we're given bad data, things should break. It seems like normalizing/clipping could hide problem; I think that's exactly what was happening when we had more than two classes.

What do you think?

TobyRoseman: @nickjong - I didn't see your message before sending my message. I believe you are correct. 

@shreyajain17 - I don't think clipping/normalizing makes sense here. What do you think?

nickjong: To be clear, I only know that the coordinates are normalized, not in image coordinates, when passed from C++. That doesn't guarantee that the are literally in [0,1], or that any of the augmentation here can cause these coordinates to fall outside this range. But even then, I believe the actual network is largely robust to that anyway??

shreyajain17: My only concern was that if the image is resized, the coordinates of the bounding box must be adjusted correctly or the network gets wrong labels. The coordinates falling outside 0 to 1 is fine as long as the coordinates after rising are correct!

PR #3174, opened 2020-05-06
reduce some warnings

guihao-liang:
will reduce roughly 1300 lines of warnings in pytest warning summary section. Save some time scrolling through the logs.

PendingDeprecattionWarning is okay to ignore for now.

### DIFF
--- a/src/python/turicreate/test/test_linear_regression.py
+++ b/src/python/turicreate/test/test_linear_regression.py
@@ -25,6 +25,7 @@ if sys.version_info.major == 3:
 
 try:
     import statsmodels.formula.api as sm
+    from sklearn import linear_model
 except ImportError as e:
     # ignore extra dependencies
     # https://github.com/apple/turicreate/pull/3156
@@ -1135,8 +1136,6 @@ class L1LinearRegressionTest(unittest.TestCase):
         self.examples = X.shape[0]
 
         # Fit the model
-        from sklearn import linear_model
-
         self.l1_penalty = 10.0
         clf = linear_model.ElasticNet(
             alpha=self.l1_penalty / (2 * self.examples), l1_ratio=1
--- a/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
+++ b/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
@@ -48,7 +48,7 @@ def hue_augmenter(
     image,
     annotation,
     random_alpha_tf,
-    max_hue_adjust=_DEFAULT_AUG_PARAMS["max_hue_adjust"]
+    max_hue_adjust=_DEFAULT_AUG_PARAMS["max_hue_adjust"],
 ):
 
     # Apply the random rotation around the color wheel.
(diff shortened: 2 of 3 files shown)


src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
```diff
@@ -565,23 +556,23 @@ def get_augmented_data(self, images, annotations, random_seed):
...
 
             # Populate feed_dict with random seed and random alpha values, used
             # to sample image perturbations. We don't use TensorFlow's built-in
             # support for random number generation, since we want to effectively
             # reset the seed for each session (batch).
             random = np.random.RandomState(seed=random_seed)
             feed_dict[self.alpha_tf] = random.rand(*self.alpha_tf.shape)
-            feed_dict[self.random_seed_tf] = random.random_integers(
-                0, 2**32 - 1, size=self.batch_size
+            feed_dict[self.random_seed_tf] = random.randint(
```
guihao-liang, line 574: this line

guihao-liang:
passed internally. ready to merge.

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
hoytak: randint is inclusive of the endpoints, so we want this to be 0, 2**32-1 to not include 2**32.

src/python/turicreate/test/test_linear_regression.py
```diff
@@ -25,6 +25,7 @@
 
 try:
     import statsmodels.formula.api as sm
+    from sklearn import linear_model
```
hoytak, line 28: In the minimal version, won't the logic below just allow this to continue, without disabling the tests that depend on linear_model or sm?  Should there be a tests there?

guihao-liang: no. Because it's not tagged with `pytest.mark.minimal`. But when pytest tries to collect all tests and filter by makers, it will execute those imports, where those non-minimal tests will contain extra dependencies that are not required by tests marked as `minimal`.

guihao-liang: `test_linear_regression.py` won't be executed by the pytest at the test stage but will be executed at test collection stage.

guihao-liang:
related to #3152

src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
hoytak: Ah, numpy.random.randint is different.

TobyRoseman:
Why are PendingDeprecationWarning "okay to ignore for now"?

guihao-liang:
> Why are PendingDeprecationWarning "okay to ignore for now"?

Because it's pending. It has another step to becoming DeprecationWarning. If I understand correctly. Besides that, nobody is going to update the APIs to the latest, why we bother to see those 1k lines of warnings every day?

PR #3175, opened 2020-05-07
minimal check for import V3

guihao-liang:
Branched from #3152.

As it is discussed with @hoytak, @TobyRoseman, and @nickjong in #3156, I provide a simpler version for minimal package lazy import in order to lower the maintenance cost. 

---

update 07/05/2020

This PR aims to provide some instructions to users who install the minimal package but somehow wind up using some ML functionalities.

For example, minimal pkg doesn't have TF installed as its dependency. When a user uses ML toolkit that relies on TF, he/she should see the instruction to install the full package's dependency.

```bash
...
ModuleNotFoundError: No module named 'tensorflow'. This is a minimal package for SFrame only, without tensorflow pinned as a dependency. You can try to install all required packages by installing the full package by:
**pip install turicreate==6.2**
```

Instead of letting them install each package individually, let them install `turicreate==VERSION` to rely on the full package's requirement list to install all required dependencies.


src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
```diff
@@ -10,11 +10,16 @@
 import numpy as _np
 from .._tf_model import TensorFlowModel
 import turicreate.toolkits._tf_utils as _utils
-import tensorflow.compat.v1 as _tf
+from turicreate._deps.minimal_package import minimal_package_import_check
 
-# This toolkit is compatible with TensorFlow V2 behavior.
-# However, until all toolkits are compatible, we must call `disable_v2_behavior()`.
-_tf.disable_v2_behavior()
+# in conjunction with minimal package
```
hoytak, line 15: Nice!  Let's move this function into _tf_utils as it's copy-pasted elsewhere, then there's only one of it.

src/python/turicreate/toolkits/sound_classifier/_tf_sound_classifier.py
```diff
@@ -9,20 +9,28 @@
...
 
-# Suppresses verbosity to only errors
-_utils.suppress_tensorflow_warnings()
-import tensorflow.compat.v1 as _tf
 
-# This toolkit is compatible with TensorFlow V2 behavior.
-# However, until all toolkits are compatible, we must call `disable_v2_behavior()`.
-_tf.disable_v2_behavior()
+# in conjunction with minimal package
+def _lazy_import_tensorflow():
```
hoytak, line 14: Move into _tf_utils

guihao-liang: no

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
guihao-liang: there's must be a reason it exists everywhere before. If we need to clean it, why now?

guihao-liang:
for 3 files (drawing, activity, and style), at top-level, I wrap

```python
    import tensorflow.compat.v1 as _tf
    # This toolkit is compatible with TensorFlow V2 behavior.
    # However, until all toolkits are compatible, we must call `disable_v2_behavior()`.
    _tf.disable_v2_behavior()
```

into,

```python
def _lazy_import_tensorflow():
    _tf = minimal_package_import_check("tensorflow.compat.v1")
    # This toolkit is compatible with TensorFlow V2 behavior.
    # However, until all toolkits are compatible, we must call `disable_v2_behavior()`.
    _tf.disable_v2_behavior()
    return _tf
```

and only for the sound classifier, I wrap

```python
    _utils.suppress_tensorflow_warnings()
    import tensorflow.compat.v1 as _tf
    # This toolkit is compatible with TensorFlow V2 behavior.
    # However, until all toolkits are compatible, we must call `disable_v2_behavior()`.
    _tf.disable_v2_behavior()
```

 into 

```python
def _lazy_import_tensorflow():
    _utils.suppress_tensorflow_warnings()

    _tf = minimal_package_import_check("tensorflow.compat.v1")
    # This toolkit is compatible with TensorFlow V2 behavior.
    # However, until all toolkits are compatible, we must call `disable_v2_behavior()`.
    _tf.disable_v2_behavior()

    return _tf
```

These should not be considered as the duplicate of code because each tool may have its own settings on tf. Leaving it at the beginning of the file will allow people to know what's the tf configuration more clearly. I only add 2 more lines around. I don't believe that these 2 more lines are considered as the liability I shouldn't put here since people may not be able to maintain it because these 2 lines increase the mental burden on them so that they can't understand it.

Second, if this is considered as the duplicate of code, we wrap all common imports into on function to remove duplicate of `import tensorflow` or` import `os` or `import sys`.

@hoytak how do you think?

guihao-liang:
passed internal tests (112066).

guihao-liang:
from #3180:
> core implementation for lazy import (including comments): 392 lines added
> client adoption code: 48 lines added

from this PR:
core implementation for lazy import (including comments): 103 lines added
client adoption code: 165 lines added

The team favors later one is reasonable.

src/python/turicreate/_deps/__init__.py
```diff
@@ -9,7 +9,17 @@
...
 import re as _re
-from turicreate._deps.minimal_package import is_minimal_pkg
+from turicreate import __version__
+from turicreate._deps.minimal_package import (
+    is_minimal_pkg,
+    is_py_interactive,
+)
+
+if is_py_interactive() and is_minimal_pkg():
+    print(
```
TobyRoseman, line 19: When do the two print statements in this file get printed out? When the user imports TuriCreate if they have a minimal install?

src/python/turicreate/_deps/minimal_package.py
```diff
@@ -6,11 +6,93 @@
 from __future__ import print_function as _  # noqa
 from __future__ import division as _  # noqa
 from __future__ import absolute_import as _  # noqa
-
+import logging as _logging
```
TobyRoseman, line 9: You're not using this.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
```diff
@@ -13,6 +13,7 @@
 from turicreate.toolkits._model import Model as _Model
 import turicreate.toolkits._internal_utils as _tkutl
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
+from turicreate._deps.minimal_package import minimal_package_import_check
```
TobyRoseman, line 16: You should hid this import, i.e. change it to:
`from turicreate._deps.minimal_package import minimal_package_import_check as _minimal_package_import_check`

General rule: if the file you're in is not hidden (i.e. the file name doesn't start with an underscore), then you need to hide anything the user should never use.

src/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -18,6 +18,7 @@
 from turicreate.toolkits import _coreml_utils
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
 from turicreate.toolkits._model import PythonProxy as _PythonProxy
+from turicreate._deps.minimal_package import minimal_package_import_check
```
TobyRoseman, line 21: hide this import

src/python/turicreate/toolkits/object_detector/object_detector.py
```diff
@@ -24,6 +24,7 @@
     _raise_error_if_not_iterable,
 )
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
+from turicreate._deps.minimal_package import minimal_package_import_check
```
TobyRoseman, line 27: hide.

src/python/turicreate/toolkits/sound_classifier/sound_classifier.py
```diff
@@ -21,6 +21,7 @@
 from turicreate.toolkits._model import CustomModel as _CustomModel
 from turicreate.toolkits._model import PythonProxy as _PythonProxy
 from turicreate.toolkits import _coreml_utils
+from turicreate._deps.minimal_package import minimal_package_import_check
```
TobyRoseman, line 24: hide

src/python/turicreate/toolkits/style_transfer/style_transfer.py
```diff
@@ -10,6 +10,7 @@
 import turicreate.toolkits._internal_utils as _tkutl
 from turicreate.toolkits import _coreml_utils
 from turicreate.toolkits._internal_utils import _raise_error_if_not_sframe
+from turicreate._deps.minimal_package import minimal_package_import_check
```
TobyRoseman, line 13: hide

guihao-liang: one is enough. no need to repeat.

src/python/turicreate/_deps/minimal_package.py
guihao-liang: cool. good to know.

src/python/turicreate/toolkits/drawing_classifier/drawing_classifier.py
guihao-liang: sounds reasonable. Will change it when I have time.

src/python/turicreate/_deps/__init__.py
guihao-liang: when they use this minimal package in interactive mode.

TobyRoseman: I don't think that makes sense. I think we should remove these print statements and instead only print out these messages if they try to use some of the non-minimal functionality.

guihao-liang: Sure. I follow whatever team decides.

hoytak: Toby is right.  The users should only know this is the minimal package if they try to use some of the non-mimimal functionality.  Otherwise this will be a really bad experience for users like trove.

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
hoytak: Didn't you add _lazy_import_tensorflow?  It was added in this PR.

src/python/turicreate/_deps/__init__.py
guihao-liang: This will only print when they use interactive mode during first-time import. So it will only print once. But anyway,

guihao-liang:
@hoytak @TobyRoseman updated in https://github.com/apple/turicreate/pull/3175/commits/ed670ffe74dd901cf414fe32e95d97c232b440ea, the version of `turicreate` is also modified to resolve the complaint that,

```bash
ERROR: <package> has requirement turicreate==6.1.1+minimal, but you'll have turicreate 6.1.1 which is incompatible.
```

guihao-liang:
this https://github.com/apple/turicreate/pull/3175/commits/6d80701e18ffdfc1e37250ce573817f73f4111fe will trim the "+minimal" from [local version](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers) (the downstream version) to prompt the upstream version to user if they use any ML toolkits from the minimal version.

src/python/turicreate/_deps/__init__.py
```diff
@@ -38,6 +42,13 @@ def __get_version(version):
     from . import pandas_mock as pandas
 
 
+if not HAS_PANDAS and is_py_interactive() and is_minimal_pkg():
+    print(
```
TobyRoseman, line 46: We shouldn't be printing out this message. We should be handling panda imports the same way we're handling TensorFlow imports.

src/python/turicreate/_deps/minimal_package.py
```diff
@@ -6,11 +6,101 @@
...
+            pos = __version__.rfind("+")
+
+            if pos != -1:
+                version = __version__[:pos]
+            else:
+                version = __version__
+
+            # append more information
+            e.msg = (
+                "{}. This is a minimal package for SFrame only, without {} pinned"
```
TobyRoseman, line 99: I just tried the pip command you mention here. It doesn't work. pip doesn't record the `+` portion of the version. So it thinks you already have that version installed. So it doesn't do anything.

To make this work, the user needs to run:
`pip install --force-reinstall turicreate=={}`

I'm not sure printing out the original error message makes sense. I think it might just be confusing to the user. At the very least we probably want a newline between that error message and our error message.

I would just remove this text:
```
nNote that this exmple may not use the latest"
                 " release version. Modify based on your needs."
```

src/python/turicreate/_deps/__init__.py
guihao-liang: The way we handle pandas is different. For other packages, we use lazy-import. But for `pandas`, we use feature flags. Once it's turned off by `HAS_PANDAS`, it will be turned off everywhere. So let the user know if they don't have pandas, they won't have pandas functionality support.

However, remove this is also reasonable, if they don't have pandas, why they should use pandas?

I will remove this prompt.

src/python/turicreate/_deps/minimal_package.py
guihao-liang: That's right. Nice catch!

src/python/turicreate/toolkits/drawing_classifier/_tf_drawing_classifier.py
guihao-liang: Yeah, it's a function name to wrap the existing code.

hoytak:
Good work!  Ready to go.

PR #3176, opened 2020-05-07
Fix color augmentation in TensorFlow object detection

### LINKED ISSUE #3148: Object detector doesn't find matches

TobyRoseman:
Fixes #3148 

#3093 incorrectly set the range for saturation and contrast. This reverts to the previous behavior.

### DIFF
--- a/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
+++ b/src/python/turicreate/toolkits/object_detector/_tf_image_augmenter.py
@@ -76,13 +76,11 @@ def color_augmenter(
     image = tf.image.adjust_brightness(image, brightness_delta)
 
     # Apply the random adjustment to contrast.
-    log_sat = np.log(max_saturation)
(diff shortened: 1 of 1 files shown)


TobyRoseman:
Internal tests pass.

Benchmarking show accuracy is inline with 6.1 release for two class datasets. For a dataset with six classes accuracy is more than three times as good.

Issue #3124, opened 2020-04-14
Review "GPU Support" Documentation


TobyRoseman:
I don't think the [GPU Support Section](https://github.com/apple/turicreate#gpu-support) of our top level README is up to date.

- [x] Verify all of the min macOS version listed there are accurate. 
- [ ] Verify that "discrete GPU only" information is still accurate.

- [x] List info for One Shot Object Detector.
- [x] List info for Sound Classifier.
- [x] List info for Drawing Classier.
- [x] List info for Style Transfer.

- [ ] Clarify that we only ever use one GPU.

tagy:
With turicreate 6.1 in Ubuntu with a GPU, I have to run this command after booting my pc:

export TF_FORCE_GPU_ALLOW_GROWTH=true

Otherwise I get a memory crash as soon as I start training. It seems to be common problem people hit with tensorflow. When using tensorflow directly there is a way set this in the configuration but I could not see how to that via turicreate. More here: [https://stackoverflow.com/questions/43843208/is-it-possible-to-make-allow-growth-true-set-by-default-in-tensorflow](https://stackoverflow.com/questions/43843208/is-it-possible-to-make-allow-growth-true-set-by-default-in-tensorflow)

I don't know if this is helpful to anyone, but I spent a long time resolving this.

PR #3179, opened 2020-05-08
Update GPU section of README

TobyRoseman:
Fixes: #3124

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,15 +127,19 @@ more details on how to use Turi Create.
 GPU Support
 -----------
 
-Turi Create **does not require a GPU**, but certain models can be accelerated 9-13x when utilizing a GPU. 
+Turi Create **does not require a GPU**, but certain models can be accelerated 9-13x by utilizing a GPU.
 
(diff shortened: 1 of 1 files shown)


TobyRoseman:
@nickjong - is the "discrete GPU only" information still valid?

nickjong:
> @nickjong - is the "discrete GPU only" information still valid?

Sigh, it's complicated. The currently intended support matrix is implemented here:
https://github.com/apple/turicreate/blob/ebc837077adc83c0dbff9b5cfbf2774d459aa119/src/ml/neural_net/mps_compute_context.mm#L57

MacOS 10.15 Catalina resolved some underlying issues, and we currently expect to support integrated GPUs from `MTLFeatureSet_macOS_GPUFamily2_v1` which in practice translates to Intel HD 515 or later.

TobyRoseman:
@nickjong - I think the markdown is finally as good as it going to get. Results can are viewed [here](https://github.com/TobyRoseman/turicreate/tree/gpu-readme#gpu-support).

PR #3180, opened 2020-05-08
minimal package prompt - v2

guihao-liang:
just used to compare with #3175


guihao-liang:
core implementation for lazy import (including comments): 392 lines
client adoption code: 48 lines.

Issue #2633, opened 2019-11-14
Add shuffle to SC data iterator


fareeha:
The `_NumPyDataIterator` needs shuffle implemented.

TobyRoseman:
@fareeha - please give more information here. Is this shuffling the order of the deep feature, prior to training the custom neural network? Is this something we had prior to migrating to TensorFlow?

fareeha:
`_NumPyDataIterator` is a data iterator used to create batches for TF to perform train. We have yet to add the option to `shuffle`. This issue is created to track that.

MXNet data iterator had the option to `shuffle`.

TobyRoseman:
I think it might make more sense to implement a `shuffle` method for SFrame (i.e. issue #942) and then just use that for this method.

fareeha:
> I think it might make more sense to implement a `shuffle` method for SFrame (i.e. issue #942) and then just use that for this method.

I agree! Shuffle on SFrame would be the right place to put this effort!

TobyRoseman:
SFrame now has a [shuffle method](https://apple.github.io/turicreate/docs/api/generated/turicreate.SFrame.shuffle.html#turicreate.SFrame.shuffle). So this should be fairly straightforward.

However let's make sure to benchmark this change, making sure the runtime increase is worth the performance improvements.

PR #3188, opened 2020-05-12
Sound Classifier: shuffle data before training

TobyRoseman:
Fixes: #2633


TobyRoseman:
Internal test pass. The benchmark indicates this improves performance (as expected).

PR #3199, opened 2020-05-18
fix for minimal pkg not throw

guihao-liang:


guihao-liang:
Passed internal tests, #114746

Issue #3196, opened 2020-05-16
Turi Create not installing


mushi296:
Python is working fine on Jupyter Notebook, and I thought I installed Turi Create using the Terminal. But when I run "import turicreatei", I get the following error: 
Please help
![Screenshot 2020-05-16 at 4 36 54 PM](https://user-images.githubusercontent.com/65428028/82118890-7f4d9900-9793-11ea-8a9b-fd2aba039bde.png)

TobyRoseman:
You have most likely installed TuriCreate in a different environment than the environment you are using to run Jupyter.

The easiest way to make sure things get installed in the right environment is to install them using Jupyter Notebook. You can run system commands from inside of Jupyter Notebook by starting them with a !. So in a Jupyter code cell, run this command:
! pip install turicreate.

Let us know if that works.

mushi296:
Yes that worked, thank you so much!! 
I spent so long trying to figure this out.

I don't understand why all this has to be so complicated. Isn't there an all-in-one installer that manages all these aspects for python?

PR #3201, opened 2020-05-19
Update INSTALL_ISSUES.md

ornob011:
This fix is for those want to use turicreate with Jupyter Notebook but can't import turicreate.


INSTALL_ISSUES.md
```diff
@@ -22,3 +22,25 @@ ___
...
 You need to install Blas and Lapack. See [LINUX\_INSTALL.md](LINUX_INSTALL.md) for details.
+
+___
+
+`ModuleNotFoundError: No module named 'turicreate'` with Jupyter Notebook
+This is for Ubuntu>= 17.10. If you're using another distro, your command will be slightly different according to your package manager.
+
+sudo apt update
+sudo apt remove virtualenv python jupyter
+sudo apt install -y libstdc++6 python-setuptools libgconf-2-4
```
TobyRoseman, line 33: If you are just installing turicreate, not trying to build it from source, I don't think you need these.

INSTALL_ISSUES.md
```diff
@@ -22,3 +22,25 @@ ___
...
+sudo apt install python-pip python3-pip
+sudo pip install virtualenv notebook jupyter
```
TobyRoseman, line 35: You seem to be installing everything twice. Once for Python2 and once for Python3. I don't think this makes sense. Users should just install for which ever one they are using.

INSTALL_ISSUES.md
```diff
@@ -22,3 +22,25 @@ ___
 `ImportError: liblapack.so.3: Cannot open shared object file: No such file or directory`
 
 You need to install Blas and Lapack. See [LINUX\_INSTALL.md](LINUX_INSTALL.md) for details.
+
+___
+
+`ModuleNotFoundError: No module named 'turicreate'` with Jupyter Notebook
```
TobyRoseman, line 28: When users are having this issue, I recommend that they just install turicreate from inside of Jupyter Notebooks. I think this is approach is simpler. My instructions fo that are [here](https://github.com/apple/turicreate/issues/3196#issuecomment-629832556). I think it might make more sense to just document this approach. @ornob011 - what do you think?

INSTALL_ISSUES.md
ornob011: Yes. Both for python2 and python3. I should have commented that inside the instructions. I found that pip was only working for python2 packages. So I installed pip3 also to run python3 packages.

INSTALL_ISSUES.md
ornob011: I think you're right. Document my approach, and say that my approach should only be tried if no other approach is working.

INSTALL_ISSUES.md
ornob011: It is just for safety. I didn't want to have virtualenv, jupyter through package manager if virtualenv, jupyter will be installed by pip/pip3. So I removed them.

INSTALL_ISSUES.md
ornob011: Should I update the INSTALL_ISSUES.md file according to your suggestion and pull a new request?

TobyRoseman:
@ornob011 - thanks for you contribution here. However I'm not sure it make sense to add this change. Just telling users to `! pip install turicreate` from inside of their notebooks seems much easier. Is there a situation where installing from inside the notebook not would not work, but your solution would?

ornob011:
Nope, your `! pip install turicreate` command seems to be working in any situation. Sorry, I wasted a lot of time of yours. Actually, I didn't find this command before. So when I solved my problem I thought it would be worth sharing. Again I'm deeply sorry that I wasted your valuable time.

TobyRoseman:
No worries @ornob011. This was **not** a waste of time. Again, thanks for your contribution. Please feel free to contribute again in the future.

ornob011:
I would be glad if I could contribute in any way. Thanks for your kind word. Bye, stay safe.

PR #3208, opened 2020-05-21
Bug fix: empty SArray can not contain audio data

TobyRoseman:


TobyRoseman:
Thanks @shantanuchhabra for the review.

Issue #3126, opened 2020-04-14
tc.image_classifier.create should have a way to use extracted features


NeerajKomuravalli:
The model building with turi create takes most of its time building features and if someone wants to retrain the model with the same data or add just some amount of data it goes through the process of extracting features again and this is a waste of time. Turi Create should have a method of extracting the features and saving the same in sFrame data format and using the same to create the model with either the extracted features or the image data itself.
I could not really find a method which addresses this anywhere in the code or documentation and I would really think this a great feature to have. Please let me know if this is already there and I am not able to find it and if it's not there I would not mind working on it and send a merge request.

TobyRoseman:
This is a great idea. It's something we have talked about doing before, see #1916.

The main reason we haven't done it already is that we're not sure where to put this new function. The same deep features are used by both [the image classifier](https://apple.github.io/turicreate/docs/userguide/image_classifier/) and [the image similarity toolkit](https://apple.github.io/turicreate/docs/userguide/image_similarity/).

So do we add the `get_deep_features` method under `tc.image_classifier.`? Or add it under `tc.image_similarity.`? We could add it in both places but I think that could be confusing.

My current thinking is that the best places to add it might be under `tc.image_analysis.`. 

If you have any opinions about where to add it, please let me know.

NeerajKomuravalli:
I think using the same piece of code at two places will be both confusing and redundant so yes that might be a bad idea but why not create a new module where we can just extract features and use the same in both places. In addition we can make the tc.image_classifier.create accept data in both formats, sframe with image_path or sframe with extracted features and will have an option of saving features after model creation.

What are your thoughts?

TobyRoseman:
I agree we'll need to modify the `tc.image_classifier.create` (and `tc.image_similarity.create`) to accept deep features rather than just images.

I think adding a new module (say we call it `deep_features`) would be a little strange since there would only be one thing inside that of that module. If we add:
`tc.deep_features.get_deep_features(image_sarray, model_name)`.
There isn't going to be anything else under `tc.deep_features`.

Another option would be to add a new module and make each model have its own function, i.e. remove the `model_name` parameter. Then we would have three method: 
`tc.deep_features.Resnet(image_sarray)`
`tc.deep_features.Squeezenet(image_sarray)`
`tc.deep_features.VisionFeaturePrint_Scene(image_sarray)`
This might be a cleaner solution.

Let me know which of these two options you had in mind. Or if I misunderstood and you had something else in mind.

[...]

PR #3210, opened 2020-05-21
Feature extraction

NeerajKomuravalli:
@TobyRoseman, please have a look and let me know what you think.
Thank you!

### DIFF
--- a/src/python/turicreate/test/test_image_classifier.py
+++ b/src/python/turicreate/test/test_image_classifier.py
@@ -172,9 +172,9 @@ class ImageClassifierTest(unittest.TestCase):
     def test_select_correct_feature_column_to_train(self):
         # sending both, the correct extracted features colum and image column
         if self.feature == "awesome_image":
-            test_data = data.select_columns([self.feature, self.target, self.feature+"_deep_features"])
+            test_data = data.select_columns([self.feature, self.target, self.pre_trained_model+"_deep_features"])
             test_model = tc.image_classifier.create(test_data, target=self.target, model=self.pre_trained_model)
-            self.assertTrue(test_model.feature == self.feature+"_deep_features")
+            self.assertTrue(test_model.feature == self.pre_trained_model+"_deep_features")
 
     def test_predict(self):
         model = self.model
@@ -347,11 +347,11 @@ class ImageClassifierTest(unittest.TestCase):
 class ImageClassifierResnetTestWithDeepFeatures(ImageClassifierTest):
     @classmethod
     def setUpClass(self):
-        super(ImageClassifierSqueezeNetTest, self).setUpClass(
+        super(ImageClassifierResnetTestWithDeepFeatures, self).setUpClass(
             model="resnet-50",
-            input_image_shape=(3, 227, 227),
-            tol=0.005,
-            num_examples=200,
+            input_image_shape=(3, 224, 224),
+            tol=0.02,
+            num_examples=100,
             feature="resnet-50_deep_features",
         )
 
@@ -370,7 +370,7 @@ class ImageClassifierSqueezeNetTest(ImageClassifierTest):
 class ImageClassifierSqueezeNetTestWithDeepFeatures(ImageClassifierTest):
     @classmethod
     def setUpClass(self):
-        super(ImageClassifierSqueezeNetTest, self).setUpClass(
+        super(ImageClassifierSqueezeNetTestWithDeepFeatures, self).setUpClass(
             model="squeezenet_v1.1",
             input_image_shape=(3, 227, 227),
             tol=0.005,
@@ -403,7 +403,7 @@ class VisionFeaturePrintSceneTest(ImageClassifierTest):
 class VisionFeaturePrintSceneTestWithDeepFeatures(ImageClassifierTest):
     @classmethod
     def setUpClass(self):
-        super(VisionFeaturePrintSceneTest, self).setUpClass(
+        super(VisionFeaturePrintSceneTestWithDeepFeatures, self).setUpClass(
             model="VisionFeaturePrint_Scene",
             input_image_shape=(3, 299, 299),
             tol=0.005,
--- a/src/python/turicreate/test/test_image_classifier.py
+++ b/src/python/turicreate/test/test_image_classifier.py
@@ -19,12 +19,14 @@ from turicreate.toolkits._internal_utils import (
     _raise_error_if_not_sframe,
     _raise_error_if_not_sarray,
 )
+from turicreate.toolkits.image_analysis.image_analysis import MODEL_TO_FEATURE_SIZE_MAPPING
 
 from . import util as test_util
 
 import coremltools
 import numpy as np
 import turicreate as tc
+from array import array
 
 
 def get_test_data():
@@ -74,7 +76,13 @@ def get_test_data():
         images.append(tc_image)
 
     labels = ["white"] * 5 + ["black"] * 5
-    return tc.SFrame({"awesome_image": images, "awesome_label": labels})
+    data_dict = {"awesome_image": images, "awesome_label": labels}
+    possible_feature_length_list = list(set(MODEL_TO_FEATURE_SIZE_MAPPING.values()))
+    no_of_columns = len(data_dict[[*data_dict][0]])
+    for feature_length in possible_feature_length_list:
+        data_dict[str(feature_length)] = tc.SArray(np.random.rand(no_of_columns, feature_length)*100, dtype=array)
+
+    return tc.SFrame(data_dict)
 
 
 data = get_test_data()
@@ -133,13 +141,17 @@ class ImageClassifierTest(unittest.TestCase):
             self.assertAlmostEqual(a, b, delta=tol)
 
     def test_create_with_missing_value(self):
+        from array import array
+
+        data_dict = {
+            self.feature: tc.SArray([None], dtype=tc.Image),
+            self.target: [data[self.target][0]],
+        }
+        possible_feature_length_list = list(set(MODEL_TO_FEATURE_SIZE_MAPPING.values()))
+        for feature_length in possible_feature_length_list:
+            data_dict[str(feature_length)] = tc.SArray([None], dtype=array)
         data_with_none = data.append(
-            tc.SFrame(
-                {
-                    self.feature: tc.SArray([None], dtype=tc.Image),
-                    self.target: [data[self.target][0]],
-                }
-            )
+            tc.SFrame(data_dict)
         )
         with self.assertRaises(_ToolkitError):
             tc.image_classifier.create(
--- a/src/python/turicreate/test/test_image_classifier.py
+++ b/src/python/turicreate/test/test_image_classifier.py
@@ -88,7 +88,7 @@ class ImageClassifierTest(unittest.TestCase):
     def setUpClass(
         self,
         model="resnet-50",
-        feature="resnet-50_WithDeepFeature", 
+        feature="awesome_image", 
         input_image_shape=(3, 224, 224),
         tol=0.02,
         num_examples=100,
@@ -172,9 +172,12 @@ class ImageClassifierTest(unittest.TestCase):
     def test_select_correct_feature_column_to_train(self):
         # sending both, the correct extracted features colum and image column
         if self.feature == "awesome_image":
-            test_data = data.select_columns([self.feature, self.target, self.pre_trained_model+"_WithDeepFeature"])
+            test_data = data.select_columns([self.feature, self.target])
+            deep_features_col_name = self.pre_trained_model+"_WithDeepFeature"
+            test_data[deep_features_col_name] = get_deep_features(data["awesome_image"], 
+                self.pre_trained_model)
             test_model = tc.image_classifier.create(test_data, target=self.target, model=self.pre_trained_model)
-            self.assertTrue(test_model.feature == self.pre_trained_model+"_WithDeepFeature")
+            self.assertTrue(test_model.feature == deep_features_col_name)
 
     def test_predict(self):
         model = self.model
@@ -360,7 +363,7 @@ class ImageClassifierResnetTestWithDeepFeatures(ImageClassifierTest):
             input_image_shape=(3, 224, 224),
             tol=0.02,
             num_examples=100,
-            feature="awesome_image",
+            feature="resnet-50_WithDeepFeature",
         )
 
 
@@ -372,7 +375,7 @@ class ImageClassifierSqueezeNetTest(ImageClassifierTest):
             input_image_shape=(3, 227, 227),
             tol=0.005,
             num_examples=200,
-            feature="squeezenet_v1.1_WithDeepFeature",
+            feature="awesome_image",
         )
 
 class ImageClassifierSqueezeNetTestWithDeepFeatures(ImageClassifierTest):
@@ -383,7 +386,7 @@ class ImageClassifierSqueezeNetTestWithDeepFeatures(ImageClassifierTest):
             input_image_shape=(3, 227, 227),
             tol=0.005,
             num_examples=200,
-            feature="awesome_image",
+            feature="squeezenet_v1.1_WithDeepFeature",
         )
 
 
@@ -400,7 +403,7 @@ class VisionFeaturePrintSceneTest(ImageClassifierTest):
             tol=0.005,
             num_examples=100,
             label_type=str,
-            feature="VisionFeaturePrint_Scene_WithDeepFeature",
+            feature="awesome_image",
         )
 
 
@@ -417,5 +420,5 @@ class VisionFeaturePrintSceneTestWithDeepFeatures(ImageClassifierTest):
             tol=0.005,
             num_examples=100,
             label_type=str,
-            feature="awesome_image",
+            feature="VisionFeaturePrint_Scene_WithDeepFeature",
         )
--- a/src/python/turicreate/test/test_image_classifier.py
+++ b/src/python/turicreate/test/test_image_classifier.py
@@ -77,12 +77,6 @@ def get_test_data():
     data_dict = {"awesome_image": images, "awesome_label": labels}
     data = tc.SFrame(data_dict)
 
-    # for model_name in MODEL_TO_FEATURE_SIZE_MAPPING:
-    #     if model_name == "VisionFeaturePrint_Scene" and _mac_ver() < (10, 14):
-    #         # VisionFeaturePrint_Scene is not support on Linux
-    #         continue
-    #     data[model_name + "_deep_features"] = get_deep_features(data["awesome_image"], model_name)
-
     return data
 
 
--- a/src/python/turicreate/test/test_image_classifier.py
+++ b/src/python/turicreate/test/test_image_classifier.py
@@ -77,11 +77,11 @@ def get_test_data():
     data_dict = {"awesome_image": images, "awesome_label": labels}
     data = tc.SFrame(data_dict)
 
-    for model_name in MODEL_TO_FEATURE_SIZE_MAPPING:
-        if model_name == "VisionFeaturePrint_Scene" and _mac_ver() < (10, 14):
-            # VisionFeaturePrint_Scene is not support on Linux
-            continue
-        data[model_name + "_deep_features"] = get_deep_features(data["awesome_image"], model_name)
+    # for model_name in MODEL_TO_FEATURE_SIZE_MAPPING:
+    #     if model_name == "VisionFeaturePrint_Scene" and _mac_ver() < (10, 14):
+    #         # VisionFeaturePrint_Scene is not support on Linux
+    #         continue
+    #     data[model_name + "_deep_features"] = get_deep_features(data["awesome_image"], model_name)
 
     return data
 
@@ -94,7 +94,7 @@ class ImageClassifierTest(unittest.TestCase):
     def setUpClass(
         self,
         model="resnet-50",
-        feature="awesome_image",
+        feature="resnet-50_deep_features", 
         input_image_shape=(3, 224, 224),
         tol=0.02,
         num_examples=100,
@@ -105,6 +105,10 @@ class ImageClassifierTest(unittest.TestCase):
         self.input_image_shape = input_image_shape
         self.pre_trained_model = model
         self.tolerance = tol
+
+        # Get deep features if needed
+        if self.feature != "awesome_image":
+            data[self.feature] = get_deep_features(data["awesome_image"], self.feature.split('_deep_features')[0])
 
         self.model = tc.image_classifier.create(
             data, target=self.target, feature=self.feature, model=self.pre_trained_model, seed=42
@@ -362,7 +366,7 @@ class ImageClassifierResnetTestWithDeepFeatures(ImageClassifierTest):
             input_image_shape=(3, 224, 224),
             tol=0.02,
             num_examples=100,
-            feature="resnet-50_deep_features",
+            feature="awesome_image",
         )
 
 
@@ -374,7 +378,7 @@ class ImageClassifierSqueezeNetTest(ImageClassifierTest):
             input_image_shape=(3, 227, 227),
             tol=0.005,
             num_examples=200,
-            feature="awesome_image",
+            feature="squeezenet_v1.1_deep_features",
         )
 
 class ImageClassifierSqueezeNetTestWithDeepFeatures(ImageClassifierTest):
@@ -385,7 +389,7 @@ class ImageClassifierSqueezeNetTestWithDeepFeatures(ImageClassifierTest):
             input_image_shape=(3, 227, 227),
             tol=0.005,
             num_examples=200,
-            feature="squeezenet_v1.1_deep_features",
+            feature="awesome_image",
         )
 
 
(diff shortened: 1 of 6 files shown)


TobyRoseman:
Address #3126

src/python/turicreate/test/test_image_classifier.py
```diff
@@ -19,12 +19,14 @@
...
     _raise_error_if_not_sarray,
 )
+from turicreate.toolkits.image_analysis.image_analysis import MODEL_TO_FEATURE_SIZE_MAPPING, get_deep_features
 
 from . import util as test_util
 
 import coremltools
 import numpy as np
 import turicreate as tc
+from array import array
```
TobyRoseman, line 29: It doesn't look like you're using this import in this file.

src/python/turicreate/toolkits/image_analysis/__init__.py
```diff
@@ -7,6 +7,6 @@
 from __future__ import division as _
 from __future__ import absolute_import as _
 
-__all__ = ["image_analysis"]
+# __all__ = ["image_analysis"]
 
-from . import image_analysis
+from .image_analysis import *
```
TobyRoseman, line 12: We generally try to void `import *`. I think it's generally considered bad practice.

src/python/turicreate/toolkits/image_analysis/image_analysis.py
```diff
@@ -177,3 +190,109 @@ def resize(image, width, height, channels=None, decode=False, resample="nearest"
...
+        downloaded, the models are cached for future use.
+
+    Returns
+    -------
+    out : SArray
+        Returns an SArray with all the extracted features.
+
+    Examples
+    --------
+
```
TobyRoseman, line 229: We'll need to file this in once everything is working.

src/python/turicreate/toolkits/image_analysis/image_analysis.py
```diff
@@ -177,3 +190,109 @@ def resize(image, width, height, channels=None, decode=False, resample="nearest"
...
+        raise ValueError("'batch_size' must be greater than or equal to 1")
+
+    # Extract features
+    feature_extractor = _image_feature_extractor._create_feature_extractor(model_name)
+    images_sf = _tc.SFrame({"image":images})
+    return feature_extractor.extract_features(images_sf, "image", verbose=verbose, 
+        batch_size=batch_size)
+
+
+def find_only_image_extracted_features_column(sframe, model_name):
```
TobyRoseman, line 254: You probably want to hide this function. i.e. rename it `find_only_image_extracted_features_column` -> `_find_only_image_extracted_features_column`.

src/python/turicreate/test/test_image_classifier.py
```diff
@@ -330,6 +352,29 @@ def test_evaluate_explore(self):
...
+        super(ImageClassifierSqueezeNetTest, self).setUpClass(
+            model="resnet-50",
+            input_image_shape=(3, 227, 227),
+            tol=0.005,
+            num_examples=200,
+            feature="2048",
+        )
+
+
+class ImageClassifierSqueezeNetTest(ImageClassifierTest):
```
TobyRoseman, line 367: I'm a bit confused by the changes in the rest of this file. We already have classes named `ImageClassifierSqueezeNetTest` and `VisionFeaturePrintSceneTest`. Why are you defining them?

src/python/turicreate/test/test_image_classifier.py
```diff
@@ -74,7 +76,15 @@ def get_test_data():
...
-    return tc.SFrame({"awesome_image": images, "awesome_label": labels})
+    data_dict = {"awesome_image": images, "awesome_label": labels}
+    data = tc.SFrame(data_dict)
+
+    for model_name, feature_length in MODEL_TO_FEATURE_SIZE_MAPPING.items():
+        if str(feature_length) in data_dict.keys():
+            continue
+        data[str(feature_length)] = get_deep_features(data["awesome_image"], model_name)
+
+    return data
```
TobyRoseman, line 87: So changing `get_test_data` to return deep features probably isn't what we want. We want test cases for creating model with deep features and test cases for creating models with images. If we make this change, then we'll only have test cases for creating models with deep features.

I would probably create new sub-classes that call `get_deep_features`, in their `setUpClass` method, before calling `tc.image_classifier.create`.

src/python/turicreate/test/test_image_similarity.py
```diff
@@ -12,9 +12,13 @@
...
 import tempfile
 from . import util as test_util
+
+from turicreate.toolkits._main import ToolkitError as _ToolkitError
+from turicreate.toolkits.image_analysis.image_analysis import MODEL_TO_FEATURE_SIZE_MAPPING, get_deep_features
+
 import coremltools
 import numpy as np
-from turicreate.toolkits._main import ToolkitError as _ToolkitError
+from array import array
```
TobyRoseman, line 21: I don't think this is being used.

src/python/turicreate/test/test_image_similarity.py
```diff
@@ -63,19 +67,27 @@ def get_test_data():
...
-    return tc.SFrame({"awesome_image": images})
+    data_dict = {"awesome_image": images}
+    data = tc.SFrame(data_dict)
+
+    for model_name, feature_length in MODEL_TO_FEATURE_SIZE_MAPPING.items():
+        if str(feature_length) in data_dict.keys():
+            continue
+        data[str(feature_length)] = get_deep_features(data["awesome_image"], model_name)
+
+    return data
```
TobyRoseman, line 78: Same comment here as I made in `test_image_classifier.py`

src/python/turicreate/toolkits/image_analysis/image_analysis.py
```diff
@@ -177,3 +190,109 @@ def resize(image, width, height, channels=None, decode=False, resample="nearest"
...
+    Finds the only column in `sframe` with a type of array.array and has
+    the length same as the last layer of the model in use.
+    If there are zero or more than one image columns, an exception will
+    be raised.
+    """
+    from array import array
+
+    try:
+        feature_columns, _ = zip(*list(filter(lambda x: x[1] == array, list(zip(sframe.column_names(), sframe.column_types())))))
+    except ValueError:
```
TobyRoseman, line 265: I'm struggling to see how the above code could ever raise a `ValueError`.

src/python/turicreate/toolkits/image_analysis/image_analysis.py
```diff
@@ -177,3 +190,109 @@ def resize(image, width, height, channels=None, decode=False, resample="nearest"
...
+
+def is_image_deep_feature_sarray(feature_sarray, model_name):
+    """
+    Finds if the given `SArray` has extracted features for a given model_name.
+    """
+    from array import array
+
+    if feature_sarray.dtype != array:
+        return False
+    if type(feature_sarray[0]) != array:
```
TobyRoseman, line 294: Before executing this line, you should check `len(feature_sarray) > 0`.

src/python/turicreate/test/test_image_classifier.py
NeerajKomuravalli: You are right, this will be removed

src/python/turicreate/toolkits/image_analysis/__init__.py
NeerajKomuravalli: Then only reason I did this was because:
1. If I don't do this I don't think I can access the functions from inside image_similarity.py or image_classifier.py.
2. Also, I looked up `__init__.py` file in both image_similarity and image_classifier and there I saw `from .image_similarity import *` and `from .image_classifier import *` respectively and that's why I did this.

src/python/turicreate/toolkits/image_analysis/image_analysis.py
NeerajKomuravalli: That's what even I thought, will add examples once everything is working.

src/python/turicreate/toolkits/image_analysis/image_analysis.py
NeerajKomuravalli: My bad, will change it in my next PR

src/python/turicreate/test/test_image_similarity.py
NeerajKomuravalli: My bad, will remove it and push the corrected on in my next PR

src/python/turicreate/toolkits/image_analysis/image_analysis.py
NeerajKomuravalli: This indeed will raise a `ValueError` when the list is empty and you try to unpack the list using `zip(*...)`.
You can recreate this error with the code below:
```
import turicreate as tc
sframe_path = <Chosse a sframe with no column of type array>
sframe = tc.SFrame(sframe_path)
try :
    feature_columns, _ = zip(*list(filter(lambda x: x[1] == array, list(zip(sframe.column_names(), sframe.column_types())))))
except ValueError:
    print('It is a value error')
```

src/python/turicreate/toolkits/image_analysis/image_analysis.py
NeerajKomuravalli: You are right, I should have checked for the `feature_sarray` length. I will update this in my next PR

src/python/turicreate/test/test_image_classifier.py
NeerajKomuravalli: I agree, that's why I added one more column to the existing sframe data to use this for testing models with both Images and deep features. 
If you look at the columns that are being generated by the function `get_test_data` it has columns `awesome_image`, `awesome_label`, `2048` and `1000`. So that we can use the same data for testing both models with Images and deep features.

src/python/turicreate/test/test_image_similarity.py
NeerajKomuravalli: Same comment as above.

src/python/turicreate/test/test_image_classifier.py
NeerajKomuravalli: Yeah, this was very stupid of me. 
I wanted to create 3 more classes that would inherit properties of `ImageClassifierTest` and would use deep features to create model and then do the same test but instead of giving them different names I just ended up giving them same names. I will correct it and change it accordingly and put it in me next PR.

src/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -250,29 +251,74 @@ def create(
...
     if feature is None:
-        feature = _tkutl._find_only_image_column(dataset)
+        try:
+            feature = image_analysis.find_only_image_extracted_features_column(dataset, model)
+            feature_type = "extracted_features_array"
+        except:
+            feature = None
+
+        if feature is None:
+            try:
```
TobyRoseman, line 261: Why not do all this in the `except` block above?

src/python/turicreate/toolkits/image_analysis/image_analysis.py
```diff
@@ -177,3 +190,111 @@ def resize(image, width, height, channels=None, decode=False, resample="nearest"
...
+    """
+    Finds the only column in `sframe` with a type of array.array and has
+    the length same as the last layer of the model in use.
+    If there are zero or more than one image columns, an exception will
+    be raised.
+    """
+    from array import array
+
+    try:
+        feature_columns, _ = zip(*list(filter(lambda x: x[1] == array, list(zip(sframe.column_names(), sframe.column_types())))))
```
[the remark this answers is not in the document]
NeerajKomuravalli: This makes sense, Will use `_find_only_column_of_type`

src/python/turicreate/toolkits/image_classifier/image_classifier.py
NeerajKomuravalli: Basically that part of the code is responsible for figuring out which column is a feature column in case the feature column in not passed by the user. 
And in that case we first see if there is a deep feature column in the sframe and if yes, we use it and if it's not found we go ahead and look for Image column. 
The try block will fail for two reasons: 
1.  There is no deep feature column
2. More than one deep feature columns were present in the sframe. (This I am assuming is by design because we throw a similar error even when more than one Image columns were found in the dataset.)
And when the `try` and `except` fail that means one of the above two cases would have happened and in that case we ignore the exception and we move ahead to see if we can find a Image column in the sframe dataset.
So this choice was by design. Please let me know if I am missing something here.

src/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -250,29 +251,74 @@ def create(
...
+                feature = _tkutl._find_only_image_column(dataset)
+                feature_type = "image"
+            except:
```
TobyRoseman, line 264: You shouldn't need this `except` block. `_find_only_image_column` will raise it's own exception, if it doesn't find what it needs.

src/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -835,13 +890,26 @@ def evaluate(self, dataset, metric="auto", verbose=True, batch_size=64):
...
+        elif dataset[self.feature].dtype is _tc.Image:
+            return _tc.SFrame(
+                {
+                    "__image_features__": self.feature_extractor.extract_features(
+                        dataset, self.feature, verbose=verbose, batch_size=batch_size
+                    )
+                }
+            )
+        else:
+            raise _ToolkitError('The "{feature}" column of the sFrame neither has the dataype image or extracted features array.'.format(feature=feature)
```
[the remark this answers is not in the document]
NeerajKomuravalli: Will make this change.

src/python/turicreate/toolkits/image_classifier/image_classifier.py
NeerajKomuravalli: You are right, will remove it.

NeerajKomuravalli:
I will make all the changes requested and test it properly and make sure all the tests are passing and will give you a detailed report of the same.

TobyRoseman:
I've pulled down these changes, built them and ran the unit tests. All the failures seem related to passing deep features to exported CoreML models. This might be something we want to support in the future, but I certainly don't think this is something we should worry about in for pull request.

Let's just not run the `test_export_coreml`/`test_export_coreml_predict` methods for the deep feature test classes. So you probably want to revert your last two commits. Then you could just return right away in those functions if deep features are being used. I'm totally open to doing this another way, but I think this going to be the simplest way.

Other than that, we need to fill in the example text for the `get_deep_features` docstring. It also looks like there is a git conflict which needs to be resolved for `test_image_similarity.py`.

@NeerajKomuravalli - Thanks for your work here!

NeerajKomuravalli:
Hi @TobyRoseman 

1. Made changes in `test_export_coreml/test_export_coreml_predict` functions in `test_image_similarity/test_image_classifier` to not test when deep_features are used
2. Added examples for `get_deep_features` in it's docstring
3. Resolved the conflict in `test_image_similarity.py`

Please have a look and let me know if anything else is needed.

Thanks!

src/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -390,7 +440,7 @@ def query(self, dataset, label=None, k=5, radius=None, verbose=True, batch_size=
 
         if isinstance(dataset, _tc.SArray):
             dataset = _tc.SFrame({self.feature: dataset})
-        elif isinstance(dataset, _tc.Image):
+        elif isinstance(dataset, _tc.Image) or isinstance(dataset, array):
```
nishantnath, line 443:
```suggestion
        elif isinstance(dataset, (_tc.Image, array)):
```

src/python/turicreate/toolkits/image_classifier/image_classifier.py
```diff
@@ -444,12 +482,15 @@ def _canonize_input(self, dataset):
         along with an unpack callback function that can be applied to
         prediction results to "undo" the canonization.
         """
+        from array import array
+
         unpack = lambda x: x
         if isinstance(dataset, _tc.SArray):
             dataset = _tc.SFrame({self.feature: dataset})
-        elif isinstance(dataset, _tc.Image):
+        elif isinstance(dataset, _tc.Image) or isinstance(dataset, array):
```
nishantnath, line 490:
```suggestion
        elif isinstance(dataset, (_tc.Image, array)):
```

src/python/turicreate/toolkits/image_similarity/image_similarity.py
NeerajKomuravalli: Made the suggested changes

src/python/turicreate/toolkits/image_classifier/image_classifier.py
NeerajKomuravalli: Made the suggested changes

NeerajKomuravalli:
Hi @nishantnath  @TobyRoseman, 
I have made all the suggested changes, let me know if any more work or changes are needed.

src/python/turicreate/test/test_image_classifier.py
```diff
@@ -264,17 +264,8 @@ def test_export_coreml_predict(self):
...
-                classes = self.model.classifier.classes
-                ret = coreml_model.predict({self.feature: deep_features})
-                coreml_values = [ret[self.target + "Probability"][l] for l in classes]
-
-                self.assertListAlmostEquals(
-                    coreml_values,
-                    list(self.model.predict(deep_features, output_type="probability_vector")),
-                    self.tolerance,
-                )                
+            # If the code came here that means the type of the feature used is deep_deatures and the predict fwature in coremltools doesn't work with deep_features yet so we will ignore this specific test case unitl the same is written.
```
TobyRoseman, line 270: Should "fwature" be "function"? We try to maintain a 80 or 100 character length line limit. I would be nice if the text wrapped to the next line after 80 or 100 characters.

src/python/turicreate/test/test_image_similarity.py
```diff
@@ -276,17 +276,8 @@ def get_psnr(x, y):
...
-
-            if _mac_ver() >= (10, 13):
-                coreml_ret = coreml_model.predict({self.feature: deep_features})
-
-                # Compare distances
-                coreml_distances = np.array(coreml_ret["distance"])
-                tc_distances = tc_ret.sort("reference_label")["distance"].to_numpy()
-                psnr_value = get_psnr(coreml_distances, tc_distances)
-                self.assertTrue(psnr_value > 50)
+            # If the code came here that means the type of the feature used is deep_deatures and the predict fwature in coremltools doesn't work with deep_features yet so we will ignore this specific test case unitl the same is written.
```
TobyRoseman, line 281: same comment

TobyRoseman:
@NeerajKomuravalli - Your most recent changes look good, just one minor comment. I'm running your changes now on an internal testing system. If that all passes, I think we're ready to merge this change.

TobyRoseman:
The unit tests are failing on Linux. The `VisionFeaturePrint_Scene` model is not available on Linux.

We should skip the tests for that model if `_mac_ver() < (10, 14)` evaluates to True. Also we should not being trying to extract deep features for that model.

NeerajKomuravalli:
@TobyRoseman 
I went through both, `test_image_classifier.py` and `test_image_similarity.py` and in both the files we are only testing for `VisionFeaturePrint_Scene` model if `_mac_ver() < (10, 14)`. 
Can you point me to the code where this might be failing?

src/python/turicreate/test/test_image_classifier.py
```diff
@@ -73,7 +74,13 @@ def get_test_data():
         images.append(tc_image)
 
     labels = ["white"] * 5 + ["black"] * 5
-    return tc.SFrame({"awesome_image": images, "awesome_label": labels})
+    data_dict = {"awesome_image": images, "awesome_label": labels}
+    data = tc.SFrame(data_dict)
+
+    for model_name in MODEL_TO_FEATURE_SIZE_MAPPING:
+        data[model_name + "_deep_features"] = get_deep_features(data["awesome_image"], model_name)
```
TobyRoseman, line 81: I think you need to add a check like this here:
```
if model_name == "VisionFeaturePrint_Scene" and _mac_ver() < (10, 14):
    # VisionFeaturePrint_Scene is not support on Linux
    continue
```

TobyRoseman:
> @TobyRoseman
> I went through both, `test_image_classifier.py` and `test_image_similarity.py` and in both the files we are only testing for `VisionFeaturePrint_Scene` model if `_mac_ver() < (10, 14)`.
> Can you point me to the code where this might be failing?

Ok, if we're already skipping those classes on Linux, then we just need to skip extracting the deep features on Linux (since that is done independently of the test classes). I've just added comments in the code about that.

NeerajKomuravalli:
Hi @TobyRoseman ,

I have made the requested changes, let me know if there is anything else.

TobyRoseman:
Thanks @NeerajKomuravalli. Those changes look good. I've just kicked off another internal test run. I'll let you know the results.

src/python/turicreate/test/test_image_similarity.py
```diff
@@ -12,8 +12,12 @@
 from turicreate.toolkits._internal_utils import _mac_ver
 import tempfile
 from . import util as test_util
-import numpy as np
+
 from turicreate.toolkits._main import ToolkitError as _ToolkitError
+from turicreate.toolkits.image_analysis.image_analysis import MODEL_TO_FEATURE_SIZE_MAPPING, get_deep_features
+
+import coremltools
```
TobyRoseman, line 19: This is no longer getting used.

TobyRoseman: Unfortunately due to the (brittle) way that we test the minimal version of TuriCreate, we are going to need to remove this unnecessary line. `coremltools` is not a dependency for the minimal version of our TuriCreate. So having it imported at a top level break our unit tests even though the tests in this file are not ran for the minimal version.

I believe removing this line should be the final required change for this pull request. All of our other internal tests are passing.

NeerajKomuravalli: Makes sense, will remove the `import coremltools` line and push changes

NeerajKomuravalli:
Made the required changes!

TobyRoseman:
Thanks @NeerajKomuravalli - I'm rerunning the internal tests now.

TobyRoseman:
Our unit tests for the minimal version of our package is still failing with this change. When our minimal package is installed we can't call `get_deep_features`. Since `get_test_data` is called just by importing the unit test files, that means `get_deep_features` gets called. We need to have to make it so that `get_deep_features` is only called from inside of a `setUpClass` method.

NeerajKomuravalli:
Makes sense, I shifted the `get_deep_features` call from `get_test_data` to `setUpClass` in both `test_image_similarity.py`and `test_image_classifier.py`. 
Let me know if there is anything else that is required!

src/python/turicreate/test/test_image_classifier.py
```diff
@@ -84,19 +88,24 @@ class ImageClassifierTest(unittest.TestCase):
     def setUpClass(
         self,
         model="resnet-50",
+        feature="resnet-50_deep_features",
```
TobyRoseman, line 91: I don't think this right. If the class name doesn't end in `WithDeepFeature`, then I think it should be `feature="awesome_image"`. This code is only calling `get_deep_features` when `self.feature != "awesome_image"`.

I think you need to make this change to all of your test classes.

NeerajKomuravalli: Changed the deep feature column name in test data to the one suggested by you and made all the requested changes.

TobyRoseman: I think you misunderstood my previous comment. I was **not** suggesting you need to change the name of the feature. I was basically just trying to say that it's important for the name of the class to represent what that class is actually testing.

For example, if the name of the class is `ImageClassifierResnetTestWithDeepFeatures`, then it should be `feature="resnet-50_WithDeepFeature"` or `feature="resnet-50_deep_features",` but not `feature="awesome_image",`. Similarly, if the name of the class is `VisionFeaturePrintSceneTest` then it should not be creating the model from deep features, so it should be `feature="awesome_image",` not `feature="VisionFeaturePrint_Scene_WithDeepFeature"`.

Does that make sense? Let me know if you have any questions.

NeerajKomuravalli: Oh yeah, you are right I did not see it there. It was correct before but during one of my recent pushes I changed it to test something and forgot to revert it back. My bad, will change it and push it. 
And about changing the name of feature columns in the test data in `test_image_classifier.py` and `test_image_similarity.py` to have a suffix `_WithDeepFeature` instead of `_deep_features` makes more sense to me. So I am keeping that change as it is.

NeerajKomuravalli:
I made the required changes. Let me know if you need anything else.

TobyRoseman:
@NeerajKomuravalli - Those changes look good. Thank you. I'm rerunning our internal tests now.

NeerajKomuravalli:
That's a great news @TobyRoseman ! It was my first contribution to an open source project so I am really excited.

It was great working on this, let me know if I can help contribute more!

Thanks!

PR #3211, opened 2020-05-23
In set_runtime_config, if bad SSL cert path is provided, exception FileNotFoundError will be thrown. Additionally, relative file path is converted to absolute file path.

kritikadusad:
This PR resolves the issue of the following code completing without an error message when a bad SSL cert filepath is provided.
```
import turicreate as tc
tc.config.set_runtime_config("TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE",
    "/this_file_does_not_exist")  
```
With an incorrect path provided, a FileNotfoundError is thrown with the message: `</this_file_does_not_exist> does not exist.`
The PR also addresses relative paths and paths with `~` not working with the function. Now such paths will be expanded to absolute paths. 
Also added a test to `test_util.py` that tests whether an incorrect filepath does throw an error. The test also checks whether relative paths are working with the function. 
This PR fixes #3211

### DIFF
--- a/src/python/turicreate/config/__init__.py
+++ b/src/python/turicreate/config/__init__.py
@@ -330,11 +330,12 @@ def set_runtime_config(name, value):
 
     unity = _glconnect.get_unity()
 
-    # Expands relative path to absolute path
-    value = _os.path.abspath(_os.path.expanduser(value))
-    # Raises exception if path for SSL cert does not exist
-    if not _os.path.exists(value):
-        raise FileNotFoundError(f"{value} does not exist.")
+    if name == "TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE":
+        # Expands relative path to absolute path
+        value = _os.path.abspath(_os.path.expanduser(value))
+        # Raises exception if path for SSL cert does not exist
+        if not _os.path.exists(value):
+            raise FileNotFoundError(f"{value} does not exist.")
 
     ret = unity.set_global(name, value)
     if ret != "":
--- a/src/python/turicreate/config/__init__.py
+++ b/src/python/turicreate/config/__init__.py
@@ -335,7 +335,7 @@ def set_runtime_config(name, value):
         value = _os.path.abspath(_os.path.expanduser(value))
         # Raises exception if path for SSL cert does not exist
         if not _os.path.exists(value):
-            raise FileNotFoundError("{} does not exist.".format(value))
+            raise ValueError("{} does not exist.".format(value))
 
     ret = unity.set_global(name, value)
     if ret != "":
--- a/src/python/turicreate/config/__init__.py
+++ b/src/python/turicreate/config/__init__.py
@@ -329,6 +329,13 @@ def set_runtime_config(name, value):
     from .._connect import main as _glconnect
 
     unity = _glconnect.get_unity()
+
+    # Expands relative path to absolute path
+    value = _os.path.abspath(_os.path.expanduser(value))
+    # Raises exception if path for SSL cert does not exist
+    if not _os.path.exists(value):
+        raise FileNotFoundError(f"{value} does not exist.")
+
     ret = unity.set_global(name, value)
     if ret != "":
         raise RuntimeError(ret)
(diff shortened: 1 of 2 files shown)


src/python/turicreate/test/test_util.py
```diff
@@ -243,3 +243,35 @@ def test_make_temp_filename(self):
         self.assertTrue(location.startswith(tmp_root))
         self.assertTrue(isinstance(location, str))
         self.assertTrue(os.path.basename(location).startswith(prefix))
+
+    def test_alternate_ssl_path(self):
+        from ..util import _get_temp_file_location
```
TobyRoseman, line 248: You are not using either of these two imports.

src/python/turicreate/config/__init__.py
```diff
@@ -329,6 +329,14 @@ def set_runtime_config(name, value):
     from .._connect import main as _glconnect
 
     unity = _glconnect.get_unity()
+
+    if name == "TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE":
+        # Expands relative path to absolute path
+        value = _os.path.abspath(_os.path.expanduser(value))
+        # Raises exception if path for SSL cert does not exist
+        if not _os.path.exists(value):
+            raise FileNotFoundError(f"{value} does not exist.")
```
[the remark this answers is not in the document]
kritikadusad: Thank you! I have addressed this issue.

src/python/turicreate/test/test_util.py
kritikadusad: Thank you! I have removed them from the test.

src/python/turicreate/config/__init__.py
```diff
@@ -329,6 +329,14 @@ def set_runtime_config(name, value):
     from .._connect import main as _glconnect
 
     unity = _glconnect.get_unity()
+
+    if name == "TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE":
+        # Expands relative path to absolute path
+        value = _os.path.abspath(_os.path.expanduser(value))
+        # Raises exception if path for SSL cert does not exist
+        if not _os.path.exists(value):
+            raise FileNotFoundError("{} does not exist.".format(value))
```
TobyRoseman, line 338: Sadly, it looks like `FileNotFoundError` also isn't in Python 2.7. I'm open to other ideas, but I would probably just change `FileNotFoundError ` to `ValueError`.

src/python/turicreate/test/test_util.py
```diff
@@ -243,3 +243,32 @@ def test_make_temp_filename(self):
...
+            "TURI_FILEIO_ALTERNATIVE_SSL_CERT_FILE"
+        ]
+
+        tmp = tempfile.mkdtemp(prefix="test_gl_util")
+        tmp_relative_path = os.path.relpath(tmp)
+        incorrect_path = "/this_file_does_not_exist"
+
+        try:
+            self.assertRaises(
+                FileNotFoundError,
```
TobyRoseman, line 258: You'll only need to change this line.

src/python/turicreate/config/__init__.py
kritikadusad: Ah! My bad. Fixed it. 
For the future, to test for Python2.7, how would you recommend doing it? 
I am thinking of using a virtualenv with Python2.7 and using that to make and test.

src/python/turicreate/test/test_util.py
kritikadusad: Changed

TobyRoseman:
Internal tests passed.

src/python/turicreate/config/__init__.py
TobyRoseman: In general testing multiple Python versions using virtualenv is the way to go. However it's difficult to test local TuriCreate changes across multiple Python versions; when setting up your local environment it gets bound to a particular Python version (this happens when you run `./configure`).

I think what happened here is a bit of an anomaly. Usually getting code to run in both Python 2 and 3 is pretty easy. Usually if code runs in Python 3, it will run in Python 2.

TobyRoseman:
@kritikadusad - thanks for the contribution.

PR #3213, opened 2020-05-26
Adding functions to darknet yolo model 

shreyajain17:


src/toolkits/object_detection/object_detector.cpp
```diff
@@ -419,18 +416,20 @@ variant_type object_detector::evaluate(gl_sframe data, std::string metric,
     log_and_throw("Annotations column " + annotations_column_name +
                   " does not exist");
   }
+  // If called during training, synchronize the model first.
```
TobyRoseman, line 419: I'm not really understanding this comment here, or the two other places this comment appears.

Issue #745, opened 2018-06-25
SArray should have an implementation of median


srikris:
A approx/exact median is very useful for SArrays

```
sa.median(approx = False) # default
sa.median(approx = True)  # uses the sketch summary to compute the median
```

TobyRoseman:
I like this idea. It's pretty simple to do efficiently at the Python layer. As a proof of concept, the following works:

```python
def median(sa, approx = False):
    if not approx:
        temp = tc.SFrame({'x': sa})
        temp['dummy'] = 1
        result = temp.groupby('dummy', {'y': tc.aggregate.QUANTILE('x', 0.5)})
        return result['y'][0][0]
    else:
        sketch_summary = tc.Sketch(sa)
        return sketch_summary.quantile(.5)
```

srikris:
@TobyRoseman We should do exactly that in the C++ layer :). I think the code is just as easy and we can also put it in the C-APIs.

TobyRoseman:
[The code I shared previously](https://github.com/apple/turicreate/issues/745#issuecomment-408253208) is not going to work when `approx = False`. It turns out we can not use `tc.aggregate.QUANTILE` if we want exact answers. 

The docstring for `tc.aggregate.QUANTILE` contains this surprising information:
```
The returned quantiles are guaranteed to have 0.5% accuracy. That is to say,
if the requested quantile is 0.50, the resultant quantile value may be
between 0.495 and 0.505 of the true quantile.
```

hoytak:
Let's use the sketch for getting the exact method by using the bound guarantees of the sketch.  Because it guarantees the value to be within eps * N of the original value, we set eps based on memory (it's pretty conservative, so that can be set small), then get the quantiles of the data at (0.5 - eps) *N and (0.5 + eps) *N.  Call these values a and b.  We then know that the true median lies between a and b. 

Then, we do a second pass on the data using an apply: 
```
atomic<size_t> n_below_a = 0; 
std::vector<flexible_type> candidates;  // could be replaced with internal lock free stack.
std::mutex candidate_lock; 

auto count_median = [&](flexible_type x) { 
   if(x < a) { ++n_below_a; }
   else if(x <= b) {
      std::lock_guard<std::mutex> lg(candidate_lock); 
      candidates.push_back(x);
   }
}); 
data.apply(count_median);   // May be reduce_with_lambda or something like that. 

Now, what's the 50% mark?  
[...]

PR #3222, opened 2020-06-06
SArray Median

TobyRoseman:
implements #745

### DIFF
--- a/src/core/storage/sframe_interface/unity_sarray.cpp
+++ b/src/core/storage/sframe_interface/unity_sarray.cpp
@@ -1109,22 +1109,23 @@ flexible_type unity_sarray::median(bool approx) {
   if(type != flex_type_enum::INTEGER && type != flex_type_enum::FLOAT) {
     log_and_throw("Can not calculate median on non-numeric SArray.");
   }
-  if (size() == 0) {
-    return flex_undefined();
-  }
 
   // Use sketch to get a fast approximate answer
   turi::unity_sketch* sketch = new turi::unity_sketch();
   gl_sarray data = std::static_pointer_cast<unity_sarray>(shared_from_this());
   data = data.dropna();
+  if (data.size() == 0) {
+    return flex_undefined();
+  }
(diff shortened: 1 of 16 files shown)


src/core/storage/sframe_interface/unity_sarray.cpp
```diff
@@ -1102,6 +1102,59 @@ flexible_type unity_sarray::mean() {
...
 }
 
+flexible_type unity_sarray::median(bool approx) {
+  log_func_entry();
+
+  flex_type_enum type = dtype();
+  if(type != flex_type_enum::INTEGER && type != flex_type_enum::FLOAT) {
+    log_and_throw("Can not calculate median on non-numeric SArray.");
+  }
+  if (size() == 0) {
```
hoytak, line 1112: Put after the dropna()

src/core/storage/sframe_interface/unity_sarray.cpp
```diff
@@ -1102,6 +1102,59 @@ flexible_type unity_sarray::mean() {
...
+  gl_sarray data = std::static_pointer_cast<unity_sarray>(shared_from_this());
+  data = data.dropna();
+  sketch->construct_from_sarray(data);
+  double approx_median = sketch->get_quantile(0.5);
+
+  flexible_type result;
+  if (!approx) {
+    const double epsilon = sketch->numeric_epsilon();
+    const double upper_bound = approx_median + (epsilon * data.size());
+    const double lower_bound = approx_median - (epsilon * data.size());
```
hoytak, line 1127: Unfortunately, the guarantees on the quantile don't work by value, but by element count.  The approximate median returned is the element that is in position +/- eps*data.size() of the true median.  Thus to get the upper and lower bounds, you need to actually do: 
```
const double upper_bound = sketch->get_quantile(0.5 + epsilon * data.size()); 
const double lower_bound = sketch->get_quantile(0.5 - epsilon * data.size()); 
```

PR #3228, opened 2020-06-17
Update a couple of python dependencies

TobyRoseman:
Update dependencies (TensorFlow and Pillow). Must still rely on the old version TensorFlow for Python 2.7; it is the last version to support 2.7.

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -7,7 +7,7 @@ scipy==1.2.1
 resampy==0.2.1
 scikit-learn==0.20.3
 statsmodels==0.9.0
-tensorflow==2.0.0
+tensorflow==2.0.1
 UISoup==2.5.7
 pyobjc==5.2; sys_platform == 'darwin'
 llvmlite==0.31.0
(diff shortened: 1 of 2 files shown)


TobyRoseman:
Internal test pass

scripts/requirements.txt
```diff
@@ -7,7 +7,8 @@ scipy==1.2.1
 resampy==0.2.1
 scikit-learn==0.20.3
 statsmodels==0.9.0
-tensorflow==2.0.0
+tensorflow==2.0.1; python_version > '2.7'
+tensorflow==2.0.0; python_version < '3.5'   # Last version to support Python 2.7
```
nickjong, line 11: Did they really drop 2.7 support between TF2.0.0 and TF2.0.1?

nickjong: Hmm, this page on PyPI suggests that TF2.0.1 supports Python 2.7? https://pypi.org/project/tensorflow/2.0.1/

TobyRoseman: Sadly, TensorFlow did drop support for Python 2.7 between 2.0.0 and 2.0.1. I've tried and you can't install 2.0.1 in 2.7.

Yes, Python 2.7 is listed at the bottom on the left hand side on the PyPI page for 2.0.1. However if you click on [Download Files](https://pypi.org/project/tensorflow/2.0.1/#files), there are no wheel for 2.7.

nickjong: Wow

PR #3233, opened 2020-06-22
kwargs flow through for Image Similarity -> Nearest Neighbors

nishantnath:
This will allow users to use methods for nearest neighbor model apart from 'auto' along with other params like distance & nearest neighbor's kwargs

Nearest neighbor already checks for validity of kwargs downstream.

### DIFF
--- a/src/python/turicreate/toolkits/image_similarity/image_similarity.py
+++ b/src/python/turicreate/toolkits/image_similarity/image_similarity.py
@@ -22,7 +22,13 @@ from .. import _image_feature_extractor
 
 
 def create(
-    dataset, label=None, feature=None, model="resnet-50", verbose=True, batch_size=64
+    dataset,
+    label=None,
+    feature=None,
+    model="resnet-50",
+    verbose=True,
+    batch_size=64,
+    **kwargs
 ):
     """
     Create a :class:`ImageSimilarityModel` model.
@@ -63,6 +69,9 @@ def create(
     batch_size : int, optional
         If you are getting memory errors, try decreasing this value. If you
         have a powerful computer, increasing this value may improve performance.
+        
+    **kwargs : optional
+        Options for downstream methods like tc.nearest_neighbors.create().
 
     Returns
     -------
@@ -148,6 +157,7 @@ def create(
         label=label,
         features=["__image_features__"],
         verbose=verbose,
+        **kwargs
     )
 
     # set input image shape


TobyRoseman:
@nishantnath - thanks for the pull request. I like this idea and the change looks good. Would it be possible to add a unit test for this new functionality?

nishantnath:
@TobyRoseman I did ponder over a unit test but considering kwargs would be passing through to downstream functions and each downstream function would (or should) have its own checks for params & unit tests - seemed redundant.

Downstream methods:
feature_extractor.extract_features (doesn't support kwargs)
nearest_neighbors.create (supports kwargs): has own unit test + kwargs param check

Happy to add an unit test if you can provide some pointers on the direction you are envisioning for the unit test.

src/python/turicreate/toolkits/image_similarity/image_similarity.py
```diff
@@ -63,6 +69,9 @@ def create(
     batch_size : int, optional
         If you are getting memory errors, try decreasing this value. If you
         have a powerful computer, increasing this value may improve performance.
+        
+    **kwargs : optional
+        Options for downstream methods like tc.nearest_neighbors.create().
```
TobyRoseman, line 74: Please change `tc.nearest_neighbors.create()` to `:py:func:`turicreate.nearest_neighbors.create()`. That will cause the text, in our API docs, to be linked to the appropriate documentation page for that function.

TobyRoseman:
> @TobyRoseman I did ponder over a unit test but considering kwargs would be passing through to downstream functions and each downstream function would (or should) have its own checks for params & unit tests - seemed redundant.
> 
> Downstream methods:
> feature_extractor.extract_features (doesn't support kwargs)
> nearest_neighbors.create (supports kwargs): has own unit test + kwargs param check
> 
> Happy to add an unit test if you can provide some pointers on the direction you are envisioning for the unit test.

@nishantnath - I agree a unit test here is largely redundant. However I still think it's worth doing since it will ensure we never accidentally remove this functionality. 

If I was going to add a unit test for this, I would probably add a new add a new testing class to our [image similarity unit test](https://github.com/apple/turicreate/blob/master/src/python/turicreate/test/test_image_similarity.py). The `setUpClass` method could create a model with a small dataset and pass it a kwarg value. 
Something like this:
```python
self.model = tc.image_similarity.create(data.head(), distance='squared_euclidean')
```

Then we could have a single test method that just checks the similarity model has that distance. I think it would be something like:
```python
self.assertEqual(
    self.model.similarity_model..distance[0][1],
    'squared_euclidean'
)
```

src/python/turicreate/test/test_image_similarity.py
```diff
@@ -287,6 +291,60 @@ def test_save_and_load(self):
             print("Export coreml passed")
 
 
+class ImageSimilarityTestWithKwargs(unittest.TestCase):
+    @classmethod
+    def setUpClass(self, input_image_shape=(3, 224, 224), model="resnet-50"):
+        """
+        The setup class method for the basic test case with all default values.
+        """
+        self.feature = "awesome_image"
```
TobyRoseman, line 300: I don't think you need the majority of the code in this `setUpClass`. It looks like the only member variable that's actually getting used is `self.model`. I think everything else should be removed.

src/python/turicreate/test/test_image_similarity.py
```diff
@@ -251,21 +251,25 @@ def get_psnr(x, y):
...
-
-            pil_img = _PIL_Image.fromarray(img_fixed.pixel_data)
-            coreml_ret = coreml_model.predict({"awesome_image": pil_img})
-
-            # Compare distances
-            coreml_distances = np.array(coreml_ret["distance"])
-            tc_distances = tc_ret.sort("reference_label")["distance"].to_numpy()
-            psnr_value = get_psnr(coreml_distances, tc_distances)
-            self.assertTrue(psnr_value > 50)
+        if self.feature == "awesome_image":
```
TobyRoseman, line 254: What are you trying to accomplish by this change?

PR #3245, opened 2020-07-01
Quantize weights from fp32 to fp16

shreyajain17:
Adds option to quantize weights post training fp32 to fp16 (only for MacOS)
Enables it for Object Detector model 
Also adds pooling types to model_spec

### DIFF
--- a/src/ml/neural_net/float16.h
+++ b/src/ml/neural_net/float16.h
@@ -1,0 +1,20 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#pragma once
+
+#ifdef __APPLE__
+#define TURI_USE_FLOAT16
+#endif
+
+/*
+ Source of the numeric limits -
+ https://github.com/apple/coremltools/blob/a63319be4c70271f229c9f5c3423a97c7535f08b/coremltools/models/utils.py#L172
+ */
+#if defined(TURI_USE_FLOAT16)
+const constexpr __fp16 FLOAT16_NUMERIC_LIMIT_MIN = 0.000061035;
+const constexpr __fp16 FLOAT16_NUMERIC_LIMIT_MAX = 65504;
+#endif
--- a/src/ml/neural_net/float16.h
+++ b/src/ml/neural_net/float16.h
@@ -7,14 +7,18 @@
 #pragma once
 
 #ifdef __APPLE__
+
 #define TURI_USE_FLOAT16
+
 #endif
 
 /*
  Source of the numeric limits -
  https://github.com/apple/coremltools/blob/a63319be4c70271f229c9f5c3423a97c7535f08b/coremltools/models/utils.py#L172
  */
-#if defined(TURI_USE_FLOAT16)
+#ifdef TURI_USE_FLOAT16
+
 const constexpr __fp16 FLOAT16_NUMERIC_LIMIT_MIN = 0.000061035;
 const constexpr __fp16 FLOAT16_NUMERIC_LIMIT_MAX = 65504;
+
 #endif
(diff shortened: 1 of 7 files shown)


src/ml/neural_net/float16.h
```diff
@@ -0,0 +1,20 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#pragma once
+
+#ifdef __APPLE__
```
brtal, line 9: please be consistent (especially within the same file) with `ifdef` vs `if defined`. i tend to prefer `if defined`.

src/ml/neural_net/model_spec.cpp
```diff
@@ -49,6 +51,9 @@ using CoreML::Specification::TransposeLayerParams;
 using CoreML::Specification::UniDirectionalLSTMLayerParams;
 using CoreML::Specification::UpsampleLayerParams;
 using CoreML::Specification::WeightParams;
+using CoreML::Specification::PoolingLayerParams_PoolingType::PoolingLayerParams_PoolingType_AVERAGE;
```
brtal, line 54: i don't recall this from your original PR?

src/ml/neural_net/model_spec.hpp
```diff
@@ -122,7 +125,7 @@ class model_spec {
    * \throw If a float_array's shape does not match the corresponding
    *        NeuralNetworkLayer.
    */
-  void update_params(const float_array_map& weights);
+  void update_params(const float_array_map& weights, bool use_quantization = false);
```
brtal, line 128: don't forget to update the documentation for `use_quantization` in the above comment block

src/ml/neural_net/quantization_utils.cpp
```diff
@@ -0,0 +1,37 @@
...
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#include <ml/neural_net/quantization_utils.hpp>
+
+#include <algorithm>
+#include <cmath>
+#include <numeric>
+
+#include <ml/neural_net/float16.h>
```
brtal, line 14: include `float16.h` in the `quantization_utils.hpp` and drop it here

src/ml/neural_net/quantization_utils.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace turi {
+namespace neural_net {
+
+#ifdef TURI_USE_FLOAT16
```
brtal, line 19: i would do a single check for float16 being enabled across both methods. leave a blank line between the `#if`/`#endif` around function declarations please.

src/ml/neural_net/quantization_utils.cpp
```diff
@@ -0,0 +1,37 @@
...
+std::vector<__fp16> get_half_precision_weights(Span<const float> output)
+{
+  std::vector<__fp16> weights_fp16 = std::vector<__fp16>(output.begin(), output.end());
+  return weights_fp16;
+}
+#endif
+
+}  // namespace neural_net
+}  // namespace turi
```
brtal, line 37: github reminds you to add a blank line at the end of this file

src/ml/neural_net/quantization_utils.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma once
+
+#include <core/util/Span.hpp>
+#include <vector>
+
+namespace turi {
+namespace neural_net {
+
+#ifdef TURI_USE_FLOAT16
```
brtal, line 15: do a single `#if` like i suggest above.

src/ml/neural_net/float16.h
shreyajain17: I think in our repo, we mostly `#if def`

shreyajain17: I changed all of them to be `#ifdef`

src/ml/neural_net/quantization_utils.cpp
shreyajain17: Okay let me change it everywhere

src/ml/neural_net/float16.h
shreyajain17: Sorry I didn't change it in that file. Changing it now

PR #3253, opened 2020-07-09
Activity Classifier use non-deprecated APIs

### LINKED ISSUE #3041: Activity Classifier Using Deprecated TensorFlow Method
The activity classifier call `tf.compat.v1.nn.dynamic_rnn`. According to its [API docs](https://www.tensorflow.org/api_docs/python/tf/compat/v1/nn/dynamic_rnn): `Warning: THIS FUNCTION IS DEPRECATED. It will be removed in a future version. Instructions for updating: Please use keras.layers.RNN(cell), which is equivalent to this API`.

TobyRoseman:
Fixes #3041

### DIFF
--- a/src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
+++ b/src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
@@ -207,7 +207,7 @@ class ActivityTensorFlowModel(TensorFlowModel):
             prob, (prob.shape[0], prob.shape[1] * prob.shape[2])
         )
 
-        result = {"loss": loss, "output": probabilities}
+        result = {"loss": _np.array(loss), "output": _np.array(probabilities)}
         return result
 
     def predict(self, feed_dict):
--- a/src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
+++ b/src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
@@ -21,8 +21,9 @@ def _lazy_import_tensorflow():
     # This toolkit is compatible with TensorFlow V2 behavior.
     # However, until all toolkits are compatible, we must call `disable_v2_behavior()`.
     _tf.disable_v2_behavior()
-    return _tf
 
+    _tf = _minimal_package_import_check("tensorflow")
+    return _tf
 
 # Constant parameters for the neural network
 CONV_H = 64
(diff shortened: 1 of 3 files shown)


TobyRoseman:
I believe these changes are ready to go. Before merging, I will confirm these changes with a GitLab and benchmark run.

src/python/turicreate/toolkits/activity_classifier/_tf_model_architecture.py
```diff
@@ -41,219 +42,128 @@ def __init__(
...
-        Function to load LSTM biases from the C++ implementation into TensorFlow
-
-        Parameters
-        ----------
-        net_params: Dictionary
-            Dict with weights from the C++ implementation and  its names
+        # Set LSTM weights
+        i2h, h2h, bias = [], [], []
+        for i in ('i', 'f', 'c', 'o'):
+            i2h.append(eval('net_params["lstm_i2h_%s_weight"]' % i))
```
shreyajain17, line 131: Can you make this logic a utility function like earlier?

TobyRoseman: Sure, I'm happy to do that. This new function would only be called from one place, so I don't think it makes sense to put it in a separate file like it was before this change. If we want to put this code in a separate function, I would probably just add it as an inner function here, i.e. add this code as an inner function to this `__init__` method. I guess maybe that only really makes sense if it improves readability.

@shreyajain17 - what do you think?

TobyRoseman:
Internal tests show there is an issue with these changes and Python 2.7. I'm investigating. All other tests pass.

Benchmarks show this change is slower. Training typically takes close to twice as long. Prediction takes more than twice as long. This warrants some investigation.

All performance metrics (e.x. AUC, F1 score) are either flat or slightly improved. Backwards compatibility of saved AC models was check and is fine.

TobyRoseman:
The Python 2.7 issue was fixed by the [last commit](https://github.com/apple/turicreate/pull/3253/commits/7147ab859f94497777964036e007951271ad376e).

Slow LSTM runtime in Keras seems to be a common issue. It **seems** our issue is that we have to use a pure-TensorFlow LSTM rather than a cuDNN-based LSTM. From [the Keras documentation](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LSTM):
```
Based on available runtime hardware and constraints, this layer will choose different
implementations (cuDNN-based or pure-TensorFlow) to maximize the performance.
If a GPU is available and all the arguments to the layer meet the requirement of the
CuDNN kernel (see below for details), the layer will use a fast cuDNN implementation.

The requirements to use the cuDNN implementation are:

activation == tanh
recurrent_activation == sigmoid
recurrent_dropout == 0
unroll is False
use_bias is True
Inputs are not masked or strictly right padded.
```
Our usage satisfies all of those requirements, except we're using `use_bias=False`. Using a bias here would probably make things a lot faster. However, if we use a bias, then additional parameters will need to be serialized when saving a model. This would prevent models trained in TuriCreate 6.4+ from being loaded in an earlier version of TuriCreate.

Although the percentage increase is significant, the total time is still small.

PR #3266, opened 2020-07-15
Linking python include dirs in the build

abhishekpratapa:
In previous versions of virtualenv this would happen automatically. In the current versions this doesn't occur so we manually do it. This shouldn't break previous versions of virtualenv because of the flags passed into `ln`. Details to follow soon.


znation:
You're getting a build error on Travis:
```
In file included from /home/travis/build/apple/turicreate/src/external/pybind11/pytypes.h:12:0,
5646                 from /home/travis/build/apple/turicreate/src/external/pybind11/cast.h:13,
5647                 from /home/travis/build/apple/turicreate/src/external/pybind11/attr.h:13,
5648                 from /home/travis/build/apple/turicreate/src/external/pybind11/pybind11.h:49,
5649                 from /home/travis/build/apple/turicreate/src/ml/neural_net/tf_compute_context.cpp:18:
5650/home/travis/build/apple/turicreate/src/external/pybind11/detail/common.h:112:20: fatal error: Python.h: No such file or directory
5651compilation terminated.
```

znation:
@abhishekpratapa Any update on this? If we can solve this issue more generally it would help with getting CI set up on Gitlab.com as well.

abhishekpratapa:
> @abhishekpratapa Any update on this? If we can solve this issue more generally it would help with getting CI set up on Gitlab.com as well.

@znation looki

scripts/requirements.txt
```diff
@@ -11,7 +11,7 @@ tensorflow==2.0.1; python_version > '2.7'
 tensorflow==2.0.0; python_version < '3.5'   # Last version to support Python 2.7
 UISoup==2.5.7
 pyobjc==5.2; sys_platform == 'darwin'
-llvmlite==0.31.0
+llvmlite==0.33.0
```
TobyRoseman, line 14: I don't think this version support either Python 2.7 or 3.5. I recently ran into this issue myself (while trying to support Python 3.8). Here is the change I used:
```
llvmlite==0.33.0; python_version == '3.8'
llvmlite==0.31.0; python_version != '3.8'
```

shreyajain17:
It still doesn't work with Xcode 12.0 Public Beta.

PR #3273, opened 2020-07-23
Stop Disabling TensorFlow V2 Behavior

### LINKED ISSUE #2928: Do not call tensorflow.disable_v2_behavior()
TuriCreate calls `tensorflow.disable_v2_behavior()`. For consistency, we're doing this every time we `import tensorflow`.

Calling this function disables v2 behavior **globally**. It prevents other parts of the TuriCreate from using v2 behavior. If users are also using TensorFlow directly in the same Python session as TuriCreate, this could break user code too.

TobyRoseman:
Fixes #2928

### DIFF
--- a/src/python/turicreate/toolkits/_tf_utils.py
+++ b/src/python/turicreate/toolkits/_tf_utils.py
@@ -59,7 +59,6 @@ def suppress_tensorflow_warnings():
 
     _tf = _minimal_package_import_check("tensorflow.compat.v1")
 
-    _tf.disable_v2_behavior()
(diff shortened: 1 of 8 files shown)


TobyRoseman:
Internal test pass. 

I've verified that the TensorFlow graph is not eagerly evaluated before this change and is eagerly evaluated after this change.

Benchmark results look good. There doesn't seem to be any meaningful changes to any performance metric. Runtime metrics aren't quite as clear; some toolkits might be a bit faster and some might be a bit slower, but nothing significant.

I think these changes are ready merge.

PR #3274, opened 2020-07-27
Add NMS Layer to the OD Neural Network

shreyajain17:


src/ml/neural_net/model_spec.cpp
```diff
@@ -105,7 +107,7 @@ void update_weight_params(const std::string& name, const float_array& value, Wei
   }
 
   Span<const float> out(value.data(), value.size());
-
+  weights->mutable_floatvalue()->begin();
```
brtal, line 110: what's this for?

shreyajain17: Removed it.

PR #3276, opened 2020-07-27
Adding MLCompute as a training backend for OD 

shreyajain17:
Merge after https://github.com/apple/turicreate/pull/3266

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,7 @@ if(APPLE)
   find_library(CORE_VIDEO CoreVideo)
   find_library(METAL NAMES Metal)
   find_library(METAL_PERFORMANCE_SHADERS NAMES MetalPerformanceShaders)
+  find_library(MLCOMPUTE NAMES MLCompute)
   set(_TC_APPLE_DEPENDENCIES
       ${ACCELERATE}
       ${CORE_GRAPHICS}
@@ -510,6 +511,11 @@ if(NOT TC_DISABLE_OBJECT_BUILDS)
         set(_TC_COMMON_REQUIREMENTS
             ${_TC_COMMON_REQUIREMENTS} ${ACCELERATE} ${CORE_IMAGE} ${METAL}
             ${METAL_PERFORMANCE_SHADERS})
+      endif()
+      if(HAS_ML_COMPUTE)
+        set(_TC_COMMON_OBJECTS ${_TC_COMMON_OBJECTS} "$<TARGET_OBJECTS:tcmlc>")
+        set(_TC_COMMON_REQUIREMENTS
+            ${_TC_COMMON_REQUIREMENTS} ${ACCELERATE} ${CORE_IMAGE} ${MLCOMPUTE})
       endif()
     endif()
   endif()
--- a/cmake/SetupCompiler.cmake
+++ b/cmake/SetupCompiler.cmake
@@ -110,6 +110,12 @@ if(APPLE)
     add_definitions(-DHAS_CORE_ML)
     set(HAS_CORE_ML TRUE)
   endif()
+  
+  # MLCompute is only present on macOS 11.0 or higher.
+  if(NOT TC_BASE_SDK_VERSION VERSION_LESS 10.16)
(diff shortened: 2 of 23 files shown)


CMakeLists.txt
```diff
@@ -222,6 +222,7 @@ if(APPLE)
   find_library(CORE_VIDEO CoreVideo)
   find_library(METAL NAMES Metal)
   find_library(METAL_PERFORMANCE_SHADERS NAMES MetalPerformanceShaders)
+  find_library(MLCOMPUTE NAMES MLCompute)
```
shreyajain17, line 225: @abhishekpratapa

cmake/SetupCompiler.cmake
```diff
@@ -110,6 +110,12 @@ if(APPLE)
     add_definitions(-DHAS_CORE_ML)
     set(HAS_CORE_ML TRUE)
   endif()
+  
+  # MLCompute is only present on macOS 11.0 or higher.
+  if(NOT TC_BASE_SDK_VERSION VERSION_LESS 10.16)
```
TobyRoseman, line 115: Should this be `11.0` rather than `10.16`? Or does the comment on the line above it need to be updated?

shreyajain17: Changing to 10.16 for now.

brtal: i think comparing for `>= 10.16` makes sense (as opposed to `11.0`). is there any reason not to write this more clearly as:

```
if (TC_BASE_SDK_VERSION GREATER_EQUAL 10.16)
```

src/ml/neural_net/TCMLComputeObjectDetectorDescriptor.m
```diff
@@ -0,0 +1,144 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+#ifdef HAS_MACOS_10_16
```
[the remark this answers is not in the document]
shreyajain17: It doesn't build without these guards. If you look at 'ml/neural_network/style_transfer', those files also have guards inspite of building a certain files.

src/ml/neural_net/mlc_compute_context.hpp
```diff
@@ -0,0 +1,62 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+#ifdef HAS_MACOS_10_16
```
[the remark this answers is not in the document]
shreyajain17: Will do

src/ml/neural_net/TCMLComputeObjectDetectorDescriptor.h
```diff
@@ -0,0 +1,39 @@
+/* Copyright © 2020 Apple Inc. All rights reserved.
+ *
+ * Use of this source code is governed by a BSD-3-clause license that can
+ * be found in the LICENSE.txt file or at
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+#ifdef HAS_MACOS_10_16
```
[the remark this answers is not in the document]
shreyajain17: Okay

src/ml/neural_net/TCMLComputeObjectDetectorDescriptor.m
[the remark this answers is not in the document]
brtal: then it sounds like something is wrong with the build system. we shouldn't be compiling something specific to an MLCompute implementation on a system that doesn't have MLCompute. we should get to the root of this rather than throwing up guards all over the place.

cmake/SetupCompiler.cmake
brtal: sorry, this should be `VERSION_GREATER_EQUAL` not `GREATER_EQUAL`

PR #3285, opened 2020-07-31
Add l2_penalty property to text_classifier's create() method

matyasbohacek:
The change proposed in this merge request adds `l2_penalty` property to the `create()` method of `text_classifier`, allowing anyone to set custom weight on l2 regularization of the created models.

This property is optional, and **0.2** is left as the default value (this has previously been the hard-coded value passed once creating the `logistic_classifier`).

Relevant description has also been added to the **docstring**.


TobyRoseman:
This is a great change. Thank you, @matyasbohacek!

PR #3292, opened 2020-08-11
Fix bulid for newer versions of virtualenv

TobyRoseman:
* Newer `virtualenv` don't include header files.
* Also add back debugging output.


TobyRoseman:
internal tests pass

Issue #3099, opened 2020-04-08
Support Python 3.8


ChrisLane:
_Issue details copied/modified from #788_

As of October 14th, 2019 Python 3.8 has been officially released. We should support it.

At an absolute minimum these things will need to be done:
- [ ] Ensure everything builds with 3.8 and that the built package can be imported.
- [ ] All unit test pass on both macOS and Linux.
- [ ] Upload 3.8 wheel to PyPI on next release.

TobyRoseman:
Before we can support Python 3.8 all of our Python dependencies must support it. Looks like there are still two dependencies that don't support Python 3.8.

Adding a few more progress check boxes:
- [ ] [TensorFlow supports Python 3.8](https://github.com/tensorflow/tensorflow/issues/33374)
- [ ] [CoreMLTools supports Python 3.8](https://github.com/apple/coremltools/issues/684)
- [ ] Update [the versions of our dependencies](https://github.com/apple/turicreate/blob/6fa89a994c471194b5f09ae80fd1d5a951f52638/src/python/setup.py#L128)

igi-chris:
This put me off using this library. I had an issue with Pandas running out of memory for large data frames so wanted to try the `sframe` from this library, but don't really want to install an old Python version just to try this out.

laksh225:
CoreMLTools does support Python 3.8. I installed coremltools using pip install coremltools and imported it.

![Screenshot from 2020-05-19 18-56-54](https://user-images.githubusercontent.com/56289983/82332441-03f21e80-9a03-11ea-886e-c09c49020c69.png)

OUTPUT : (venv) laksh@L-ubuntu18:~$ python
Python 3.8.2 (default, Apr 27 2020, 15:53:34) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import coremltools
>>>

TobyRoseman:
Supporting Python 3.8 on Linux is blocked by #3003. In order to support Python 3.8 we need to upgrade our TensorFlow dependency and not have things segfault.

TobyRoseman:
TuriCreate 6.4 has been released with Python 3.8 support.

ankur290109:
Getting below error while installing Turicreate, I have installed Python 3.8 and Anaconda 3:-

ERROR: Failed building wheel for turicreate
Running setup.py clean for turicreate
Failed to build turicreate
Installing collected packages: turicreate
Running setup.py install for turicreate ... error
[...]

PR #3295, opened 2020-08-13
Python 3.8 Support

TobyRoseman:
#3099

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -385,7 +385,6 @@ test_python_linux_python27:
     - build_wheel_linux_python27
   script:
     - apk add bash
-    - export TC_ENABLE_S3_TESTS=1
     - bash -e scripts/test_wheel.sh --docker-python2.7
   artifacts:
     when: always
@@ -408,7 +407,6 @@ test_python_linux_python27_minimal:
     - build_wheel_linux_python27
   script:
     - apk add bash
-    - export TC_ENABLE_S3_TESTS=1
     - bash -e scripts/test_wheel.sh --docker-python2.7 --minimal
   artifacts:
     when: always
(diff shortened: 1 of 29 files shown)


.gitlab-ci.yml
```diff
@@ -349,6 +385,7 @@ test_python_linux_python27:
     - build_wheel_linux_python27
   script:
     - apk add bash
+    - export TC_ENABLE_S3_TESTS=1
```
znation, line 388: I don't think we want `TC_ENABLE_S3_TESTS=1` in master; last I checked this will cause test failures in public CI.

src/python/turicreate/util/lambda_closure_capture.py
```diff
@@ -256,7 +258,13 @@ def visit_FunctionDef(self, node):
             # it actually shows up in the ast as a Expr.str
             # so we need to catch that and skip it
             try:
-                if type(next_node) is ast.Expr and type(next_node.value) is ast.Str:
+                # Docstrings are constants in Python 3.8 and strings in all other
+                # supported Python versions.
+                if python_version.major == 3 and python_version.minor == 8:
```
znation, line 263: Should this be `python_version.major == 3 and python_version.minor >= 8` to account for future Python3 versions that will also need this behavior? Presumably Python 4+ will as well but I wouldn't worry about those yet.

.gitlab-ci.yml
TobyRoseman: Good catch. This was the result of a bad merge. I've removed all three occurrences.

src/python/turicreate/util/lambda_closure_capture.py
TobyRoseman: That's a good idea. Done.

Issue #3236, opened 2020-06-24
Drawing classifier segmentation fault occurs when creating model


hipwelljo:
Running `model = tc.drawing_classifier.create(train_data, target='label')` results in a crash.

```
(venv) ... % python3 createmodel.py
Automatically generating validation set from 5% of the data.
WARNING:tensorflow:From /Users/Jordan/venv/lib/python3.7/site-packages/tensorflow/python/compat/v2_compat.py:96: disable_resource_variables (from tensorflow.python.ops.variable_scope) is deprecated and will be removed in a future version.
Instructions for updating:
non-resource variables are not supported in the long term
Using CPU to create model.
+--------------+----------------+---------------------+--------------+
| Iteration    | Train Accuracy | Validation Accuracy | Elapsed Time |
+--------------+----------------+---------------------+--------------+
zsh: segmentation fault  python3 createmodel.py
```

It also generates a Python crash report. Please see attached zip that includes a script you can use to replicate the issue. I've included the images, about 230 classes with 15 images for each, a small subset of my real data.
[Drawing.zip](https://github.com/apple/turicreate/files/4826855/Drawing.zip)

 Thanks!

Turi Create 6.3
macOS Catalina 10.15.5 (19F101)
Python 3.7

TobyRoseman:
This issue reproduces with in a debug build of master. 

We're getting a `stop reason = EXC_BAD_ACCESS` [here](https://github.com/apple/turicreate/blob/af567ff5ba333f39ff9b7dac44e74a7b558c0da7/src/model_server/lib/image_util.cpp#L63).

The most recent part of the call stack looks like this:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x7fda3e331000)
  * frame #0: 0x000000011fb71ab0 libunity_shared.dylib`void turi::image_util::copy_image_to_memory_impl<float>(input=0x00007ffee957fce8, outptr=0x00007fda3e32a1c0, outstrides=size=3, outshape=size=3, channel_last=true) at image_util.cpp:63:60
    frame #1: 0x000000011fb7176e libunity_shared.dylib`turi::image_util::copy_image_to_memory(input=0x00007ffee957fce8, outptr=0x00007fda3e32a1c0, outstrides=size=3, outshape=size=3, channel_last=true) at image_util.cpp:77:3
[...]

PR #3313, opened 2020-09-03
Drawing Classifier: resize input images

TobyRoseman:
Fixes: #3236


TobyRoseman:
This fixes the bug, but there is likely a better/faster way. This change will resize the images on each batch. Ideally, we should only resize each image once.

TobyRoseman:
Internal test pass

shantanuchhabra (approved):

PR #3328, opened 2020-09-24
New pip dependency resolver

TobyRoseman:
* Add `--use-feature=2020-resolver` to our Python build/test setup to ensure we stay compatible with the new dependency resolver.
* Also in our Python build/test setup, pin TensorFlow version (for a few Python versions and OS combinations). Without this one of our packages used for unit testing causes a SciPy version conflict.


TobyRoseman:
Internal tests pass

nickjong (approved):

PR #3329, opened 2020-09-25
Remove tc.config.get_server_log_location()

TobyRoseman:


src/python/turicreate/data_structures/image.py
```diff
@@ -226,7 +226,7 @@ def save(self, filename):
 
     def show(self):
         """
-        Displays the image. Requires PIL/Pillow.
```
znation, line 229: Why remove this comment? I think it is still true.

znation (approved):

src/python/turicreate/data_structures/image.py
TobyRoseman: It is still true, but PIL/Pillow is now a hard dependency. So I don't think it makes sense to mention.

znation: I see. Makes sense, thanks!

PR #3333, opened 2020-09-29
Update to SFrame serialization to fix issues with python 2.7 and cloudpickle.

hoytak:


src/python/turicreate/data_structures/serialization.py
```diff
@@ -8,6 +8,7 @@
 # SFrames require a disk-backed file or directory to work with.  This directory 
 # has to be present to allow for serialization or deserialization. 
 __serialization_directory = None 
+from sys import version_info
```
TobyRoseman, line 11: You're already importing this in the function which it is used. So you don't need this one. 

If you do want to keep this one, please hide.

TobyRoseman (approved):

TobyRoseman (approved):

PR #3337, opened 2020-10-01
Fixed issues in python toolchain setup script.

hoytak:
Current python uses `python -m venv` to run virtualenv.  This updates our toolchain setup to reflect this. 

Also fixes a bug in which creating softlinks can cause the script to fail if they are already present.

### DIFF
--- a/scripts/install_python_toolchain.sh
+++ b/scripts/install_python_toolchain.sh
@@ -3,7 +3,7 @@
 set -exo pipefail
 
 if [[ -z $VIRTUALENV ]]; then
-  VIRTUALENV=virtualenv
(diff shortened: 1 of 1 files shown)


scripts/install_python_toolchain.sh
```diff
@@ -3,7 +3,7 @@
 set -exo pipefail
 
 if [[ -z $VIRTUALENV ]]; then
-  VIRTUALENV=virtualenv
+  VIRTUALENV='python3 -m venv'
```
TobyRoseman, line 6: We still need to support Python 2.7. I don't think this is going to work for Python 2.7

Issue #3342, opened 2020-10-02
Published 6.4.1 requires_dist has inconsistent coremltools version


malcolmgreaves:
The PyPi published package for `turicreate` version `6.4.1` has inconsistent dependency information. 

The package's `requires_dist` from [pypi/turicreate/json](https://pypi.org/pypi/turicreate/json ) has `"coremltools (==3.3)"`. (From `curl -sL https://pypi.org/pypi/turicreate/json | jq '.info.requires_dist'`). 

However, the wheel's `requires_dist` is different: executing `$ pkginfo requires_dist`[`turicreate-6.4.1-cp38-cp38-manylinux1_x86_64.whl`](https://pypi.org/project/turicreate/6.4.1/#files) lists the dependency as 'coremltools (==4.0b3)'. 

This means that installation of `turicreate` with tools that use the published `requires_dist` will fail for Python 3.8 environments. For instance, I ran into this when using `poetry`: https://github.com/python-poetry/poetry/issues/3011.

Given that [`setup.py`](https://github.com/apple/turicreate/blob/master/src/python/setup.py#L182) puts the version at `4.0b3` for Python 3.8 and `3.3` for Python 3.7, I believe that the `install_requires` in `setup.py` should specify the compatible versions explicitly.

malcolmgreaves:
I have this PR https://github.com/apple/turicreate/pull/3343, which I think will fix the publishing issue going forward.

TobyRoseman:
Thanks for the information and the pull request. 

So to summarize: in our [setup.py](https://github.com/apple/turicreate/blob/a918b2d76c165f62c19895465b6ec443d7ada9fd/src/python/setup.py#L166-L205) we need to put all version requirements in `install_requires`, not just the requirements for that particular wheel. Is that correct?

I think the coremltools difference may just be the first issue you're going to hit here. Looking at [setup.py](https://github.com/apple/turicreate/blob/a918b2d76c165f62c19895465b6ec443d7ada9fd/src/python/setup.py#L166-L205) there are a few other dependencies that will be different for different wheels.

malcolmgreaves:
Hi @TobyRoseman thanks for the assistance!

I believe that is the right course of action, to have the `install_requires` specify all dependencies alongside Python version requirements, if applicable. (I'm no Python packaging expert -- I got the PR solution from python-poetry/poetry#3011 and some googling).

malcolmgreaves:
[...]

PR #3343, opened 2020-10-02
Make coremltools is python-dependent in requires_dist

malcolmgreaves:
The`coremltools` dependency depends on the user's Python version. The logic for specifying this already exists in `setup.py`. This PR makes the `coremltools` version explicitly dependent on the Python version in the `install_requires` entry.

This change fixes https://github.com/apple/turicreate/issues/3342 .


src/python/setup.py
```diff
@@ -179,11 +179,12 @@ def do_not_install(require):
     ):
         install_requires.append("llvmlite == 0.31.0")
 
-    if sys.version_info[0] == 3 and sys.version_info[1] == 8:
```
TobyRoseman, line 182: I would much prefer a change like the following:
```python
    if sys.version_info[0] == 3 and sys.version_info[1] == 8:
         # Only 4.0 betas support Python 3.8
         install_requires.append("coremltools==4.0b3; python_version >= 3.8.0',")
     else:
         install_requires.append("coremltools==3.3; python_version < 3.8.0,")
```
I think that is much more readable and accomplishes the same thing.

malcolmgreaves: Sure thing!

malcolmgreaves: Changed.

PR #3344, opened 2020-10-02
Update random number generator to use C++11 standard library

hoytak:
Currently, our random number generator uses boost random library, which has largely been supplanted by C++11.  This PR updates the random number tools to use the C++11 intrinsics. 

This change is needed to allow us to transition to C++17.  In c++17, std::random_shuffle is removed in place of std::shuffle, which is no longer compatible with the boost random generators.  This requires us to change the base random generators to the c++11 versions, and thus swap out the distributions to the c++11.   After this PR, we no longer depend on boost random. 

The c++11 versions allow us to simplify a couple of things: 

1.  C++11 provides a true-random generator, which can replace our complicated, platform-specific version. 
2. C++11's mt19937_64 benchmarks as one of the fastest random number generators on modern hardware (e.g. https://gist.github.com/MikeLing/d258cf0c7e18d17639c577a695dbec41), so there is no point in having separate fast_ versions.  This PR consolidates these.  This change is in the second commit.

### DIFF
--- a/src/external/boost/CMakeLists.txt
+++ b/src/external/boost/CMakeLists.txt
@@ -7,7 +7,6 @@ set(__boost_lib_dir "${__boost_src_dir}/libs/")
 file(GLOB_RECURSE boost_src_files
   CONFIGURE_DEPENDS
   ${__boost_lib_dir}/system/src/*.cpp
-  ${__boost_lib_dir}/random/src/*.cpp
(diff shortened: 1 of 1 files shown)


znation (approved):

TobyRoseman:
@hoytak - will this change impact our users? Is there functionality where the same data and the same seed will now produce different results? If so, what functionality is affected?

PR #3346, opened 2020-10-03
Use new coremltools beta

TobyRoseman:
Now that apple/coremltools#372 is fixed, we can use the latest coremltools beta.


TobyRoseman:
Internal test pass

abhishekpratapa (approved):

PR #3398, opened 2021-03-29
update configuration 

rschilling:
update configuration file to support cmake 3.20.0
It is not possible for me to validate if this change works on all versions of cmake used by this project.
Please reject this PR if the change breaks other builds.d
Is there a development branch I could submit pull requests into?


TobyRoseman:
Internal tests look good. @rschilling - Thanks for the pull request.

TobyRoseman (approved):

PR #3479, opened 2023-11-01
Optimizing Performance and Resource Utilization with Concurrent Asynchronous Execution 🚤

sanjaiyan-dev:
Hi, this PR introduces a minor performance improvement by running the asynchronous function concurrently using `Promise.all`, which enhances the speed and optimizes the utilization of hardware resources.


pratyaksh123:
@sanjaiyan-dev Any idea how can we run this lib on colab? It doesn't support python 3.10.

sanjaiyan-dev:
> @sanjaiyan-dev Any idea how can we run this lib on colab? It doesn't support python 3.10.

Hi @pratyaksh123 , try running it by downgrading the python version

